Issue #138, opened 2014-02-11
A neutral grouping object.


samuelgarcia:
I detached the end of this issue #132.
I think it is also link to issue #96 and #137

There is a global needs for grouping things together.
Not only Unit or RC.

use case are quite specific:

Andrew wants : Parameters and ParametersSet for navigation throuth SpikeTrain with same parameters at simulation time
Robert and Todd want : Condition for grouping Segment (=trial)
Samuel wants : RCG for grouping RC for spike sorting
Other users wants : RCG for grouping Units that represent a population.
Roberts wants : gourping some Epochs.

Could we have some kind of very neutral Group that could link neo objects of the same type.

g1 = neo.Group(type = 'RecordingChannel', name = 'tetrode1', items = [rc1, rc2, rc3, rc4])
g2 = neo.Group(type = 'RecordingChannel', name = 'tetrode2', items = [rc3, rc7, rc8, rc9])

or 
g1 = neo.Group(type = 'Segment', name = 'trial with condition A', items = [seg1, seg2, seg3])
g1.annotate(intensity = 'heavy')

or 
g1 = neo.Group(type = 'SpikeTrain', name = 'parameters set', items = [st1, st2, st3])
g.annotate(param1 = 8, param2 = 12)

And group would attached to Block: 
Block.groups

It could solve in one simple way many issues ?

PR #144, opened 2014-03-25
sphinx documentation on Collection object

toddrjen:
Here is the requested sphinx docs on the proposed collection (see issue #138).  Please DO NOT merge this, at this stage it is simply a proposal for what could be done.


samuelgarcia:
I guess we can close this. Since it was a proposal for discussion and it was anterior to new apibreak branch of Andrew.

PR #187, opened 2015-01-06
Plexon IO class changes

maxgillett:
- Times and labels (i.e. event codes) for event data are now accessible in an EventArray object that belongs to a given Segment object. 
- Annotations for segment and spiketrain objects now include all properties found in their respective file and spike channel headers (see [documentation](http://hardcarve.com/wikipic/PlexonDataFileStructureDocumentation.pdf)).
- A host of typos and style have been fixed to make the code more readable.


samuelgarcia:
Hi,
thank you.
Travis is unhappy but I think it has nothing todo with that commit.

It is OK for me.

PR #190, opened 2015-02-13
Some attempted fixes for failing tests in the current master branch

apdavison:
(presumably ones which are skipped by Travis). Please could @maxgillett and @samuelgarcia take a look, as the failures were in `AxonIO` and `PlexonIO`.

### DIFF
--- a/neo/io/axonio.py
+++ b/neo/io/axonio.py
@@ -296,7 +296,7 @@ class AxonIO(BaseIO):
 
                     anaSig = AnalogSignal(signal, sampling_rate=sampling_rate,
                                           t_start=t_start,
-                                          name=name.decode('utf-8'),
+                                          name=str(name),
                                           channel_index=int(num))
                     if lazy:
                         anaSig.lazy_shape = subdata.shape[0]
--- a/neo/io/plexonio.py
+++ b/neo/io/plexonio.py
@@ -241,17 +241,18 @@ class PlexonIO(BaseIO):
         for chan, h in iteritems(eventHeaders):
             if lazy:
                 times = []
+                labels = None
             else:
                 times = evarrays[chan]['times']
-
+                labels = evarrays[chan]['labels']
             ea = EventArray(
                 times*pq.s,
-                labels=evarrays[chan]['labels'],
+                labels=labels,
                 channel_name=eventHeaders[chan]['Name'],
                 channel_index=chan
             )
             if lazy:
-                ea.lazy_shape = nb_events[chan]['times']
+                ea.lazy_shape = nb_events[chan]
             seg.eventarrays.append(ea)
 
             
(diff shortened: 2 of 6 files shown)


neo/io/axonio.py
```diff
@@ -296,7 +296,7 @@ def read_block(self, lazy=False, cascade=True):
 
                     anaSig = AnalogSignal(signal, sampling_rate=sampling_rate,
                                           t_start=t_start,
-                                          name=name.decode('utf-8'),
+                                          name=str(name),
```
samuelgarcia: What happen if na me contain non ascii ?
better name.decode('utf-8').encode('ascii', 'ignore') no ?

neo/io/plexonio.py
```diff
@@ -241,17 +241,18 @@ def read_segment(self, lazy=False, cascade=True, load_spike_waveform=True):
         for chan, h in iteritems(eventHeaders):
             if lazy:
                 times = []
+                labels = None
```
samuelgarcia: why not labels = [ ]

neo/io/axonio.py
apdavison: I'm not sure what the restrictions are on the type of `AnalogSignal.name`. The tests require it to be of type `str`, which means effectively ASCII for Python 2 and unicode for Python 3. Couldn't we allow unicode in Python 2?

samuelgarcia: Yes. it would be better. It is not a big API break.
The thing is that we need to hack the test engine which test that each attribute have the good type (only one).
If we change AnalogSignal.name to unicode we need to change for all IOs. The better would to allow several types for each attributes whihch is not the case see AnalogSignal._necessary_attrs and BaseNeo._necessary_attrs.

neo/io/plexonio.py
apdavison: The constructor for `EventArray` is :

```python
EventArray(times=None, labels=None, name=None, description=None,
                   file_origin=None, **annotations)
```

samuelgarcia: Yes but in case of lazy load the len need to be 0 because in case of full load labels is not None but an array.
This is specifyed nowhere but in for me in case of lazyload : each array need to be of size 0 when exist at full load.

The IO support labels for EventArray so labels = [ ].
For other IOs that have EventArray.labels = None even at fullload so when lazyload EventArray.labels = None.

I am not really sur to clear, sorry.

apdavison: But inside the `EventArray` constructor is the following:

```python
if labels is None:
    labels = np.array([], dtype='S')
```

so the result of passing `labels=None` is that `labels` becomes an empty array of strings.

samuelgarcia: OK sorry. i have not touch the code for a while.

neo/io/axonio.py
apdavison: ok, changing to unicode for all IOs is a big effort, so I think that should be done in a separate PR. If you're otherwise ok with these fixes, can you merge this?

PR #191, opened 2015-02-24
Don't force EArray type to float64.

matham:
Right now, no matter the `dtype` of the `AnalogSignal`, the EArray is created as a float64. This leads to unnecessarily large files. This PR allows the user to set whether to respect the dtype of the Arrays.


samuelgarcia:
Hi.
OK for me.
Thank you.

PR #194, opened 2015-04-09
Optionally include Spike2 sorted unit tags in SpikeTrain.

matham:
Right now, if you sorted spikes with spike2, Spike2 will attach a unit number to each spike using the first marker byte, indicating which unit generated the spike. However, when reading the smr file, neo ignores the units and only creates a single SpikeTrain pretending as if the spikes all come from a single unit.

This adds a `ced_units` keyword, which when true, will create a SpikeTrain for each unit. Otherwise, the default, will just create a single SpikeTrain ignoring units like before. It also adds a `ced_unit` keyword to the annotations distracting from which Spike2 unit this SpikeTrain was generated.

### DIFF
--- a/neo/io/spike2io.py
+++ b/neo/io/spike2io.py
@@ -66,15 +66,22 @@ class Spike2IO(BaseIO):
 
     mode = 'file'
 
-    def __init__(self , filename = None) :
+    ced_units = False
+
+    def __init__(self , filename = None, ced_units = False) :
         """
         This class read a smr file.
 
         Arguments:
             filename : the filename
+            ced_units: whether a spike trains should be added for each unit
+                as determined by Spike2's spike sorting (True), or if a spike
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi,
thank you for this.
In neo the good way to manage unit is the neo.Unit object.
Which for SpikeTrain what RecordingChannel is for AnalogSignal.

Block.units is a list of neo.Unit
Unit.spiketrains is a list of SpikeTrain along Segment.
It is bit harder to construct in the IO than simple annotation but more compatible with neo schema.

Few IOs deal correctly with this. So no problem.

apdavison:
> So no problem

@sgarcia - when you say "no problem", do you mean we should accept this PR, or do you mean that @matham should rewrite the IO to use `Unit`?

samuelgarcia:
I guess it was meaning : *OK for merging*

PR #196, opened 2015-04-30
time_slice of analogsignalarray also accepts None as t_start and t_stop values

JuliaSprenger:
Analogsignalarray.time_slice() does not require two quantities as input (t_start and t_stop), but also accepts None for any of them. If None is provided as t_start (t_stop) no shortening of the signal is performed at the beginning (end) of the signal.


apdavison:
Many thanks for this. Please could you add a unit test?

JuliaSprenger:
I added a unit test to cover 'None' as input parameter.

PR #197, opened 2015-05-21
Merged recent changes in "master" into "apibreak"

apdavison:
With this update, I think we are ready for the next round of api-breaking changes.


apdavison:
In the absence of comments, I will go ahead and merge this.

PR #209, opened 2015-08-19
Fix lazy loading for AxonIO

kh90909:
No matter whether lazy loading was requested or not, all of the data was
still read from the file. The lazy setting only affected whether the
data was put into the AnalogSignal object(s) or not.


samuelgarcia:
Oh yes. i missed that.
Thank you for the patch.
Travis is complaning but not because of this patch I propose to merge.

kh90909:
You're welcome. Glad I could help.

PR #212, opened 2015-09-18
Changed "== None" to "is None" to avoid element-wise comparison warning.

rgerkin:
Numpy warns that in version 1.9 all comparisons of the type "== None" are element-wise.  Since t_start and t_stop can be quantities arrays, they trigger this warning when they are compared to None using "==".  Since t_stop can be None, but can never be a quantity array with None (e.g. [None] * ms), element-wise comparison should never be needed so this change can eliminate the warning.


samuelgarcia:
Thanks.
Ok for me.

PR #213, opened 2015-09-23
Implementation of read_segment for TdtIO

npyoung:
I had some experimental collaborators who were saving all their data in a single TDT DataTank, so that `read_block`, the only read method for TdtIO, attempted to read > 100GB into memory. To get around this, I needed a method for reading individual segments. So in my fork I have

1. Implemented `TdtIO.read_segment`, which takes a `blockname` argument (block being the TDT block, the Neo segment) to specify which block. This was done by factoring the segment-reading code out of a for loop in read_block into its own function.
2. Made some formatting changes for better PEP8 compliance.
3. Used try/except blocks instead of conditionals when checking if a file exists. This is better practice (avoids a race condition).

The new code passes the TdtIO unit test.

Alternatively, I considered implementing `load_lazy_object`, which would have you call `read_block(lazy=True, cascade=True)`, then pass in segment objects to `load_lazy_object`. However it turned out that the call to `read_block` is _very_ slow since it parses all the data files in the tank. It's much faster to `os.listdir` the DataTank to see the segment names. and use `read_segment`, as I have implemented it.

### DIFF
--- a/neo/io/tdtio.py
+++ b/neo/io/tdtio.py
@@ -35,7 +35,7 @@ def get_chunks(sizes, offsets, big_array):
     # offsets are octect count
     # sizes are not!!
     # so need this (I really do not knwo why...):
-    sizes = (sizes -10)  * 4 # 
+    sizes = (sizes -10)  * 4 #
     all = np.concatenate([ big_array[o:o+s] for s, o in itertools.izip(sizes, offsets) ])
     return all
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Thank you. I write long time ago but never used it.
This looks good, sorry to not have done that from beginning.
I propose to merge.

PR #215, opened 2015-10-09
t_stop and t_start property for segments

JuliaSprenger:
I recently discovered the absence of t_start and t_stop properties of segments. Since segments are meant to cover a specific time window, I think it could be useful to include the t_start and t_stop properties also for segments.
 My implementation calculates t_start and  t_stop for a segment based on the earliest (latest) signal, which is linked to this segment. Probably there are more efficient implementations, but this could be a starting point.

### DIFF
--- a/neo/core/segment.py
+++ b/neo/core/segment.py
@@ -103,6 +103,40 @@ class Segment(Container):
         self.rec_datetime = rec_datetime
         self.index = index
 
+    # t_start attribute is handled as a property so type checking can be done
+    @property
+    def t_start(self):
+        '''
+        Time when first signal begins.
+        '''
+        t_starts = [sig.t_start for sig in self.analogsignalarrays + self.analogsignals +  self.spiketrains]
+        t_starts += [e.times[0] for e in self.epocharrays + self.eventarrays + self.irregularlysampledsignals]
+        t_starts += [e.time for e in self.events + self.epochs + self.spikes]
+
+        # t_start is not defined if no children are present
+        if len(t_starts)==0:
+            return None
+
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Agree with this but I would give another name to avoid confusion with OtherObject.t_start.
No specific idea, but something like absolut_t_start/absolut_t_stop.

No ?

JuliaSprenger:
I thought it would be more intuitive to stick to the existing property names instead of adding another vocable to learn. But if you plan to introduce another type of t_start and t_stop for segments in the future this could be a reason for selecting a different name. Should I then use 'absolute_t_start' and 'absolute_t_stop' instead or should we come up with a different (perhaps shorter) name?

samuelgarcia:
Maybe you are right, it is more intuituitive to keep with t_start/t_stop it the doc clearly indicate that it is an a properties and the value is able change depending of object in the segment.

Lets wait for Andrew.

asobolev:
Julia, thanks for the change. As we discussed a couple of year ago in
Juelich, I think it is very logical and I completely support it.

The t_start and t_stop for the Segment (time domain) look very suitable.
IMHO for the signals (AnalogSignal etc.) would be friendly to have "offset"
and "duration" - these are short and hold clear what they mean (relative
values).

It's clear this would break existing scripts. However one could introduce
this with the new Neo 0.4 as there are non-compatible changes anyway.

Bests,
Andrey


2015-10-09 11:49 GMT+02:00 Garcia Samuel <notifications@github.com>:

samuelgarcia:
OK for me.
Andrew ?

PR #218, opened 2015-10-20
Update neuroshareapiio.py

amchagas:
There was a small bug in read_analogsignal function. When only a chunk of a file was read (say from second 5 to second 11), the function would read more than the chunk specified. It would start at second 5, but then read 5+6 seconds of data, instead of only 6 seconds.


samuelgarcia:
Ok for me.

Issue #109, opened 2013-12-19
Add IgorIO


samuelgarcia:
base on matlab and R we could add an igor IO

http://www.inside-r.org/packages/cran/IgorR/docs/read.ibw

https://www.mathworks.fr/matlabcentral/fileexchange/42679-igor-pro-file-format-ibw-to-matlab-variable/content/readIBWheaders.m

wvangeit:
Just want to mention there is also a python reader for igor you could use to implement this: https://pypi.python.org/pypi/igor

apdavison:
Leaving this open until .pxp support is added.

PR #225, opened 2015-11-30
Added an IO for IGOR Pro. Limited functionality so far.

apdavison:
cf #109

This PR is against the apibreak branch (0.4 dev), but the code should also work with the master branch (0.3.x) if someone wants to use it.

I will add some tests when I have time.

### DIFF
--- a/neo/io/igorproio.py
+++ b/neo/io/igorproio.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+"""
+Class for reading data created by IGOR Pro (WaveMetrics, Inc., Portland, OR, USA)
+
+Depends on: igor (https://pypi.python.org/pypi/igor/)
+
+Supported: Read
+
+Author: Andrew Davison
+
+"""
+
+from __future__ import absolute_import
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi,
to avoid dependency could we copy/paste the internal code of igor inside neo directly (with author autorisation ?

Samuel

apdavison:
I don't think there is a big advantage to having the code inside Neo; it is a simple Python dependency, `pip install igor` works fine.

samuelgarcia:
Hi Andrew.
Have you a small file to put here https://portal.g-node.org/neo/ for running tests ?

jakirkham:
It seems `igor` last saw a new release in 2012. There is also `igor.py`, which saw a release in 2015. Though haven't evaluated other merits of the two.

apdavison:
On the `igor.py` page it states that it has been superseded by `igor`. The 2015 'release' was apparently just to fix a typo in the README.

apdavison:
Closing this, since the apibreak branch has been merged to master. Replaced by #251

PR #235, opened 2016-02-22
Rename `RecordingChannelGroup` to `ChannelIndex`

apdavison:
This follows the discussion in #210. Since individual `RecordingChannel` objects have been removed from the proposed Neo 0.4 API, the name `RecordingChannelGroup` no longer seems appropriate.


samuelgarcia:
Hi andrew.

OK for this.

The only litlle comment I could do is that we lost here the semantic of **Group**.
Maybe this not important.


A typical use case even if noby use I guess is:
We have 128 channels that come from 2 diffrents arrays of 64 electrodes in the same recording.
```
signals = neo.AnalogSignal(np.zeros(nb_sample, 128), units='uV', sampling_rate=10*pq.kHz)
indexes_A = neo.ChannelIndex(name='shank A', channel_indexes=np.arange(0,64))
indexes_B = neo.ChannelIndex(name='shank B', channel_indexes=np.arange(64,128))
```

Maybe the semantic of ChannelGroup is easier to catch. I do known.

This is just to have your opinion on that I don't care so much. 
So OK for me for merge.

Thank again for all the work in apibreak.

PR #239, opened 2016-03-24
Added fourth iteration of BlackrockIO module with new features.

lzehl:
This pull request contains a fourth iteration of the BlackrockIO module to read files stored in the Blackrock file format.

The module is a combination from the previous two iterations combining the optimized, efficient, and more Neo compliant code of iteration 3 by @samuelgarcia with the important extended functionality of iteration 2. In addition, this IO adds several new features, in particular support for files of specification 2.1 and 2.3 and recording pauses (in 2.2 and 2.3). 

Also the code has been restructured to make it easier to add upcoming file specifications of Blackrock, by grouping read operations common to multiple file specifications (termed 'variants', e.g. 'variant_a').

Two new specific unit tests complement the generic IO unit tests and check the content of the read data objects, in part based on the Matlab implementation of the loading routine provided by the manufacturer.

### DIFF
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -2192,11 +2192,12 @@ class BlackrockIO(BaseIO):
         # read recordingchannelgroup
         if channels:
             for i, ch_idx in enumerate(channels):
-                if ch_idx in units.keys() and not \
+                if units and ch_idx in units.keys() and not \
                         isinstance(units[ch_idx], types.NoneType):
                     ch_units = units[ch_idx]
                 else:
                     ch_units = None
+
                 rcg = self.__read_recordingchannelgroup(
                     channel_idx=ch_idx,
                     index=i,
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -1035,10 +1035,8 @@ class BlackrockIO(BaseIO):
         """
         tp = []
         if self._avail_files['nev']:
-            print 'nev', self.__get_nev_rec_times()[0]
             tp.extend(self.__get_nev_rec_times()[0])
         for nsx_i in self._avail_nsx:
-            print nsx_i, self.__nsx_rec_times[self.__nsx_spec[nsx_i]](nsx_i)[0]
             tp.extend(self.__nsx_rec_times[self.__nsx_spec[nsx_i]](nsx_i)[0])
 
         return min(tp)
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -1744,8 +1744,9 @@ class BlackrockIO(BaseIO):
             st.left_sweep = self.__get_left_sweep_waveforms()
 
         # add additional annotations
-        st.annotate(ch_idx=int(channel_idx))
-        st.annotate(unit_id=int(unit_id))
+        st.annotate(
+            ch_idx=int(channel_idx),
+            unit_id=int(unit_id))
 
         return st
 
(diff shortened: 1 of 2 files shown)


lzehl:
Current travis fail seems to be related to nose test runner, but the reason is unclear to us.

mdenker:
To be compatible with the new Neo hierarchy, the last commits have changed the structure from creating one RecordingChannelGroup object for the complete array to creating one RCG per electrode (since Unit objects are directly attached to one particular RCG, but it is not possible to identify on which of the channels of the group the unit was recorded).

samuelgarcia:
Hi.
Many thanks for continuing the effort on blackrockio!! A long story.

If I anderstand correctly this PR only modify the tests file to match modifications in APi break branch ?

Ok for me.

mdenker:
Hi Samuel,

no, the PR contains basically a rewrite that tries to merge the best of the two previous versions of the IO (the diff for the main file is not shown on github since it's too large). It also provides additional unit tests on the test files.  Please see the initial message by lzehl for a summary of the most important changes. The main motivation for this work is that we could not get the current IO in the main branch to work for the majority of our data files (in part because of the zoo of different file format versions that are around -- hopefully the PR improves the situation).

Best,
Michael

samuelgarcia:
Thanks, I haven't seen it... (I am stupid)
This is a hudge work!
Woooaa!

lzehl:
with the last commit I corrected an error in the blackrockio (introduced with the previous changes) and updated the corresponding tests. The travis test is still failing though. I guess for the same reason as before where it seemed to be related to nose test runner, but the reason is still unclear to us.

lzehl:
Last two commits solves a bug concerning the selection of data blocks given the user n_starts and n_stop times. This needs further testing if a dataset with more than one recording block is available for tests.

Current solution: 
Parameters of data_block_i are used if at least n_start or n_stop are within the intrinsic time range (t_start_i, t_stop_i) of the data block. The first data_block for which one of the conditions is True will be returned. The case that the n_start lies within one data block and the n_stop within another data block should not occur, since the n_start and n_stop times of the user are directly validated in the beginning against possible recording pauses and correspondingly extended by stop and start points to avoid this situation. (In such a case the user would automatically get back two segments and a corresponding warning in the verbose print out).

jakirkham:
I don't know if you have tried Python 3.x with this, but the existing Blackrock support needed a couple minor tweaks for Python 3.x, which I have added with this PR ( https://github.com/NeuralEnsemble/python-neo/pull/244 ). Not sure if that PR is relevant in this context or not.

samuelgarcia:
Hi.
Do you want a merge into apibreak now or do you plan more work on this ?

apdavison:
I have manually merged this. This required a number of modifications, mainly

- `RecordingChannelGroup` --> `ChannelIndex` (since I recently merged the `channelindex` branch into `apibreak`)
- changes to get the tests to pass with Python 3

Please can @lzehl @JuliaSprenger and/or @mdenker check that this now works as expected (in both Python 2.7 and 3.x), in case I've broken something.

mdenker:
@apdavison Great, thanks for the hard work. We are currently still in the process of testing this IO, to make sure it works well -- we'll use the updated version as a basis for this now.

@jakirkham Thanks for pointing out the PR, we were not aware of it -- the IO has received quite some changes, so, we'll check that the issues you point to are addressed.

jakirkham:
I think that PR got merged in then end. Though it is probably still worthwhile to go back and check Python 3.x compatibility. I'll try to help out in this effort as I have time.

PR #241, opened 2016-04-25
Make TdtIO more robust

SummitKwan:
Make TdtIO more robust:
Add a function “is_block” to class TdtIO. This function checks the
directory to read is indeed a tdt block, to make this IO more robust to
unexpected files in the tank folder, like temporary files such as
“.DS_Store” on the Mac Platform


samuelgarcia:
Thank you for this.
OK for me.
@apdavison : ok for you ?

PR #247, opened 2016-06-24
Feature 'duplicate_with_new_data' for spiketrains, events and epochs

JuliaSprenger:
Currently only analogsignals offer the option to be duplicated keeping the same metadata (annotations, t_start...) but replacing the main signal (times). I added this feature also for spiketrains, events and epochs in order to simplify the modification of these objects, eg when shuffeling spiketimes or shifting all signals in time eg. for alignment of different trials. The current implementation uses helper functions (_copy_data_complement) analogous to the implementation for analogsignals and also contains some basic tests.
Currently only specific attributes are copied (shallow) to the new objects, which could be extended or implemented by a deepcopy in the future.


apdavison:
Many thanks for this. Now that `apibreak` has been merged into `master`, please could you close this PR and open a new one against the master branch?

PR #248, opened 2016-07-12
Small modification in io.axonio

Silmathoron:
I set version to string to avoid floating point surprises.
I used int on division to avoid floating point division in py3 and get rid of deprecation warning with numpy reshape.


samuelgarcia:
Tank you.
OK for me.

PR #249, opened 2016-07-21
apibreak-compatible NeoHdf5IO

apdavison:
The current implementation of `NeoHdf5IO` (in the 0.4.x branch) is very dependent on the Neo object model.

This causes problems for reading files created with Neo 0.3/0.4 in Neo 0.5.

This pull request introduces a new, read-only implementation of `NeoHdf5IO`, which is capable of reading files created with Neo 0.3/0.4. (Because of the tight coupling between the current IO implementation and the object model, it seemed easier to reimplement from scratch rather than adapt the original code).

Write support could be added, but since other, more widely-supported HDF5-based formats (NIX, NWB) have IO classes under development for Neo 0.5, it would be better, in my opinion, to save data in one of these formats.

Note that the tests rely on a data file created with Neo 0.4.0, which needs to be uploaded to G-Node.


samuelgarcia:
This sounds good. 
OK for me.

PR #256, opened 2016-09-05
NIX IO and tests

achilleas-k:
This is a pull request to add an IO for [NIX](https://github.com/G-Node/nix). The IO uses the Python version of the library, [NIXPy](https://github.com/G-Node/nixpy) in pure Python mode (no compiled dependencies).

As of today, the version of NIXPy that is compatible with this IO should be on PyPI. The name of the package is `nixio` and the compatible version is `1.2.0`.

Note: NIX is a format based on HDF5 and uses the default file extension `.h5`. This may cause issues with file type detection if both the NIX IO and the HDF5 IO are used.

### DIFF
--- a/neo/io/__init__.py
+++ b/neo/io/__init__.py
@@ -66,6 +66,8 @@ Classes:
 
 .. autoclass:: neo.io.WinWcpIO
 
+.. autoclass:: neo.io.NixIO
+
 """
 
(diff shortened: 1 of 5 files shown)


achilleas-k:
Not sure how those Igor commits ended up here. I'll close and rebase.

PR #257, opened 2016-09-05
New IO for NIX

achilleas-k:
This is a pull request to add an IO for [NIX](https://github.com/G-Node/nix). The IO uses the Python version of the library, [NIXPy](https://github.com/G-Node/nixpy) in pure Python mode (no compiled dependencies).

As of today, the version of NIXPy that is compatible with this IO should be on PyPI. The name of the package is `nixio` and the compatible version is `1.2.x`.

Note: NIX is a format based on HDF5 and uses the default file extension `.h5`. This may cause issues with file type detection if both the NIX IO and the HDF5 IO are used.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -1,0 +1,1311 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016, German Neuroinformatics Node (G-Node)
+#                     Achilleas Koutsou <achilleas.k@gmail.com>
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted under the terms of the BSD License. See
+# LICENSE file in the root of the Project.
+"""
+Module for reading data from files in the NIX format.
+
+Author: Achilleas Koutsou
+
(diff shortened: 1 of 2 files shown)


apdavison:
The Travis failure is unlikely to have anything to do with this PR.

neo/io/nixio.py
```diff
@@ -0,0 +1,1311 @@
...
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted under the terms of the BSD License. See
+# LICENSE file in the root of the Project.
+"""
+Module for reading data from files in the NIX format.
+
+Author: Achilleas Koutsou
+
+This IO supports both writina and reading of NIX files. Reading is supported
```
apdavison: typo: writing

achilleas-k:
I thought so too. It was strange. The 2.6 build was the only one that was failing before (due to unittest import) and after I fixed it, the 3.5 build failed.

apdavison:
Many thanks for this @achilleas-k !

achilleas-k:
You're very welcome.

PR #267, opened 2016-12-15
Some fix for BlackrockIO v4

samuelgarcia:
* lazy very fast
 * fix lazy problem (because anasig.signal=[] do not work)
 * put float32 as default instead of float
 * fix int overflow

### DIFF
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -1824,7 +1824,7 @@ class BlackrockIO(BaseIO):
         mask = (data_times >= n_start) & (data_times < n_stop)
         
         if lazy:
-            lazy_shape = (np.sum(maks), )
+            lazy_shape = (np.sum(mask), )
             sig_ch =  np.array([], dtype='float32')
             t_start = n_start.rescale('s')
         else:
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -623,7 +623,7 @@ class BlackrockIO(BaseIO):
             ('nb_ext_headers', 'uint32')]
 
         nev_basic_header = np.fromfile(filename, count=1, dtype=dt0)[0]
-        #print(nev_basic_header) ###
+        
         # extended header
         # this consist in N block with code 8bytes + 24 data bytes
         # the data bytes depend on the code and need to be converted
@@ -970,7 +970,7 @@ class BlackrockIO(BaseIO):
                     ('node_id', 'uint16'),
                     ('node_count', 'uint16'),
                     ('point_count', 'uint16'),
-                    ('tracking_points', 'uint16', ((data_size - 14) / 2, ))]},
+                    ('tracking_points', 'uint16', ((data_size - 14) // 2, ))]},
             'ButtonTrigger': {
                 'a': [
                     ('timestamp', 'uint32'),
@@ -1817,19 +1817,17 @@ class BlackrockIO(BaseIO):
         description = \
             "AnalogSignal from channel: {0}, label: {1}, nsx: {2}".format(
                 channel_idx, labels[idx_ch], nsx_nb)
+
+        data_times = np.arange(
+            t_start.item(), t_stop.item(),
+            self.__nsx_basic_header[nsx_nb]['period']) * t_start.units
+        mask = (data_times >= n_start) & (data_times < n_stop)
         
         if lazy:
-            period = self.__nsx_basic_header[nsx_nb]['period']
-            length = int(((n_stop.rescale(nsx_time_unit)-n_start.rescale(nsx_time_unit))/\
-                            self.__nsx_basic_header[nsx_nb]['period']).magnitude)
-            lazy_shape = (length, )
-            sig_ch = []
+            lazy_shape = (np.sum(maks), )
+            sig_ch =  np.array([], dtype='float32')
             t_start = n_start.rescale('s')
         else:
-            data_times = np.arange(
-                t_start.item(), t_stop.item(),
-                self.__nsx_basic_header[nsx_nb]['period']) * t_start.units
-            mask = (data_times >= n_start) & (data_times < n_stop)
         
             data_times = data_times[mask].astype(float)
             sig_ch = signal[dbl_idx][:, idx_ch][mask].astype('float32')
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -623,7 +623,7 @@ class BlackrockIO(BaseIO):
             ('nb_ext_headers', 'uint32')]
 
         nev_basic_header = np.fromfile(filename, count=1, dtype=dt0)[0]
-        print(nev_basic_header) ###
+        #print(nev_basic_header) ###
         # extended header
         # this consist in N block with code 8bytes + 24 data bytes
         # the data bytes depend on the code and need to be converted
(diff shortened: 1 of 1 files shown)


samuelgarcia:
@apdavison @mdenker   : Is it OK for merging ?
I need this code for having tridesclous working for some collaborators.

apdavison:
Apart from the commented-out `print` statements, this looks fine to me.

neo/io/blackrockio.py
```diff
@@ -538,8 +538,9 @@ def __read_nsx_dataheader_variant_b(
                 'offset_to_data_block': offset + dh.dtype.itemsize}
 
             # data size = number of data points * (2bytes * number of channels)
-            data_size = dh['nb_data_points'] * \
-                self.__nsx_basic_header[nsx_nb]['channel_count'] * 2
+            #int for avoid overflow problem
+            data_size = int(dh['nb_data_points']) * \
+                int(self.__nsx_basic_header[nsx_nb]['channel_count']) * 2
             # define new offset (to possible next data block)
```
mdenker: Yes, we didn't realize this could be problematic with large file sizes, good point.

neo/io/blackrockio.py
```diff
@@ -622,7 +623,7 @@ def __read_nev_header(self, ext_header_variants):
             ('nb_ext_headers', 'uint32')]
 
         nev_basic_header = np.fromfile(filename, count=1, dtype=dt0)[0]
-        print(nev_basic_header) ###
+        #print(nev_basic_header) ###
```
mdenker: In our version of fixes, we removed this print statement entirely. I think both is fine, however, if logging, maybe one should identify first the best places to put logger.debug() statements.

neo/io/blackrockio.py
```diff
@@ -1816,43 +1817,54 @@ def __read_analogsignal(
...
-        sym_dig = (max_dig[idx_ch] == -min_dig[idx_ch])
-        if sym_ana and sym_dig:
-            sig_ch *= float(max_ana[idx_ch]) / float(max_dig[idx_ch])
+        
+        if lazy:
+            period = self.__nsx_basic_header[nsx_nb]['period']
+            length = int(((n_stop.rescale(nsx_time_unit)-n_start.rescale(nsx_time_unit))/\
+                            self.__nsx_basic_header[nsx_nb]['period']).magnitude)
+            lazy_shape = (length, )
+            sig_ch = []
```
mdenker: Not really important, but maybe make an array np.array([]) for better consistency/legibility with the else part.

neo/io/blackrockio.py
```diff
@@ -1816,43 +1817,54 @@ def __read_analogsignal(
...
-        # transform dig value to pysical value
-        sym_ana = (max_ana[idx_ch] == -min_ana[idx_ch])
-        sym_dig = (max_dig[idx_ch] == -min_dig[idx_ch])
-        if sym_ana and sym_dig:
-            sig_ch *= float(max_ana[idx_ch]) / float(max_dig[idx_ch])
+        
+        if lazy:
+            period = self.__nsx_basic_header[nsx_nb]['period']
+            length = int(((n_stop.rescale(nsx_time_unit)-n_start.rescale(nsx_time_unit))/\
+                            self.__nsx_basic_header[nsx_nb]['period']).magnitude)
```
mdenker: You should use the `period` variable here (or remove line 1822, alternatively).

neo/io/blackrockio.py
```diff
@@ -1816,43 +1817,54 @@ def __read_analogsignal(
...
-
-        # transform dig value to pysical value
-        sym_ana = (max_ana[idx_ch] == -min_ana[idx_ch])
-        sym_dig = (max_dig[idx_ch] == -min_dig[idx_ch])
-        if sym_ana and sym_dig:
-            sig_ch *= float(max_ana[idx_ch]) / float(max_dig[idx_ch])
+        
+        if lazy:
+            period = self.__nsx_basic_header[nsx_nb]['period']
+            length = int(((n_stop.rescale(nsx_time_unit)-n_start.rescale(nsx_time_unit))/\
```
mdenker: Possibly, one could shorten this line via
(n_stop-n_start).rescale(nsx_time_unit)

neo/io/blackrockio.py
mdenker: We believe that this expression does not necessarily reflect the shape of the resulting array.
Assume a period of 1 ms.

It works correctly for nstart=0.1 ms and nstop=5.0 ms. In this case the `length` computes above is 4. The statement in line 1834 covers samples at index 1, 2, 3, and 4. So the length is 4 (correct).

It works incorrectly for nstart=0.2 ms and nstop=5.1 ms. In this case the `length` computes above is still 4. The statement in line 1834 covers samples at index 1, 2, 3, 4, and 5. So the length is 5 (incorrect).

In our current version of the code, where we fixed the lazy loading in a somewhat similar fashion as you propose here, we used a very unelegant approach to circumvent this problem and get a consistency: We calculate `data_times` regardless of lazy or not, and use its length. This is not very efficient or nice, but maybe one can find a calculation that always works correctly.

neo/io/blackrockio.py
```diff
@@ -2330,6 +2342,7 @@ def __str__(self):
                     "====================================\n"
                 for i, el in enumerate(avail_el):
                     output += "Electrode ID %i: " % el
+                    output += "label %s: " % self.__nev_params('channel_labels')[el]
```
mdenker: Fine with us, more info is better info.

mdenker:
The only major issue to us seems the calculation of `lazy_size` / `length`, see above. Otherwise, only minor points.

apdavison:
@samuelgarcia can you reply to @mdenker's last comment?

samuelgarcia:
I will try to have a better lazy_size / length.
It is not so easy without computing data_times vector. I was trying to avoid this when lazy.

mdenker:
Hi Samuel, yes, we also had the problem of correctly calculating it, and don't have a solution yet. Our intermediate solution was just that to always calculate the data_times vector, which is not nice -- but maybe it can serve as first version of the IO until we get a piece of code and a test that performs the calculation correctly.

samuelgarcia:
Hi, yes maybe the best is to stick with time vertoc for the moment.

apdavison:
@samuelgarcia will you be able to make the changes to `lazy_size / length` today or tomorrow? My goal it to release Neo 0.5.0 this week, and it would be nice to have the BlackrockIO fixes included.

samuelgarcia:
This is a big challenge. But I will try!!!

apdavison:
If you are too busy, it's not a problem, we can make a 0.5.1 release for the BlackrockIO fixes.

samuelgarcia:
I have tried something.

neo/io/blackrockio.py
```diff
@@ -1817,19 +1817,17 @@ def __read_analogsignal(
...
+            self.__nsx_basic_header[nsx_nb]['period']) * t_start.units
+        mask = (data_times >= n_start) & (data_times < n_stop)
         
         if lazy:
-            period = self.__nsx_basic_header[nsx_nb]['period']
-            length = int(((n_stop.rescale(nsx_time_unit)-n_start.rescale(nsx_time_unit))/\
-                            self.__nsx_basic_header[nsx_nb]['period']).magnitude)
-            lazy_shape = (length, )
-            sig_ch = []
+            lazy_shape = (np.sum(maks), )
```
apdavison: typo - should be `mask`, not `maks`, no?

samuelgarcia:
Very strange that tests did not catch this ?!!?
Does tests still try to read all files in lazy mode ?

samuelgarcia:
Travis is unhappy it is unrelated to this PR, no ?

apdavison:
Yes, unrelated. It is this intermittent bug with the PyNNIO. I should probably just go ahead and remove PyNNIO, since it is anyway only applicable to old versions of PyNN.

apdavison:
@mdenker can I merge this?

apdavison:
Going to merge anyway, to get it into 0.5.0 release.

PR #268, opened 2016-12-16
py3 changes to plexonio

kylerbrown:


samuelgarcia:
Thank you for that clean.
Are you suing PlaexonIo for real ? isnt't it too slow ?

kylerbrown:
@samuelgarcia Sorry for the delayed response, our data loads in a few seconds with the reader in python3. There is no continuous data, only spike data. The major limitation I found was that plexon units are defined by a unit number and a channel, but this metadata didn't seem to be read into the NEO spiketrain object. Ultimately we switched to exporting our data from the Plexon Offline Sorter to CSV instead of reading an exported .plx with NEO.

PR #269, opened 2017-01-03
Kwikio

lepmik:
Hi, I have rewritten the kwikio to be dependent on phy. That is it is now only supposed to read kwik output from klusta and the phy project. It is not very well tested yet, but I thought it would be nice to discuss with you guys how it should look like in the end. In addition, I wonder if you would consider having a group "id" attribute to the ChannelIndex class, such that one can index e.g. shank id.

### DIFF
--- a/neo/io/kwikio.py
+++ b/neo/io/kwikio.py
@@ -92,7 +92,7 @@ class KwikIO(BaseIO):
     def read_block(self,
                    lazy=False,
                    cascade=True,
-                   get_waveforms=False,
+                   get_waveforms=True,
                    cluster_metadata='all',
                    raw_data_units='uV',
                    get_raw_data=False,
@@ -122,7 +122,7 @@ class KwikIO(BaseIO):
                 group_meta.update(model.metadata)
                 chx = ChannelIndex(name='channel_group #%i' % group_id,
                                    index=group_id,
-                                   channel_ids=model.channels
+                                   channel_ids=model.channels,
                                    **group_meta)
                 blk.channel_indexes.append(chx)
                 clusters = model.spike_clusters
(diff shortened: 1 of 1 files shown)


lepmik:
Update: I changed the ChannelIndex.index to represent the group id, and then channel_ids to be channel_indexes. If this is how it is designed to work, then the previous question can be ignored.

samuelgarcia:
Hi,
thank you for this.
I am also following the discussion on klusta list.

This PR looks find to me. OK for merging so.

I guess you already known but doing spike sorting for hardware threshold is a bad pratice...
I also guess that you don't have choice for the dataset you are playing with.


ChannelIndex.index are the index of the channel corrsesponding to an analogsignal

For what you want ChannelIndex.name would represent the group id (of a tetrode for instance).

apdavison:
> ChannelIndex.index are the index of the channel corrsesponding to an analogsignal
> 
> For what you want ChannelIndex.name would represent the group id (of a tetrode for instance).

As Samuel says, `ChannelIndex.index` should be an array of integers such that `signal[unit.channel_index.index]` gives the channels from which the spikes associated with `unit` were detected.

lepmik:
Ok, then I will change the index back to channel_indexes and name to group id. Would you consider requiring the ChannelIndex.name attribute to only take int? The reason I ask is that I'm making a NeoGUI to phy, and there one explicitly has to set the channel_group, if some IO has a weird name for their channel group this will produce an error in the GUI.

> I guess you already known but doing spike sorting for hardware threshold is a bad pratice...
I also guess that you don't have choice for the dataset you are playing with.

Yes, I know, but we have tons of data with threshold detected spikes where raw data is not saved, and the system is still in use until we get a new pipeline set up in the lab. But I mean it's not the worst practice, you loose allot of data on low amplitude spikes, but the waveforms that are clear are not a problem to sort out.

lepmik:
Update: I'm not agreeing with my self, `ChannelIndex.name` should of course not be `int`, but I would then vote for having a `ChannelIndex.name_id` or something similar which is `int` and representing the channel_group id.

samuelgarcia:
All neo object have a annotations dict for non standard use:
ci = ChannelIndex()
ci.annotate(unit_id=456)
print(ci.annotations)

Start with this, no ?

This id could discuss and added in next version of neo maybe, one day!! try a ticket for that, we are not very active at the moment.

lepmik:
ok, then I will annotate for now. Please don't merge before my next commit.

samuelgarcia:
OK for merging ?

PR #271, opened 2017-01-12
Fix nixio infinite reading.

matham:
Currently it transposes the list of elements, which leads to infinite processing by numpy. The first element of the list should be used instead, as is done in the surrounding code, since all the list elements are supposed to represent the same data.

The symptom is that that line causes cpu to get stuck at 100% and never stops.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -308,7 +308,7 @@ class NixIO(BaseIO):
             signaldata = pq.Quantity(np.empty(0), unit)
             lazy_shape = (len(nix_da_group[0]), len(nix_da_group))
         else:
-            signaldata = pq.Quantity(np.transpose(nix_da_group), unit)
+            signaldata = pq.Quantity(np.transpose(nix_da_group[0]), unit)
             lazy_shape = None
         timedim = self._get_time_dimension(nix_da_group[0])
         if (neo_type == "neo.analogsignal" or
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -308,7 +308,8 @@ class NixIO(BaseIO):
             signaldata = pq.Quantity(np.empty(0), unit)
             lazy_shape = (len(nix_da_group[0]), len(nix_da_group))
         else:
-            signaldata = pq.Quantity(np.transpose(nix_da_group[0]), unit)
+            signaldata = np.array([d[:] for d in nix_da_group]).transpose()
+            signaldata = pq.Quantity(signaldata, unit)
             lazy_shape = None
         timedim = self._get_time_dimension(nix_da_group[0])
         if (neo_type == "neo.analogsignal" or


apdavison:
@achilleas-k can you look at this?

achilleas-k:
In the surrounding code we use the first dimension because we're just extracting features from the data (e.g., length of the data). On the line in question, we're copying the actual signal data. Theoretically we would want the entire DataArray to be transposed and converted to a quantity. Using only dim0 is probably fine given that nix files created with the IO only create one-dimensional DataArrays. 

I'm more curious about why the processing takes so long. If it's because the DataArray is big? Is it two-dimensional? If it's not, would transposing the first element make a difference?

matham:
Hmm, when I looked at the contents `nix_da_group` was actually a list containing a analog signal. So when numpy transposes it it transposes the list of an array.

This should be easily replicable from: https://github.com/matham/MCDConv/blob/master/MCDConv/__init__.py#L117. Some similar code based on the one that broke for me:

```py
    ofile = NixIO(output, mode='ow')

    blk = Block()
    for group_name in map(str, range(3)):
        seg = Segment(name=group_name)
        for chan_name in map(str, range(20)):
            seg.analogsignals.append(AnalogSignal(
                np.zeros((512 * 1024,)), units=uV,
                sampling_rate=10000 * Hz, name=chan_name))
    blk.segments.append(seg)

    nix_file.close()
```

Now, just open this file in neo and do `blocks = f.read()`. You should see the CPU being fully utilized and memory increasing very slowly.

achilleas-k:
Thanks for the example. I'll have a look.

A correction on my previous comment (sorry, I misunderstood the position of the change):
You are correct that `nix_da_group` is a list. It should be a list of NIX DataArrays though, not AnalogSignals. When writing, NIX IO converts a single AnalogSignal to multiple DataArrays, one for each signal contained in the AS. On read, the DataArrays are collected into `nix_da_group`, the first DataArray is used to determine things like units, type, and metadata, since all DataArrays share this information.
On the line in question, `signaldata = pq.Quantity(np.transpose(nix_da_group), unit)`, the DataArrays are eventually converted to a 2D Quantity, which will be used to reconstruct the original AnalogSignal. DataArrays behave like numpy arrays when you index or slice them, so the transpose call will result in a numpy array. The transposition in necessary because the dimensionality of a list of DataArrays is inverse from how neo uses AnalogSignals.

Now for the real bad news: The conversion isn't broken. It's working as intended. The issue with the infinite processing is simply an issue with the size. I'm open to suggestions on how to make it faster. It's possible that there's some inefficiency in how nixpy accesses data and so the reading takes too long during the call to transpose.

achilleas-k:
I wrote a sample script based on your code:
```python
import sys
from neo import Block, Segment, AnalogSignal
from neo.io.nixio import NixIO
import numpy as np
from quantities import Hz, uV


def create():
    ofile = NixIO("somefile.h5", mode="ow")

    blk = Block("signal block")
    seg = Segment("signal group")
    blk.segments.append(seg)

    sig = AnalogSignal(np.random.random((10000,)), units=uV,
                       sampling_rate=10000 * Hz, name="signal data")
    seg.analogsignals.append(sig)

    print(np.shape(sig))
    print(sig[30])  # arbitrary value in the signal
    print(sig[90])
    ofile.write_block(blk)
    del ofile


def read():
    ifile = NixIO("somefile.h5", mode="ro")
    print("Reading file")
    blocks = ifile.read_all_blocks()
    print("Done")

    sig = blocks[0].segments[0].analogsignals[0]
    print(np.shape(sig))
    print(sig[30])
    print(sig[90])


if __name__ == "__main__":
    if len(sys.argv) < 2:
        print("Specify either 'read' or 'create'")
    elif sys.argv[1] == "read":
        read()
    elif sys.argv[1] == "create":
        create()
    else:
        print("Specify either 'read' or 'create'")
```

Calls and output:
```
▶ python asignal.py create
(10000, 1)
[ 0.73751036] uV
[ 0.30869742] uV

▶ python asignal.py read  
Reading file
Done
(10000, 1)
[ 0.73751036] uV
[ 0.30869742] uV
```

If you change the size of the data to the value you used in your example `np.random.random((512 * 1024,))`, it does indeed take too long to process.

I'll need to investigate whether the bottleneck is the reading of the data from nix or the transpose function. Perhaps we can skip the transpose when the data is one-dimensional and construct the AnalogSignal another way when necessary.
As I mentioned earlier, there is the possibility that the reading of the data from the NIX objects is the issue.

matham:
One part that isn't clear to me. Are you intending to call transpose on the list or on each item of the list? Because the problem seems to be the calling of transpose on a list containing the array which numpy probably interprets in a funny way leading to the issues.

In fact, in the debugger I did `a = nix_da_group[0] + 1` and it took under a second, meaning it reads the data quickly, but is slow because numpy has issues transposing a list object which contains the data. It's probably because it's trying to slice the list and then the underlying data array, which is slow.

Shouldn't the fix then be `signaldata = [pq.Quantity(np.transpose(d), unit) for d in nix_da_group]`?

achilleas-k:
Thanks @matham, you're absolutely right that the issue is with transposing the items. Your proposed solution wont work since we want to transpose the entire set of data.
This works and is fast:
```python
signaldata = np.array([d[:] for d in nix_da_group]).transpose()
signaldata = pq.Quantity(signaldata, unit)
```

I just tested the script I posted earlier and it's fast even with data with shape `(512 * 1024, 200)` (not instantaneous, but it doesn't take forever).

matham:
Thanks! I updated the code with your suggested change.

I also looked at `np.concatenate` but it doesn't seem like it would be faster as it also does just a copy of the underlying arrays before building the resulting array.

apdavison:
Is this ready to merge now?

achilleas-k:
Yes. The change is correct and conversion should be much faster now.

PR #276, opened 2017-02-01
New feature for tdt io: using sort code generated by offline sorting

SummitKwan:
The TDT format allows offline sorting and the sort code generated by offline sorting is stored as a separate file in their data folder (every data folder corresponding to a neo segment).

I the pull request, I implemented a feature to the tdtio.py: an option 'sortname' is offered so that the user could either use the original online sort code (by default) or the offline sort code if it matches the name in the folder.  This does not affect the behavior of this io of no sortname is given by the user

Best,
Shaobo


apdavison:
The 'sortname' feature looks fine.

There is a problem with the pull request, however, as it contains more than 50 or so files with no changes to their content, only a change in permissions. Please can you remove them from the PR, or create a new PR with only the relevant changes?

samuelgarcia:
This look good. Thank you.

To avoid changes to all files I do ```git config core.fileMode false```, so that chmod are not taken in account for git.

apdavison:
Manually merged in 376c97cc4231ca1a3b064951df2953568ea604ff to avoid the problems with the changes to file permissions.

Issue #270, opened 2017-01-05
Making spiketrains with more times than waveforms does not produce error


lepmik:
```python
neo.SpikeTrain(times=np.arange(4), waveforms=np.ones((10,6,50)), units='s', t_stop=4)
```
Shouldn't this produce an error since `len(times) != waveforms.shape[0]`

apdavison:
Yes, this should produce an error.

(Note that it is planned in future to have `waveforms` be a 3D `AnalogSignal`)

apdavison:
Fixed in #279

PR #279, opened 2017-02-07
Issue270

jgonin:
#270 has been fixed and a corresponding test has also been added. Note that the import bug we had is not fixed but it does not block us and we can continue to move forward. 

Small midifications has also been done in the examples in order to make it run and learn from it.

### DIFF
--- a/examples/generated_data.py
+++ b/examples/generated_data.py
@@ -24,7 +24,8 @@ def generate_block(n_segments=3, n_channels=8, n_units=3,
     # Create container and grouping objects
     segments = [neo.Segment(index=i) for i in range(n_segments)]
 
-    chx = neo.ChannelIndex(name='T0')
+    chx = neo.ChannelIndex(index =1, name='T0')
+
     for i in range(n_channels):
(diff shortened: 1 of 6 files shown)


jgonin:
(I know I should have done the changes of the examples in an other branch but it was already done and I thought it was not critical. Next time I will stay specifically on the considered issue)

apdavison:
Unfortunately, as you can see from the [Travis CI build](https://travis-ci.org/NeuralEnsemble/python-neo/builds/199206587), the `test/generate_datasets.py` script does not respect the constraint that the number of waveforms be equal to the number of spikes, so there are lots of test failures. Please can you amend `generate_datasets.py` to avoid this?

jgonin:
Ok, thank you for making me notice that. Before the pull request, I ran the
nosetest and got 0 error but 150 skips, I am trying to see why in the same
time.

2017-02-07 14:59 GMT+01:00 Andrew Davison <notifications@github.com>:

apdavison:
Nice work!

Issue #166, opened 2014-06-25
Inconsistency in handling `self.annotations`


apdavison:
In the `__array_finalize__` method of `AnalogSignal` and `SpikeTrain`, a default value of `None` is used for the `annotations` attribute. This is inconsistent with the expectation in many other parts of the code that `annotations` be a dict.

I propose to give `annotations` a default value of `{}`.

grg2rsr:
I think this is relevant to this, but I might be mistaken. I just found that performing in place operations on a SpikeTrain removes it's annotations, I assume because the `__array_finalize__` set's it to `None`
e.g.
```python
import neo
import quantities as pq
st = neo.core.SpikeTrain([1,2,3]*pq.s,t_stop=4*pq.s,label='foo')
print(st.annotations) // prints: {'label': 'foo'}
st2 = st - 2 * pq.s
print(st2.annotations) // prints: None
```
for `neo = '0.4.1.1'` and both python2.7 and 3.5

apdavison:
Fixed in #280

PR #280, opened 2017-02-07
Issue166 In the __array_finalize__ method of AnalogSignal and SpikeTrain,

jgonin:
the default value "None" is now a dict "{}"


jgonin:
Is there a reason why the issue #166 has not been closed ?

apdavison:
Now closed

Issue #275, opened 2017-02-01
nixio file not closed


JuliaSprenger:
When using the nixio in an ipython notebook, written files are only valid after the kernel is restarted. If one tries to load a nix file created in a separate notebook without explicitly stopping the kernel of that notebook, the file turns out to be corrupted and can not be read.

Suggestions
- A quick fix for this is to manually close the file after writing to file by calling nixio.nix_file.close(). However, I think a nixio.close() function, which closes all files would be more intuitive.
- In the long run it would be nice to use the nixio as a context manager, so one does not need to close files explicitly. Currently python returns an AttributeError: __exit__ when calling
```
with neo.NixIO('data.nix','ro') as nix:
    pass
```

achilleas-k:
Hello @JuliaSprenger.

The lack of an explicit `close()` method is indeed an issue. I'm not sure how I feel about a `nixio.close()` function on the entire namespace. Given that the general convention with importing IOs is `from neo.io.Xio import X` or as you mentioned in the last example, simply importing neo and using `neo.NixIO(...)`, a method that closes all open NixIO files might not be directly accessible. 

I like the context manager idea, though. This would be especially nice given how there's no real reason to keep the file open after reading or writing. So a common use case would be:
```python
with neo.NixIO("data.nix", "ro") as nix:
    blocks = nix.read_all_blocks()

...
<work with blocks>
...

with neo.NixIO("newdata.nix", "ow") as nix:
    nix.write_all_blocks(blocks)
```

and there would be no issue with open files.

PR #283, opened 2017-02-08
NIX IO small fixes

achilleas-k:
Some fixes and maintenance for the NIX IO:
- Remove support for C++ backend
The NIX IO was meant to rely on the pure Python backend of NIX. There was some code in the IO which made it compatible with the C++ backend before the Python backend was fully complete and polished, but now it's unnecessary and has been removed.
- Alias `nixio` import to `nix`
By convention, we `import nixio as nix`. This has the added benefit that it avoids any confusion that might arise from using the `nixio` package in the Neo-NIXIO.
- Remove dependency on `six`
It was used just for checking `string_types` so now the variable is set in the same way `six` sets it and the dependency is removed.
- Default file mode set to `rw`
Was read-only. Change is compatible with NIX and Neo tests.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -117,6 +117,12 @@ class NixIO(BaseIO):
         self._object_hashes = dict()
         self._block_read_counter = 0
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
     def read_all_blocks(self, cascade=True, lazy=False):
         blocks = list()
         for blk in self.nix_file.blocks:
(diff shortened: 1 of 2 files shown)


achilleas-k:
Not sure why the tests failed here. It failed for py2.6 on the PyNNIO.
Is there any chance it's related to my previous PR or something I changed in this one?

achilleas-k:
Pushed another small fix and rebased off current master.

apdavison:
The PyNNIO module has an intermittent test failure, which you can ignore.

achilleas-k:
Good to know, thanks.

I have a couple of more small changes/fixes coming, one of them to address #275. I'll push them to this PR if you don't mind keeping it open.

achilleas-k:
Close method and context management added to PR.

apdavison:
Is this ready to merge?

achilleas-k:
Yes, it's good to go.
Thanks!

Issue #60, opened 2013-12-16
Slice `AnalogSignal`, `AnalogSignalArray` and `SpikeTrain` by time


apdavison:
In addition to slicing data by index as at present, we should support slicing by time, e.g. if `s` is an `AnalogSignal` and the sampling period is 10 ms, then

```
s[4:10]
```

would be equivalent to

```
s[40*ms:100*ms]
```

This would be a particularly nice feature for `SpikeTrain`. I'm not sure this can be implemented in Python 2, since numpy array has a `__getslice__` method, so that `s[i:j]` requires `i` and `j` to be integers, but it should work in Python 3 (where `__getslice__` has been removed).

---
### Imported from Trac ticket:27

**Opened**: 2011-09-23 13:58:20
**Last modified**: 2011-09-24 00:07:41
**Component**: core
**Priority**: major
**Owner**: somebody
**Reporter**: apdavison

apdavison:
Current situation: `SpikeTrain` and `AnalogSignal` have `time_slice()` methods. `Event`, `Epoch` and `IrregularlySampledSignal` do not.

grg2rsr:
I started with this. Concerning epochs, what do you think the expected behavior should be?
Say we have 3 epochs
```
epc = Epoch(times=(0,1,2) * pq.s, durations=(0.5,0.5,0.5) * pq.s)
```
I can think of two ways how time slice could be applied. A call to `epc.time_slice(0.25*pq.s, 2.25*pq.s)` should return an `Epoch` with 
```
epc.times = (0.25,1,2) * pq.s
epc.durations =  (0.25,0.5,0.25) * pq.s
```
or alternatively, `time_slice` could act on the individual epochs, re-slicing each epoch in the same way, for example `epc.time_slice(0.25*pq.s, 2.25*pq.s)` would then return an `Epoch` with 
```
epc.times = (0.25,1.25,2.25) * pq.s
epc.durations =  (2,2,2) * pq.s
```
Which one do you think makes more sense? The first one would be more intuitive as all other `time_slice` act on the absolute time axis, but I could imagine the second being more useful.

apdavison:
@grg2rsr I'm sorry for not replying earlier. I think the second would be useful, but should probably have a different name.

apdavison:
Implemented in #284 and #290

PR #290, opened 2017-03-08
#60 Time_slice() added in epoch, event and irregularsampledsignal

jgonin:
With tests for each.


apdavison:
Merged by hand, due to conflicts with earlier PR (#284) addressing part of the same issue

Issue #287, opened 2017-03-03
Spike2IO and numpy v1.12


dEvasEnApati:
Hi all

Using Spike2IO of neo v0.4.1with numpy v1.12 raises a type error

> Traceback (most recent call last):
  File "tmp/1.py", line 4, in <module>
    blk = spike2Reader.read()
  File "/home/aj/intel/intelpython27/envs/GJEMS_pip/lib/python2.7/site-packages/neo/io/baseio.py", line 121, in read
    seg = self.read_segment(lazy=lazy, cascade=cascade,  **kargs)
  File "/home/aj/intel/intelpython27/envs/GJEMS_pip/lib/python2.7/site-packages/neo/io/spike2io.py", line 128, in read_segment
    anaSigs = self.readOneChannelContinuous( fid, i, header, take_ideal_sampling_rate, lazy = lazy)
  File "/home/aj/intel/intelpython27/envs/GJEMS_pip/lib/python2.7/site-packages/neo/io/spike2io.py", line 255, in readOneChannelContinuous
    if channelHeader.unit in unit_convert:
  File "/home/aj/intel/intelpython27/envs/GJEMS_pip/lib/python2.7/site-packages/neo/io/spike2io.py", line 472, in __getattr__
    return self.array[name][1:l+1]
TypeError: only integer scalar arrays can be converted to a scalar index

With numpy 1.11, we get a Deprecation Warning:
>/home/aj/intel/intelpython27/envs/GJEMS/lib/python2.7/site-packages/neo/io/spike2io.py:472: VisibleDeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future
  return self.array[name][1:l+1]

From what I could figure out, the value of 'l' at the line mentioned above with numpy 1.11 is a one member numpy.ndarray object, while with numpy 1.12 is a one member list. Hence the problem with indexing.

Could you please take a look at this?

gulpgiulia:
Hi,
I got exactly the same. I was here to put a note for neo developers, and I found the issue already raised.

samuelgarcia:
I have push 2 PR for  0.4.x and master that should fix this.
PR #294 and PR #293 

Could you test ?

gulpgiulia:
Yes, thank you samuelgarcia. I have just tried, and now I get exactly the same I had with numpy 1.11, i.e. the Deprecation Warning, but who cares because it that does not block the processing :)

```
[...]

PR #293, opened 2017-03-10
Should fix issue #287

samuelgarcia:


samuelgarcia:
Travis is broken for 0.4.x because of pip
```
no such option: --use-mirrors
The command "pip install -r requirements.txt --use-mirrors" failed and exited with 2 during .
Your build has been stopped.
```

Issue #289, opened 2017-03-07
neo, blackrockio, for version 0.41, comments mask is 0xFFFF, not 0xFFF


neoana:
neo, blackrockio, for version 0.41, comments mask is 0xFFFF, not 0xFFF.
and for version Neo 0.5.0-alpha 1, blackrockio, comments data is not read

samuelgarcia:
I can push 0xFFFF in 0.4.x

In neo-0.5.x (master) the blackrockIO is a different beast completly rewritten to deal with many version of the file format.

We would be happy to get a contribution for including again event comments .

I ask authors.

@JuliaSprenger :
@mdenker :
In old BlackrockIO event generated by comments were catch by this (line 267 in neo0.4.x):
```
        mask = (data['id']==0xFFF) 
        dt_comments = [('samplepos', 'uint32'),
                    ('id', 'uint16'), 
                    ('charset', 'uint8'), 
                    ('reserved0', 'uint8'), 
                    ('color', 'uint32'), 
                    ('comment', 'S{0}'.format(h['packet_size']-12)),
                ]
        data_comments = data.view(dt_comments)[mask]
```
I don't see equivalent things in the new BlackrockIO.
Maybe I am missing it.
Do you have this in your dev version of this IO ?

mdenker:
Hi, I'll have to look into it, but on line 947, there is at least corresponding structure defined:

            'Comments': {
                'a': [
                    ('timestamp', 'uint32'),
                    ('packet_id', 'uint16'),
                    ('char_set', 'uint8'),
                    ('flag', 'uint8'),
                    ('data', 'uint32'),
                    ('comment', 'S{0}'.format(data_size - 12))]},

and on line 899, there is a similar bug with the 0xFFF

            'Comments': {
                'a': (packet_ids == 0xFFFF)},

However, I am not yet sure, whether these fields are acutally used -- I'll dig deeper in the code later.

samuelgarcia:
Thank you. I was not able to make the link.

@Deckard123321: could you provide a file were comments is read in 0.4.X and not read in 0.5.x ?

mdenker:
I think that one would have to enter a bit of new code around line 2059 following in function `read_segment()`, which only reads `nev_data["NonNeural"]` Events so far...

Deckard123321:
1. For version 0.4.x, first ,you should change this "mask = (data['id']==0xFFFF) "
    you can read "comments" data by this:
                from neo import io
[...]

PR #295, opened 2017-03-10
This should fix issue #289

neoana:
[blackrockio.py.zip](https://github.com/NeuralEnsemble/python-neo/files/834155/blackrockio.py.zip)


apdavison:
Are these changes for the 0.4.x branch? In that case, please could you make a new PR targeting that branch rather than master?

apdavison:
Closing, as it seems this should be targeting the 0.4.x branch

PR #298, opened 2017-03-17
fix some python 3 incompatibilities in axonio

dmeliza:
Some values need to be converted to ints in order to work as indices

Some strings need to have .decode() called, not str(), or they have a spurious "b" at the start.


apdavison:
@dmezila Please could you add your name and affiliation to `doc/source/authors.rst`

dmeliza:
Sure, will do.

PR #299, opened 2017-03-20
Add NESTSpikeIO and NESTMultimeterIO

apdavison:
building on work by @JuliaSprenger, Maximilian Schmidt, Johanna Senk, Jakob Jordan

PR #301, opened 2017-03-21
Add nestio

JuliaSprenger:
Adding an IO, which reads gdf and dat files as they are generated by NEST. This is an alternative implementation to the IOs suggested by @apdavison in PR #299.


apdavison:
This is great, much more complete than my version in #299.

The only thing missing is the `read_block()` method; to save time I will merge this now and then manually copy over the `read_block()` implementation from #299.

PR #315, opened 2017-04-20
Add NeuralynxIO

JuliaSprenger:
Adding an IO which reads standard Neuralynx output files in ncs, nev, nse and ntt format. First ones are  tested, for later ones there are currently no test files available.


samuelgarcia:
Waooo.
This is impresive!!
Congratulation.
I try to have a look soon.

PR #316, opened 2017-04-26
NeuroExplorerIO: correct offset for continuous signal

merky:
When parsing continuous signal from a NeuroExplorer file, the same byte offset was being used for indices, fragment starts, and data signal (i.e. `entity_header['offset']`). However, these are positioned sequentially in the file. This corrects the offsets such that the correct fragment starts, and signal are returned with the IO object. 

Tested against a local continuous signal file of `.nex` format.

### DIFF
--- a/neo/io/neuroexplorerio.py
+++ b/neo/io/neuroexplorerio.py
@@ -3,13 +3,13 @@
 Class for reading data from NeuroExplorer (.nex)
 
 Documentation for dev :
-http://www.neuroexplorer.com/code.html
+http://www.neuroexplorer.com/downloads/HowToReadAndWriteNexAndNex5FilesInMatlab.zip
 
 Depend on:
 
 Supported : Read
 
-Author: sgarcia,luc estebanez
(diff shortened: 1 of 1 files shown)


apdavison:
@samuelgarcia you implemented NeuroExplorerIO originally. Can you check this PR?

samuelgarcia:
I must confess that I really don't rember how I get the spec of this file format.
The link in the code is not working anymore.

I must also confess that I never use this code myself.

From what I see in the patch, we have made a quite big mistake for a while... This is sad.
This is suprising that nobody complain before!!

I guess that Mark is doing this patch with an official document from NeuroExplorer.

@merky: could also add your name in the neuroexplorerio.py and cite the new source of documentation ?

merky:
Yeah, I'm hoping too many people didn't rely on the continuous signal code. Found this mistake from looking at the actual waveforms. 

@samuelgarcia I couldn't find an online version of the format specification, but pointed instead to instructions for reading/writing .nex format via Matlab. Let me know if that works.

PR #319, opened 2017-05-02
Added `extras_require` field to setup.py

apdavison:
to clearly document the requirements for different io modules.

For example, this allows you to run:
```pip install neo[neomatlabio]```
and have the extra dependency needed for the `neomatlabio` module (scipy in this case) be automatically installed.


wvangeit:
Great, 
eFEL now use neo[neomatlabio]

PR #325, opened 2017-05-08
Minor fixes to the Blackrock reader

rohanshah:
The Blackrock reader errors when some of the (optional) extended headers are missing (like `NEUEVLBL`). It should instead handle these cases gracefully and still be able to read the rest of the file information that is available to it.

These are some minor fixes I made while a parsing a Blackrock file with missing extended headers but full data. With these changes I was able to read all the spike and waveform data within the Blackrock files without issue.

Note: I've also included a few spacing fixes, removed extraneous `print` lines, etc.


neo/io/blackrockio.py
```diff
@@ -1765,7 +1766,7 @@ def __read_spiketrain(
             wf_size = self.__nev_params('waveform_size')[channel_idx]
 
             waveforms = spikes['waveform'].flatten().view(wf_dtype)
-            waveforms = waveforms.reshape(spikes.size, 1, wf_size)
```
rohanshah: depending on your exact version of python2.7 this will either give the following warning or error out completely:
```
neo/io/blackrockio.py:1769: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  waveforms = waveforms.reshape(spikes.size, 1, wf_size)
```

samuelgarcia:
Hi.
Thank you.
This is OK for me.

@mdenekr : is OK for you ?

rohanshah:
@samuelgarcia @mdenker I've moved these changes to this pull request: https://github.com/NeuralEnsemble/python-neo/pull/330.

It has a cleaner commit history but is otherwise equivalent.

Thank you for looking at this!

Issue #333, opened 2017-05-22
Confusing logic in BlackrockIO


rohanshah:
Hi I have a question about some logic in the BlackrockIO reader that I'm a little confused by.

The piece of code in question is here: https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/blackrockio.py#L1466-L1480

reproduced here is:
```
def __transform_channels(self, channels, nsx_to_load):
        """
        Transforms the input argument channels to a list of integers.
        """
        all_channels = []
        nsx_to_load = self.__transform_nsx_to_load(nsx_to_load)
        if nsx_to_load is not None:
            for nsx_nb in nsx_to_load:
                all_channels.extend(
                    self.__nsx_ext_header[nsx_nb]['electrode_id'].astype(int))
        else:
            hdr = self.__nev_ext_header
            elec_id = self.__nev_ext_header[b'NEUEVWAV']['electrode_id']
            all_channels.extend(elec_id.astype(int))
        all_channels = np.unique(all_channels).tolist()
```

It seems that when `nsx_to_load` is specified such as with `"all"`, then only the analog signal channels are loaded, even if `channel` is set to `"all`".

It currently isn't possible to load both the spike and analog channels in this function.

So when I do `read_block(..., channels='all', nsx_to_load='all')` I can only read the analog signals and only when I do `read_block(..., channels='all', nsx_to_load='none')` can I get the spike channels.

It seems like the channels returned by `_transform_channels` should contain both the analog and spike channels when `channel='all'` and `nsx_to_load='all'`.

Is there a reason this isn't the case? If not, I can fix it. But it seems like the logic was specifically written that way so I wanted to double check that I wasn't missing something.

CC: @apdavison

rohanshah:
PR to address this: https://github.com/NeuralEnsemble/python-neo/pull/334

JuliaSprenger:
Hi rohanshah, I think the BlackrockIO was implemented this way since in the testing data the spiking channels are a subset of the the analogsignal channels. I guess this is not the case for your data. I will have a look at your PR.

PR #334, opened 2017-05-24
[BlackrockIO] Refactor __transform_channels

rohanshah:
### Motivation
The `__transform_channels` function only returns continuous data channels or spike data channels but not both, even when both `nsx_to_load='all'` and `channels='all'`. This shouldn't be the case.

This change also cleans up the logic in the function by grouping conditions based on what they'll return and does some minor edge case handling.

Closes https://github.com/NeuralEnsemble/python-neo/issues/333

### Notes
It might be easier to just compare the entire new function with the old one than look at the diff since git has it split it up weirdly.

### Testing
- Imported and validated data from multiple Blackrock data sets.

### DIFF
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -1923,7 +1923,7 @@ class BlackrockIO(BaseIO):
 
         if self._avail_files['nev']:
             channel_labels = self.__nev_params('channel_labels')
-            if channel_labels is not None:
+            if channel_labels is not None and channel_idx in channel_labels:
                 rcg.channel_names = channel_labels[channel_idx]
 
             # additional annotations from nev
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -1464,42 +1464,55 @@ class BlackrockIO(BaseIO):
 
         return nsx_to_load
 
+    def __channels_from_nsx(self, nsx_to_load):
+        channels = []
+
+        nsx_to_load = self.__transform_nsx_to_load(nsx_to_load)
+
+        if nsx_to_load is not None:
+            for nsx_nb in nsx_to_load:
+                channels.extend(
+                    self.__nsx_ext_header[nsx_nb]['electrode_id'].astype(int))
+
+        return channels
+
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
I opened a PR to fix the merge issues in this PR (https://github.com/Blackfynn/python-neo/pull/7). Can you have a look at it? If you have a minimal dataset where the spiketrain channels are not a subset of the analogsignal channels, could you provide this to / and implement a unittest covering this?

rohanshah:
@JuliaSprenger thanks! I've been trying to fix the merge conflict myself today but I've been running into a lot of errors. I'll check out your branch and see if it works with my datasets, and let you know the result!

rohanshah:
@JuliaSprenger I looked through your merge and I'm a little hesitant to approve it as there are some logical inconsistencies between the new changes since my PR and my my changes. For instance, `master` branch now returns a subset of `channel_ids` if `all_channels` is a superset of `channels` whereas my changes did not have that at the time, and your merge removes this update.

I have committed my own merge of the master branch here with the logic as it was at the time of my pull request.

However, I am getting errors trying to validate my dataset with my updated PR. I think there have been too many changes at this point to safely recover the original PR I made. I'll play around the `master` branch and see if I can re-write my PR in a way that is consistent with all the updates.

Thank you for looking at this though! I hope to have a new PR out soon. Would I be able to tag you for a review at that time?

JuliaSprenger:
@rohanshah Sounds like a good plan. Thanks for putting your effort into the BlackrockIO! I would be happy to review the next PR.

samuelgarcia:
@rohanshah Do you means that we can safetly close this PR without merging ?

@rohanshah I hope to merge more or less soon this PR #355. This is complete rewrite of many IOs included BlackrockIO. Before doing any effort in the actual BlackrockIO could you have a look in the next neo.rawio.BlackrockRawIO to avoid duplicated work.

Best

Samuel

rohanshah:
@samuelgarcia yes, I will close this PR now. I will also take a look at the new reader and make sure that I'm not conflicting with any of the changes you've made. Cheers!

PR #335, opened 2017-06-02
Cumulative set of fixes for blackrockio

mdenker:
This PR includes a package of several fixes to the revised blackrockio that is available since Neo 0.5, which were contributed by various people. In particular, this PR fixes:

* New parameter `scaling` to indicate whether analog signals are scaled to mV ('voltage'), or left as raw integer values ('raw'). Note that the default was set to "raw", since for file spec 2.1 and below, only in the presence of a `nev` file the conversion to voltage can be done. However, note that this may lead to different behavior since the initial inclusion of the new blackrockio. Changing the default to `voltage` is of course discussable.
* Fixed issue where one unit test was not actually testing correctly
* Fix issue where available channels were pooled only from nsX files, not the nev.
* Fixed issue where AnalogSignals were not connected to the ChannelIndex
* Updates to docstrings
* Corrected use of "index" in ChannelIndex object (now always set to 0, since the IO creates one ChannelIndex object for each recording channel)
* Additional annotations for ChannelIndex objects
* Removal of print statements
* PEP8 issies, code cleaning
* Minor fixes

### DIFF
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -131,9 +131,9 @@ class BlackrockIO(BaseIO):
     is_writable = False
 
     # This IO can only manipulate continuous data, spikes, and events
-    supported_objects  = [Block, Segment, Event, AnalogSignal, SpikeTrain,
-                          Unit, ChannelIndex]
-    readable_objects    = [Block, Segment]
+    supported_objects = [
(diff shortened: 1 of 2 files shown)


rohanshah:
Code looks good to me, tested it locally with some of my Blackrock files and it works fine.

@apdavison @mdenker any chance of this getting merged in the near future? I have a couple of changes as well I'd like to get reviewed and merged if possible.

PR #339, opened 2017-06-14
New IO for NSDF format

mgr0dzicki:
This is to add a new IO class for [NSDF](https://github.com/nsdf/nsdf). It supports reading and writing Blocks, Segments and AnalogSignals with all relationships and meta data. Support for another neo objects will be added in the near future.


mgr0dzicki:
I'm sorry, there were few things that I haven't notices before creating this PR, but now it seems that all is right, so can you have a look at it and merge if all is OK please?

apdavison:
Thank-you for contributing this! Please could you add your name and affiliation (if appropriate) to `doc/source/authors.rst` before I merge this?

mgr0dzicki:
Thank you for the reply. I updated authors.rst file as you said.

PR #345, opened 2017-06-21
Minor fixes for NeuralynxIO

JuliaSprenger:
This PR fixes

1. An incorrect link within the generated neo structure
2. A missing autoclass entry for the NeuralynxIO

### DIFF
--- a/neo/io/__init__.py
+++ b/neo/io/__init__.py
@@ -46,6 +46,8 @@ Classes:
 
 .. autoclass:: neo.io.NestIO
 
+.. autoclass:: neo.io.NeuralynxIO
+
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Great you are back to this!!

I very recently try to use this IO but get into problems with cachedir. I don't remember exaclty.

Since I need only to acces AnalogSignal with original dtype (int16) I rewrote this by myself.

My code is really fast because it use mmap and is able to read chunk very fastly.
The code is here at the end of the file:
https://github.com/samuelgarcia/tridesclous/blob/master/tridesclous/datasource.py

maybe the 2 codes could be merge one day.

Best.

Samuel

Issue #285, opened 2017-02-17
PickleIO: Unit of SpikeTrain not stored


schmitts:
I'm playing a bit with neo and spykeviewer. I've noticed that the Unit
of a Spiketrain is not loaded from a pickled neo block:

```python
#!/usr/bin/env python

import logging

logger = logging.getLogger('neo')
logger.setLevel(logging.DEBUG)

from neo import *
import pickle
from quantities import s

train = SpikeTrain([3, 4, 5]*s, t_stop=10.0)
unit = Unit()
train.unit = unit
unit.spiketrains.append(train)

blk = Block()
seg = Segment()
seg.spiketrains.append(train)
blk.segments.append(seg)

reader = PickleIO(filename="blk.pkl")
reader.write(blk)

reader = PickleIO(filename="blk.pkl")
r_blk = reader.read_block()
r_seg = r_blk.segments[0]
print r_seg.spiketrains[0].unit
```

results in None.

tristanstoeber:
similar problem.

mczerwinski:
Similar with Block, Segment, analogsignal

I try to save using this module, and read it afterwards.
Depending on the reading function it is displayed on the screen or not.
Depending on the function it is returning None or is not returning anything.

PR #349, opened 2017-06-22
Ensure that SpikeTrains do not lose their parent objects on pickle

apdavison:
fix for #285


HFragnaud:
i will make another pull request for Epoch, Events and IrregularlySampleEvent, same fixes.

Issue #346, opened 2017-06-21
Slicing a ChannelIndex should keep the links to data objects


JuliaSprenger:
This issue is related to changes introduced with issue #320. Currently the slicing of a ChannelIndex object is possible, but all references to AnalogSignals, Units and IrregularlySampledSignals are lost in the process. Only the reference to the block is kept.
I think it would make sense to keep at least the references to the Analog- and IrregularlySampledSignals since the information about the indexes and channel_ids and channel_names in the ChannelIndex object only makes sense if it is connected to these data objects.

PR #351, opened 2017-06-22
Slicing a ChannelIndex now preserves lists of child signals (cf #346)

apdavison:


JuliaSprenger:
What about Units? They should not be copied? How do they relate to the index and channel_id and channel_name attribute of the ChannelIndex?

apdavison:
I don't think units should be copied. The original channelindex indicates which channels contributed to the spikes detected in a given unit. If we take a subset of the channels by slicing the index, this subset no longer represents all the contributing channels.

Issue #354, opened 2017-06-23
Refactoring IO with 2 levels


samuelgarcia:
I plan a refactoring of some (many?) (all ?) IO classes in neo.

I have a strong need in [tridesclous](https://github.com/tridesclous/tridesclous) to acces data:
  1. In Raw format (very often int16)
  2. Fast acces
  3. chunk access

I also need this for writting fast viewer for dataset.

They actual neo.io API is far from that because for almost all IOs evrything is copied in memory (and sometimes twice) and converted to float32 or float64 without rules or control on that. Actual output of our IOs is a tree of neo object. We try in the past to make to lazyload API but this is difficult for users and developpers. I guess that very few of us are using it.

With the size of actual dataset in ephys expiremt we need a strong change.

Here 2 examples of 2 (anonymous) guys involved in neo team but re-coding there own readers only for themself (to bad!!):
  * https://github.com/spyking-circus/spyking-circus/tree/master/circus/files
  * https://github.com/samuelgarcia/tridesclous/blob/master/tridesclous/datasource.py

So, I would like to entirely refactor the io API to be able to stream and load chunk or selected channel easily and in an efficient way.
To not break everything and to have a slow transition I propose to have 2 levels for IOs:
  1. one, top level, that give neo objects as today.
  2. another, low level, that stream buffer in the original format (int16, int32, float32). 

The layer 2 would something that expose something like that:
  * read_header(...) and every stuff we need later one to have a fast acces.
  * get_nb_block()
  * get_nb_segment(block_num)
  * get_nb_analogsignals(block_num, seg_num)
  * get_nb_spiketrain(block_num, seg_num)
  * read_analogsignal_chunk(block_num, seg_num, channel_indexes, i_start, i_stop) this give raw data (often int16)
  * rescale_to_float(..)
  * rescale_to_quantities(..)
  * channel_name_to_index()
  * channel_index_to_name()
  * channel_id_to_index()
  * channel_index_to_id()
  * equivalent stuff for spiketrain

With this kind of level 2 API we could have one level 2 class for an IO and the level 1 would auto generated.

Example:
```python


# low level class that accces to chunks in raw format:
class BlackrockIORaw:
     pass

# Class to bridge new raw API to old neo obect API
class GenericNeoLegacy:
    pass

# Class the behave as before and read neo objects.
class BlackrockIO(GenericNeoLegacy, BlackrockIORaw):
    pass

```

[...]

PR #355, opened 2017-06-23
Proposal for new neo.rawio API.

samuelgarcia:
This PR is for discussion around a new neo.rawio API see issue #354.


samuelgarcia:
Note that all test are OK except for python2.6.

JuliaSprenger:
I compared the output of the NeuralynxIO with the new implementation and noticed a couple of differences when loading the Neuralynx test datasets used in the unittests. Some of the differences are improvements coming with the new implementation, but for some others the old implementation seems better. Here a summary of the latter ones:

- all times should be rescaled by the time at which the recording started
- the shapes of the extracted AnalogSignals are not the same
- AnalogSignal objects are missing annotations
- Event generation is different in terms of the number of event objects generated as well as their content and annotation
- ChannelIndex.name could be more telling than just 'Channel group 0'
- ChannelIndex.index can be negative (eg. [-2]) using the new implementation when no analogsignal is attached. Shouldn't it rather be []?
- Spiketrain objects do have less annotations with the new IO
- the number of waveforms read by the different versions can vary by 1.
- the new IO uses the unit of mV to the waveforms. Where does this unit come from?
- the left_sweep attribute should be None if it is not known and not set to 0.
- the unit of the left_sweep attribute should be time and not 's/Hz'
- the 'file_origin' attribute is not set for spiketrains, but should refer to the data file

The code used for comparison of the neo structures is available here: https://github.com/JuliaSprenger/python-neo/tree/fix/raw_io_api at  python-neo/neo/test/iotest/test_neuralynxio.py compare_neo_content(). I will extend the list above when starting to test the new IO also on different files.

samuelgarcia:
@JuliaSprenger:
For NeuralynxIO, I have fix some problems of the list:

> *  all times should be rescaled by the time at which the recording started

Done

> * the shapes of the extracted AnalogSignals are not the same

This need more investigation. My guess is that inthe very last data block the timestamp is the first sample of block and so this times is not the t_stop. The real t_start is last_time+nb_sample/sampling_rate. This could lead to t_stop difference between the two IOs. i need to check.

> * Event generation is different in terms of the number of event objects generated as well as their content and annotation

I have keep the same logic as before. One event channel=(event_id, ttl_input).
Except that a special channel was added for 'Starting Recording' and 'Stop Recording'.
This is not kept since they already have a specific channel.

> * the number of waveforms read by the different versions can vary by 1.

This must be also a limit problem link to t_stop see before.

> * the new IO uses the unit of mV to the waveforms. Where does this unit come from?

This wa s a mistake. It is in fact uV. For spiketrain.waveforms same logic than signals is applied with "bit_to_microVolt" from header.


> * the left_sweep attribute should be None if it is not known and not set to 0.

At RawIO level None is not possible because it is in a np.array. Maybe we can put -1 whch could be None at neo.io level.

> * the unit of the left_sweep attribute should be time and not 's/Hz'

Done.

samuelgarcia:
I have added a small cache system.
Only NeuralynxRawIO is using it at the moment.

@JuliaSprenger  : could have an eye on it ?

cboulay:
I'm playing with your raw_io_api branch now.
setup.py needs `, 'rawio'` added to the list of packages to install it properly.

I'll post more when I've had a chance to play with the other importers, but so far the BlackrockRawIO is much faster and requires less code on my application's end.

samuelgarcia:
Thank you for feedback.

Have you seen the small doc in doc/source/rawio.rst ?

Maybe some methods name will change before the merge in the master. Do not code too much on top of this before the merge in master.

samuelgarcia:
neo 0.5.2 have been release to gather many PR and bug fixes.
Since this present PR is big change.
I propose to merge this soon in master before arrival of other PR.

The place for improvement will be directly in the master so.

Any remark before this ?

bjoern1001001:
I tested the new BlackRockIOI and compared the output to the old implementation, based on how @JuliaSprenger tested the NeuraLynxIO. I used the test files that are used for the UnitTests. The following is a list of the differences I found, sorted by objects. The points marked by "!" are the ones I think are most important.

- Parameters for calling the IO
	- units ('units') and channels to load ('channels') cannot be specified any more (old parameters not accepted any more)
	- In old IO all nsX were loaded by default, now it's just the highest. This might be a problem, because sometimes not all values are in the highest nsX but rather in lower nsX, so they should be loaded by default as well
	- parameter 'scaling' is now 'voltage' by default, was 'raw' before
- General
	!- Unit and SpikeTrains are not linked, i.e. for every Unit object, unit.spiketrains is empty and for every SpikeTrain object, spiketrain.unit is empty as well
	!- Times are not saved in sampling resolution but in seconds (=> keeping the native sampling resolution prevents accumulating rounding errors early on)
	- file origin and description missing most of the time
	- inconsistent values of attribute name, in ChannelIndex name is 'Channel group X', in AnalogSignal name is 'chanX'

- AnalogSignals
	- more AnalogSignals/channels loaded by default, in old IO only channels to 128 were loaded by default, if found
	- no nsX value in annotations, so that it's difficult to understand from which nsX file the data was loaded
	- in anasig.description signals are now numbered, maybe start index at 1, like numbering for channel_id and label. It looks like this: 'AnalogSignal X from channel_id: Y label: chanY, nsx: 5', but X is always one less than Y: X=Y-1

- Block
	- missing annotations, earlier there was more meta-information

- ChannelIndexes
	- In old IO Units were linked to the ChannelIndex that the AnalogSignal with the same number was linked to. Why do you create extra ChannelIndexes for the Units?
	- missing annotations, earlier there was more meta-information
	- index becomes array with number instead of 0 => numbers too large, every AnalogSignal only has 1 channel


- Events
	- additional annotations (copy and id), what are they for?
	- one new Event that is not found by old IO, has empty times and labels (called 'serial_input_port' (0,))

- SpikeTrains
	- names look differently: 'ch1#0' was 'Unit 1000'
	- There is an additional annotation 'id', that contains the same content as SpikeTrain.name in old IO (e.g 'Unit 1000')
	- description contained 'unit: unclassified', is now 'unit_id: 0'

- Unit
	- Annotations Channel_ID and Unit_ID don't exist any more, instead the information is combined in annotation 'ID'. So there is no possibility to access them seperately any more

Some of those differences are quite similar to what @JuliaSprenger reported for the NeuraLynxIO.

samuelgarcia:
Thank you for feed back.

The short answer is that the behavior of the old BlackrockIO will not be exactly similar that the new one. So choices we made specifically to the IO. In the new branch the same behavior should work for all IOs.


>        units ('units') and channels to load ('channels') cannot be specified any more (old parameters not accepted any more)

This can be done at rawio API. For io API it is easy to add channel_indexes key. channel_ids is more ambiguous for some cases.

>        In old IO all nsX were loaded by default, now it's just the highest. This might be a problem, because sometimes not all values are in the highest nsX but rather in lower nsX, so they should be loaded by default as well

You can do ```reader = BlackrockIO(filename='...',nsx_to_load=3)``` but it is not at instanciation not in read_block. This was too specific to blackrock. Specific kargs are now in __init__. This is easier and more consistent.

>        parameter 'scaling' is now 'voltage' by default, was 'raw' before

At the moment reading raw data is only available at raw level but since the class is made by compisition. You can do  ```reader.get_signal_chunk(...)``` and you get directly numpyint16.



>         Unit and SpikeTrains are not linked, i.e. for every Unit object, unit.spiketrains is empty and for every SpikeTrain object, spiketrain.unit is empty as well

Oups, this is a bug in BaseFromRaw.

> Times are not saved in sampling resolution but in seconds (=> keeping the native sampling resolution prevents accumulating rounding errors early on)

At the io level, yes, but you can get timestamp (original format) at the rawio level with ```spike_timestamps(..)```

> file origin and description missing most of the time

Oups. I will fix.

> inconsistent values of attribute name, in ChannelIndex name is 'Channel group X', in AnalogSignal name is 'chanX'

This is due to ```signal_group_mode="group-by-same-units" or "split-all"```. I guess we need some discussion here.


>        more AnalogSignals/channels loaded by default, in old IO only channels to 128 were loaded by default, if found

OK. I was not aware of that. We can clip the channel exploration to 128. Is it the case for all blackrock system ?

>        no nsX value in annotations, so that it's difficult to understand from which nsX file the data was loaded

I will add this.

>        in anasig.description signals are now numbered, maybe start index at 1, like numbering for channel_id and label. It looks like this: 'AnalogSignal X from channel_id: Y label: chanY, nsx: 5', but X is always one less than Y: X=Y-1

I don't get this. X is the channel_index (0 based) can't be sparse. Y is the channel_id (1 base in the case of Blackrock) can be sparse. What is the expected descr ?

>        missing annotations, earlier there was more meta-information

I need to check.

>        In old IO Units were linked to the ChannelIndex that the AnalogSignal with the same number was linked to. Why do you create extra ChannelIndexes for the Units?

This cannot be done here at the moment because the BaseFromRaw is generic.
If someone have a solution for this. The IOs are able to do that. Some other not....

>        missing annotations, earlier there was more meta-information

I need to check.

>        index becomes array with number instead of 0 => numbers too large, every AnalogSignal only has 1 channel

This behavior can be fixed with ```signal_group_mode="group-by-same-units" or "split-all"```.

>        additional annotations (copy and id), what are they for?

This is something general at rawio level. All channel have an id.

>        one new Event that is not found by old IO, has empty times and labels (called 'serial_input_port' (0,))

I need to check. Is it a problem ?

>        names look differently: 'ch1#0' was 'Unit 1000'

This is my choice. I thin this more explicit to have both channel and unit id. No ?

>        There is an additional annotation 'id', that contains the same content as SpikeTrain.name in old IO (e.g 'Unit 1000')

Same answer.

>        description contained 'unit: unclassified', is now 'unit_id: 0'

I can add this but not in name or id. Maybe another key ?



>        Annotations Channel_ID and Unit_ID don't exist any more, instead the information is combined in annotation 'ID'. So there is no possibility to access them seperately any more

Yes but with other key.

> Some of those differences are quite similar to what @JuliaSprenger reported for the NeuraLynxIO.

Yes they are!


I will fix most of theses soon when it is easy. Then I propose to merge this branch in the master.
It will easier for other developers related to previous IO implementation to propose a PR that could enhance everybody needs.

It is now more clear were to put changes : in the RawIO directly when it is specific or in BaseFromRaw when we want a global behavior for all IO.

JuliaSprenger:
@samuelgarcia thanks for the quick relpy. We are currently looking into your comments and will come back to you soon.  However, before merging this branch, we would like to test the new BlackrockIO not only on a single test data set (as done above) but also confirm that the new IO is working for real datasets containing more data and also recordings using different versions of Blackrock files. This might take a couple of days, but we will keep you informed.

samuelgarcia:
@JuliaSprenger. Yes of course we can wait a before this merge. Merging soon is nevertheless important because any incoming PR could lead to conflict with rawio.

I propose:
  * keep explicitly OldBlackockIO and OldNeuralynxIO in the master for a long overlapping time to not break everyhting in existing pipeline you have.
  * add some commit in rawio branch this week.
  * merge rawio let's say next week. (but refusing any PR that concern theses IOs in old fashion)
  * discuss this new API and specific IO stuff directly in the master with PR and reviews.

In any case the release of 0.6.0 will be in a long time.

Is it OK for you ?

samuelgarcia:
> Unit and SpikeTrains are not linked, i.e. for every Unit object, unit.spiketrains is empty and for every  SpikeTrain object, spiketrain.unit is empty as well

This is strange, because the link is already done. Do you have my latest latest raw_io_api branch ?

>  missing annotations, earlier there was more meta-information

Could you be more specific ?

>description contained 'unit: unclassified', is now 'unit_id: 0'

unit_id still is unit_id but I have added unit_tag for this  = 'unclassified', 'noise' or str(unit_id)

> names look differently: 'ch1#0' was 'Unit 1000'
We now have at line 252
```python
                    name = "ch{}#{}".format(channel_id, unit_id)
                    _id = "Unit {}".format(1000 * channel_id + unit_id)
```
if you prefer we can swap name<-> _id.

JuliaSprenger:
@samuelgarcia: I agree to your proposed plan. Keeping both versions of the IO would help a lot. Maybe we can do the same for the NeuralynxIO for some time?

samuelgarcia:
OldNeuralynxIO is now in neo.io.neuralynxio_v1

The __init__ look like this
```
from neo.io.neuralynxio import NeuralynxIO
from neo.io.neuralynxio_v1 import NeuralynxIO as OldNeuralynxIO
...
from neo.io.blackrockio import BlackrockIO
from neo.io.blackrockio_v4 import BlackrockIO as OldBlackrockIO
```

If there is another IO you want to keep in double version teel me.

bjoern1001001:
I worked through your comments and will write a detailed answer to them soon. In addition, I just investigated whether it is possible to load different real Blackrock datasets. I noticed that it seems that some version checks are missing. blackrockrawio.py fails for file version 2.1 in _parse_header at line 244, because 'NEUEVFLT' cannot always be read as a key from the header (i.e., this Block is not existent in version 2.1). In the old IO 'NEUEVFLT' is only referenced in a function that would be run for file versions 2.2 and 2.3, whereas _parse_header in the new IO is called in any case.

As a somewhat related issue, could you please explain a bit more in detail how to load only specific channels in your new IO? I don't quite understand what you meant and don't want to crash everything by loading a complete ns5 dataset.

samuelgarcia:
Thank you Björn for this investigation across the code.

I have done a fast patch for NEUEVFLT changed NEUEVWAV. Could you test it ?

For channel you have several possibilities for selecting (only at rawio level for the moment):
```python
reader = BlackrockRawIO(filename='FileSpec2.3001')
reader.parse_hedaer()

#read all signals channels
raw_sigs = reader.get_analogsignal_chunk(channel_indexes=None)

#select by channel_indexes
raw_sigs = reader.get_analogsignal_chunk(channel_indexes=[0,  2, 4])

#select by channel_ids (note tha in Blackrock id is one based and this is specific to blackrock and some other
raw_sigs = reader.get_analogsignal_chunk(channel_ids=[1, 3, 5])

#select by channel_names
raw_sigs = reader.get_analogsignal_chunk(channel_names=['chan1', 'chan3', 'chan5'])
```

The file for testing is 2.3. If you have the small dataset <10Mo for 2.1 and 2.2, we could put them at gnode portal and use them in the test suite.

Since the neo.io Blackrock is done by inheritance with BlackrockRawIO ```get_analogsignal_chunk()``` is also available.

Have you seen the doc doc/source/rawio.rst ?

I will add in doc the channel selection stuff.

samuelgarcia:
I guess it is time to merge this rawio.
Lets improve rawio directly in master.

Issue #360, opened 2017-06-30
BUG? micromed io


dengemann:
Hi neighbors,

we're taking another look at reading micromed data into MNE

https://github.com/mne-tools/mne-python/issues/4359
https://github.com/mne-tools/mne-python/issues/1605

And I just realized that the micromed io seems to be broken in Python 3:
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/micromedio.py#L79

Here the file object is replaced by a BufferedReader which expects apparently an integer as input.

Trying to use example code:
```Python
filename =  'EEG_1.TRC'
seg = neo.io.MicromedIO(filename=filename)
seg.read_segment(lazy=False, cascade=True)
```

gives ugly errors:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-39-e6a3cafb8ed5> in <module>()
      5     seg = neo.io.MicromedIO(filename=filename)
      6     seg.filename
----> 7     seg.read_segment(lazy=False, cascade=True)

/Users/dengemann/anaconda/lib/python3.5/site-packages/neo/io/micromedio.py in read_segment(self, cascade, lazy)
     77         Arguments:
     78         """
---> 79         f = StructFile(self.filename, 'rb')
     80 
     81         # Name

TypeError: 'str' object cannot be interpreted as an integer
```

Any thoughts?

dengemann:
I confirm that it does not crash in Python 2.7

samuelgarcia:
This is anoying.
I see in neo.tests.iotest.test_micromedio that it is not py3 compatible!!
We need to fix it.

samuelgarcia:
The Io is very old it should be written entirely...
I made a quite and dirty fix here PR #361.

Could you test ?

dengemann:
Thanks! I will have a look in a moment.

johntukker:
I have a similar issue with neuroexplorerio.py, don't know if I should open this as a new issue but seems to be same so would be very happy if someone could have a look (I'm new to Python myself, so sorry if this is an obvious fix but even after looking at PR#361 still can't figure it out). This is running Python 3.4.5 with Jupyter.

r = io.NeuroExplorerIO(filename)
seg = r.read_segment(lazy=False, cascade=True)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-29-f75e95c4a5b3> in <module>()
[...]

PR #361, opened 2017-06-30
Quite and dirty fix for Denis bug.

samuelgarcia:
aka #360

### DIFF
--- a/neo/io/micromedio.py
+++ b/neo/io/micromedio.py
@@ -14,12 +14,7 @@ import datetime
 import os
 import struct
 
-# file no longer exists in Python3
-try:
-    file
-except NameError:
-    import io
-    file = io.BufferedReader
+from io import BufferedReader
 
 import numpy as np
 import quantities as pq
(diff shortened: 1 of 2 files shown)


dengemann:
@samuelgarcia sorry for delay. I just tried your branch and it looks good.

neo/io/micromedio.py
```diff
@@ -76,16 +71,16 @@ def read_segment(self, cascade=True, lazy=False, ):
         """
         Arguments:
         """
-        f = StructFile(self.filename, 'rb')
+        f = StructFile(open(self.filename, 'rb'))
```
dengemann: how does this actually work? Will there be an open file handle floating around? In that case using a with-statement could help.

samuelgarcia: StructFile is a very old wrapper arround struct module of python. And very old code in neo.io was using good old "file()" function. This function disapear with open(). So this is a quite and dirty fix.

For newly coded IOs. It really better to use the numpy.dtype with fields

So NeuroExplorer IO should be rewritten with this in mind.

Note that I am starting a strong refactoring here #354 around neo.io.

Maybe you could have a look there because tthis new neo.rawio could used in mne.

PR #365, opened 2017-07-03
Merge function for SpikeTrains

JuliaSprenger:
This PR implements the merge function (+tests) for SpikeTrain objects in an analogue way to the AnalogSignal merge function.
A new SpikeTrain object is generated containing the times (and waveforms) of both original SpikeTrains. If possible it is automatically linked to the parent segment (if identical for both original SpikeTrains). However, a link to a Unit is not created as I am not sure what should be the default behaviour for this: Create a new Unit and somehow merge the attributes of two original parent Units or link to one of the two existing ones?


apdavison:
Do you have a real world use case for such merging? In the case of `AnalogSignal`, there are a couple of good use cases for merging based on the possibility of having multiple channels in a single object. A `SpikeTrain` seems more like a single channel.

JuliaSprenger:
Here is one possible use case I came across:
In many cases the IOs directly yield multiple spiketrains each containing single-unit activity if spike sorting was performed on the data and do not offer the option to load all spikes in a single spiketrain. The generation of a single spiketrain containing multi-unit activity to simplify further multi-unit activity analysis. 
Eg. calculating the binned multi-unit activity based on single-unit activity spiketrains requires more resources than doing the same based on a single neo.spiketrain containing multi-unit activity.

PR #367, opened 2017-07-04
Multiple fixes for BlackRockIO

JuliaSprenger:
This PR fixes multiple issues of the BlackRockIO:

- previously manually provided locations of specific files were ignored if the files were not found. This is now checked explicitly and an error is raised if the specified file is not found.
- some checks on the provided time ranges were only executed on the first and last provided time range. Checks are now applied to all time ranges.
- the exact number of spikes loaded from a nev file depended on the other recording files loaded. This dependency is now removed.


samuelgarcia:
@JuliaSprenger @mdenker is it ready for merging in master ?
I would like to make a realease 0.5.2 more or less soon.

After 0.5.2, if you are OK this BlackrockIO would either:
  * replace by BlackrockRawIO.
  * live in an other name like BlackrockIO_V4 along side with futur BlackrockIO based on BlackrockRawIO

PR #368, opened 2017-07-12
Fixed a bug in BlackrockIO ChannelIndex annotations, and adds a few new ones.

mdenker:
This PR fixes a small bug in BlackrockIO, where the filter settings of the spike wave form extraction had units of uV instead of Hz. Also, this PR adds new annotations to the Channelindex that indicates the filters of the nsX file [x in 1..5], if present. The ns6 file, that may exist in addition to one ns1-ns5 file, has no filters.


apdavison:
I presume you mean the PR adds a few new annotations, not adds a few new bugs ;-)

Issue #381, opened 2017-09-09
Create a parent mixin class for `AnalogSignal` and `IrregularlySampledSignal`


apdavison:
There is a lot of repeated code in many of the methods of `AnalogSignal` and `IrregularlySampleSignal`. This should be moved to a mixin class that can be a parent of both classes.

apdavison:
as @JuliaSprenger suggested in the discussion on #382, we should also harmonize the unit tests, and develop a suite of tests common to both classes.

apdavison:
This is almost complete, but the common implementation of `merge()` wasn't working for `IrregularlySampledSignal`, so I had to re-add a sub-class-specific `merge()` implementation.

PR #382, opened 2017-09-09
Preserve parent objects on `rescale()`

### LINKED ISSUE #353: AnalogSignal.rescale() does not preserve parent objects
The same is true for IrregularlySampledSignal and SpikeTrain.

apdavison:
for `AnalogSignal`, `IrregularlySampledSignal`, `SpikeTrain`. Also fixed `IrregularlySampledSignal.copy()`

Fixes #353

### DIFF
--- a/neo/test/coretest/test_irregularysampledsignal.py
+++ b/neo/test/coretest/test_irregularysampledsignal.py
@@ -19,7 +19,7 @@ else:
     HAVE_IPYTHON = True
 
 from neo.core.irregularlysampledsignal import IrregularlySampledSignal
-from neo.core import Segment
+from neo.core import Segment, ChannelIndex
 from neo.test.tools import (assert_arrays_almost_equal, assert_arrays_equal,
                             assert_neo_object_is_compliant,
                             assert_same_sub_schema)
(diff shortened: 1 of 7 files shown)


samuelgarcia:
OK

neo/test/coretest/test_irregularysampledsignal.py
```diff
@@ -550,6 +562,11 @@ def test_as_quantity(self):
         self.assertIsInstance(sig_as_q, pq.Quantity)
         assert_array_equal(self.data1, sig_as_q.magnitude.flat)
 
+    def test__copy_should_preserve_parent_objects(self):
```
JuliaSprenger: in test_analogsignals, this test has a different name (test__copy_should_let_access_to_parents_objects). Maybe adjust one of the two for consistency?

JuliaSprenger:
Maybe one should start writing a battery of common tests for analogsignals and irregularsampledsignals, since many tests are similar or identical for both classes? But besides of the minor comment, the PR looks fine to me.

apdavison:
Good idea, I've added a comment ro #381 to this effect.

Issue #388, opened 2017-09-15
Creating epochs from events


mdenker:
An array of events (class `Event`) `[t1, t2, t3]` can be transformed into an array of epochs (class `Epoch`), `[(t1, t2), (t2, t3)]`.

It would be useful to add this as a method `Event.to_epoch()`

Example:
```python
>>> import numpy as np
>>> from neo import Epoch, Event, Segment

>>> seg = Segment(name="test")
>>> event = Event(times=np.array([5.0, 12.0, 23.0, 45.0]), units="ms", labels=np.array(["A", "B", "C", "D"]))
>>> event.segment = seg

>>> epoch = event.to_epoch()
>>> epoch.times
array([5.0, 12.0, 23.0]) * ms
>>> epoch.durations
array([7.0, 11.0, 22.0]) * ms
>>> epoch.labels
array(['A-B', 'B-C', 'C-D'])
>>> epoch.segment.name
"test"
```

JuliaSprenger:
some example code exists already here: https://web.gin.g-node.org/INT/multielectrode_grasp/src/master/code/neo_utils.py

apdavison:
Replaced #404 by #586

zm711:
This appears to be closed by #586. Reopen if I am wrong.

PR #404, opened 2017-09-28
add to_epoch method in Event class: Transform Events into an array of Epoch

jonathanduperrier:
add to_epoch method in Event class: Transform Events into an array of Epoch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,5 @@ install:
 # command to run tests, e.g. python setup.py test
 script:
   nosetests --with-coverage --cover-package=neo
-  assert_array_equal(epoch.magnitude, np.array([7.0, 11.0, 22.0]))
 after_success:
   coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,6 @@ install:
 # command to run tests, e.g. python setup.py test
 script:
   nosetests --with-coverage --cover-package=neo
+  assert_array_equal(epoch.magnitude, np.array([7.0, 11.0, 22.0]))
 after_success:
   coveralls
--- a/neo/core/event.py
+++ b/neo/core/event.py
@@ -243,9 +243,14 @@ class Event(BaseNeo, pq.Quantity):
     def to_epoch(self, durations=None):
         from neo.core import Epoch, Event, Segment
         if durations is None:
-            times = self.times[:]
-            durations = np.diff(self.times)[:]
-            labels = self.labels[:]
+            # times = self.times[:]
+            # durations = np.diff(self.times)[:]
+            # labels = self.labels[:]
+            
+            times = self.times[:-1]
+            durations = np.diff(self.times)
+            # labels = self.labels[:-1]
+            labels = np.array(["{}-{}".format(a, b) for a, b in zip(self.labels[:-1], self.labels[1:])])
         else:
             times = self.times
             labels = self.labels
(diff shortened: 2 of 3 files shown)


neo/core/event.py
```diff
@@ -221,3 +221,25 @@ def as_quantity(self):
...
+        """
+        Transform Events into an array of Epoch 
+        """
+        from neo.core import Epoch, Event, Segment
+
+        print("self.time : " + str(self.times))
+        print("self.label : " + str(self.labels))
+        i = 0
+        duration = []
+        for t in self.times:
```
JuliaSprenger: variable t is never used. Maybe it would be better to use 'for i in range(len(self.times)):'?

neo/core/event.py
```diff
@@ -221,3 +221,25 @@ def as_quantity(self):
...
+            if i==0 :
+                duration.append(self.times[i])
+            else :
+                duration.append(self.times[i] - self.times[i-1])
```
JuliaSprenger: for a better performace you could use np.diff(self.times) to calculate the durations instead of looping.

JuliaSprenger:
It looks as if you are generating epochs with starting at the times of the events, but having a duration corresponding to the time difference since the previous event. I would have rather expected epochs spanning the time between the events. Is this intended and under which conditions is it useful?
It would also help to add some documentation clarifying this as well as a unittest.
I am also not sure if a function like this should be a method of Events or rather should be available in an extended fashion as a general helper function. @apdavison: Any opinions on that?

apdavison:
@JuliaSprenger yes, I think the indexing is off by one.

We discussed in Lausanne the question of whether to make this a method or a helper function. There are arguments for both. In the end, I think having it as a method gives better discoverability - it is attached to the Event class, which I think makes it easier to find if you already have an Event object, compared to having to search through a potentially long list of helper functions. Of course we could have both, where the helper function calls the method or vice versa.

@jonathanduperrier I think the top priority now is to add some unit tests (in the file `neo/test/coretest/test_event.py`). The example given in #388 would be a good starting point.

samuelgarcia:
This looks also very strange to me.

I add in mind that to_epoch would be:
```python
Epoch(time=event.times[:-1], durations=np.diff(event.times), labels=event.labels[:-1])
```
Here, in thisimplementation, the duration is with the previsous event. I don't get the use case.

JuliaSprenger:
I agree with @samuelgarcia, this should be the default behaviour of the function. However, do you think it would make sense to introduce the option to provide durations in case you don't want to cover the whole recording time with the epochs but only specific time windows after the events?
```python
def to_epoch(durations=None):
    if durations is None:
        times = self.times[:-1]
        durations = np.diff(self.times)[:-1]
        labels = self.labels[:-1]
    else:
        times = self.times
        labels = self.labels
    return Epoch(time=times, durations=durations, labels=labels)
```

.travis.yml
```diff
@@ -13,5 +13,6 @@ install:
 # command to run tests, e.g. python setup.py test
 script:
   nosetests --with-coverage --cover-package=neo
+  assert_array_equal(epoch.magnitude, np.array([7.0, 11.0, 22.0]))
```
apdavison: I suppose you meant to put this line in test_epoch.py?

apdavison: sorry, I meant test_event.py

samuelgarcia:
I propose to postpone thie PR for 0.7.

apdavison:
Since this now has conflicts, I have updated @jonathanduperrier's code and made a new PR: #586

PR #409, opened 2017-10-04
Fixed typo preventing correct nev to be querried for file format version

mdenker:
This PR fixes a small typo/bug in the BlackrockIO of Neo.


neo/io/blackrockio.py
```diff
@@ -391,7 +391,7 @@ def __extract_nev_file_spec(self):
         """
         Extract file specification from an .nsx file
```
JuliaSprenger: There is also a typo in the docstring here

mdenker: True... fixing

samuelgarcia:
@mdenker: make teh same patch but for the file blackrockio_v4.py wich is now the old BlackrockIO.

Otherwise we have a big conflict here since rawio merge.

mdenker:
Hi @samuelgarcia 
I transferred the changes to v4 blackrockio, and to the rawio of Blackrock, which had the same bug.

samuelgarcia:
OK. thanks can I merge ?

mdenker:
Yes please go ahead. I'm very confident this small fix is ok.

Issue #420, opened 2017-10-19
Missing check for spike times before t_start in BlackRockIO


bjoern1001001:
Trying to load real datasets with the latest BlackRockIO, I noticed that some of them would raise the error that spike times are before t_start or after t_stop respectively. This seems to happen because these t_start and t_stop are determined based on the nsX files available. However, in rare cases also spike times just outside of the nsX recording time range can be present, therefore leading to a 'ValueError: The first spike (..) is before t_start (..)'. When no nsX file is available the IO works just fine.

So @JuliaSprenger 's and my suggested solution is to explicitly check for this and correct t_start and t_stop for the SpikeTrains accordingly by setting it to the min_nev_time (first spike observed) or max_nev_time (last spike observed) respectively.

I'd try implementing this myself tomorrow if there are no other suggestions.

bjoern1001001:
So I tried implementing this, but found a few difficulties, especially regarding the time_slices parameter in read_block. I'd like to ask you, @samuelgarcia, what you think should be done here.

What happens in BaseFromRawIO is that all t_starts (for AnalogSignals and SpikeTrains) are the same. However, because spike times can be before this t_start, that yields a few problems:
It's not possible to set all t_start to the smallest existing value (in this case the first spike) because this would shift the AnalogSignals towards earlier times which would be incorrect.
On the other hand setting SpikeTrain t_start to a lower value independently makes it impossible to find those spikes when setting the time_slices parameter in read_block, because the check whether the values are correct is done earlier.
Apart from that it's not always known when segments start or stop to people working with the data. In order to prevent frequent errors from happening, earlier versions of BlackRockIO would automatically set n_starts and n_stops to correct values, i.e. too small values to the minimal available time and too high values to the highest possible value. This also holds true, if 'None' was specified for either of them. So @JuliaSprenger and I think this behavior should be kept.

To sum all of this up, times should probably be corrected automatically and this should happen independently for AnalogSignals and SpikeTrains, so AnalogSignals can keep their correct t_start and the spikes occurring earlier than this can be read and shown as well.

samuelgarcia:
Hi Bjorn.
time_slices in read_block() is really ambiguous, 
I would prefer to discuss this case in #374 to avoid mixing all discusions.

In the new master, in BaseFromRawIO, I have ported more or less the behiavior of old BlackrockIO.
I don't really like this because there are special where it do not make sens for instance when all segment start at 0s.
[...]

PR #421, opened 2017-10-20
Fixed bug with BlackrockIO not loading because of spike times before/after t_start/t_stop

bjoern1001001:
I fixed a bug (see #420) that caused BlackRockIO to fail loading certain files because of spikes otuside of the time range set by the nsX start and stop times. Now _segment_t_start in BlackrockRawIO returns the smaller of t_start (read from available nsX files) and the first spike timestamp instead of t_start. Also the general t_stop is now set to the higher of t_stop from nsX and the last spike timestamp, thus  _segment_t_stop now returns this.

Apart from that I changed the value of self._t_starts in BlackrockRawIO from 0 to the first spike timestamp if no nsX file is available.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -309,8 +309,8 @@ class BlackrockRawIO(BaseRawIO):
                                                             units, gain, offset,group_id,))
             
             #t_start/t_stop for segment are given by nsx limits or nev limits
-            self._t_starts, self._t_stops = [], []
-            self._seg_t_starts = []
+            self._sigs_t_starts = []
+            self._seg_t_starts, self._seg_t_stops = [], []
             for data_bl in range(self._nb_segment):
                 length = self.nsx_data[data_bl].shape[0]
                 if self.__nsx_data_header[self.nsx_to_load] is None:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
This looks very good. Thanks.

To avoid any confusion for the day we will be back on the code I would explicitly change:
  * self._t_starts to self._sigs_t_starts
  * self._t_stops to self._seg_t_stops
  * keep self._seg_t_starts

Otherwise, just curious : why to np.math.ceil() instead of int() in _get_timestamp_slice ?
Note that I am OK with that.

bjoern1001001:
Thanks for your comment.

I think you're right here and I will change this (probably on Monday).

I needed to do this because if it rounds downwards then spikes are loaded from the lower value. Then the time of the first spike might be lower than t_start. Let's say there is a timestamp at 40 samples and t_start is set by the user to (40.5/30000)*pq.s then SpikeTrain will raise an error because that spike at 40 samples is earlier than t_start. I admit this is probably are very rare situation but I managed to create it and thus noticed and changed this.

samuelgarcia:
I guess this is ready to merge now.

Did you run noestests ?

bjoern1001001:
Yes, I just ran them. 
The result is the same as for the Unittests: Two tests fail when testing BlackRockIO because it expects Unit.name to be a regular string but it's a Unicode string now. The other tests pass.

samuelgarcia:
I guess you are in python27. Have you tried on py3 ?

The problem come from basefromrawio.py I have forgotten 
```unit_annotations = check_annotations(self.raw_annotations['unit_channels'][c])
```
at line 145 and 151.

I will fiw it in a separated PR.

PR #422, opened 2017-10-23
Fixed channel_id in Unit annotations always 144 and unit_id always 1 in BlackrockIO

bjoern1001001:
The channel_id and unit_id values were determined in a for loop and saved to a list. When reading them to set the annotations the temporary variables that contain only the last values were read. Now the values are read from the list as originally intended.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -222,6 +222,7 @@ class BlackrockRawIO(BaseRawIO):
         event_channels = []
         unit_channels = []
         sig_channels = []
+        unit_channel_ids = []
 
         # Step1 NEV file
         if self._avail_files['nev']:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Oh, I am stupid.
Thank you for the vigilance.
Tell me when ready to merge.

JuliaSprenger:
@samuelgarcia would it make sense to start adding unittests for the new BlackRockIO at this point or do you already have a plan on how to adjust the unittests of the old IO to the new implementation?

samuelgarcia:
There are already tests at rawio level here : neo/rawio/tests/...

For Blackrock, I try to port all possible old tests when it was possible.
Some of them are in neo/rawio/tests/test_blackrockrawio.py and some of them  are in neo/tests/iotests/test_blackrockio.py. Tell me if I missed some tests.

Note that in the old implementation it was ```files_to_test = []``` and now it is ```files_to_test = ['FileSpec2.3001']```. This means that many common tests are launch and this avoid some lines of codes. Commons tests try to open everything and tests if all obejct are conform (hard to pass).

Comparaison with matlab raw signals are done at rawio tests level.

JuliaSprenger:
Thanks for pointing to the rawio tests, I was looking for them at neo/test/rawiotest. The BlackRockIO test were completely ported to the new IO. I think we can merge this PR.

samuelgarcia:
OK!!

Issue #423, opened 2017-10-26
annotation field name consistency in baserawio


cboulay:
[Here](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/baserawio.py#L207) it should probably be `event_channels` instead of `event_channel`.

samuelgarcia:
Absolutely.
Thank you very much for this fine inspection.

PR #426, opened 2017-10-27
Added file origin to all neo objects when using RawIO API

bjoern1001001:
When creating the objects in BaseFromRawIO.py I added the parameter file_origin everytime an object is created. The value of this is taken from self._source_name(). Now every object contains the attribute file_origin and this is not None.

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -143,10 +143,8 @@ class BaseFromRaw(BaseIO):
         unit_channels = self.header['unit_channels']
         if units_group_mode=='all-in-one':
             if unit_channels.size>0:
-                chidx_annotations = self.raw_annotations['signal_channels'][-1]
-                chidx_annotations = check_annotations(chidx_annotations)
                 channel_index = ChannelIndex(index=np.array([], dtype='i'),
-                                        name='ChannelIndex for all Unit', **chidx_annotations)
+                                        name='ChannelIndex for all Unit')
                 bl.channel_indexes.append(channel_index)
             for c in range(unit_channels.size):
                 unit_annotations = self.raw_annotations['unit_channels'][c]
@@ -155,12 +153,10 @@ class BaseFromRaw(BaseIO):
                 
         elif units_group_mode=='split-all':
             for c in range(len(unit_channels)):
-                chidx_annotations = self.raw_annotations['signal_channels'][-(len(unit_channels)-c)]
-                chidx_annotations = check_annotations(chidx_annotations)
                 unit_annotations = self.raw_annotations['unit_channels'][c]
                 unit = Unit(**unit_annotations)
                 channel_index = ChannelIndex(index=np.array([], dtype='i'),
-                                        name='ChannelIndex for Unit', **chidx_annotations)
+                                        name='ChannelIndex for Unit')
                 channel_index.units.append(unit)
                 bl.channel_indexes.append(channel_index)
         
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -127,7 +127,8 @@ class BaseFromRaw(BaseIO):
             for i, (ind_within, ind_abs) in self._make_signal_channel_subgroups(channel_index, 
                                                         signal_group_mode=signal_group_mode).items():
                 neo_channel_index = ChannelIndex(index=ind_within, channel_names=all_channels[ind_abs]['name'].astype('S'),
-                                channel_ids=all_channels[ind_abs]['id'], name='Channel group {}'.format(i))
+                                channel_ids=all_channels[ind_abs]['id'], name='Channel group {}'.format(i),
+                                                 file_origin=self.source_name())
                 bl.channel_indexes.append(neo_channel_index)
         
         #ChannelIndex and Unit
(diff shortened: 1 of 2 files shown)


samuelgarcia:
For having this, I really would prefer to add 3 lines here
BaseRawIO._generate_minimal_annotations(...)
```python
        for c in range(signal_channels.size):
            #use for ChannelIndex.annotations
            d = {}
            d['name'] = signal_channels['name'][c]
            d['channel_id'] = signal_channels['id'][c]
            d['file_origin'] = self._source_name()#<<<<<<Patch
            a['signal_channels'].append(d)

        for c in range(unit_channels.size):
            #use for Unit.annotations
            d = {}
            d['name'] = unit_channels['name'][c]
            d['id'] = unit_channels['id'][c]
            d['file_origin'] = self._source_name()#<<<<<<Patch
            a['unit_channels'].append(d)

        for c in range(event_channels.size):
            #not used in neo.io at the moment could usefull one day
            d = {}
            d['name'] = event_channels['name'][c]
            d['id'] = event_channels['id'][c]
            d['file_origin'] = self._source_name()#<<<<<<Patch
            a['event_channel'].append(d)

```

**_generate_minimal_annotations** is the place where annotations should be minimal for **all** IO.

Making this in BaseFromRaw make things spread and less clear.

samuelgarcia:
While you are in annotations corner, could you also change a small inconsistency see #423 mentioned by @cboulay.

'event_channel' must be change to 'event_channels' at lines 207 and 257.

neo/io/basefromrawio.py
```diff
@@ -119,15 +119,19 @@ def read_block(self, block_index=0, lazy=False, cascade=True, signal_group_mode=
...
-        
+
+
         #ChannelIndex ofr AnalogSignals
         all_channels = self.header['signal_channels']
         channel_indexes_list = self.get_group_channel_indexes()
         for channel_index in channel_indexes_list:
             for i, (ind_within, ind_abs) in self._make_signal_channel_subgroups(channel_index, 
                                                         signal_group_mode=signal_group_mode).items():
+                chidx_annotations = self.raw_annotations['signal_channels'][i]
```
samuelgarcia: To be compatible with py2 and py3 you must call here:
```
chidx = check_annotations(chidx)
```

neo/io/basefromrawio.py
```diff
@@ -138,8 +142,9 @@ def read_block(self, block_index=0, lazy=False, cascade=True, signal_group_mode=
         unit_channels = self.header['unit_channels']
         if units_group_mode=='all-in-one':
             if unit_channels.size>0:
+                chidx_annotations = self.raw_annotations['signal_channels'][-1]
```
samuelgarcia: To be compatible with py2 and py3 you must call here:
```
chidx = check_annotations(chidx)
```

neo/io/basefromrawio.py
```diff
@@ -148,10 +153,11 @@ def read_block(self, block_index=0, lazy=False, cascade=True, signal_group_mode=
                 
         elif units_group_mode=='split-all':
             for c in range(len(unit_channels)):
+                chidx_annotations = self.raw_annotations['signal_channels'][-(len(unit_channels)-c)]
```
samuelgarcia: To be compatible with py2 and py3 you must call here:
```
chidx = check_annotations(chidx)
```

neo/rawio/baserawio.py
```diff
@@ -233,28 +233,38 @@ def _generate_minimal_annotations(self):
...
             d['id'] = unit_channels['id'][c]
+            d['file_origin'] = self._source_name()
             a['unit_channels'].append(d)
 
+        for c in range(unit_channels.size):
+            #use for ChannelIndex.annotations
+            d = {}
+            d['file_origin'] = self._source_name()
+            a['signal_channels'].append(d)
+
```
samuelgarcia: This lines 255>260 are strange.

You loop over unit_channel.size and you fill annotations for signal_channels.
With this we will have len(raw_annotations['signal_channels'])>signal_channels.size this is not consistent.

My guess is that you do this for having annotations in ChannelIndex that describe a Unit (which is a weakness is actual neo.core and that should change). If so I would prefer no doing this here and remove that lines.
In that case you could use raw_annotations['unit_channels'] to ChannelIndex that are linked to Unit.

bjoern1001001: Yes, that's what I was trying to do. 
So what you mean is adding something like raw_annotations['unit_channels'][channel_id]['channel_index'] as a dictionary for the annotations of ChannelIndex?

bjoern1001001:
Yeah, you are right. I only understand that system correctly now. As you noticed I just fixed it and hope it is better now.
One more question: What to do when signal_group_mode="group-by-same-units" or units_group_mode="all-in-one"? Should I implement a test for this and if it's like this then group the annotations as well? I mean that there would be a list of values for each key of the annotations.

neo/rawio/baserawio.py
samuelgarcia: line 255 to 260 must be removed.

What I meant is in basfromrawio.py:
```python
        elif units_group_mode=='split-all':
            for c in range(len(unit_channels)):
                unit_annotations = self.raw_annotations['unit_channels'][c]
                unit = Unit(**unit_annotations)
                channel_index = ChannelIndex(index=np.array([], dtype='i'),
                                        name='ChannelIndex for Unit')
                channel_index.units.append(unit)
                bl.channel_indexes.append(channel_index)
```
could be change by

```python
        elif units_group_mode=='split-all':
            for c in range(len(unit_channels)):
                unit_annotations = self.raw_annotations['unit_channels'][c]
                unit = Unit(**unit_annotations)
                channel_index = ChannelIndex(index=np.array([], dtype='i'),
                                        name='ChannelIndex for Unit', **unit_annotations)###patch here
                channel_index.units.append(unit)
                bl.channel_indexes.append(channel_index)
```

But finally it is a bad idea. raw_annotations['unit_channels'] must be routed to Unit.annotations only.
ChannelIndex must get raw_annotations['signal_channels'] in case of AnalogSignal and nohting in case of SpikeTrain to avoid confusion (which is already high).

samuelgarcia:
Hi,
I am back on this old PR.
Line 255 to 260 are still here, they must be removed.
Could you remove them ? So we could finalize this PR before others.
Best

samuelgarcia:
thank you very much.
I guess this is ready for merge.

bjoern1001001:
Yes, I think it can be merged.

PR #427, opened 2017-10-28
blackrockrawio _get_event_timestamps make use of t_start, t_stop

cboulay:
I promise I'm not combing through the code looking for tiny bugs. I'm writing a new io and I'm using the existing rawio's as examples. Sometimes I find something that I think might be a bug but it may just as well be there on purpose because of some weird idiosyncrasy of the data format. This is one of those times. I think it's a bug but I could easily be wrong.

PR passes test_blackrockio in intel python3.5 on Ubuntu 16.04


samuelgarcia:
Waou!!
Yes it is bug.
Thank you.

PR #428, opened 2017-10-30
Improvement for the parent class BaseSignal: adding other methods (see issue #381)

msenoville:


msenoville:
Improvement for the parent class BaseSignal: adding other methods (see issue #381)

PR #429, opened 2017-11-03
Added back list of available files as annotations to block in BlackRockRawIO

bjoern1001001:
The latest version of BlackrockIO using the RawIO API was missing a few annotations for the Block objects that contained information on available files and recording pauses. Using the already existing structure for annotations in RawIO I added the checks for those annotations like it was done in earlier versions of BlackrockIO.


samuelgarcia:
I guess we can merge.
OK ?

bjoern1001001:
I think so, yes.

PR #431, opened 2017-11-03
BCI2000 IO

cboulay:
This adds a new IO to Neo for importing BCI2000 dat files.

How do I upload the test data files to g-node?


neo/rawio/bci2000rawio.py
```diff
@@ -0,0 +1,346 @@
...
+        # Calculate the dtype for a single timestamp of data. This contains the data + statevector
+        self._read_info['line_dtype'] = [('raw_vector', self._read_info['sample_dtype'], self._read_info['n_chans']),
+                                         ('state_vector', np.uint8, self._read_info['state_vec_len'])]
+        import os
+        self._read_info['n_samps'] = int((os.stat(self.filename).st_size - file_info['HeaderLen'])
+                                         / np.dtype(self._read_info['line_dtype']).itemsize)
+
+        # TODO: Do we just memmap it now, or elsewhere?
+        self._memmap = np.memmap(self.filename, dtype=self._read_info['line_dtype'],
+                                 offset=self._read_info['header_len'], mode='r')
```
samuelgarcia: I think this is teh good place to create the memmap.

neo/rawio/bci2000rawio.py
```diff
@@ -0,0 +1,346 @@
...
+        return 0
+    
+    def _get_spike_timestamps(self,  block_index, seg_index, unit_index, t_start, t_stop):
+        return []
+    
+    def _rescale_spike_timestamp(self, spike_timestamps, dtype):
+        return spike_timestamps
+
+    def _get_spike_raw_waveforms(self, block_index, seg_index, unit_index, t_start, t_stop):
+        return np.array((0, 0, 0))
```
samuelgarcia: When _spike_count return 0 you can to simplify have
```python
def _get_spike_timestamps(...):
    return None
def _rescale_spike_timestamp(...):
    return None
def _get_spike_raw_waveforms(...):
    return None

```
because normally this never will be called.

neo/rawio/bci2000rawio.py
```diff
@@ -0,0 +1,346 @@
...
+            group_id = 0 
+            sig_channels.append((ch_name, chan_id, sr, dtype,  units, gain,offset, group_id))
+        self.header['signal_channels'] = np.array(sig_channels, dtype=_signal_channel_dtype)
+
+        self.header['unit_channels'] = np.array([], dtype=_unit_channel_dtype)
+        
+        # creating event channel for each state variable
+        event_channels = []
+        for st_ix, st_tup in enumerate(state_defs):
+            event_channels.append((st_tup[0], 'ev_' + str(st_ix), 'event'))
```
samuelgarcia: In your code it seems that here it should be 'epoch' and not 'event' because duration is not None.

samuelgarcia:
This new io looks, the code is really clear.

For the file you need, a G-Node account (there are base in Munchen, Germany).
I don't remember how to get it. I will ask.

@achilleas-k is it possible to create a ccount for Chadwick ?

I can also send me theses files and I will put them there myself, it will be faster.

achilleas-k:
If it's just a one-time addition of files I (or you, @samuelgarcia) can add them. If @cboulay will need to update the files frequently, we can create an account, yes. Let me know if an account is necessary.

cboulay:
Sorry, the last few days have been busy and somehow I didn't get notifications about your comments. The data files are publicly available as part of the BCI2000 distribution but it's probably still easier for me to send them to Samuel.

neo/rawio/bci2000rawio.py
```diff
@@ -0,0 +1,346 @@
...
+            if t_stop is not None:
+                keep = np.logical_and(keep, ts <= t_stop)
+        return ts[keep], dur[keep], labels[keep]
+    
+    def _rescale_event_timestamp(self, event_timestamps, dtype):
+        event_times = (event_timestamps / float(self._read_info['sampling_rate'])).astype(dtype)
+        return event_times
+    
+    def _rescale_epoch_duration(self, raw_duration, dtype):
+        # really easy here because in our case it is already seconds
```
cboulay: oops. I should get rid of this old comment that happens to be incorrect.

neo/rawio/bci2000rawio.py
cboulay: Almost all state changes in BCI2000 are stimulus-related; the state will go from 0 to some value when it turns on, then back to 0 when it turns off. Sometimes the transition back to 0 is nonsense, e.g. when the state change triggers an animation in a game engine, the state will change back to 0 after a minimum duration of one data block, but the animation might continue for some unknowable time. The duration is often not useful, but sometimes it is, so I would like to hold on to the 'duration' information for experts who need it.

Another issue is that I predict calling these 'epochs' will introduce confusion due to context. BCI2000 is more popular in EEG and ECoG than in microelectrode neurophys, so many people that might use Neo to import BCI2000 will be coming from those backgrounds and might already be familiar with EEGLAB. In EEGLAB, the BCI2000 importer (to which I contribute) imports the zero-to-nonzero state transitions as events, and the word 'epoch' is strongly associated with the concept of an entire 'trial'.

I don't mind calling them epochs instead but that's easy because I'm aware of the issue. Or do you think it would be better to keep them as events but maybe give each an annotation holding the duration, for the expert users?

samuelgarcia: I am aware of that confusion for people from eeg.

'event' or 'epoch' is not the name it refer to the type.
_event_channel_dtype = [('name','U64'),('id','U64'),('type', 'S5')]   # type = epoch or event
It is part of tha API to tell neo.io if this event channel is in fact a neo.Event (no duration) or a neo.Epoch (with duration).
So in your case, if you put 'event' this will never be translate to neo.Epoch and so loose the duration at neo.io level.

Maybe you can create 2 channel which duplicate artificially the same triggers.
One is 'event' and will be a neo.Event and one is 'epoch' and will be a neo.Epoch at neo.io.
Depending of chanel_index you can set duration is None or not.

cboulay: Thanks for the clarification about data types. I've been working almost exclusively at the rawio level.

Maybe I'm overthinking this. I think I will try keep the stimulus-off transitions as events, then the user can calculate their own duration if they need it, or they can discard/ignore these events if they don't.

cboulay:
I've addressed your comments and sent you the test data. I will await further instructions or comments.

samuelgarcia:
Everything looks good.

I have put files at gnode there are here https://portal.g-node.org/neo/bci2000/

So you shoudl able to lauch  ```nosetest  neo/test/iotest/test_bci2000.py```

Could you also make a test file in neo/rawio/tests/test_bci2000rawio.py
See other test in the dir.
It is important, some other tests are launch at low level.

Could you test both on py and py3 ?

cboulay:
I added the rawio test then made some py2 compatibility fixes. It is now py2 and py3 tested.

I had to comment out `from __future__ import unicode_literals` because it was interfering with my ability to define column names for my structured arrays. Nothing else seems to be affected so far.

Sorry I took so long to get to this. I still can't figure out why I'm not getting notifications for comments on this PR. If you have any further comments or requests for me, please @ mention me by name to make sure I get the notification.

samuelgarcia:
I think it is OK.
Tell us if we can merge.
Thank you for this new RawIO.

samuelgarcia:
@cboulay: we now have a circle-ci that do test also on IOs . BCI2000 fails because of key 'event_channel' that have change to 'event_channels' to be consistent.

Could you check this ? So we could merge soon.

cboulay:
The 3.6 test failed because `TestTdtIOIO` (SIC) timed out after 10 min.
The 2.7 test is taking a long time, probably for the same reason.

BTW, I discovered why I wasn't getting e-mail notifications about this PR. I hope I won't take another 2 weeks to answer you next time.

samuelgarcia:
Hi.
thanks for the fix. 
I guess this is ready for merge.

PR #433, opened 2017-11-14
Open Ephys file format

Palaeocortex:
Added support for Open Ephys file format
OpenEphys.py from https://github.com/open-ephys/analysis-tools.git, with minor changes
openephysio.py
Tested with Python 3.5.2 and 2.7

### DIFF
--- a/neo/io/OpenEphys.py
+++ b/neo/io/OpenEphys.py
@@ -1,0 +1,720 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Aug  3 15:18:38 2014
+
+@author: Dan Denman and Josh Siegle
+
+Loads .continuous, .events, and .spikes files saved from the Open Ephys GUI
+
+Usage:
+    import OpenEphys
+    data = OpenEphys.load(pathToFile) # returns a dict with data, timestamps, etc.
+
+"""
+
+import os
+import numpy as np
+import scipy.signal
+import scipy.io
+import time
+import struct
+import json
+from copy import deepcopy
+import re
+import math
+
+# constants for pre-allocating matrices:
+MAX_NUMBER_OF_SPIKES = 1e6
+MAX_NUMBER_OF_EVENTS = 1e6
+
+def load(filepath):
+    
+    # redirects to code for individual file types
+    if 'continuous' in filepath:
+        data = loadContinuous(filepath)
+    elif 'spikes' in filepath:
+        data = loadSpikes(filepath)
+    elif 'events' in filepath:
+        data = loadEvents(filepath)
+    else:
+        raise Exception("Not a recognized file type. Please input a .continuous, .spikes, or .events file")
+        
+    return data
+
+def loadFolder(folderpath,**kwargs):
+
+    # load all continuous files in a folder
+
+    data = { }
+
+    # load all continuous files in a folder  
+    if 'channels' in kwargs.keys():
+        filelist = ['100_CH'+x+'.continuous' for x in map(str,kwargs['channels'])]
+    else:
+        filelist = os.listdir(folderpath)   
+
+    t0 = time.time()
+    numFiles = 0
+    
+    for i, f in enumerate(filelist):
+        if '.continuous' in f:
+            data[f.replace('.continuous','')] = loadContinuous(os.path.join(folderpath, f))
+            numFiles += 1
+
+    print(''.join(('Avg. Load Time: ', str((time.time() - t0)/numFiles),' sec')))
+    print(''.join(('Total Load Time: ', str((time.time() - t0)),' sec')))
+            
+    return data
+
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Thank you very much for this PR.
I also wanted to add openephys in neo.

Neo is io is moving slowly toward rawio for some format that need chunk acces and speed.

Format integrated in neo.rawio have automatically neo.io.

[tridesclous](https://github.com/tridesclous/tridesclous) and [ephyviewer](https://github.com/NeuralEnsemble/ephyviewer) are already based on this rawio low level API.

I think having openephys in rawio will be important for spike sorting on long datasets.

I will have a look on thsi io but after would you mind if I try to port this Io to rawio ?

Palaeocortex:
No problem, port it to whatever you find more appropriate. Happy to contribute.

apdavison:
Pending a rawio-compatible implementation, I think we should merge this.

@Palaeocortex could you provide us with a test file?

Palaeocortex:
I've just uploaded a test data file (2017-05-19_20-25-32 DAT wt 271 testfile.tar.gz
) to my python-neo branch.

samuelgarcia:
@Palaeocortex:
files must not be upload directly in the neo git.

They should be uploaded here at [gin gnode](https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data)

See [doc](http://neo.readthedocs.io/en/latest/io_developers_guide.html).

Best.

samuelgarcia:
Hi @Palaeocortex
I am back on this.
I plan to rewrite this directly at neo.rawio level
So user of tridesclous could also use this IO.

I will do a lot od copy/paste from your code.

I will open a separate PR.
Could you send me the file for testing ?

When the new implementaion will be done could you help me by rewriwing it ?

best


Samuel

samuelgarcia:
I got your file.
It was in your branch.

samuelgarcia:
hi @Palaeocortex
I have started look at the dataset you provide in your PR.
I have read the the [format doc](https://open-ephys.atlassian.net/wiki/spaces/OEW/pages/491632/Data+format)
And I have some questions:
  * Where does the xml files come from ?
    Are they part of open ephys ?
  * Why does some files have the following naming 100_CH1 and some other 100_CH1_2 ?
    Does the _2 have a meaning of a new Segment ?
    If so it is not present in your code.

Thank you for helping.

best

Palaeocortex:
Hi Samuel! Sorry for uploading data in the wrong place, I forgot to correct.

The xml files are part of OpenEphys and carry information about GUI settings and channel info. 

The _2 says that the measurement has been interrupted and restarted.  100_CH1 is the first measuring session, then recording stopped and started again. The second session is called 100_CH1_2, the third 100_CH1_3 and so on. My code checks the timestamps and puts the snips together, adding zeros inbetween, for the duration of the brake. The example file shows one such interruption. I can happen if the experimenter wants to interrupt or when occasionaly the OpenEphys GUI breaks and needs to be restarted.

samuelgarcia:
Hi Cristian,
Thank you for clarification.
I did not get that from code nor it is explain in format documentation.
I will change this behavior to be more neo compatible with creating several neo.Segment
This will avoid to create to much zeros in between "session" (segment)
Best,

samuelgarcia:
@Palaeocortex 
If you have time please have a look to #580 
For the moment the multi segment is not taken in account but it won't be difficult.

Palaeocortex:
I'll check the code in #580. Maybe both behaviors should be implemented? In my case, it was desirable to have a single segment because data belongs to a single experiment with important timing. With several neo Segments I'd have to stick them together manually and add zeros. With software crashes you lack any other choice.

samuelgarcia:
Yep. But guleing with zeros could be done outside this IO.
It is  generic for all IO with multi segment support.

It is more tools that could be( or not) in neo but I don't think to add this option  specifically to this IO

The new way of neo is :  load from disk only the necessary chunks to save memory bandwith.
So the fragmentation is kept as close as possible from the datset.

samuelgarcia:
@Palaeocortex 
Hi.
I implementing teh multi segment stuff in openephys io.
I experience some problem with the datset you provide.

For a given segment all channel files do not have the lenght.
How is possible ?
And worst than than the filesize is a not multiple of 1024 as it should be.
How is it possible. There is no garanty of atomic buffer writing at acquisition in OE ?

samuelgarcia:
superseded by #580

Palaeocortex:
Hi! Yes, the file length is not consistent and this occurs often, especially when the OpenEphys GUI freezes. I'm still loading the data because just one data point is missing (if I remember correctly), that means an interval of 1/sampling rate. It is not the official behavior of the OpenEphys GUI though, at least from what I understand from the documentation. I made the consistency checks in my code a bit more flexible to deal with this minor events.

PR #435, opened 2017-11-17
Fix for loading Blackrock files of version 2.1

bjoern1001001:
After playing around a lot with the new BlackrockRawIO I was able to get Blackrock files of version 2.1 to load (see also issue 417). I copied some parts of the old IO over because they were referenced but not yet available and adapted them to fit into the new structure. They originally required more data being available already but this could be eliminated because the needed functionality (a lot less than the original methods had) did not require those. Using this code I was able to fully load files of version 2.1.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -297,13 +297,13 @@ class BlackrockRawIO(BaseRawIO):
             if spec in ['2.2', '2.3']:
                 ext_header = self.__nsx_ext_header[self.nsx_to_load]
             elif spec == '2.1':
-                self.__nsx_params[spec]('labels', self.nsx_to_load)
                 ext_header = []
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Thank you for this.
Have you tested with both 2.7 and 3.X ?

samuelgarcia:
Tell me OK for merging.

bjoern1001001:
I had thought so but I tested again and noticed the bug you seem to be referring to about str.decode() not being possible. I changed this to differ between version 2.1 and 2.2/2.3 as well. Now it should work with all python versions.
I also added additional tests for loading version 2.1 and will create a pull request for those soon.

bjoern1001001:
In my opinion it can be merged.

bjoern1001001:
@samuelgarcia, do you think there are any more necessary changes to this pull request or can it be merged?

samuelgarcia:
Let's merge.

PR #438, opened 2017-12-13
ChannelIndex annotations for BlackrockIO

bjoern1001001:
I copied the generation of annotations for ChannelIndex for Analogsignals from the old BlackrockIO to BlackrockRawIO and adapted it to use the annotation structure that is set in BaseRawIO.
Apart from that I also differantiated between signal_group_modes when adding the annotations to the ChannelIndex objects. When it is 'split-all' nothing changed but for 'group-by-same-units' all annotations for all AnalogSignals appended to the ChannelIndex are added there in a list. So the AnalogSignal should be at the same position in ChannelIndex.analogsignals as the corresponding annotations. These are the annotations that would be added to the ChannelIndex if it had only a single AnalogSignal.


samuelgarcia:
Outch. The big PEP8 clean break the compatibility with the master.
Sorry for not having merged before.
Could rebase on the actual master ?

bjoern1001001:
Updated

samuelgarcia:
OK.

samuelgarcia:
Can I merge ?

Issue #226, opened 2015-12-15
units should not require unit magnitude


kylerbrown:
Hello,

Right now, neo enforces the `units` attribute to have unit magnitude, e.g. `1*uV`.

There at least two situations where I disagree with this rule.
1) When using fixed gain amplifiers, such as the popular Intan chips used in the Open Ephys, there is a known constant conversion from a binary integer value to a microvolt value. The 16 bit integer is a much more efficient storage option (both in space and speed) than converting all the values to a 64 bit floating point value required to represent the true microvolt magnitude. Therefore, I'd like to store the units as `0.195 uV`.
2) For some event data, the values may be in samples, and the units therefore, are `1/sampling_rate * s`. For the reasons mentioned in (1), and to eliminate any added floating point error. I'd like to be able to store them in there original integer format.

Curious what you think! (And thanks for working on NEO)

apdavison:
I'm not sure this was a deliberate decision. Can you give a short example of code which doesn't work (together with the Traceback), which you think ought to work?

samuelgarcia:
I am agree with this.
But this depend on each IO to implement it or not.

Now the io api have been split in two part. BaseFromRaw is a wrapper for neo.io on top on neo.rawio.

So we could add an option in BaseFromRaw.read_block, for instance **quantity_mode**:
  * **quantity_mode='float_magnitude'** would lead to [0.2, 0.4, .6, .8] * pq.uV
  * **quantity_mode='raw_magnitude'** would lead to [1, 2, 3] * (0.2*pq.uV)

I guess this can be iplemenetd with quantities.CompoundUnit but I am not sur.

samuelgarcia:
This will be implemented in proxy object.

samuelgarcia:
#461 implement this feature but it is postpone to 0.8.

PR #439, opened 2017-12-13
IO specific time units

bjoern1001001:
Note: This pull request builds on pull request #435, thus should not be merged before it.
Using the new RawIO I noticed that all times are rescaled to seconds. Me and my colleagues thought it might be better to at least be able to keep the timestamps as integer values in order not to lose precision and to be able to work with them more easily.
So what I did was to pass all timestamps and t_start and t_stop with a pq.unit (e.g. 1.0/30000 * s) from BlackrockRawIO and implement a check for this in BaseFromRawIO. So all IOs that did not change are not affected, because they are treated the same way as before but it can now be decided on RawIO level, which kind of behavior is wanted.
I have not done this consequently, because many methods in the RawIOs require times passed as seconds but I tried to keep them in their original form for as long as possible.

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -273,7 +273,7 @@ class BaseFromRaw(BaseIO):
                 t_stop = seg_t_stop
             
             #in float format in second (for rawio clip)
-            t_start_, t_stop_ = t_start.rescale('s'), t_stop.rescale('s').magnitude
+            t_start_, t_stop_ = t_start.rescale('s').magnitude, t_stop.rescale('s').magnitude
             
             #new spiketrain limits
             seg_t_start = t_start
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi. I won't have time to read that code before january. I am really sorry for that.

I totally agrre with the idea of CompundUnits. See thies really old issu #226

This more global that timestamps. It also must be discuss for magnitude of signals when rescaling.

I would like to have this completly optional for all RawIO not just BlackRockIO:
  * having the true raw data in there dtyp in numpy way (RawIO level)
  * having the rescale data with magnitude in raw format with compund units at neo.io level (IO level)
  * having the reqscale data with float magnitude (rescaled like today) so that AnalogSignal.magnitude and SpikeTrain.magnitude are transparent for users. (IO level)

Note that RawIO **must not** depend depend on quantitites module because rawio will be share with mne.io soon and they certainly won't be agree to depend on an unmaintened package as we do.
So the design of RawIO is from the begenning for all rawIO without *pq* deliberatly.

The *pq* is only at IO level so if one day we switch to another units system it will be easy.

Making pq.CompundUnits should be easy for AnalogSignal with channelheader BaseFromraw.

Doing the same for times (in spike and event)  need some change at rawIO API by adding some more info in event and units headers table but should be quite easy.

I really prefer wait the meeting with mne in january before taking decision how to implement theses magnitude and compundunits at io level.

samuelgarcia:
I think, we are most or less agree with @JuliaSprenger that this magnitude and CompunfUnits should be done with proxy object when lazy loading because the implemention will be more simpler.

So, I will be back here when proxy/lazy lod will be started

bjoern1001001:
This should be implemented with proxy objects in a different way, so I think, this can be closed here.

PR #441, opened 2017-12-20
Corrected source for file_origin in BlackrockIO

bjoern1001001:
Now file_origin in BlackrockIO really points to the file that data are read from. nsx_override and nev_override are respected for this. The more general self.filename is only used for blocks and segments now.


samuelgarcia:
Outch. The big PEP8 clean break the compatibility with the master.
Sorry for not having merged before.
Could rebase on the actual master ?

bjoern1001001:
Updated

samuelgarcia:
OK.

PR #442, opened 2017-12-20
Loading only a subset of all channels

bjoern1001001:
Currently it is not possible to load only a few channels instead of the whole file. @samuelgarcia mentioned something about doing this on IO level. As far as I can tell this requires a framework at BaseFromRawIO level. Thus I tried implementing it there. This required combining the wanted channels with the existing channels on multiple places. 
What happens now is that when 'channels_to_load=...' is specified in read_block or read_segment, AnalogSignals and SpikeTrains are only loaded if they belong to one of the specified channels. 
If nothing is specified, all channels are loaded in order not to change the current default behavior.

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -71,7 +71,7 @@ class BaseFromRaw(BaseIO):
         self.parse_header()
     
     def read_block(self, block_index=0, lazy=False, cascade=True, signal_group_mode=None, 
-                units_group_mode=None, load_waveforms=False, time_slices=None):
+                units_group_mode=None, load_waveforms=False, time_slices=None, channels_to_load='all'):
         """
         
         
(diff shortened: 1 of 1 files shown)


neo/io/basefromrawio.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, *args, **kargs):
         self.parse_header()
     
     def read_block(self, block_index=0, lazy=False, cascade=True, signal_group_mode=None, 
-                units_group_mode=None, load_waveforms=False, time_slices=None):
+                units_group_mode=None, load_waveforms=False, time_slices=None, channels_to_load='all'):
```
samuelgarcia: This **channels_to_load** is that it refer the index inside the rawio.
For the general case when only one AnalogSignal with several channel at same sampling_rate this sounds goods.
But difficulties here for the user when it is not the genral case
Imagine that a IO give you for each segment 2 AnalogSignal:
  * the first shape (64, 10000) at 30kHz
   * the second is (20, 1000) at 1000Hz
Blackrock could and certainly will be have this if we read at the same time ns1 and ns5.
Internally the rawio have 84 channels so channels_to_load is in [0 ... 83].
I think this is confusing because if you when the channel 10 related to the second group of channel you specify channel_to_load=[74]

bjoern1001001:
This functionality is intended for the proxy objects and also there are problems with this pull request, so I think it should be closed.

samuelgarcia:
Ok thanks.

PR #444, opened 2017-12-20
More liberal implementation of time_slices

bjoern1001001:
@JuliaSprenger and I noticed that the expected behavior of time slices is different to the current implementation.
Thus, I made the following changes to this:
- A time slice does not need to be inside of a segments time range. Instead multiple segments are loaded if they (at least a small part of them) are in the time slice. This means that what is one segment in a file could be loaded multiple times, if more than one time slice includes it. Also if time slices overlap, some data might be loaded multiple times.
- None is  set to minimal or maximal time for t_start and t_stop respectively. So (None, 5*pq.s) loads the file from the very beginning to 5 seconds. Also (None, None) loads the whole file.
- t_start and/or t_stop may also be outside of the file, then they are treated like they are None. Thus, (float('-inf), float('inf)) will also load the whole file.
- If no segment could be loaded at all because there are no data available at this time, a ValueError is raised, indicating no data would be available.

Edit: Added Segment description to this.

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -179,6 +179,10 @@ class BaseFromRaw(BaseIO):
                 for seg_index in range(self.segment_count(block_index)):
                     seg_t_start = self.segment_t_start(block_index, seg_index) * pq.s
                     seg_t_stop = self.segment_t_stop(block_index, seg_index) * pq.s
+                    if t_start is None:
+                        t_start = seg_t_start
+                    if t_stop is None:
+                        t_stop = seg_t_stop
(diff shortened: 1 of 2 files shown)


samuelgarcia:
I think that we find agrement with @JuliaSprenger during neXmne code sprint that implementing a liberal slicing for reading could be done more easily witg proxy object and futur lazy loading stuff.

So, I propose to keep this in a corner before.

bjoern1001001:
This functionality is intended for proxy objects, so this pull request can be closed.

samuelgarcia:
Ok thanks.

Issue #440, opened 2017-12-19
Making a deep copy of a Block problems


johnpi:
I want to modify a copy of a Block and then save it to a file. So I tried making a deep copy using the copy module. Even though I can modify the copied object without problems, it fails to store the copied Block into a file. Even the following minimal test case fails:

```Python
import neo
import copy
def readSMR(filename):
        if isinstance(filename, str) and filename.endswith('.smr'):
            reader = neo.io.Spike2IO(filename=filename)
            bl = reader.read(cascade=True, lazy=False)[0]
            return bl

def saveMatlab(block, filename_mat):
        if isinstance(filename_mat, str) and filename_mat.endswith('.mat'):
            w = neo.io.NeoMatlabIO(filename=filename_mat)
            w.write_block(block)
        else:
            print('ERROR: Expected a .mat filename.')


block = readSMR('input_file.smr')
block_copy = copy.deepcopy(block)
saveMatlab(block_copy, 'output_file.mat')
```

This results to the following:

```Python
Traceback (most recent call last):
  File "minimal_test_case.py", line 20, in <module>
    saveMatlab(block_copy, 'output_file.mat')
  File "minimal_test_case.py", line 13, in saveMatlab
    w.write_block(block)
  File "/.../Python/2.7/lib/python/site-packages/neo/io/neomatlabio.py", line 255, in write_block
    ea_struct = self.create_struct_from_obj(ea)
  File "/.../Python/2.7/lib/python/site-packages/neo/io/neomatlabio.py", line 279, in create_struct_from_obj
    for i, attr in enumerate(ob._all_attrs):
AttributeError: 'Quantity' object has no attribute '_all_attrs'
```

Is there another recommended way of making a copy of the Block so that I do not modify the original one?

apdavison:
This is related to #400

apdavison:
@johnpi This should now be fixed in the master branch.

(Please let us know if you are not comfortable re-installing Neo from the Git repository - I'd be happy to give some instructions)

johnpi:
@apdavison Thanks for fixing this so quickly. Indeed, I am not familiar with what is the best way to re-install from the Git repository. Could you please give me some guidance?

apdavison:
It looks as though this is same error you are getting: https://github.com/NeuralEnsemble/python-neo/issues/466 ?

johnpi:
Yes Andrew it looks to be the same error. With Neo 0.5.2 the same files open without issues.

[...]

PR #446, opened 2018-01-02
Implements `__deepcopy__` for all data classes

### LINKED ISSUE #400: Implement deepcopy for all classes
#399 implements `__deepcopy__` for `AnalogSignal` and so fixes #220.

This should also be done for the other Neo classes.

apdavison:
Fixes #400, #440


apdavison:
The `test__deepcopy()` test in `test_block.py` should do a basic test of all classes, because in that test a Block is generated randomly containing examples of all the data classes. This test fails if you remove the `__deepcopy__` method from any of the data classes.

I agree a more in-depth test would be good, but I don't have time to write that at the moment.

HFragnaud:
ok then it is good for me

PR #452, opened 2018-01-09
[MRG] adapt coding style

massich:
This transformation has been done mostly automatically and then I read the file and changed whatever was catching my eye.

Run commands: 
```sh
sed -i 's/[[:space:]]*$//' neo/rawio/baserawio.py
docformatter --i neo/rawio/baserawio.py
yapf -i neo/rawio/baserawio.py
```
All the heavy lifting is done by `yapf`. In this case `yapf` is configured with the following `.style.yapf` file where I called `yapf`.
```sh
[style]
based_on_style: pep8
column_limit: 99
split_arguments_when_comma_terminated: true
```

### DIFF
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -40,12 +40,13 @@ or vector can be store somewhere (near the fiel, /tmp, any path)
 
 """
 
-#from __future__ import unicode_literals, print_function, division, absolute_import
-from __future__ import  print_function, division, absolute_import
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Yep.

JuliaSprenger:
I think we should do the code style adjustment using yapf (or similar) for the whole project just before releasing v0.6. I will open a separate issue for this.

PR #454, opened 2018-01-10
Implement `__getitem__` for Epoch

### LINKED ISSUE #413: Slicing epochs does only slice times not durations
When slicing an epoch only the times array is sliced and the durations array is not modified. This results in incompatible number of time stamps and durations.

apdavison:
so that durations and labels are correctly sliced. Fixes #413

### DIFF
--- a/neo/core/epoch.py
+++ b/neo/core/epoch.py
@@ -24,10 +24,10 @@ def _new_epoch(cls, times=None, durations=None, labels=None, units=None,
                name=None, description=None, file_origin=None, annotations=None, segment=None):
     '''
     A function to map epoch.__new__ to function that
-    does not do the unit checking. This is needed for pickle to work. 
+    does not do the unit checking. This is needed for pickle to work.
     '''
-    e = Epoch(times=times, durations=durations, labels=labels, units=units, name=name, file_origin=file_origin,
-              description=description, **annotations)
+    e = Epoch(times=times, durations=durations, labels=labels, units=units, name=name,
+              file_origin=file_origin, description=description, **annotations)
     e.segment = segment
     return e
 
@@ -242,10 +242,7 @@ class Epoch(BaseNeo, pq.Quantity):
             _t_stop = np.inf
 
         indices = (self >= _t_start) & (self <= _t_stop)
-
         new_epc = self[indices]
-        #new_epc.durations = self.durations[indices]
-        #new_epc.labels = self.labels[indices]
         return new_epc
 
     def as_array(self, units=None):
--- a/neo/core/epoch.py
+++ b/neo/core/epoch.py
@@ -151,6 +151,16 @@ class Epoch(BaseNeo, pq.Quantity):
                 label, time, dur in zip(labels, self.times, self.durations)]
         return '<Epoch: %s>' % ', '.join(objs)
 
+    def __getitem__(self, i):
+        '''
+        Get the item or slice :attr:`i`.
+        '''
+        obj = Epoch(times=super(Epoch, self).__getitem__(i))
+        obj._copy_data_complement(self)
+        obj.durations = self.durations[i]
+        obj.labels = self.labels[i]
+        return obj
+
     @property
     def times(self):
         return pq.Quantity(self)
@@ -234,8 +244,8 @@ class Epoch(BaseNeo, pq.Quantity):
         indices = (self >= _t_start) & (self <= _t_stop)
 
         new_epc = self[indices]
-        new_epc.durations = self.durations[indices]
-        new_epc.labels = self.labels[indices]
+        #new_epc.durations = self.durations[indices]
+        #new_epc.labels = self.labels[indices]
         return new_epc
 
     def as_array(self, units=None):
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi andrew. Could you fix conflict in the file. So it would trigered a new test on circleci ?
So we could merge.

bjoern1001001:
I have a question about this, because I have been working on `SpikeTrain.__getitem__` for the last days. It's rather general, but I noticed this here.

I noticed here that what you wrote works in Epoch for things like epoch[1], while it would raise an error in SpikeTrain, because when you create the new Epoch in `__getitem__`, times is a quantity scalar, NOT an array or Epoch. Creating a SpikeTrain like this in `__getitem__` via `SpikeTrain(times=super(...).__getitem, ...)` would raise an error here saying it's testing the length of an unsized object. Is it supposed to be okay that epochs are created with a scalar quantity instead of a quantity array? And if that is the case, shouldn't it be converted to an array then in order to ensure consistency?

A little example of what happens:

`epoch = neo.Epoch([1,2,3]*pq.s)`
`ep1 = epoch[1] # ep1 is now an Epoch with times=2*pq.s, not [2]*pq.s or array([2]*pq.s)`

This is because the following happens in `__getitem__`:
`times = super(...).__getitem__(1)` 
Times now becomes a scalar quantity, because numpy `__getitem__` returns a scalar instead of an array when called with int index. From this pq.Quantity.__getitem__ creates a quantity, because it is not yet a quantity.

`obj = Epoch(times=times, ...)`
This is not a problem, because there is no test about whether times is an array or not
 `...`
`return obj # This returns obj, with times=2*pq.s`

Should it be like this or should a test for this be introduced?

Apart from that I'd like to know if any of you has deeper knowledge of `__getitem__` because I noticed together with @JuliaSprenger that there a scalar quantity is returned instead of a SpikeTrain object. I've been trying to change this but it fails. Trying to copy what you did, I noticed that this works for Epoch, but not for SpikeTrain.

Again some code to illustrate what I mean:

What currently happens:
`st1 = new SpikeTrain(times=[1,2,3]*pq.s, ...)`
`testobj = st1[1]`
testobj is now a (scalar) quantity object, not a SpikeTrain
`print(testobject) # This prints 2 s`

`new_st = st1[0:1]`
new_st is a SpikeTrain, because numpy returns a SpikeTrain array
`print(new_st) # This prints [1] s` 



If one would do the same as you did in Epoch for SpikeTrain, it would raise an error
`st2 = st1[1] # This would raise an error`

What happens SpikeTrain `__getitem__`:

`times = super(...).__getitem__(1)`
times is now a scalar quantity just like in Epoch

`obj = SpikeTrain(times=times, ...)`
This will raise an error in `SpikeTrain.__new__` (line 220), because times is a scalar, and thus an unsized object: TypeError: len() of unsized object

On the other hand, calling things like `st1[0:1]` will work just fine, because numpy then returns an array (or a subclass, which is SpikeTrain in this case)


I then tried creating a new SpikeTrain if I get a quantity by creating a list with a single entry:

`if not isinstance(obj, SpikeTrain):`
        `obj = SpikeTrain(times=obj.magnitude, t_stop=self.t_stop, units=self.units) # Error happens here`

This works for most cases, but it fails whenever calling sorted(st1, st2), because apparently some kind of 'intermediate' SpikeTrains are generated that don't contain any attributes.
The error I get then is that SpikeTrain object [referring to `self` here] has no attribute t_stop.
If anybody knows more about this stuff, I'd really appreciate your help.

bjoern1001001:
I kept working on this problem and noticed that returning a SpikeTrain when st[int] is called would create more problems than it solves. So it seems to be fine that a scalar Quantity object is returned.

But in my opinion this should be consistent across all neo objects, which means that also here in Epoch.__getitem__ an Epoch should be returned only, when ep[slice] was called, not when ep[int] was called. Do you agree with this @apdavison @samuelgarcia?
This would also simplify the code a bit here, because no new Epoch would have to be created but instead you would only have to check if an Epoch was returned from super(...).__getitem__() and if that is the case, slice durations correctly, like it is done in SpikeTrain.__getitem__.

On a more general level, as I wrote in the previous comment, I want to ask if it makes sense to allow Epochs to be created with times being a scalar instead of an array. In class SpikeTrain there are checks that fail if times is not a list or an array, whereas in Epoch nothing similar happens. If times is a scalar, then ep.times is simply a scalar quantity. Are there reasons for this or should checks be implemented to make sure it's consistent? And if it should be allowed, wouldn't it make sense to wrap it in an array then?
Apart from that it's also not assured that ep.labels and ep.durations are the same length as ep.times. I think this should also be checked in order to ensure consistency.

apdavison:
I've rebased this onto master to remove the conflicts.

I suggest creating a separate issue for the inconsistencies noted by @bjoern1001001 with respect to spike trains, and a separate PR to fix them.

samuelgarcia:
@bjoern1001001 : could you create a sperate issue for you comment. So I could merge and close this without forgeting the issus on __getitem__ and scalar inside neo.

samuelgarcia:
I have opened an issue #490 that refer this PR so I merge.

PR #455, opened 2018-01-10
Brainvision bugfix

samuelgarcia:


neo/rawio/brainvisionrawio.py
```diff
@@ -115,6 +118,11 @@ def _parse_header(self):
         self.header['event_channels'] = event_channels
         
         self._generate_minimal_annotations()
+        for c in range(sig_channels.size):
+            coords = vhdr_header['Coordinates']['Ch{}'.format(c+1)]
+            coords = [float(v) for v in coords.split(',')]
+            if coords[0]!=0.:
```
dengemann: Maybe if coords[0] > 0: would read better.
Also add a comment why we tes this.
```# if radius is 0 we do not have coordinates.```

samuelgarcia:
This is code done during mneXneo with @dengemann to integrate neo.rawio in mnei.io.raw

Now it pass tests, I think we can merge. Never, never in my life I will mege myself my own PR again. So if someone could...

PR #460, opened 2018-01-11
Change URL for testing files.

samuelgarcia:


samuelgarcia:
@JuliaSprenger @apdavison 
I need help here.

For a reason I don't anderstand the file 'tdt/aep_05/Block-1/aep_05_Block-1.tev' is downloaded only partial from gin.g-node.org with http url.

It occurs both on my machine and circle-ci : only a small file is downloaded 6k.

But when you go to the gin web interface here the file size is OK (51.7Mo)

Could you test. test_tdtrawio.py  on your machine and tell me the size of this file ?

This problem explain the reason circle-ci is not passing because when putting manual the good file with the good size tests pass again...

JuliaSprenger:
For me the test also fails and the size of   'tdt/aep_05/Block-1/aep_05_Block-1.tev' is also only 6k.
However, the non-raw IO test work.

The reason for this phenomenon is that for files larger than a certain limit you are presented with a captcha on the website and you are only downloading this captcha, since 51MB seem to be above the threshold. You can also open the file you downloaded using a browser and see the captcha.

One way to avoid this would be to actually use gin / git-annex to download the data. Alternatively you would need to log in while downloading the files, which works but is not a nice solution since you need to put you password into the code somewhere...

samuelgarcia:
OK that make sens. But this a problem because for big files this will always fails.

TDT is maybe the last big files example we have but it is the only one tdt file that I have with both signals, spiketrain, waveforms and events.

What is the solution ? Do we skip that TDT ?

apdavison:
I think probably the best solution is to use gin or git-annex, as Julie suggests.

Logging in is possible without putting the password in the code, as CircleCI can store the password and then make it available to the script as an environment variable, but I think this is more complex, if gin can download the files without needing to be logged in.

samuelgarcia:
@apdavison @JuliaSprenger : Yes we can use gin directly but we need : to write more code for our fetcher, + make a fake user for the password + install gin on circle-ci. Since we already have lot of works for neo this year, if we could change the file size limit before have a captcha we could directly click on merge with almost no code change.

@achilleas-k: we are changing file location for neo unitest from portal.g-node.org/neo to your new gin.g-node.org. This new service is very convinient because anybody will be abable to deposite files there. We are now able to run the full test suite on circle-ci in addition to travis (kept only for fast ci on neo.core)
The only drawback is that our cocde used to download all files with the http url. gin.g-node also offer http public url but for big files this lead to a captcha. So the questions are : what are the file sie limit before having a captcha ?  Is there a way to change this limits by project ? Best.

cgars:
Hey there
Great that you use GIN!!

For the page listing a file (e.g. `https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/alphaomega/File_AlphaOmega_1.map ` the limit is ~5MB - mainly to fend off bots.

For **"raw" downloads** (e.g. `https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data/raw/master/alphaomega/File_AlphaOmega_1.map`) **there is currently no captcha**. 

I removed the captcha from raw recently, and this change is (in all the GINs) in place by now. 

Be aware, however, that we might have to "capturize" raw downloads in the future again - so **logging in via password or token would still be a good idea**- but for the time being its strictly speaking not necessary.  In any case, we will keep in touch, should the need arise.

Hope that helps.

Best

samuelgarcia:
Thank you for the answer.
The biggest we have are around 50Mo.
For instance this one:
https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data/raw/master/tdt/aep_05/Block-1/aep_05_Block-1.tev

I will try a rebuild and see if tdt test are failing because of the size.

JuliaSprenger:
@samuelgarcia: so you are suggesting a two step change, first using the 'raw' downloads for now and setting up gin + login at a later time? For the second one we should have a look here (https://github.com/G-Node/gin-cli-test/blob/master/login.sh)

samuelgarcia:
@JuliaSprenger : yes exactly. As long as the raw dowload is avaible we do not need any change in our code. So we can postpone writting a new fetcher base on gin + lgon for later.

If cricle-ci pass tests I suggest to merge this PR (wich is basically a change on URL and some docs) as soon as possible.

cgars:
BTW: Instead of setting up gin-cli (which needs to login too) you might also consider using any of the HTTP authentication methods supported (i.e. uname/password or access tokens maybe taken from  CircleCIs password storage as @apdavison suggested)

samuelgarcia:
Yeah.
tests are now passing!!!!

JuliaSprenger:
Test are also working locally on my side. I think we can merge this one, right?

PR #464, opened 2018-01-12
Remove mutated copy of `AnalogSignal.splice()`

apdavison:
Ironically, this seems to have been spliced in during a merge


samuelgarcia:
OK.

legouee:
I just looked at the code. There are two splice functions. The difference comes from the definition of signal.channel_index which creates an error in the test function test_splice_2channels_inplace. I agree to delete the first definition of splice and keeping the second, like the current pull request.

msenoville:
OK for me. I confirm that the error comes from a merge two months ago at the time of the creation of the class BaseSignal.

Issue #447, opened 2018-01-02
Error in NeuralynxRawIO parse_header()


brykko:
Hi,

I have just been trying to load some Neuralynx tetrode recordings, and I've been reliably getting an error from the `parse_header()` method with various datasets that I've tested. Every time, the error occurs when processing the first .ntt file.

Could it perhaps be down to differences in file format? I'm using Neuralynx Cheetah version 5.6.3. I'm happy to provide some sample data if that helps.
Thanks!
Rich

```python
from neo import rawio

pth = "N:\\richarga\\rec\\2015-11-24_11-58-34"
io = rawio.NeuralynxRawIO(dirname=pth)
io.parse_header()
```
Traceback:

```
N:\richarga\rec\2015-11-24_11-58-34\CSC1.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC10.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC11.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC12.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC13.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC14.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC15.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC2.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC3.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC4.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC5.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC6.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC7.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC8.ncs
N:\richarga\rec\2015-11-24_11-58-34\CSC9.ncs
N:\richarga\rec\2015-11-24_11-58-34\CheetahLogFile.txt
N:\richarga\rec\2015-11-24_11-58-34\CheetahLostADRecords.txt
N:\richarga\rec\2015-11-24_11-58-34\Events.nev
N:\richarga\rec\2015-11-24_11-58-34\TT1.ntt

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-4f2dd9ade201> in <module>()
      7 io = rawio.NeuralynxRawIO(dirname=pth)
[...]

PR #465, opened 2018-01-16
[MRG] Fix expected number of features for ntt and nse files (issue #447)

JuliaSprenger:
... and typo corrections


samuelgarcia:
Hi Julia, thank you taking care of ntt files.

Because of typos and pep8 modifications I am not able at all to see what real diffrence in the code you made.

What was wrong with ntt ?
Did you push some ntt files on gin ?

JuliaSprenger:
The main change for the ntt files was done in the first commit (https://github.com/NeuralEnsemble/python-neo/pull/465/commits/f690bb5d512b346dec728f86cdae41c1bde9b7e6). The problem was the number of features expected to be in the header. Previously this was extracted based on the info from the txt header, but this mechanism does not actually work for ntt files. I did not yet add test to cover ntt files, but will do so soon.

samuelgarcia:
OK thank you. I get it. Sorry.

Are we sure that the nb_feature is always 8 ?
Is it user parmas dependent ?

JuliaSprenger:
No, we can not be sure that nb_feature is always 8 for all files, because these are the first ntt files we are testing this on. The format specifications of revision 1.1 (5/31/2013) state that nb_feature is always equal to 8, but this might have changed for newer format versions. I guess we will solve this problem when we encounter a ntt file which does not fit the specifications we cover...

samuelgarcia:
Yep.
Good approach lets keep 8 and see when a new bug with ntt.

But could we keep somewhere in some comments theses lines:
```
# count feature
   nb_feature = 0
   for k in info.keys():
       if k.startswith('Feature '):
           nb_feature += 1
    dtype += [('features', 'int32', (nb_feature,))]
```
to have a memory in the code of this ?

JuliaSprenger:
I am not sure if this makes sense since I think this was just a guess from my side on how it should work, without ever having access to an ntt file.
Do you have an idea what's going wrong regarding the coverage error on circleci?

JuliaSprenger:
I added test for covering Neuralynx files of version 5.6.3, including the ntt format. PR for the corresponding test files is https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/1 and needs to be accepted first for these test to pass.

samuelgarcia:
This looks good. Tell me when you want to merge.

@apdavison @JuliaSprenger  : I propose to turn off pep8 for a while. And turn it on after having cleaning the whole code. Because now, patch on IOs are really hard to read, the main logic is lost in many changes that satisfy pep8.

apdavison:
ok for ignoring pep8 for now

JuliaSprenger:
I am not sure regarding pep8, since I am not convinced we will it turn on again. I rather think we should try to keep pep8 and other changes in different commits in the future (as I also did only partially in this PR), so it the diffs are still telling and we are still reminded to adjust the style.

neinreview:
I have no idea why the tests were queuing on circleci to be cancelled at the end. @samuelgarcia @apdavison can you restart the circleci tests?

JuliaSprenger:
All errors occurring with python2.7 are due to missing or empty files. Somehow some files which were recently added via PR are not available using the raw download mechanism (eg. https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/neuralynx/Cheetah_v5.6.3/plain_data/CSC1.txt). @cgars: is this expected?

For python3 only the coveralls don't work.

apdavison:
Note that the coveralls failure is probably because you haven't enabled coveralls for the INM-6 fork of python-neo

samuelgarcia:
@JuliaSprenger: you have push news file with a PR in gin (I merged) and not directly as I did.
If this make impossible the raw download with https, we have a strong problem...

Have you used gin-annex or gin command ?

JuliaSprenger:
I have used gin v0.11 and not git-annex.

achilleas-k:
The issue with the missing data is that Pull Requests on GIN don't merge annexed files. The annex metadata is merged, since this is tracked by git, but any data in the annex (which in the backend is tracked in separate branches) can't be merged as easily.

For the current data, the best solution right now would be to download the two data files and push them to the NeuralEnsemble repository directly.

For the future, to be able to get new data files via Pull Requests, it would be better to add files into git directly. If contributors are going to be using the gin client, this can be achieved by putting a config file in the root of the repository which sets the pattern for excluded files to `*`.

The following configuration forces all files to be checked into git instead of annex:
```yml
annex:
    exclude: ["*"]
```

If you don't want this, and would rather use annex and manually merge files on the rare occasion when they are very big, you could instead use the following configuration to raise the annex threshold to e.g. 1 GiB:
```yml
annex:
    minsize: 1GB
```

If you put either of these into a file called `config.yml` in the root of the `ephys_testing_data` repository, the configuration will be picked up by anyone forking the repository and using the gin client in order to push new files. Additionally, for people who don't want to use the gin client, they could be instructed to simply `git add` their data and not worry about annexing.

JuliaSprenger:
Thanks for clarifying! I think for us the best solution would be to set up a config.yml specifying a minimum filesize for annexed files to something like 100MB since the testing files should not exceed this size.
For the current problem: @samuelgarcia, @apdavison: can one of you push the 'plain_data/CSC*.txt' files directly?

samuelgarcia:
I don't understand everything except the fact that we must not use fork/merge process. Is it ?

@JuliaSprenger : Hi Julia, could you send me /ephy_testing_data/neuralynx/Cheetah_v5.6.3/plain_data/CSC*
so I can try to push them directly ? Don't you have acces directly on web.gin.g-node.org/NeuralEnsemble/ephy_testing_data ?

JuliaSprenger:
I don't think I have the rights to push to Neuralenseble/ephy_testing_data, because I only get 'Uploading... ERROR Error uploading files' when setting my remote to the Neuralenseble repository and running 'gin upload'. If I remember correctly only @apdavison and @samuelgarcia have rights to push there, correct?

Regarding the forking/pull requests: As soon as we implement the config.yml, there is no reason not to use PRs and forks, I will try to do so via PR, as soon as https://web.gin.g-node.org/G-Node/Info/issues/9 is solved.

apdavison:
Julia, I've just given write permission to "sprenger" on Neuralensemble/ephy_testing_data  - I hope that's you!

JuliaSprenger:
@apdavison: Yes, that's me. I pushed the files to gin. Now someone needs to restart the circleci build, for which again I don't have the rights :/

JuliaSprenger:
Ok, I think this is complete. Can anybody review this?

samuelgarcia:
It was already OK for me.
If nobody complain I will merge today.

Issue #463, opened 2018-01-12
Remove lazy in read_XXX() for neo.io that are derivative from neo.rawio


samuelgarcia:
I think before introduce new lazy system we should first remove lazy=True for all IOs that are not rawio based.

This could be done at the same time we remove cascade  to avoid sevral readinf of old neo.io. see #405

samuelgarcia:
WIp in #468.

samuelgarcia:
Close by #468.

Issue #405, opened 2017-10-02
Remove cascade in read_XXX()


samuelgarcia:
cascade keyword in read_block() or read_segment() or read_XXX was introduced to not load the entire tree with cascade=False.

Since:
  * this add complexity for developping an IO
  * There are few real usecase
  * Noboby use it (I huess)

I propose to remove all cascade=.. in all IOs.

samuelgarcia:
WIP in #468.

samuelgarcia:
Close by #468

PR #468, opened 2018-01-18
Clean cascade and lazy

samuelgarcia:
For smooth transition toward futur lzay mode and proxy object I am depreciating cascade for all IOs and
lazy for all of them except one based on rawio.

See #463  and #405.

Some more Ios need to be done.

This is quite fastidious I hope that I am not doing it wrong.

### DIFF
--- a/doc/source/releases/0.6.0.rst
+++ b/doc/source/releases/0.6.0.rst
@@ -1,0 +1,14 @@
+=======================
+Neo 0.6.0 release notes
+=======================
+
+DD MM 2018
+
+  * remove cascade support from all IOs
+  * remove lazy loading axcept for IOs based on rawio
+
+
+
+
+
+
--- a/examples/read_files_neo_io.py
+++ b/examples/read_files_neo_io.py
@@ -16,7 +16,7 @@ urllib.request.urlretrieve(distantfile, localfile)
 # create a reader
 reader = neo.io.PlexonIO(filename='File_plexon_3.plx')
 # read the blocks
-blks = reader.read(cascade=True, lazy=False)
+blks = reader.read(lazy=False)
 print (blks)
 # access to segments
 for blk in blks:
@@ -36,7 +36,7 @@ urllib.request.urlretrieve(distantfile, localfile)
 # create a reader
 reader = neo.io.Spike2IO(filename='File_spike2_1.smr')
 # read the block
-bl = reader.read(cascade=True, lazy=False)[0]
+bl = reader.read(lazy=False)[0]
 print (bl)
 # access to segments
 for seg in bl.segments:
--- a/examples/simple_plot_with_matplotlib.py
+++ b/examples/simple_plot_with_matplotlib.py
@@ -26,7 +26,7 @@ urllib.request.urlretrieve(distantfile, localfile)
 reader = neo.io.PlexonIO(filename='File_plexon_3.plx')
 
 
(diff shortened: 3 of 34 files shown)


samuelgarcia:
@achilleas-k: I am in a process of removing 'cascade' from neo.io API. Few people us it, there is not real use case and many IOs do not implement it well. NixIO implement. Do you mind if I remove all theses cascade in neo.io.NixIO ? Do you prefer to do it yourself ?
I am also doing the same for lazy : here the idea is to go toward proxy object and real lazy loading, this will be implemented in neo.io for class than inherits with neo.rawio. The idea is to keep the actual lazy in neo 0.6 but only for some IOs (depending on rawio) and depreciate all lazy in many IOs. Then in neo 0.7 (or more) change to the new system. It is easier to temporally depreciate lazy for many IO now and then to integrate then again. DO you mind if I also remove lazy in NixIO ?

achilleas-k:
Hey @samuelgarcia. As luck has it, I am in fact in the process of simplifying the NIXIO and I was planning on removing lazy loading while doing that.

I saw the recent issue about lazy loading being deprecated (#463) and mentioned it briefly in a meeting I had last week with @JuliaSprenger and @mdenker. I'll remove the `cascade` keyword as well in the meantime.

samuelgarcia:
Hi @achilleas-k. If you are back in NixIO and want to simplify, maybe you could gain to directly use the neo.rawio API for reading and keep th neo.io for writting. With inheritance it is easy to mix neo.rawio and neo.io. In that case cascade and lazy would naturally disappear. The neo.rawio is quite easy to implement, you just need to render numpy.array. I can help you.

achilleas-k:
Part of the plan was to use the rawio, yes, but I thought that might come later, after the simplification. I'll admit I haven't fully looked at how rawios are made, so I didn't consider doing everything (simplification, removing lazy/cascade, rawio) in the same step. If it's as straightforward as you say, I could take a stab at it while simplifying and let you know if there's anything I need help with.

Thanks.

samuelgarcia:
@achilleas-k: we need to merge this PR soon because @apdavison is cleaning pep8 everywhere. So I have remove cascade from function signature but the logic of cascade is kept inside. I let you do the clean.

samuelgarcia:
If tests pass I think we can merge this soon.
So we will be able to merge also soon #471.

achilleas-k:
Yes, that's the best option for now. Are you removing the `lazy` keyword as well?

I will be cleaning up the lazy and cascade logic in an upcoming PR. It's turning into a pretty major rewrite, but I think it will all be for the best.

samuelgarcia:
The lazy is kept in signature for all IOS (so in NixIO), but lazy=False for all IOs except the ones that hinerits rawio.

In short:
  * BaseIO.support_lazy = False
  * BaseFromRaw.support_lazy = True

Test for lazy=True are not done for BaseIO.support_lazy.
It means that for neo 0.6 the cascade disapear, the lazy is kept as before but is not supported by all IOs (less than before).

The transition towards ProxyObject will be smoother then.

samuelgarcia:
Test passes. Except the coverall for 3.6.
I guess it is OK.

achilleas-k:
Ah, I see. I understand now. Thanks for cleaning up the IO and the tests.

apdavison:
Phew, that's quite some work!

Issue #390, opened 2017-09-15
Creating annotations on AnalogSignals, Events and Epochs that relate to individual data elements ('array-valued annotations')


mdenker:
as discussed in Lausanne, details to follow

JuliaSprenger:
The idea is to have a second type of annotation, which contains an array of the same length as the actual data contained in the object. This way individual traces of an AnalogSignal can be annotated with their channel ids or additional information can be provided to individual entries of an Event or Epoch. The advantage would be that the data objects can also be sliced based on the array valued annotations (multiannotations) and filtering for specific properties of data object elements gets simpler.

samuelgarcia:
#606

PR #472, opened 2018-01-29
Array annotations

bjoern1001001:
I have so far implemented the structure needed for array annotations in neo.core. It is not complete yet. At the very least I need to make sure a few things are really correct. The IOs have not been modified at all yet, so the current changes should go unnoticed when using neo the way it is now. 

The concept is that in an AnalogSignal every signal trace has its own annotations and SpikeTrain, Event and Epoch have different annotations per timestamp, see also issue #390 . The structure of array annotations is obj.array_annotations={'key1': np.array(), 'key2': np.array()} etc. So it's a dict that maps strings  to numpy arrays.

The current changes include:

- New class DataObject, from which BaseSignal, SpikeTrain, Event and Epoch inherit
- as_array and as_quantity have been moved there instead of being implemented in all classes
- Array annotations are dealt with in every method that changes the object or returns a new one:
    - Copying the object with new data omits array annotations completele
    - When only parts of the object are returned, the array annotations are sliced accordingly
    - When sorting, the array annotations stay bound to their respective timestamp or signal trace

It's important to note that array annotations cannot be treated like normal annotations or other attributes, because they need to be changed, whenever the data are changed. So array_finalize only contains fallback functionality that adds the attribute if it is not added before (i.e. when slicing etc.).


neo/core/dataobject.py
```diff
@@ -0,0 +1,154 @@
...
+
+        index_annotations = {}
+
+        # Use what is given as an index to determine the corresponding annotations,
+        # if not possible, numpy raises an Error
+        for ann in self.array_annotations.keys():
+            index_annotations[ann] = self.array_annotations[ann][index]
+
+        return index_annotations
+
```
samuelgarcia: I am wondering here if in case of type(index)==int what we must do.
In catual implementation we get a scalar in array_annotation, waybe we should have a array of size 1 because the analogsignal.shape is (N,1) when slicing on channel dim.
So having array_annotations as a scalar is not very consistent.
@apdavison @JuliaSprenger  :What do you think ?

bjoern1001001:
As I believe this is getting close to being finished, I'm going to explain a bit more in depth what changes with this PR.

The most important aspect is of course the introduction of array_annotations. This is a new type of annotations that works analogous to the regular annotations but may only contain numpy arrays. The length of those arrays is tested to be equal to the length of the objects data, afterwards the correct length will always be ensured on every operation. It's a new structure for annotations that allows annotations to be added for every data point. So every object containing data now has a new attribute called 'array_annotations' which is a dict that maps the 'name' of the array_annotation to a numpy array (or any subclass, like pq.Quantity) containing the information. This array needs to have the same length as the data of the object. A "data point" in this context is a timestamp for SpikeTrains, Events and Epochs and one signal (NOT every single sample) for AnalogSignal and IrregularlySampledSignal. The length of the array_annotations thus needs to be the number of those data points and is the same as `len(obj)`.
Example:
```python
# On object creation
sptr = SpikeTrain(times=[1, 2, 3], array_annotations={'test': [4, 5, 6], 'abc':['a', 'b', 'c']})

# Afterwards
sptr.array_annotate(xyz=np.array(['x', 'y', 'z']))

# Print all of them
print(sptr.array_annotations)
# Result: {'test': array([4, 5, 6]), 'abc': array(['a', 'b', 'c'], dtype='<U1'), 'xyz': array(['x', 'y', 'z'], dtype='<U1')}

# Accessing:
print(sptr.array_annotations['abc'])
# Result: array(['a', 'b', 'c'], dtype='<U1')

# Slicing the object slices array annotations as well
print(sptr[0:1].array_annotations)
# Result: {'test': array([4]), 'abc': array(['a'], dtype='<U1'), 'xyz': array(['x'], dtype='<U1')}
```

This new structure is basically a generalization of 'labels' and 'durations' in Event and Epoch. So these attributes now use that generalized structure and are just shortcuts for the corresponding array_annotations. E.g. calling `event.labels = [...]` will just do `event.array_annotate(labels=[...])`.
It is important to note here that the array_annotations should only contain arrays, thus single values are wrapped into an array in any case. This also means that even if e.g. `event.times` is a scalar Quantity, `event.labels` will be an array. The same holds true for Epoch labels and durations.
(Note: This required a fix in BrainwaresrcIO which assumed them to be scalar)

There is also a possibility to create 'dummy' array_annotations by passing an empty numpy array. This will result in an array_annotation that is an array of correct length and the same dtype initialized with zeros. So e.g. caling `obj.array_annotate(test=np.array([], dtype='S'))` will create an array annotation with name 'test' that contains n empty strings where n is `len(obj)`.

What's important to know as well, is how array_annotations are handled in different situations.
Usually they remain the same when the object is changed. Of course, when slicing the object, the corresponding slice of array_annotations is used. When data consistency cannot be assured however, they are omitted. This happens in `copy_data_complement`, because the length of the data could change, and when merging, only the keys contained in both objects can be kept. In the latter case a warning will be shown, indicating what was deleted.
Array_annotations are deepcopied any time a new object is created, so basically in any circumstances that an object is changed, except if it is accessed via slicing (i.e. 'time_slice' or 'obj[...]', where ` obj.__getitem__` gets called), in which case array_annotations are passed as a view (normal numpy behavior) in order to be able to change values.
&nbsp;

The next notable change is the implementation of a `DataObject` class that all data objects inherit from. This was mostly used to create the required infrastructure for array_annotations but some other things could be generalized as well. What's important here is `rescale`. This could easily be adapted to work for all data objects, but it required `duplicate_with_new_data`. In BaseSignal this was called `duplicate_with_new_array` which was very inconsistent. So, although it was not directly related to this PR, I changed that to match the others. Note that this \*might\* cause problems in code using this functionality.
&nbsp;

Apart from that, when implementing tests, I noticed that the generated test datasets sometimes did not contain consistent data. Events and Epochs had labels (and durations) that were not as long as times. This failed due to the consistency checks for array_annotations. So now the length of those is hardcoded (5) as well, just like times.

doc/source/core.rst
```diff
@@ -251,3 +251,23 @@ limitations on the data types of annotations: they must be "simple" types or
...
+
+Array Annotations
+-----------------
+
+Next to "regular" annotations there is also a way to annotate arrays of values
+in order to create annotations with one value per data point. Using this feature,
+called Array Annotations, the consistency of those annotations with the actual data
+is ensured.
+Apart from adding those on object construction, Array Annotations can also be added
+using the :meth:`array_annotate` method possessed by all Neo data objects, e.g.::
```
JuliaSprenger: instead of 'possessed' better use 'provided' ?

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+    - duplicating with new data
+    - rescaling the object
+    - copying the object
+    - returning it as pq.Quantity or np.ndarray
+    - handling of array_annotations
+
+    Array_annotations are a kind of annotations that contain metadata for every data point,
+    i.e. per timestamp (in Spiketrain, Event and Epoch) or signal channel (in AnalogSignal
+    and IrregularlySampledSignal).
+    They may contain the same data as regular annotations, except they are arrays containing these
```
JuliaSprenger: This sentence needs clarification, do you mean something like 'Array annotations can contain the same data types as regular annotations, but are always represented as numpy arrays of the same length as the number of data points of the annotated neo object'?

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+        # If array annotation, check for correct length, only single dimension and allowed data
+        else:
+            # Get length that is required for array annotations, which is equal to the length
+            # of the object's data
+            try:
+                own_length = self._get_arr_ann_length()
+            # FIXME This is because __getitem__[int] returns a scalar Epoch/Event/SpikeTrain
+            # To be removed when __getitem__[int] is 'fixed'
+            except IndexError:
+                    own_length = 1
```
JuliaSprenger: this line is over indented

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+
+            # Escape check if empty array or list and just annotate an empty array
+            # This enables the user to easily create dummy array annotations that will be filled
+            # with data later on
+            if len(value) == 0:
+                if isinstance(value, np.ndarray):
+                    # Uninitialized array annotation containing default values (i.e. 0, '', ...)
+                    # Quantity preserves units
+                    if isinstance(value, pq.Quantity):
+                        value = np.zeros(own_length, dtype=value.dtype)*value.units
```
JuliaSprenger: Wouldn't it be better to use np.NaN here instead of zeros?

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+
+        """
+        Return dictionary of array annotations at a given index or list of indices
+        :param index: int, list, numpy array: The index (indices) from which the annotations
+                      are extracted
+        :return: dictionary of values or numpy arrays containing all array annotations
+                 for given index
+
+        Example:
+        >>> obj.array_annotate(key1=[value00, value01, value02], key2=[value10, value11, value12])
```
JuliaSprenger: same as for the other examples: be more explicit

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+        '''
+        Merges array annotations of 2 different objects.
+        The merge happens in such a way that the result fits the merged data
+        In general this means concatenating the arrays from the 2 objects.
+        If an annotation is only present in one of the objects, it will be omitted
+        :return Merged array_annotations
+        '''
+
+        # Make sure the user is notified for every object about which exact annotations are lost
+        warnings.simplefilter('always', UserWarning)
```
JuliaSprenger: Do we need to reset the warnings.filter settings at the end of this method?

neo/core/epoch.py
```diff
@@ -211,13 +218,20 @@ def merge(self, other):
         merged_annotations = merge_annotations(self.annotations,
                                                other.annotations)
         kwargs.update(merged_annotations)
+
+        kwargs['array_annotations'] = self._merge_array_annotations(other)
+        labels = kwargs['array_annotations']['labels']
```
JuliaSprenger: are you sure the keys 'labels' and 'durations' are always present here?

neo/core/epoch.py
```diff
@@ -236,13 +250,20 @@ def __deepcopy__(self, memo):
...
-        new = self.__class__(times=signal)
+
+        if units is None:
+            units = self.units
+        else:
+            units = pq.quantity.validate_dimensionality(units)
+
+        new = self.__class__(times=signal, units=units)
         new._copy_data_complement(self)
+        # Note: Array annotations can not be copied here because length of data can change
```
JuliaSprenger: This should be also mentioned in the docstring of  ```duplicate_with_new_data```

neo/core/event.py
```diff
@@ -62,6 +64,10 @@ class Event(BaseNeo, pq.Quantity):
         :description: (str) Text description.
         :file_origin: (str) Filesystem path or URL of the original data file.
 
+    *Optional attributes/properties*:
+        :array_annotations: (dict) Dict mapping strings to numpy arrays containing annotations \
```
JuliaSprenger: no '\' here

neo/core/spiketrain.py
```diff
@@ -173,6 +177,8 @@ class SpikeTrain(BaseNeo, pq.Quantity):
         :dtype: (numpy dtype or str) Override the dtype of the signal array.
         :copy: (bool) Whether to copy the times array.  True by default.
             Must be True when you request a change of units or dtype.
+        :array_annotations: (dict) Dict mapping strings to numpy arrays containing annotations \
+        for all data points
```
JuliaSprenger: indent this line

neo/core/spiketrain.py
```diff
@@ -641,6 +669,54 @@ def merge(self, other):
...
+        If an annotation is only present in one of the objects, it will be omitted.
+        Apart from that the array_annotations need to be sorted according to the sorting of
+        the spikes.
+        :return Merged array_annotations
+        '''
+
+        assert sorting is not None, "The order of the merged spikes must be known"
+
+        # Make sure the user is notified for every object about which exact annotations are lost
+        warnings.simplefilter('always', UserWarning)
```
JuliaSprenger: Same as for DataObject: Should this filter setting be reset at the end of the method?

neo/test/coretest/test_analogsignal.py
```diff
@@ -579,6 +606,37 @@ def test_splice_1channel_invalid_units(self):
...
 
+    def test_array_annotations_getitem(self):
+        data = np.arange(15).reshape(5, 3)*pq.mV
+        arr_ann1 = [10, 15, 20]
+        arr_ann2 = ['abc', 'def', 'ghi']
+        arr_anns = {'index': arr_ann1, 'label': arr_ann2}
+        signal = AnalogSignal(data, sampling_rate=30000*pq.Hz, array_annotations=arr_anns)
+        # data2 = np.arange(15)*2*pq.mV
+        # signal2 = AnalogSignal(data2, sampling_rate=30000*pq.Hz)
+        # r = signal.merge(signal2)
```
JuliaSprenger: are these lines still needed?

JuliaSprenger:
Hi @bjoern1001001!
Thanks for implementing this. I had a look at the changes and added a couple of minor comments. Besides these I noticed only one additional issue we need to discuss: The checks for the size of array annotations are only executed when using ```annotate_array()``` and not when modifying the array_annotation dictionary directly (eg via ```a.array_annotations['name'] = 1```. This might lead to inconsistencies if the user is not aware of that. Maybe it is possible to prevent the user from directly modifying the dictionary somehow? One idea would be to use properties and return a copy of the dictionary when accessing it. But this is of coure also a lot of overhead we should avoid. Do you have another idea how this could be achieved?

JuliaSprenger:
@bjoern1001001: Sorry, the latest updates in master are conflicting with this PR. Could you have a look and resolve them?

samuelgarcia:
Hi all.
thank you very much @bjoern1001001 for this.
My apoligies: I won't have time to review the code itself.
The idea good and clearly written.

I totally trust @JuliaSprenger for the review.

To answer @JuliaSprenger last suggestion : about the possible modification directtly at the *array_annatations* dict level.
I think we should not make the code too complicated for the moment an let the user be adult.

neo/core/basesignal.py
```diff
@@ -91,10 +98,16 @@ def _rescale(self, signal, units=None):
                 raise ValueError("Units must be specified")
         elif isinstance(signal, pq.Quantity):
             # could improve this test, what if units is a string?
-            if units != signal.units:
+            # This test should be safe now
+            if pq.quantity.validate_dimensionality(units) != signal.dimensionality:
```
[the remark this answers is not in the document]
bjoern1001001: You're right, it does not work for CompoundUnits causing signals with CompoundUnits to always be rescaled. This is fine because rescaling still does the right thing. So I'll for now change the comment because I don't know a better way to check this.

neo/core/basesignal.py
```diff
@@ -91,10 +98,16 @@ def _rescale(self, signal, units=None):
...
                 signal = signal.rescale(units)
         return signal
 
+    def rescale(self, units):
+        obj = super(BaseSignal, self).rescale(units)
+        obj.channel_index = self.channel_index
```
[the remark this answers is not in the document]
bjoern1001001: Segment is set in DataObject.rescale, because this needs to be done for all DataObjects anyway. ChannelIndex is more specific and not required for all DataObjects, so it needs to be treated separately.

neo/core/dataobject.py
bjoern1001001: Yes, you are right. I'll change this.

neo/core/dataobject.py
```diff
@@ -0,0 +1,328 @@
...
+        return value
+
+    def array_annotate(self, **array_annotations):
+
+        """
+        Add annotations (non-standardized metadata) as arrays to a Neo data object.
+
+        Example:
+
+        >>> obj.array_annotate(key1=[value00, value01, value02], key2=[value10, value11, value12])
```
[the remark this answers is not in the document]
bjoern1001001: Yes, I will do this.

neo/core/event.py
bjoern1001001: The '\' is masking the newline for building the documentation. If it is removed, this sentence will not be formatted correctly.

bjoern1001001: The '\' is masking the newline for building the documentation. If it is removed, this sentence will not be formatted correctly, so I think it's better to keep it here.

neo/test/coretest/test_analogsignal.py
bjoern1001001: No, I'll remove them.

bjoern1001001:
Thanks for your reply, @samuelgarcia.
Regarding the checks at dict level, I don't think it makes the code more complicated. The only thing I needed to do for this was creating my own dict subclass and using this for array_annotations. The code for this is not very pretty (due to problems with pickle), but also not too complicated. 
What you are suggesting is the more pythonic approach to this, but I believe most of the users would actually prefer to have checks performed, so mistakes won't go unnoticed.

I'm going to push this so you can have a look and afterwards we can decide whether to use it or not.

samuelgarcia:
@JuliaSprenger @mdenker @apdavison 
This is the big one for neo 0.7, maybe we should merge before the sprint next Thursday to avoid to much rebase on other PR.
What do you think ?

neo/core/dataobject.py
```diff
@@ -0,0 +1,365 @@
...
+            # Because a np.ndarray cannot contain scalars and sequences simultaneously
+            try:
+                # Perform check on first element
+                _check_single_elem(value[0])
+            except IndexError:
+                # If length of data is 0, then nothing needs to be checked
+                pass
+            return value
+
+        # In case of list, it needs to be ensured that all data are of the same type
```
JuliaSprenger: Doesn't it make more sense to first convert to an np.ndarray? Converting elements to a consistent dtype would then be already handled by this.

JuliaSprenger:
There is also still a number of pep8 issues included. Could you run autopep8 or manually fix the issues for the changed files?

neo/core/dataobject.py
```diff
@@ -0,0 +1,365 @@
...
+        Example:
+
+        >>> obj.array_annotate(code=['a', 'b', 'a'], category=[2, 1, 1])
+        >>> obj.array_annotations['code'][1]
+        'b'
+        """
+
+        self.array_annotations.update(array_annotations)
+
+    def array_annotations_at_index(self, index):
```
JuliaSprenger: I think in future versions it would make sense to also support slices as index here

apdavison:
As we are trying to get the 0.7 release out during the CodeJam, I've merged #606, which is based on this one but resolves the conflicts. Many thanks to @bjoern1001001 and @JuliaSprenger for your work on this.

Issue #473, opened 2018-01-30
Remove time_slices args in BaseFromRaw.read_block(...)


samuelgarcia:
In rawio PR #335, time_slice wa added to read_segment() and time_slices to read_block() in BaseFromRaw. The idea was to mimic a behavior present in BlackrockIO and NeuralynxIO.

See #374 for more discussion.

**time_slice** in read_segment in OK.

**time_slices** in read_block() is ambiguous because they do not target specific segment. Some IOs do not support absolut time along segment, so assign a slice to a segment imply some magic code. Since this behavior will be useless will the upcoming new lazy mode and proxy object and since release 0.6 should be soon, I propose to remove this time_slices in read_block(..) before 0.6 and not waiting the proxy branch to be included.

@mdenker  @JuliaSprenger @apdavison @bjoern1001001: are you OK ? or do you prefer a release 0.6 with this behavior ?

JuliaSprenger:
I am ok with removing time_slices in read_block before the 0.6 release as long as the blackrockio_v4 and old neuralynxio are kept until the same functionality is implemented an tested via the proxyobject mechanism.

samuelgarcia:
Done in #474.

samuelgarcia:
Close by #474

PR #474, opened 2018-01-31
Remove time_slices from read_block in BaseFromRaw

samuelgarcia:
See #473

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -72,7 +72,7 @@ class BaseFromRaw(BaseIO):
         self.parse_header()
 
     def read_block(self, block_index=0, lazy=False, signal_group_mode=None,
-                   units_group_mode=None, load_waveforms=False, time_slices=None):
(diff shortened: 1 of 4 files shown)


samuelgarcia:
ready to merge.

JuliaSprenger:
The changes look fine for me.
Regarding the idea behind the removal of the time_slice / introduction of proxy objects: read_segment still has the functionality to load only a time slice of the data. Are you planning to change this somehow when introducing proxy objects?

samuelgarcia:
At read_segment time_slice is not amiguous and not magic at all.
We can keep it (or not).
For 0.6, it it usefull.
For next version, I think that it can be usefull for having a slice of all object without worring about lazy object and slicing them. For non adavance user, it can be cool.
When proxy object will be there, we will need something like
```python
if lazy:
   assert time_slice is None, 'time slice is not possible when lazy=True, use ProxyXXX.load(time_slice=..) instead'
```
So I vote for having time_slice in read_segment now and keep for next version.

PR #481, opened 2018-02-22
Make rescaling Events and Epochs return neo objects instead of quantity

bjoern1001001:
So far rescaling Events and Epochs returned a Quantity object, not an Event or Epoch. This code overrides the functionality from class Quantity and returns a new neo object, keeping all attributes and annotations, just like it's done in SpikeTrain. 

This works quite well but it is not my preferred solution:
As I'm also working on implementing array annotations using a generic DataObject class (#472) , I tried to unify this across all those classes and implement it only once in DataObject using duplicate_with_new_data and passing the changed data. 
Unfortunately this yields two problems: 
I can't get this to work for SpikeTrain because it's not possible to pass units there and if the units are kept  the same, the SpikeTrain will rescale itself back to these units, which Event and Epoch don't. In my opinion this behavior should be made consistent across classes.
And also in BaseSignal this method is called duplicate_with_new_array which I think should be changed as well.
If there are no objections against this, I would soon try to implement this the way I had thought and make the behavior consistent across all classes.

Note: This solves the rescaling issue, but related topics like described above still need to be dicussed, e.g. in #472.


neo/core/epoch.py
```diff
@@ -151,6 +151,20 @@ def __repr__(self):
                 label, time, dur in zip(labels, self.times, self.durations)]
         return '<Epoch: %s>' % ', '.join(objs)
 
+    def rescale(self, units):
+        '''
+        Return a copy of the :class:`SpikeTrain` converted to the specified
```
apdavison: `Epoch`, not `SpikeTrain`

neo/core/event.py
```diff
@@ -142,6 +142,19 @@ def __repr__(self):
                                                                self.times)]
         return '<Event: %s>' % ', '.join(objs)
 
+    def rescale(self, units):
+        '''
+        Return a copy of the :class:`SpikeTrain` converted to the specified
```
apdavison: `Event`, not `SpikeTrain`

bjoern1001001:
Sorry, I fixed that now.

samuelgarcia:
If ready do we tag this for 0.6 ?

bjoern1001001:
Yes, you can do this.

I hope I can change the implementation a bit for 0.7, but for now this is the best solution in my opinion.

samuelgarcia:
OK. If nobody disclaim I will merge.

Issue #480, opened 2018-02-21
BlackrockIO: faulty code path hides file-not-found error.


cboulay:
Currently if you try

```Python
from neo.io import BlackrockIO
reader = BlackrockIO(filename='incorrect file name.ns6')

Traceback (most recent call last):
  File "/home/chad/miniconda3/envs/np/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 2910, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-6d8e9f1dfa9e>", line 1, in <module>
    reader = BlackrockIO(filename='nonsense')
  File "/home/chad/miniconda3/envs/np/lib/python3.5/site-packages/neo/io/blackrockio.py", line 16, in __init__
    BaseFromRaw.__init__(self, filename)
  File "/home/chad/miniconda3/envs/np/lib/python3.5/site-packages/neo/io/basefromrawio.py", line 72, in __init__
    self.parse_header()
  File "/home/chad/miniconda3/envs/np/lib/python3.5/site-packages/neo/rawio/baserawio.py", line 154, in parse_header
    self._parse_header()
  File "/home/chad/miniconda3/envs/np/lib/python3.5/site-packages/neo/rawio/blackrockrawio.py", line 366, in _parse_header
    for k, data in self.nev_data.items():
AttributeError: 'BlackrockIO' object has no attribute 'nev_data'
```

This identifies a faulty code path, in which `self.nev_data` is never assigned because the '.nev' file is not found, and then later the search for nsx files depends on the presence of `self.nev_data`. This would be a problem if you had a data directory with nsx files but no .nev file.

I think nsx loading shouldn't depend on the presence of a .nev file.

I also think a more informative error message is needed when no matching files of any extension are found.

samuelgarcia:
I agree this is a bug. This IO should work without nev files.
I will fix it.

samuelgarcia:
Close by #482.

PR #482, opened 2018-02-23
Fix BlackrockIO unable to load nsX without nev

bjoern1001001:
This should fix issue #480, @samuelgarcia.
I implemented the missing checks for the availability of nev files and also added a correct error message for cases when no Blackrock files could be found at all.
This works for Blackrock file versions 2.2 and 2.3, but unfortunately it seems that version 2.1 does indeed require a nev file to be loaded because scaling to Volts can only be done with dig_factor which is included in the nev file. I also added an error message for this case.

Apart from that I added a test for this in test_blackrockio that makes sure correct errors are raised in these cases.


samuelgarcia:
@bjoern1001001: thanks for the patch I will have a look.

For the 2.1, case maybe we could force to read without nev file and in that case we would have not units (units='') and a wrning for that. It could be usefull for instance in spike sorting on theses old files where the units is not mandatory. 
What do you think ?

bjoern1001001:
This sounds like a good idea. The 'old' BlackrockIO had a parameter called 'scaling' where you could basically choose this. So if you agree, I would try to implement it in a more general way for all Blackrock file versions.

samuelgarcia:
For the behavior I propose:
  * scaling=None by default
  * for v>=2.2 take the internal scaling and ```assert scaling is None, 'not used ffor version>2.1```
  * for v==2.1:
    * if scaling is None raise an error
    * if scaling=='' or scaling=='realscaling' load data with the

For the keyword instead of **scalling** maybe **ext_units** or **units_scale** or **signal_units** would be more expressive.

bjoern1001001:
In this case, it would not be necessary to even introduce such a parameter, one could just issue a warning as you first suggested.
In certain cases like this there could be advantages (e.g. memory consumption) to being able to keep original scaling and not even convert to float but rather keep the data as int. For this, an explicit scaling keyword would be necessary. The important question now is, if we want to offer the option to acces all data in the original int data type as present in the file. This would require a change in rescale_signal_raw_to_float() and the introduction of a scaling keyword, which could be called something like 'amplitude_scaling' or similar, because 'units' might be misleading because this term appears in many different places with different meanings (e.g. neo.units).

samuelgarcia:
I am back.
Thank for teh code.
Finaly, I am more happy with my first comments : "with 2.1 no nev then no scaling no units".

The actual code is : "with 2.1 no nev then raise error", no ?
I think reading nsx v2.1 than lost there nev v2.1 should be able to be readable.

bjoern1001001:
I agree about version 2.1. This should be possible and thus will not be able to be scaled.
I will check my code again and report back when I'm sure it does this correctly.

For the scaling parameter and thus memory usage, I will create a new issue or PR in the future.

bjoern1001001:
I updated my code so it can now load any nsX file correctly without nev. For version 2.1 it now only shows a warning that raw data will be given instead of a voltage.

samuelgarcia:
OK for me. I merge it.

Issue #412, opened 2017-10-09
`spiketrain.times - <compatible time quantity>` raises Value error


dEvasEnApati:
I get a Value error as below with neo v0.5.0:

    >>> from neo import SpikeTrain
    >>> import quantities as qu
    >>> tr = SpikeTrain([3, 4, 5] * qu.s, t_stop=10, t_start=2.5)
    >>> tr.times - 1.5 * qu.s
    Traceback....
    ValueError: The first spike ([ 1.5  2.5  3.5] s) is before t_start (2.5 s)

From [the documentation](http://neo.readthedocs.io/en/0.5.0/api_reference.html#neo.core.SpikeTrain), spiketrain.times must return a quantities array. However, it returns a neo.spiketrain object which complicates things (as is being discussed [here](https://github.com/NeuralEnsemble/python-neo/issues/369)). I think this can be fixed by casting spiketrain to a quantities object when the attribute times is read. What do you think?

lungsi:
This error also occurs with neov05.1 ( same as [elephant issue #116](https://github.com/NeuralEnsemble/elephant/issues/116) )

samuelgarcia:
I agree that tr.times should be a quantities object.

apdavison:
Fixed in #483

PR #483, opened 2018-02-26
Made SpikeTrain.times return Quantity instead of SpikeTrain

bjoern1001001:
Issue #412 states that SpikeTrain.times should return a Quantity instead of a SpikeTrain. This pull request should fix this behavior. This makes obj.times more consistent across all neo objects, as SpikeTrain was the only one with this behavior.
I also adapted the docstrings and UnitTests accordingly.

However, I'm wondering if there was a reason for having it return a SpikeTrain, because this was tested quite often in the UnitTests and also the documentation explicitly states that a SpikeTrain should be returned.

### DIFF
--- a/neo/core/spiketrain.py
+++ b/neo/core/spiketrain.py
@@ -639,9 +639,9 @@ class SpikeTrain(BaseNeo, pq.Quantity):
     @property
     def times(self):
         '''
-        Returns the :class:`SpikeTrain` without modification or copying.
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Thanks.
I don't rememeber the reason of returning a SpikeTrain.
I prefer this new behavior.
@apdavison has certainly a better memory on this.

apdavison:
I don't remember if there was a reason for returning a spike train, but I agree that returning a Quantity is better.

PR #484, opened 2018-02-27
BlackrockRawIO: Enabled reading connector_ID and connector_pinID from

bjoern1001001:
This fixes that connector_ID and connector_pinID could only be read from a nev file although the information is present in the same way in nsX files.
Apart from that it now checks for nsx_spec when reading information from nsX instead of checking nev_spec, which might be a problem when no nev files are present.


samuelgarcia:
OK.
Can I merge ?

bjoern1001001:
Yes, I think this can be merged.

Issue #493, opened 2018-03-14
improve filter function


JuliaSprenger:
Currently it is not possible to provide a targdict implying no restrictions on the objects selected. Filtering using ```targdict=None``` or ```targdict={}``` always returns an empty list even when an object type is specified.
It would be more intuitive if ```targdict={}, object=neo.Unit``` would return all objects of type neo.Unit.

PR #495, opened 2018-03-16
[MRG] Change filter behaviour to return all objects if no filter parameters are specified

JuliaSprenger:
Closes issue #493

This allows to filter for objects of a specific type without specifying another filter attribute.
Previously filters like ```targdict={}, object=neo.SpikeTrain``` returned an empty list, where as now this returns all SpikeTrain objects available.

### DIFF
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -56,27 +56,27 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
         targdict += [kwargs]
 
     if not targdict:
-        return []
+        results = data
 
     # if multiple dicts are provided, apply each filter sequentially
-    if not hasattr(targdict, 'keys'):
+    elif not hasattr(targdict, 'keys'):
         # for performance reasons, only do the object filtering on the first
         # iteration
         results = filterdata(data, targdict=targdict[0], objects=objects)
         for targ in targdict[1:]:
             results = filterdata(results, targdict=targ)
         return results
-
-    # do the actual filtering
-    results = []
-    for key, value in sorted(targdict.items()):
-        for obj in data:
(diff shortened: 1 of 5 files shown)


JuliaSprenger:
Failing tests are not related to this PR and will be fixed in PR #496, which should be merged first.

samuelgarcia:
You can re run tests now.

samuelgarcia:
I must confess that I do not use *filterdata* so I was not aware of the old behavior.

I will put some coment.

The first impression: the doc is missing a bit for the why target and/or **kargs.
And also targetdict=None meaning *no filter* should be more explicit.

But this do not come from your patch.

neo/core/container.py
```diff
@@ -56,27 +56,27 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
         targdict += [kwargs]
 
     if not targdict:
```
samuelgarcia: I prefer the more explicit:
if targdict is not None:

neo/core/container.py
```diff
@@ -56,27 +56,27 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
-        return []
+        results = data
 
     # if multiple dicts are provided, apply each filter sequentially
-    if not hasattr(targdict, 'keys'):
+    elif not hasattr(targdict, 'keys'):
```
samuelgarcia: Maybe this would be more explcit:
```elif isinstance(targdict, list):```

samuelgarcia:
Except improving a bit the doc, I think this is OK for merging.
The two coments I made are useless.

JuliaSprenger:
I extended the documentation. Ready to merge?

PR #505, opened 2018-03-22
Proposal: add channel_names/ids in AnalogSignal when channel are grouped

samuelgarcia:
Last minut proposal for 0.6: add channel_names and channel_ids in AnalogSignal when channel are grouped.

For beginner it is easier to have directly channel_names in AnalogSignal than in ChannelIndexes.

Example:
```python
reader = neo.MicromedIO(filename='myfilen.trc')
seg = reader.read_segment()
anasig = seg.analogsignals[0] # this have all channels
print(anasig.annotations['channel_names']) # this print all channels
```

This behavior will be superseded by array_annotaiont but it is cool to have then in 0.6.

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -297,10 +297,13 @@ class BaseFromRaw(BaseIO):
                             annotations['name'] = signal_channels['name'][chan_index]
                     else:
                         # when channel are grouped by same unit
-                        # annotations are empty...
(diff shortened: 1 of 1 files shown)


apdavison:
ok for me, since we're going to redo ChannelIndex soon anyway. @JuliaSprenger what do you think?

JuliaSprenger:
Also fine for me. It works, so I am going to merge.

Issue #518, opened 2018-04-10
Master broken


samuelgarcia:
@JuliaSprenger @apdavison @bjoern1001001 
The master branch do not pass tests because of a pretty print in Epoch.
I don't remember who play with the print recently ?

We have to fix it before other PR because they all fail for py36 but not for py27.
Any idea ?

```
======================================================================
FAIL: test__pretty (neo.test.coretest.test_epoch.TestEpoch)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/repo/neo/test/coretest/test_epoch.py", line 230, in test__pretty
    self.assertEqual(prepr, targ)
AssertionError: '<Epoch: test epoch 1@1.1 ms for 20.0 ns,[62 chars] ns>' != "Epoch\nname: 'test'\ndescription: 'teste[44 chars] 2]}"
- <Epoch: test epoch 1@1.1 ms for 20.0 ns, test epoch 2@1.5 ms for 40.0 ns, test epoch 3@1.7 ms for 60.0 ns>
+ Epoch
name: 'test'
description: 'tester'
annotations: {'test1': 1.1, 'test0': [1, 2]}


======================================================================
FAIL: test__pretty (neo.test.coretest.test_event.TestEvent)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/repo/neo/test/coretest/test_event.py", line 372, in test__pretty
    self.assertEqual(prepr, targ)
AssertionError: '<Event: test event 1@1.1 ms, test event [26 chars] ms>' != "Event\nname: 'test'\ndescription: 'teste[44 chars] 2]}"
- <Event: test event 1@1.1 ms, test event 2@1.5 ms, test event 3@1.7 ms>
+ Event
name: 'test'
description: 'tester'
annotations: {'test1': 1.1, 'test0': [1, 2]}


======================================================================
FAIL: test__pretty (neo.test.coretest.test_spiketrain.TestPropertiesMethods)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/circleci/repo/neo/test/coretest/test_spiketrain.py", line 1699, in test__pretty
    self.assertEqual(res, targ)
AssertionError: '<SpikeTrain(array([3., 4., 5.]) * ms, [0[13 chars]s])>' != "SpikeTrain\nname: 'train 1'\ndescription[59 chars]1.1}"
[...]

Issue #519, opened 2018-04-11
Update requirements for `nixio`


mdenker:
In a future release, the version requirement for nixio should be increased. Neo asks for the `nix.__version__` property, which seemingly does not exist in earlier versions of the library and therefore importing Neo fails. 
@achilleas-k Do you know which version of nixio should be required by Neo?

achilleas-k:
1.4.3

Given how nixio it is the current latest release and the neo-nixio depends on a recently added feature (arbitrary unit support), I see no reason to be flexible about the version requirement and have it any lower than that..

PR #520, opened 2018-04-11
Added minimum version to nixio

mdenker:
Addresses issue #519, where using a too old version of nixio will cause importing Neo to fail.


mdenker:
@achilleas-k Thanks for the input in the issue.

samuelgarcia:
You are unlucky, you touch the setup.py and so you trigger lots of pep8!!

I think we must fix #518  before this PR because the master is still broken.

achilleas-k:
Thanks for adding this. I'll make a note to keep the min version updated whenever necessary.

PR #523, opened 2018-04-19
Adapting BlackrockIO for new features of version 2.3

bjoern1001001:
Blackrock file version 2.3 enables features like a reset of the global timestamp clock or pauses during recording. This creates multiple segments in nsX data whereas in the nev files there is usually nothing that shows the beginning of a new segment. Thus, nev segments need to be mapped from the nsX or, if this is not available, segments can be distinguished, when a reset occurred and the timestamps start from 0 again. If the recording was paused and thus the timestamps were not reset, this can only be done using the nsX file.
This pull request enables the following:
- Mapping segments found in nsX to nev:
    - If the recording was paused, a new segment border is entered using the timestamp of the nsX segment
    - If a reset occurred, the nev segment can be detected by a negative difference of timestamps.
      If this is not possible, the segments cannot be distinguished and a ValueError is raised
- Splitting nev into segments even when no nsX is available. A warning is issued, that 'paused' segments cannot be detected.
- Empty segments (0 or 1 sample) that are contained in the nsX file are discarded

I will now create test files for this and write tests.


neo/rawio/blackrockrawio.py
```diff
@@ -375,23 +391,45 @@ def _parse_header(self):
                 self._seg_t_stops.append(float(t_stop))
                 self._sigs_t_starts.append(float(t_start))
 
+            # Remap segments so they are a range again
+            for i in range(self._nb_segment):
+                try:
+                    self.nsx_data[i]
+                except KeyError:
+                    self.nsx_data = {key - 1 if key > i else key: value for (key, value) in
+                                     self.nsx_data.items()}
```
samuelgarcia: With a quick overview I don't anderstand these lines. Maybe you could add more comments.
the **key -1 ** is not obvious.

bjoern1001001:
I just added tests that make sure segments are detected correctly and their start and stop times are correct. There are multiple tests with different files in order to check all situations:
- Failing, because segments could not be separated in nev
- Warning because of spikes outside of segment
- Warning because reset was not explicitly documented in nev file
- Correctly running through with no problems
I will make a pull request to upload those files to gin now.

In my opinion this PR is thus finished.

neo/rawio/blackrockrawio.py
```diff
@@ -63,6 +63,7 @@
 import os
 import re
 import warnings
+import time
```
JuliaSprenger: I don't think this import is necessary any more...

samuelgarcia:
Sharp eye.
It was to benchmark internal performence. No longer needed.

neo/rawio/blackrockrawio.py
```diff
@@ -597,6 +640,7 @@ def _get_timestamp_slice(self, timestamp, seg_index, t_start, t_stop):
         else:
             ts = np.math.ceil(t_start * self.__nev_basic_header['timestamp_resolution'])
             ind_start = np.searchsorted(timestamp, ts)
+            # SearchSorted does not apply. timestamp is not sorted!
```
JuliaSprenger: is this comment still up to date? I guess it's not, but it's not obvious to me.

neo/rawio/blackrockrawio.py
```diff
@@ -632,12 +677,14 @@ def _get_spike_raw_waveforms(self, block_index, seg_index, unit_index, t_start,
         return waveforms
 
     def _event_count(self, block_index, seg_index, event_channel_index):
+        # TODO: Should this only return NonNeural events or also Comment events ?
```
JuliaSprenger: currently only NonNeural events are counted. Is there a reason to exclude Comment events here?

neo/rawio/blackrockrawio.py
samuelgarcia: I don't remember who write this comment.
Is it me ?

neo/rawio/blackrockrawio.py
samuelgarcia: I don't remember.
I guess here both nonneural + comments should count in different event channel.

neo/rawio/blackrockrawio.py
JuliaSprenger: No, this stuff is introduced in this PR. I probably entered it in an early version of this PR.

neo/rawio/blackrockrawio.py
```diff
@@ -648,8 +695,9 @@ def _event_count(self, block_index, seg_index, event_channel_index):
 
     def _get_event_timestamps(self, block_index, seg_index, event_channel_index, t_start, t_stop):
         name = self.header['event_channels']['name'][event_channel_index]
-        events_data = self.nev_data['NonNeural']
+        events_data, event_segment_ids = self.nev_data['NonNeural']
         ev_dict = self.__nonneural_evtypes[self.__nev_spec](events_data)[name]
+        mask = ev_dict['mask'] & (event_segment_ids == seg_index)
```
JuliaSprenger: mask is never used -> can be removed here

neo/rawio/blackrockrawio.py
```diff
@@ -648,8 +695,9 @@ def _event_count(self, block_index, seg_index, event_channel_index):
...
 
         timestamp = events_data[ev_dict['mask']]['timestamp']
         labels = events_data[ev_dict['mask']][ev_dict['field']]
```
samuelgarcia: mask should be used here instead of ev_dict['mask'] for the 2 previsous lines

neo/rawio/blackrockrawio.py
```diff
@@ -1045,12 +1094,166 @@ def __read_nev_data(self, nev_data_masks, nev_data_types):
...
+        # NSX required for matching, if not available, warn the user
+        if not self._avail_nsx:
+            warnings.warn("No nsX available so it cannot be checked whether "
+                          "the segments in nev are all correct. Most importantly, "
+                          "recording pauses will not be detected", UserWarning)
+            return
+
+        # Only needs to be done for nev version 2.3
+        if self.__nev_spec == '2.3':
+            # TODO: Set earlier for 'self' and find out values!!!
```
JuliaSprenger: @bjoern1001001:  please update these todos and comments

neo/rawio/blackrockrawio.py
bjoern1001001: Yes, as far as I remember this came from you. But as it only refers to single segments, searchsorted should work anyway because spikes are sorted for each segment. So I just removed the comment.

neo/rawio/blackrockrawio.py
bjoern1001001: Yes, you are right. I fixed this.

neo/rawio/blackrockrawio.py
```diff
@@ -346,52 +352,94 @@ def _parse_header(self):
...
             self._sigs_t_starts = []
             self._seg_t_starts, self._seg_t_stops = [], []
+            nb_empty_segments = 0
             for data_bl in range(self._nb_segment):
                 length = self.nsx_data[data_bl].shape[0]
+                # Discard empty segments
+                if length < 2:
+                    nb_empty_segments += 1
+                    self.nsx_data.pop(data_bl)
+                    # TODO: CHECK WHAT HAPPENS HERE AND IF -1 IS CORRECT
```
JuliaSprenger: is this comment still valid?
I think subtracting 1 of data_bl is not the right solution since it is only providing the reset of the counter in the current run of the loop, but in the next run the index is continued as usual. Maybe it would be better to use while here instead of a for loop.

neo/test/iotest/test_blackrockio.py
```diff
@@ -334,6 +335,7 @@ def test_segment_detection_pause(self):
         # Path to nev that has spikes that don't fit nsX segment
         filename_nev_outside_seg = self.get_filename_path(
             'segment/PauseSpikesOutside/pause_spikes_outside_seg')
+        # Path to nsX and nev that are correct
```
JuliaSprenger: what exactly is 'correct' here? non-corrupted?

Issue #387, opened 2017-09-15
Cutting Blocks/Segments by Epoch


mdenker:
as discussed in Lausanne, details to follow

JuliaSprenger:
some example code exists already here: https://web.gin.g-node.org/INT/multielectrode_grasp/src/master/code/neo_utils.py

samuelgarcia:
Do we want this for 0.8 ?

samuelgarcia:
Already impleemented. If not please re-open.

PR #525, opened 2018-04-19
extract epoch based signal

lungsi:
Added two methods within the class `BaseSignal` these are: (staticmethod) `_rescale_epoch_time` and `extract_for_epoch`. Therefore when a neo object signal is created (which in turn is an instance of the child class of `BaseSignal`) it has the method `extract_for_epoch`. Passing the epochs (created neo.Epoch) into this method will return a list of extracted signals (specific to the epochs).
```
>>> import neo
>>> import quantities as pq
>>> import numpy as np
>>> sigarr = neo.AnalogSignal([[1], [2], [3], [4], [5], [6]], units='mV',
                              sampling_rate=1*pq.Hz)
>>> epc = neo.Epoch(times=np.array([0, 3000])*pq.ms, durations=[1000, 2000]*pq.ms,
                    labels=np.array(['btn0', 'btn1'], dtype='S'))
>>> epochsignals = sigarr.extract_for_epoch(epc)
>>> epochsignals
[<AnalogSignal(array([[1]]) * mV, [0.0 s, 1.0 s], sampling rate: 1.0 Hz)>,
 <AnalogSignal(array([], shape=(0, 1), dtype=int64) * mV, [3.0 s, 3.0 s],
 sampling rate: 1.0 Hz)>]
>>> len(epochsignals)
2
>>> epochsignals[0]
<AnalogSignal(array([[1]]) * mV, [0.0 s, 1.0 s], sampling rate: 1.0 Hz)>
```

### DIFF
--- a/neo/core/basesignal.py
+++ b/neo/core/basesignal.py
@@ -303,3 +303,65 @@ class BaseSignal(BaseNeo, pq.Quantity):
             signal.channel_index = ChannelIndex(index=np.arange(signal.shape[1]))
 
         return signal
+
+    @staticmethod
+    def _rescale_epoch_times(a_signal, times_of_an_epoch):
+        """
+        Checks epoch.times.units against signal.times.units
+        
+        Arguments:
+        a_signal
+        times_of_an_epoch; epc.times or epc.durations of the created epoch, epc.
+
+        Returns:
+        same times_of_an_epoch if units are the same
+        or
+        rescaled (to signal.times.units) times_of_an_epoch if units are different.
+        """
+        if times_of_an_epoch.units is not a_signal.times.units:
+            return times_of_an_epoch.rescale(a_signal.times.units)
+        else:
+            return times_of_an_epoch
+
+    def extract_for_epoch(self, epoch):
+        """
+        Checks self (which is the instance, neo.AnalogSignal or neo.SpikeTrain) for
+        specified neo.Epoch and returns signals for respetive epochs within neo.Epoch
+
+        Arguments:
+        epoch; a created neo.Epoch
+
+        Returns:
+        signal from start of an epoch to stop (start+duration).
+        This is done for respective epoch. Therefore if there are three epochs within
+        the epoch neo object (in the argument) then it returns a list of three signals.
+        But if there's only one epoch it returns just the signal (not a list).
+
+        Usage:
+           >>> import neo
+           >>> import quantities as pq
+           >>> import numpy as np
+           >>> sigarr = neo.AnalogSignal([[1], [2], [3], [4], [5], [6]], units='mV',
+                                         sampling_rate=1*pq.Hz)
+           >>> epc = neo.Epoch(times=np.array([0, 3000])*pq.ms, durations=[1000, 2000]*pq.ms,
+                               labels=np.array(['btn0', 'btn1'], dtype='S'))
+           >>> epochsignals = sigarr.extract_for_epoch(epc)
+           >>> epochsignals
+           [<AnalogSignal(array([[1]]) * mV, [0.0 s, 1.0 s], sampling rate: 1.0 Hz)>,
+            <AnalogSignal(array([], shape=(0, 1), dtype=int64) * mV, [3.0 s, 3.0 s],
+            sampling rate: 1.0 Hz)>]
+           >>> len(epochsignals)
+           2
+           >>> epochsignals[0]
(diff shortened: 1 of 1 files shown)


neo/core/basesignal.py
```diff
@@ -303,3 +303,65 @@ def merge(self, other):
...
+        """
+        if times_of_an_epoch.units is not a_signal.times.units:
+            return times_of_an_epoch.rescale(a_signal.times.units)
+        else:
+            return times_of_an_epoch
+
+    def extract_for_epoch(self, epoch):
+        """
+        Checks self (which is the instance, neo.AnalogSignal or neo.SpikeTrain) for
+        specified neo.Epoch and returns signals for respetive epochs within neo.Epoch
```
JuliaSprenger: typo: this should be 'respective'

JuliaSprenger:
Hi Lungsi!
Thanks for your contribution.
The utility function you are proposing is indeed part of a collection of functions we are missing for quite some time. However, I am not sure if this type of utility functions should be a method of the BaseSignal class or should go into a separate neo_utility function collection.
@apdavison  @samuelgarcia I think we discussed this already once, based on the neo_utils collection here (https://web.gin.g-node.org/INT/multielectrode_grasp/src/master/code/neo_utils.py), which we planned to include at some point. What is your current opinion on the topic?

If this is going to stay a BaseSignal method, I would suggest to include the code of the _rescale_epoch_time directly in the extract_for_epoch, since it does not make sense to have an epoch specific function in the BaseSignal class.

neo/core/basesignal.py
```diff
@@ -303,3 +303,65 @@ def merge(self, other):
...
+        same times_of_an_epoch if units are the same
+        or
+        rescaled (to signal.times.units) times_of_an_epoch if units are different.
+        """
+        if times_of_an_epoch.units is not a_signal.times.units:
+            return times_of_an_epoch.rescale(a_signal.times.units)
+        else:
+            return times_of_an_epoch
+
+    def extract_for_epoch(self, epoch):
```
JuliaSprenger: maybe `cut_by_epoch` or `slice_by_epoch` is more telling?

lungsi: Hi Julia,
Thanks for the comments.
The point regarding the function performing a utility is a good one. I started out thinking that way but since there were no current `utility.py` script/source fille I took the current approach which is to make signal have the method where one just had to pass the epoch. If there's going to be a script containing the utility methods I thinks its ok to have it there.

With regards to why I have `_rescale_epoch_time` and `extract_for_epoch`, not just one is because it avoids code duplication and also one has only 1-reason to change/edit either methods.

apdavison:
I believe this is an implementation of #387 ?

neo/core/basesignal.py
apdavison: I prefer `slice_by_epoch` to match `time_slice`

neo/core/basesignal.py
```diff
@@ -303,3 +303,65 @@ def merge(self, other):
...
+        Arguments:
+        a_signal
+        times_of_an_epoch; epc.times or epc.durations of the created epoch, epc.
+
+        Returns:
+        same times_of_an_epoch if units are the same
+        or
+        rescaled (to signal.times.units) times_of_an_epoch if units are different.
+        """
+        if times_of_an_epoch.units is not a_signal.times.units:
```
apdavison: shouldn't this be "!=" rather than "is not"?

neo/core/basesignal.py
apdavison: I agree with @JuliaSprenger on moving `_rescale_epoch_time` inside the function where it is used. This does not imply code duplication, since you can define a function inside a function.

Alternatively, it could be moved out of the BaseSignal definition into a utility function, as checking two Neo objects have the same units must surely be a common need.

samuelgarcia:
Dear all.
I made some coments about slicing here #526

I have implemented proxy objects here #461.
This is not merged yet, but it is almost ready (except some side effect errors and pep87).

For doing this I would prefer to write external function as proposed by @JuliaSprenger and map them to method as proposed by @apdavison.

So with proxy objects the external function could be hacked easily to work with standard/proxy objects.

My proposal is AnalogSignal.extract_for_epoch should call external neo.io.utils_slice.slice_sig_by_epoch (or whatever)

External function will also help for global slicing at segment/block level.

apdavison:
I'm postponing this to the next milestone due to lack of time to work on it.

apdavison:
closing now as this PR is very old, but we will revisit the idea

PR #527, opened 2018-04-23
Strong bug in Spike2IO.

### LINKED ISSUE #522: previously readable .smr is not readable anymore
I encountered a weird problem. I have `.smr` files which were definitely readable by `neo.io.Spike2IO` in the past (`python=2.7`, `neo=0.4`). If I attempt to read them now (`python=3.6`, `neo=0.6`) I can only get the events from the file, but no analogsignals.

The strange part is: the same files are not readable after downgrading to the old versions of neo and python, hence it might not even be directly neo related. I have a) no clue at all where to start looking around what else to downgrade, b) ask here if somebody has encountered something similar.

Thanks for help

samuelgarcia:
This should fix #522.


JuliaSprenger:
Looks fine for me.

PR #539, opened 2018-06-14
nixio doesn't support lazy loading.

werthiness:
After installing Neo 0.6.1, I tried to load a Nix file using Neo with the following lines:

```py
neo_file = NixIO(filename, "ro")
block = neo_file.read()
```

which prompted the following error:
```
  File "C:/Users/CPL/PycharmProjects/ephys_analysis/ephys_analysis/free_playground.py", line 85, in <module>
    raw_signal = get_electrode_data(h5_filename, LFP_electrode_name)
  File "C:\Users\CPL\PycharmProjects\ephys_analysis\ephys_analysis\elephant_main.py", line 21, in get_electrode_data
    block = neo_file.read()
  File "C:\python_venv\lib\site-packages\neo\io\baseio.py", line 116, in read
    return self.read_all_blocks(lazy=lazy, **kargs)
TypeError: read_all_blocks() got an unexpected keyword argument 'lazy'
```
I fixed this error by adding the lazy keyword to the `read_all_blocks()` function. While this bypasses the error, the lazy keyword is still going to raise an exception if set to `True`, because Nix does not support lazy loading.
Additionally, we changed the assert to an exception because we are value checking, and an assert is not appropriate for this.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -199,7 +199,7 @@ class NixIO(BaseIO):
         elif neoname is not None:
             for blk in self.nix_file.blocks:
                 if ("neo_name" in blk.metadata
-                    and blk.metadata["neo_name"] == neoname):
+                        and blk.metadata["neo_name"] == neoname):
                     nix_block = blk
                     break
             else:
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -156,7 +156,9 @@ class NixIO(BaseIO):
     def __exit__(self, *args):
         self.close()
 
-    def read_all_blocks(self):
+    def read_all_blocks(self, lazy=False):
+        if lazy:
+            raise Exception("Lazy loading is not supported for NixIO")
         return list(self._nix_to_neo_block(blk)
                     for blk in self.nix_file.blocks)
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Thank you for the patch.
Effectivelly, we removed lazy load for all IO that are not based on raw.
So nixio dot not support lazy. But there are some plan to modify NixIO to NixRawIO, so that it will support again in futur version.

The main goal is to implement a new lazy load API with proxy object see #461.
It more or less already working in that branch but need more work.

This futur lazy load will behave like this:

The first example is without lazy, so it consume more memory:
```
    
    lim0, lim1 = -500*pq.ms, +1500*pq.ms
    seg = reader.read_segment(lazy=False)
    triggers = seg.events[0]
    anasig = seg.analogsignals[0]  # here anasig contain the whole recording in memory
    all_sig_chunks = []
    for t in triggers.times:
        t0, t1 = (t + lim0), (t + lim1)
        anasig_chunk = anasig.time_slice(t0, t1)
        all_sig_chunks.append(anasig_chunk)
    apply_my_fancy_average(all_sig_chunks)
 
```

The second example use lazy so it consume fewer memory::
```
    lim0, lim1 = -500*pq.ms, +1500*pq.ms
    seg = reader.read_segment(lazy=True)
    triggers = seg.events[0].load(time_slice=None)  # this load all trigers in memory
    anasigproxy = seg.analogsignals[0]  # this is a proxy
    all_sig_chunks = []
    for t in triggers.times:
        t0, t1 = (t + lim0), (t + lim1)
        anasig_chunk = anasigproxy.load(time_slice=(t0, t1))  # here real data are loaded
        all_sig_chunks.append(anasig_chunk)
    apply_my_fancy_average(all_sig_chunks)

```

samuelgarcia:
Let's merge.

PR #541, opened 2018-06-21
Fix bug where wrong TypeError was caught with spike trains of length 1

lkoelman:
I got a bug when loading a previously saved Neo file in matlab format that contains spiketrains of length 1. The problem was that the try/except statement caught a TypeError in a call to len() at a deeper level than intended. I fixed it by limiting the extend of the try-clause to only the indended len() call and adding a new try/except at the other place the error was occurring.

### DIFF
--- a/neo/io/neomatlabio.py
+++ b/neo/io/neomatlabio.py
@@ -327,6 +327,12 @@ class NeoMatlabIO(BaseIO):
             if "sampling_rate" in (at[0] for at in cl._necessary_attrs):
                 # put fake value for now, put correct value later
                 data_complement["sampling_rate"] = 0 * pq.kHz
+            try:
+                len(arr)
+            except TypeError:
+                # strange scipy.io behavior: if len is 1 we get a float
+                arr = np.array(arr)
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Thanks for adding this improvement. I confirmed the original bug and that your fix is working. While testing I added a small test function as you can see here: https://github.com/lkoelman/python-neo/pull/1
Could you include this or a similar test to ensure this is will not occur in future neomatlabio versions again? And maybe fix the PEP8 issues while you are on it?
Thanks!

lkoelman:
Hi there, I fixed the PEP8 issues too, should I create a new pull request for this? Not super familiar with this process as you can see ... :)

samuelgarcia:
HI. thank you for the patch.
No need to do new pull request just commit on the same place until evrything is OK.

JuliaSprenger:
Can I merge or do you plan to fix the remaining PEP8 issues? Sorry, one of them is due to my changes is the tests...

lkoelman:
Sorry I did not have a PEP8 linter at hand at the moment so might have missed some. I think a few were also present before the changes.

PR #543, opened 2018-06-25
To fix bug 533

samuelgarcia:


JuliaSprenger:
Is this PR complete? Would it make sense to add a test covering this?

samuelgarcia:
I will fix PEP8.
I don't have idea a case for testing, it was a missintepreting of format specifications.

JuliaSprenger:
So if we have test files with different format specifications we could test this, no?

samuelgarcia:
I think we already have theses files.
We could test them against the real magnitude given by clampex but I don't have this software and it run only on windows. So the extra work for this bug is high compared to the patch.

JuliaSprenger:
Ok, then we can merge this?

samuelgarcia:
Yes, I have just push pep8 fixes.

samuelgarcia:
Maybe we can wait for @Emerson207 and @lbologna if the patch fix there magnitude problem.

lbologna:
Hello,

thanks for the fix.
I am testing it right now.
I'll be back soon with a feedback on how it behaves on our data.

lbologna:
This worked for me!
Thank you.

Emerson207:
Hello!
I tested the fix and it worked as well.
Thank you!

Issue #545, opened 2018-07-02
Inconsistency in writing and reading AnalogSignals with NixIO


JuliaSprenger:
The order of values stored in an AnalogSignal is not preserved when writing to the Nix format and reading the data again.

Here is a minimal code example

```python
from neo import AnalogSignal,NixIO, Segment, Block
import quantities as pq
import numpy as np
import os

arr = np.array([[-0.36, -0.28, -0.36, -0.28, -0.2 , -0.12, -0.04,  0.04,  0.12, 0.2 ,  0.28,  0.36, -0.2 , -0.36, -0.28, -0.2 , -0.12, -0.04, 0.36]])

anasig = AnalogSignal(arr*pq.dimensionless,t_start=0*pq.s, sampling_rate=1*pq.Hz)

bl1 = Block()
bl1.segments.append(Segment('testseg'))
bl1.segments[0].analogsignals.append(anasig)

filename='testfile.nix'
if os.path.exists(filename):
    os.remove(filename)

with NixIO(filename) as io:
    io.write(bl1)

with NixIO(filename) as io:
    bl2 = io.read_block()

print(bl.segments[0].analogsignals[0].magnitude == bl2.segments[0].analogsignals[0].magnitude)
```

JuliaSprenger:
It seems when loading the nix file the AnalogSignal values are sorted along axis=1.

achilleas-k:
I've fixed the issue and I'm adding a test based on your code. PR incoming.

THANKS!

Read below for a description of the issue and solution.

---

So this isn't exactly a value sorting issue, but it is a sorting issue at its core. At first I thought it was an issue with order tracking not being enabled at the HDF5 layer, but that's something I explicitly test for in nixpy.

There are a few things of note

### The following creates an AnalogSignal with 19 channels and one sample per channel

```python
arr = np.array([[-0.36, -0.28, -0.36, -0.28, -0.2 , -0.12, -0.04,  0.04,  0.12, 0.2 ,  0.28,  0.36, -0.2 , -0.36, -0.28, -0.2 , -0.12, -0.04, 0.36]])

anasig = AnalogSignal(arr*pq.dimensionless,t_start=0*pq.s, sampling_rate=1*pq.Hz)
```

<details>
    <summary>Expand for ipython example</summary>

```python
In [1]: from neo import AnalogSignal

In [2]: import quantities as pq

In [3]: import numpy as np

[...]

PR #546, opened 2018-07-02
NIXIO data ordering fix

achilleas-k:
Sorting creates issues with lexicographical ordering of numbers.
This PR removes all name-based sorting of DataArrays (signals) and lets the HDF5 creation order tracking take care of the signal order.


JuliaSprenger:
Thanks for the quick fix for #545!
It's tested and working.

PR #551, opened 2018-07-18
[nixio] Use Neo object names in NIX file

achilleas-k:
Last year I changed the NixIO to use UUIDs as names for objects as a way to get around a couple of issues. Issue #311 is where the discussion happened and #331 is the relevant PR where this was implemented.

Some time later, there was a [discussion on the mailing list](https://groups.google.com/d/topic/neuralensemble/u0m79CbibA0/discussion) regarding the write performance but, as a side topic, the issue of the naming also came up. In short, with the current naming scheme, the NIX/HDF5 file that is generated has names that make it difficult to work with and can be a hindrance (and an annoyance) for people who purposefully name their objects in order to be able to navigate a file.

This PR adds a new feature which somewhat backtracks from the design decision of automatically naming objects. The `write_all_blocks()` and `write_block()` methods now accept an optional, boolean argument, `use_obj_names` (false, by default). If enabled, the objects in the NIX file will use the names of the corresponding Neo objects. Before writing, however, the whole Neo object tree is checked for conflicts, based on the requirements of NIX and HDF5. If a conflict is found, a ValueError is raised and no objects are written.

I felt it was important to not check the names while writing and instead check names beforehand. This avoids potential issues with partially writing data with names that later turn out to have conflicts. It's also good for avoiding situations where a write will fail half way through the saving of a large file that could take minutes (or hours). I've clocked the name checking function and even with thousands of objects (the saving of which took over 20 mins), it only needs a few seconds to run.

Tests for the new functionality are also included.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -150,6 +150,9 @@ class NixIO(BaseIO):
         self._ref_map = dict()
         self._signal_map = dict()
 
+        # _names_ok is used to guard against name check duplication
+        self._names_ok = False
+
     def __enter__(self):
         return self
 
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
Thanks for the improvements. I tested it and it works.
I will merge this now.

PR #552, opened 2018-07-19
Fix reading blackrock digital events.

histed:
Blackrock specs say packet_insertion_reason is
1 for PARALLEL and 129 for SERIAL.  (also 64 for PERIODIC, not implemented)
i.e. bit 1 is set for parallel and bits 1 and 7 are set for serial.

Because the bitfield is just a unique ID, I removed the bit testing and use equality instead.

------------------------
Apologies: I didn't write tests for this code.  But this fixes a bug where both parallel and serial events were being returned as parallel due to the bit testing above.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -1854,18 +1854,18 @@ class BlackrockRawIO(BaseRawIO):
         a 2.3 nev file.
         """
         # digital events
+        if not np.all(np.in1d(data['packet_insertion_reason'], [1,129])):
+            raise ValueError('Unknown event codes found.')  # Blackrock spec gives reason==64 means PERIODIC, but never seen this live
         event_types = {
             'digital_input_port': {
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Thank you for the patch.
I look good.
I don't have read blacrock spec.
I let @JuliaSprenger  @bjoern1001001 or @mdenker check this.
Best

bjoern1001001:
Hi @histed,

thank you for your contribution. 

The Blackrock manual that I have available is quite ambiguous regarding this issue. 
I talked to @JuliaSprenger and we both think what you did makes more sense than the current implementation.

I would only suggest a small change: Instead of raising an error when an unknown event code is found, a warning would fit better. Then the user knows there might be more data in the file that cannot be processed, but the rest of the content remains available.

JuliaSprenger:
I agree the manual is ambiguous and @histed interpretation makes more sense. @histed: Could you change the ValueError to a warning?

PR #553, opened 2018-07-20
filter empty files while reading in neuralynxrawio.py

sin-mike:
We have multiple records, where some channels are missing, i.e., they have only the header in it. Therefore the pull request

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -89,7 +89,8 @@ class NeuralynxRawIO(BaseRawIO):
             if ext not in self.extensions:
                 continue
 
-            if os.path.getsize(filename)<=HEADER_SIZE:
+            if (os.path.getsize(filename)<=HEADER_SIZE) and \
+               (ext in ['nse', 'ncs', 'ntt']):
                 self._empty_files.append(filename)
                 continue
 
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -73,6 +73,7 @@ class NeuralynxRawIO(BaseRawIO):
         self._spike_memmap = {}
         self.internal_unit_ids = []  # channel_index > (channel_id, unit_id)
         self.internal_event_ids = []
+        self._empty_files = [] # this list contains filenames of empty records
 
         # explore the directory looking for ncs, nev, nse and ntt
         # And construct channels headers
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @sin-mike!
Thanks for your contribution! Indeed, we never had any empty files, so I never considered this case.
However, I would prefer not to print every time there is an empty file to avoid unnecessary output. Could you remove the print? If this information is important for you, I would suggest to rather track this somewhere in the io, so you can access it from your scripts.

samuelgarcia:
Hi @sin-mike! and @JuliaSprenger 

Maybe we could apply this case only to **ncs** files because an empty **nev** could be something valid or desired:

```python
 
          if os.path.getsize(filename)<=HEADER_SIZE and ext=='ncs':
              self._empty_files.append(filename)
              continue
```

Best,


Samuel

neo/rawio/neuralynxrawio.py
```diff
@@ -89,7 +89,8 @@ def _parse_header(self):
             if ext not in self.extensions:
                 continue
 
-            if os.path.getsize(filename)<=HEADER_SIZE:
+            if (os.path.getsize(filename)<=HEADER_SIZE) and \
+               (ext in ['nse', 'ncs', 'ntt']):
                 self._empty_files.append(filename)
                 continue
```
samuelgarcia: I would put only 'ncs' because emty spike channel could be also a relevant object. No ?

sin-mike: On the one hand, true, when we have a good channel with empty spikes, this is something informative, on the other hand, if we have an empty channel and an empty spike channel, this must be skipped.

sin-mike: So, should I go with just NCSs?

samuelgarcia: Lets wait for @JuliaSprenger opinion.

sin-mike: So, should I go with just NCSs?

JuliaSprenger: I agree with @samuelgarcia, an empty nse or ntt file is still important information we should provide. In principle this would even be true for ncs files, but there we just don't have enough info in the file header to be able to generate an empty AnalogSignal or similar.

sin-mike: @JuliaSprenger Done!

samuelgarcia:
It is OK for me.
Thank you very much for the patch.
I guess it is the very first commit from kazan!!
Welcome.

PR #554, opened 2018-07-25
New RawIO for NIX files

hkchekc:
This PR adds the NIXRawIO file and the test for it.

The raw reader assumes that all blocks have the same object tree, since the docstring for baserawio.py says 
>neo tree object is symetric and logical: same channel/units/event along all block and segments.

I noticed that there aren't any other rawios which support multiple blocks and we want to support multiple blocks in NIXRawIO. I think we can add support for asymmetrical blocks but this would require changes in the baserawio (as far as I can tell). Would this be useful or it is redundant for the rawios?

### DIFF
--- a/neo/io/__init__.py
+++ b/neo/io/__init__.py
@@ -128,6 +128,7 @@ from neo.io.neuralynxio_v1 import NeuralynxIO as OldNeuralynxIO
 from neo.io.neuroexplorerio import NeuroExplorerIO
 from neo.io.neuroscopeio import NeuroScopeIO
 from neo.io.nixio import NixIO
+from neo.io.nixio_fr import NixIO as NixIOFr
 from neo.io.nsdfio import NSDFIO
 from neo.io.pickleio import PickleIO
 from neo.io.plexonio import PlexonIO
--- a/neo/io/nixio_fr.py
+++ b/neo/io/nixio_fr.py
@@ -19,5 +19,6 @@ class NixIO(NIXRawIO, BaseFromRaw):
         return self
 
     def __exit__(self, *args):
+        self.header = None
         self.file.close()
-        NIXRawIO.header = None
+
--- a/neo/io/nixio_fr.py
+++ b/neo/io/nixio_fr.py
@@ -9,7 +9,7 @@ class NixIO(NIXRawIO, BaseFromRaw):
     name = 'NIX IO'
 
     _prefered_signal_group_mode = 'group-by-same-units'
-    _prefered_units_group_mode = 'all-in-one'
+    _prefered_units_group_mode = 'split-all'
 
     def __init__(self, filename):
         NIXRawIO.__init__(self, filename)
--- a/neo/io/nixio_fr.py
+++ b/neo/io/nixio_fr.py
@@ -1,7 +1,7 @@
 from neo.io.basefromrawio import BaseFromRaw
 from neo.rawio.nixrawio import NIXRawIO
 
-#This class subjects to limitations when there are multiple asymmetric blocks
+# This class subjects to limitations when there are multiple asymmetric blocks
 
 
 class NixIO(NIXRawIO, BaseFromRaw):
--- a/neo/io/nixio_fr.py
+++ b/neo/io/nixio_fr.py
@@ -14,3 +14,10 @@ class NixIO(NIXRawIO, BaseFromRaw):
     def __init__(self, filename):
         NIXRawIO.__init__(self, filename)
         BaseFromRaw.__init__(self, filename)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.file.close()
+        NIXRawIO.header = None
--- a/neo/io/nixio_fr.py
+++ b/neo/io/nixio_fr.py
@@ -1,0 +1,16 @@
+from neo.io.basefromrawio import BaseFromRaw
+from neo.rawio.nixrawio import NIXRawIO
+
+#This class subjects to limitations when there are multiple asymmetric blocks
+
+
+class NixIO(NIXRawIO, BaseFromRaw):
+
+    name = 'NIX IO'
+
+    _prefered_signal_group_mode = 'group-by-same-units'
+    _prefered_units_group_mode = 'all-in-one'
+
+    def __init__(self, filename):
+        NIXRawIO.__init__(self, filename)
+        BaseFromRaw.__init__(self, filename)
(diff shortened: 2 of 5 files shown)


samuelgarcia:
Hi @hkchekc.
thank you very much for this.

Yes at the moment no IOs support multi block except this one.
I let the possibilty of multi block with hdf5 familly in mind for this when coding rawio layer.

Yes block must have constant channel maps.
I understand that it could be annoying in some user cases.
I would prefer to keep basefromraw (the API) simple at the moment.
The change you mention would need to have a dict of header channel desciprtion by block.

Some questions:
  * Did you run some benchmark for speed ?
    The test should run "benchmark_speed_read_signals" which basically read everything
    by chunk of 1024. Do you have correct results ?
  * I am expert in hdf5. So did you check that no data block are read in parse_header ?
  * Do you plan to to rewrite neo.io.NixIO with inheritance on neo.rawio.NixrawIO ?


Best,


Samuel

hkchekc:
Hi Samuel,

Thanks for your prompt reply.

For the first question, I have run the test (thus also the "benchmark_speed_read_signals" ) and the size of chunk read is consistent with file itself.

Best,
Chek

samuelgarcia:
@achilleas-k @hkchekc thanks for answer.
I agree to keep separate at the moment.

best

JuliaSprenger:
Thanks for the PR!
Can you add a plain NixIO class, which uses the basefromrawio to generate an IO which actually creates Neo objects? To keep the old implementation in addition you can set up two nixios at the same time, similar as it is done for Neuralynx, where the old implementation is kept as a reference marked as '_v1'.

The new nixio class should look similar to this:
```python
from neo.io.basefromrawio import BaseFromRaw
from neo.rawio.nixrawio import NIXRawIO

class NixIO(NIXRawIO, BaseFromRaw):
    def __init__(self, filename):
        NIXRawIO.__init__(self, filename)
        BaseFromRaw.__init__(self, filename)
```

hkchekc:
Thanks for the comment.

The context manager functions are added in the last two commits. 

On the other hand, the IO compatibility test will take a while to be done.  I think it would be nice to create another pull request for the test later instead of leaving this PR open.

JuliaSprenger:
Hi @hkchekc! Thanks for your patience.
I agree, the compatiblity test should go into another PR.

The failing tests are probably due to a numpy issue fixed in the master branch. Could you rebase your branch, so the tests are run including the latest changes in master?

hkchekc:
@JuliaSprenger Thanks for your suggestion. The branch is now rebased.

JuliaSprenger:
Hi @hkchekc,
I did a bit of testing and found a couple of issues:
- The IO is assuming ```Spiketrain``` objects to always be attached to a ```Unit```. Neo however also allows ```Spiketrain``` objects to only be attached to a ```Segment``` without any ```ChannelIndex``` and ```Unit``` objects being used.
- The IO is assuming ```Spiketrain``` objects to  always contain waveforms.  This is also not required by the Neo structure.
- When a ```Spiketrain``` is attached to a ```Unit``` as well as a ```Segment``` this leads to an ```IndexError``` [here](https://github.com/NeuralEnsemble/python-neo/pull/554/files#diff-65316ad09005d5143f9f3d1e791c7176R160)

- When having a ```Block``` with a single ```ChannelIndex``` with an ```AnalogSignal``` attached, only the ```Block```, but not the attached ```ChannelIndex``` is loaded.
- The same structure, but in addition including a ```Segment``` can not be loaded due to a ```KeyError``` in nixpy (version 1.4.7):
```
  File ".../neo/rawio/nixrawio.py", line 55, in _parse_header
    chan_id = src.metadata['channel_id']
  File ".../nixio/section.py", line 143, in __getitem__
    prop = self.props[key]
  File ".../nixio/util/proxy_list.py", line 41, in __getitem__
    raise KeyError("The given key does not exist: " + key)
KeyError: 'The given key does not exist: channel_id'
```

hkchekc:
Hi, @JuliaSprenger 

Thanks for that. I am fixing it now, but I face some problems and hope that you can help me out.

Spiketrains can be attached to a segment but not a unit. Can the other way around happen? So can spiketrains and analogsignals be only attached to units (channelindex for AS) but not attached to the segment?

achilleas-k:
Something I'm not clear on: Is having a Neo structure without Segments valid for the RawIO? I know it's OK for the general case, but given the assumptions made for the RawIO (e.g., Segment symmetry), I wonder if this might be another case the RawIO can't handle.

hkchekc:
>     * The IO is assuming `Spiketrain` objects to always be attached to a `Unit`. Neo however also allows `Spiketrain` objects to only be attached to a `Segment` without any `ChannelIndex` and `Unit` objects being used.

The first problem is fixed and I will push it later with other changes

>     * The IO is assuming `Spiketrain` objects to  always contain waveforms.  This is also not required by the Neo structure.

The second problem is fixed and I will push it later with other changes

>     * When a `Spiketrain` is attached to a `Unit` as well as a `Segment` this leads to an `IndexError` [here](https://github.com/NeuralEnsemble/python-neo/pull/554/files#diff-65316ad09005d5143f9f3d1e791c7176R160)

It is probably the problem with the waveform and it works fine now.

>     * When having a `Block` with a single `ChannelIndex` with an `AnalogSignal` attached, only the `Block`, but not the attached `ChannelIndex` is loaded.

There was an issue when there is only one analogsignal and it is fixed now. However, I am not sure if I understand your question correctly. Did you mean you attach the AnalogSignal only to ChannelIndex and there was no Segment?

>     * The same structure, but in addition including a `Segment` can not be loaded due to a `KeyError` in nixpy (version 1.4.7):

I couldn't reproduce this error, but probably is related to the previous issue (single analogsignal problem) and maybe is fixed now?

JuliaSprenger:
@achilleas-k: Yes, you are right, this might be a RawIO limitation. @samuelgarcia is the expert here.

samuelgarcia:
Reading with classes that are RawIO based always return Segments see
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/basefromrawio.py#L183

hkchekc:
Hi! With this commit the following changes are made:
1. Analogsignals can be, but not necessarily be, attached to ChannelIndex
2. existence of waveforms are not any more necessary
3. The previous version had problem loading epoch.durations, now is fixed
4. Spiketrains can be, but not necessarily be, attached to Units and ChannelIndex
Some problems are observed with the NIXIO_fr:
1.  when ChannelIndex and units are not included in the original file, ChannelIndex and spiketrains are created by the NIXIO_fr
2. These ChannelIndex and units are automatically linked to Analogsignals and spiketrains

JuliaSprenger:
Thanks for the fixes.
I did some more tests comparing ```blocks``` written by the non-raw implementation of the nixio and read by the raw version of the nixio. These work fine for neo object structures generated by another ```rawio```.  For arbitrary neo structures, the ```ChannelIndex``` structure differs, due to the inherent structure enforced by the ```basefromrawio``` implementation. This is due to the nature of the switch to ```rawio``` implementations and therefore can't be avoided.
I suggest to add some more specific unittests with another PR.
I will merge this one now.

Issue #419, opened 2017-10-18
BlackRockIO can only load single nsX file


bjoern1001001:
In the latest BlackRockIO the parameter nsx_to_load in __init__ only accepts an int as parameter, so only a single nsX file can be loaded. In earlier versions multiple files could be read simultaneously which allowed for much higher flexibility and also comparison. This seems not to be possible any more.

samuelgarcia:
Yes this is True.

Loading all nsx with the same instance made the code over complicated.

For user it will also make things more complicated at RawIO level because sampling_rate change across channel. In that case the API is a bit more verbose because you need to precise which channel_index for everything. See neuroexplorerrawio.py.

Since parse_header is quite fast I think it is not so complicated to do this:
```
reader_30k = BlackrockRawIO(filename='..', nsx_to_load=5)
reader_30k.parse_header()
sigs_30k = reader_30k.get_analogsignal_chunk()

reader_2k = BlackrockRawIO(filename='..', nsx_to_load=5)
reader_2k.parse_header()
sigs_2k = reader_2k.get_analogsignal_chunk()
```

I hope this is not too annoying.

I guess it would not be too complicated to write code to have another neo.io.BlackrockIO that is composed by several BlackrockRawIO that could render all signal at all sampling rate.

bjoern1001001:
@JuliaSprenger and I think it would be good to have an IO that can load all signals from the specified (list or 'all') nsX files into a single neo.core.block. So if you could do this with your suggestion that would be good.

PR #555, opened 2018-08-07
BlackrockIO loading of multiple blackrock nsx files

JuliaSprenger:
This PR tackles #419 by allowing to specify nsx_to_load='all' when initialising the BlackrockIO. The implementation is probably not the most elegant one, but it comes with a test covering the new feature.

### DIFF
--- a/neo/io/blackrockio.py
+++ b/neo/io/blackrockio.py
@@ -4,17 +4,71 @@ from neo.io.basefromrawio import BaseFromRaw
 from neo.rawio.blackrockrawio import BlackrockRawIO
 
 
+def _move_channel_indexes_and_analogsignals(from_block, to_block):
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi @JuliaSprenger 
I feel shamefull because I promised to fix the situation.
It is still in my todo list!!

Indeed the best aproach would to add this at rawio level and use the group_id field of header['signal_channels'] to separate channel in groups.
Putting group_id=nsx would make it even more easy to implement.

Please keep #419 open.

best.

samuelgarcia:
héhé.
coming days no.
Let say coming month...
So lets go for this.
At least when I will be back on blackrock test will be done for nsx_to_load='all'.

JuliaSprenger:
Ok, then you can merge if you are happy with the current quick fix.

PR #560, opened 2018-08-30
np.isin -> np.in1d. Fixes #559

cboulay:

PR #561, opened 2018-08-30
Blackrock comments

cboulay:
(Note: First 2 commits are part of #560)

This is just to get the ball rolling on a modification to the Blackrock importer to include Comment events. Related to #289 

With this modification, comments are now showing up as a 3rd event channel, behind `digital_input_port` and `serial_input_port`.

I'm sure it can be made better by someone with a deeper understanding of the Blackrock importer.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -216,6 +216,10 @@ class BlackrockRawIO(BaseRawIO):
             '2.1': self.__get_nonneural_evtypes_variant_a,
             '2.2': self.__get_nonneural_evtypes_variant_a,
             '2.3': self.__get_nonneural_evtypes_variant_b}
+        self.__comment_evtypes = {
+            '2.1': self.__get_comment_evtypes_variant_a,
+            '2.2': self.__get_comment_evtypes_variant_a,
+            '2.3': self.__get_comment_evtypes_variant_a}
 
     def _parse_header(self):
 
@@ -264,10 +268,20 @@ class BlackrockRawIO(BaseRawIO):
                                           wf_offset, wf_left_sweep, wf_sampling_rate))
 
             # scan events
+            # NonNeural: serial and digital input
             events_data, event_segment_ids = self.nev_data['NonNeural']
             ev_dict = self.__nonneural_evtypes[self.__nev_spec](events_data)
             for ev_name in ev_dict:
                 event_channels.append((ev_name, '', 'event'))
+
+            if 'Comments' in self.nev_data:
+                comments_data, comments_segment_ids = self.nev_data['Comments']
+                ev_dict = self.__comment_evtypes[self.__nev_spec](comments_data)
+                for ev_name in ev_dict:
+                    event_channels.append((ev_name, '', 'event'))
+            # TODO: TrackingEvents
+            # TODO: ButtonTrigger
+            # TODO: VideoSync
 
         # Step2 NSX file
         # Load file spec and headers of available nsx files
@@ -540,6 +554,7 @@ class BlackrockRawIO(BaseRawIO):
 
             if self._avail_files['nev']:
                 ev_dict = self.__nonneural_evtypes[self.__nev_spec](events_data)
+                ev_dict.update(self.__comment_evtypes[self.__nev_spec](events_data))
                 for c in range(event_channels.size):
                     ev_ann = seg_ann['events'][c]
                     name = event_channels['name'][c]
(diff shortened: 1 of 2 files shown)


cboulay:
Rebased.

bjoern1001001:
Hi @cboulay 
Thank you for the pull request.
I looked through your code and I think it's already the cleanest solution to this issue. As comments are stored in the files like other events, it makes sense to treat them the same way.
I only noticed one minor issue in line 685 `if name in ['digital_input_port', 'serial_input_port']:`.
There might be other nonneural event names than those named here (see ll. 1804 ff.). Thus it would be better to test for the other possibilities as well in order to ensure compatibility with Blackrock file versions <=2.2.

neo/rawio/blackrockrawio.py
```diff
@@ -1870,6 +1888,17 @@ def __get_nonneural_evtypes_variant_b(self, data):
 
         return event_types
 
+    def __get_comment_evtypes_variant_a(self, data):
+        return {
+            'comments': {
+                'name': 'comments',
+                'field': 'comment',
+                'mask': np.asarray([len(_) > 0 for _ in data['comment']], dtype=bool)
```
JuliaSprenger: is there a reason to use this mask instead of directly comparing to the packet_id, which should be 65535 for comments?

JuliaSprenger:
@cboulay: could you provide a small test checking for the existence of the comment events. Maybe there are even already some blackrock files available on gin, which contain comment events.

cboulay:
Can anyone please tell me what the comments in `FileSpec2.3001.nev` are supposed to be?
I'm getting `[b'Comment2\x03\xda\x0b', b'Comment3\x92\x10\r']`. They are probably just `Comment2` and `Comment3`, but I want to make sure.

cboulay:
More on the comments...

There are multiple possible encodings. Do we want neo to return comments as bytes or as decoded strings? If it is the former, then I don't think there is any way for neo users to know what the original encoding was. I expect that this is not much of an issue because each user is unlikely to have many different datafiles each with different encodings, so they can probably be satisfied with hardcoded decoder string (e.g., 'latin_1', 'utf_16').

There's also some information about the comment colour that is lost.
[Here's the sig and docstrings for SdkSetComment in CereLink](https://github.com/dashesy/CereLink/blob/CereLink7.0/cbmex/cbsdk.cpp#L3235-L3252). This is stored in the events_data['data'] field and is in rgba format. I don't think there's any way for neo users to access this either.

samuelgarcia:
@cboulay, thank you very much for tthis.

  * Is there a reason to add a special dict ```__comment_evtypes``` for this and not add not use the existing``` ̀__nonneural_evtypes``` ?
  * About label, I think I prefer py3 str, so decoded label.

cboulay:
@samuelgarcia I'm finding blackrockrawio.py to be a little opaque. I'm totally happy to let someone else take over this PR. From my reading, I thought I needed the custom __comment_evtypes in order to name the channel properly and mask the trials properly, but it's entirely likely there are better ways to do this. For someone intimately familiar with the IO and the filespec this should be easy. For me, these small changes are taking more time than I have.

bjoern1001001:
@samuelgarcia Regarding your first point I'd say it's a lot cleaner to add this independent of nonneural events, because in the manual and in the existing parts of the implementation (ll. 1406 ff. in the current master) they are separated as well. So I would also treat them separately.

@cboulay What you noticed about the undecipherable (non-printable / non-ASCII) characters in the comments seems to originate from the fact that the comment strings "do not need to be null-terminated" according to the Blackrock manual. So because the actual length of valid content is unknown the complete length of data (data_size - 12) is read and these additional (random) characters are not deleted. Unfortunately it's hard, in fact I believe it's impossible, to determine how much of the string was supposed to actually be part of it. We cannot even be sure that 'Comment2' was meant to be 'Comment2' and not just 'Comm' (with 'ent2' just being random data), although that is highly unlikely.
Regarding the issue about decoding I don't think there is any need for the user to know or do anything about it. The file contains the encoding information, so it can be used internally without the user knowing about it.

I'm currently taking a closer look at all of this, but unfortunately I can't guarantee that I will be very quick.

samuelgarcia:
Hi all.

  * OK for me, to use a separate dict. Take the easiest and more intuitive solution for you.
  * For string decoding. In my mind, the label of  Event would be :
     * a str for neo py2 , so ascii
     * also a str for neo py3, so a decoded unincode

    But, for now in basefromrawio we use dtype='S' so always bytes.
    I was mistaken we I declare I prefer decoded str.
    Let's keep raw bytes.

    At some point we should add an option at basefromrawio to get a dtype='S' or dtype'U' depending the 
    user needs. And in case of 'U' we should add somewhere in rawio API a field that give the charset. 
    And if the charset is unknown the user will have to provide it explicitly.
  * For the color, I don't get it. Is it a per event color  ? In yes, the color would be a array_annotation. 
    No ? For that we should first 1) merge the array_annotation branch 2) add this at rawio level 3) 
    make the change for BlacrockRawIO. I propose to add a todo in the code for thos color stuff.

bjoern1001001:
@cboulay @samuelgarcia @JuliaSprenger 
Please have a look at #574. I implemented a few of the changes discussed here.

apdavison:
Replaced by #574

PR #562, opened 2018-08-30
Parse EpochSection in ABF2 files to get digital output waveform info

erikli:
Noticed that there was an EpochInfoDescription block sitting around that wasn't being used, so I followed the format of the other sections to parse an 'EpochInfo' block into the ABF2 header. On initial inspection, it seems to work properly and I can see the digital output epochs.

A caveat is that each DAC channel can define it's own epoch durations and the EpochInfo returned doesn't provide a duration field. So for example, if Epoch A is defined as 1 second in CMD#0 and as 5 seconds in CMD#1, it isn't necessarily obvious which length the digital output pulse should be if you were to turn the EpochInfo section into a waveform. (Similar to how read_raw_protocol converts the dictEpochInfoPerDAC into a waveform). Still, I think that parsing the EpochInfo block is useful.

### DIFF
--- a/neo/rawio/axonrawio.py
+++ b/neo/rawio/axonrawio.py
@@ -562,6 +562,21 @@ def parse_axon_soup(filename):
                 header['dictEpochInfoPerDAC'][DACNum][EpochNum] = \
                     EpochInfoPerDAC
 
+            # Epoch sections
+            header['EpochInfo'] = []
+            for i in range(sections['EpochSection']['llNumEntries']):
+                # read EpochInfo
+                f.seek(sections['EpochSection']['uBlockIndex'] *
+                       BLOCKSIZE + sections['EpochSection']['uBytes'] * i)
+                EpochInfo = {}
+                for key, fmt in EpochInfoDescription:
+                    val = f.read_f(fmt)
+                    if len(val) == 1:
+                        EpochInfo[key] = val[0]
+                    else:
+                        EpochInfo[key] = np.array(val)
+                header['EpochInfo'].append(EpochInfo)
+
         # date and time
         if header['fFileVersionNumber'] < 2.:
             YY = 1900


JuliaSprenger:
Thanks for your contribution @erikli!
Could you incorporate the latest changes from master by rebasing? This should fix the failing travis tests here.

erikli:
@JuliaSprenger thanks for the advice! Looks like all checks now pass.

Also just adding a note here to resolve the caveat I pointed out in my initial post for the pull request and provide some useful knowledge for anyone who might be trying to parse Axon ABF2 files. The information that you need to recapitulate the original stimulus waveform (both digital and analog) is contained in multiple places.

 - `AxonIO._axon_info['protocol']` -- things like number of samples in episode
 - `AxonIO.axon_info['section']['ADCSection']` | `AxonIO.axon_info['section']['DACSection']` -- things about the number of channels and channel properties
 - `AxonIO._axon_info['protocol']['nActiveDACChannel']` -- bitmask specifying which DACs are actually active
 - `AxonIO._axon_info['protocol']['nDigitalEnable']` -- bitmask specifying which set of Epoch timings should be used to specify the duration of digital outputs
- `AxonIO._axon_info['dictEpochInfoPerDAC']` -- dict of dict. First index is DAC channel and second index is Epoch number (i.e. information about Epoch A in Channel 2 would be in `AxonIO._axon_info['dictEpochInfoPerDAC'][2][0]`)
 - `AxonIO._axon_info['EpochInfo']` -- list of dicts containing information about each Epoch's digital out pattern. Digital out is a bitmask with least significant bit corresponding to Digital Out 0
 - `AxonIO._axon_info['listDACInfo']` -- information about DAC name, scale factor, holding level, etc
 - `AxonIO._t_starts` -- start time of each sweep in a unified time basis
 - `AxonIO._sampling_rate`

The current AxonIO.read_protocol() method utilizes a subset of these. In particular I know it doesn't consider `nDigitalEnable`, `EpochInfo`, or `nActiveDACChannel` and it doesn't account for different types of Epochs offered by Clampex/pClamp other than discrete steps (such as ramp, pulse train, etc and encoded by `nEpochType` in the EpochInfoPerDAC section). I'm currently parsing a superset of the properties used by read_protocol() in my analysis scripts, but that code still doesn't parse the full information and isn't in a state where it could be committed and I can't currently prioritize putting together all the code that would parse the full set of data. The `AxonIO._axon_info['EpochInfo']` section doesn't currently exist.

What this commit provides is parsing of the ABF2 file header to populate `AxonIO._axon_info['EpochInfo']`. With that populated, it is possible for a knowledgeable individual to parse out the relevant parts of the protocol for them.

apdavison:
@erikli The information in your previous comment is very useful. Could you put this information within the docstring for the `axonio` module or the `AxonIO` class, so that users can easily see the limitations?

erikli:
I'd be happy to but it will probably take me until sometime at the end of this week or next weekend. I will do my best to follow the guidelines in https://neo.readthedocs.io/en/0.6/developers_guide.html#working-on-the-documentation but it will definitely warrant a bit of review as I have never written documentation that would be parsed by something like sphinx autodoc.

samuelgarcia:
I am not a user of clampfit so you are more expert that me here.

As an author of AxonIO and AxonRawIO, the only thing I can say is : **read_protocol** is not garanty.
I port it to the new implentation but it certainly cover all cases correctly.

Adding more thing  to **_axon_info** info is really good.

Feel free to also make modification on **read_protocol** and/or add warning/raise for uncorvered case.

samuelgarcia:
Can we merge ?

erikli:
Life has run away with me a bit so I haven't had time to make any of the
doc changes. It should be safe to merge, but none of the information from
my comments in this thread will be incorporated into the relevant
docstrings.

Issue #556, opened 2018-08-16
Neo tests fail with NumPy 1.15.0's assert_array_equal


jpgill86:
I'm finding that several tests are failing on Travis CI, depending on the version of NumPy installed on the test server. The problem is isolated to the latest NumPy, version 1.15.0. When the version of NumPy is not forced, it varies depending on the Python version. All of the tests that fail rely on NumPy's `assert_array_equal`.

I ran several tests to demo my findings. From master, I added `stderr` print statements so that Travis CI build logs would display the version of NumPy installed (search for "`@@@@@`" to find them quickly). I then tried building with and without forcing the version of NumPy.

NumPy version NOT forced:
- Commit demoing behavior: 940c449
- Travis CI report: https://travis-ci.com/jpgill86/python-neo/builds/82022219
- Result: Python 2.7, 3.4, 3.6 pass because they install NumPy 1.12.1 (Py 3.4) or 1.13.3 (Py 2.7, 3.6) by default, whereas Python 3.5 fails because it installs NumPy 1.15.0 by default.

NumPy version forced to 1.15.0:
- Commit demoing behavior: dd54a60
- Travis CI report: https://travis-ci.com/jpgill86/python-neo/builds/82022633
- Result: All Python versions fail in the same way

NumPy version forced to 1.14.5:
- Commit demoing behavior: 6f25856
- Travis CI report: https://travis-ci.com/jpgill86/python-neo/builds/82023437
- Result: All Python versions pass without errors

I don't think this was a problem only a few days ago ([see latest master branch Travis CI report here](https://travis-ci.org/NeuralEnsemble/python-neo/builds/414047508)). My guess is that in the last few days the default NumPy version for Python 3.5 changed from something older to 1.15.0. I'm not sure who controls that (PyPI? Travis CI?). When this happens for other Python versions, more builds will fail.

jpgill86:
I just noticed another difference between the failing Python 3.5 build on Travis CI and the others ([see first report here](https://travis-ci.com/jpgill86/python-neo/builds/82022219)). The Python 3.5 build currently uses pip version 18.0, whereas the others use pip version 9.0.1.

This was not the case in the [latest master branch report](https://travis-ci.org/NeuralEnsemble/python-neo/builds/414047508) from a week ago, where the report shows that pip version 9.0.1 was used for all builds.

[...]

PR #563, opened 2018-08-31
Indexing of AnalogSignals and IrregularlySampledSignals using numpy arrays

JuliaSprenger:
This PR is based on #557. In addition to #557 it implements indexing of AnalogSignals and IrregularlySampledSignals via numpy arrays. The returned object is a quantity object, as the temporal information can not necessarily be captured by an AnalogSignal or IrregularlySampledSignal.
See also #556 for the discussion.

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -304,8 +304,19 @@ class AnalogSignal(BaseSignal):
         elif isinstance(i, slice):
             if i.start:
                 obj.t_start = self.t_start + i.start * self.sampling_period
+        elif isinstance(i, np.ndarray):
+            # Indexing of an AnalogSignal is only consistent if the resulting number of
+            # samples is the same for each trace. The time axis for these samples is not
+            # guaranteed to be continuous, so returning a Quantity instead of an AnalogSignal here.
+            new_time_dims = np.sum(i, axis=0)
+            if len(new_time_dims) and all(new_time_dims == new_time_dims[0]):
+                obj = obj.reshape(-1, self.shape[1])
+                obj = pq.Quantity(obj.magnitude, units=obj.units)
+            else:
+                raise IndexError("indexing of an AnalogSignals needs to keep the same number of "
+                                 "sample for each trace contained")
         else:
-            raise IndexError("index should be an integer, tuple or slice")
+            raise IndexError("index should be an integer, tuple, slice or boolean numpy array")
         return obj
 
     def __setitem__(self, i, value):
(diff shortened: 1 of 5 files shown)


JuliaSprenger:
@jpgill86 Is this the solution you had in mind?

jpgill86:
@JuliaSprenger, yes, this looks good!

Since indexing into a multi-trace `AnalogSignal` could in general result in different numbers of samples per trace, I think it makes sense to raise an `IndexError` when this happens. However, for the special case where the number of samples is the same for all traces, would it make sense to add a test to `test_analogsignalarray.py`?

    def test__comparison_as_indexing(self):
        assert_array_equal(self.signal1[self.signal1 < 10],
                           np.array([[0, 1, 2, 3, 4],
                                     [5, 6, 7, 8, 9]]) * pq.nA)

JuliaSprenger:
Yes, I agree. I added two more tests. More suggestions for improvement?

jpgill86:
Looks good to me! 👍

JuliaSprenger:
@samuelgarcia, @apdavison: could one of you have a look and merge if you agree to the changes?

apdavison:
As a general point, all of the formatting changes make it difficult to review the actual important code changes. I'm not at all sure that @pep8speaks is a net positive, we should review the decision to install this at some point.

More specific to the real changes, what was the argument for going with `Quantity` rather than `IrregularlySampledSignal` as the return type? Under what conditions is an `IrregularlySampledSignal` unable to capture the temporal information?

JuliaSprenger:
In the case of the ```IrregularlySampledsignal``` containing multiple traces in combination with arbitrary indexing, the resulting array contains entries related to different times, so it can't be represented as a (single) IrregularlySampledSignal.
Here is an example
```python
i = IrregularlySampledSignal([1,2,3] * pq.s, [[2,1],[1,1],[1,2]]*pq.V)
mask = i > 1
j = i[mask]
```
Currently a quantity array is returned containing the indexed entries.  If we strictly wanted to stick to neo objects, the only option would be to return multiple ```IrregularlySampledSignals```, which might run into more numpy related problems in the long run.

apdavison:
ok, I was thinking only of 1D indexes, but I guess for NumPy interoperability we need to support arbitrary indexing.

samuelgarcia:
I am back, sorry for the long silence.

Even if it is late, I am OK with this.

I also totally agree with @apdavison  that pep8 is total regression.
It is becoming very hard to understand  the real patch.
Specially in that case when few lines are involved but with a big impact.

Thanks for the work!!

PR #565, opened 2018-09-11
fix bug for bool key type in neuralynxrawio.py

sin-mike:
in header there is a key:

```
-InputInverted False
```

which is interpreted as True when you use `bool('False')`

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -497,7 +497,7 @@ def _to_bool(txt):
     elif txt=='False': 
         return False
     else: 
-        raise Exception('Connot convert %s to bool' %(txt))
+        raise Exception('Can not convert %s to bool' %(txt))
 
 # keys in
 txt_header_keys = [
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -490,6 +490,15 @@ class NeuralynxRawIO(BaseRawIO):
                     self._sigs_length.append(length)
 
 
+### Keys funcitons
+def _to_bool(txt):
+    if txt=='True': 
+        return True
+    elif txt=='False': 
+        return False
+    else: 
+        raise Exception('Connot convert %s to bool' %(txt))
+
 # keys in
 txt_header_keys = [
     ('AcqEntName', 'channel_names', None),  # used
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Oups. I am the guilty here.
Thank you for discover.

To stay more elegant I would add a func:

```python

...
    ('InputInverted', 'input_inverted', to_bool),  # used

...

    for k1, k2, type_conv in txt_header_keys:
        pattern = '-(?P<name>' + k1 + ') (?P<value>[\S ]*)'
        matches = re.findall(pattern, txt_header)
        for match in matches:
            if k2 == '':
                name = match[0]
            else:
                name = k2
            value = match[1].rstrip(' ')
            if type_conv is not None:
                value = type_conv(value)
            info[name] = value

...
def to_bool(txt):
    if txt=='True': 
        return True
    elif txt=='False': 
        return False
    else: 
        raise Exception('Connot convert %s to bool' %(txt))

```

sin-mike:
> 
> 
> Oups. I am the guilty here.
> Thank you for discover.
> 
> To stay more elegant I would add a func:
> 
> ```python
> ...
>     ('InputInverted', 'input_inverted', to_bool),  # used
> 
> ...
> 
>     for k1, k2, type_conv in txt_header_keys:
>         pattern = '-(?P<name>' + k1 + ') (?P<value>[\S ]*)'
>         matches = re.findall(pattern, txt_header)
>         for match in matches:
>             if k2 == '':
>                 name = match[0]
>             else:
>                 name = k2
>             value = match[1].rstrip(' ')
>             if type_conv is not None:
>                 value = type_conv(value)
>             info[name] = value
> 
> ...
> def to_bool(txt):
>     if txt=='True': 
>         return True
>     elif txt=='False': 
>         return False
>     else: 
>         raise Exception('Connot convert %s to bool' %(txt))
> ```

yep, fixed

neo/rawio/neuralynxrawio.py
```diff
@@ -490,6 +490,15 @@ def read_ncs_files(self, ncs_filenames):
...
 
 
+### Keys funcitons
+def _to_bool(txt):
+    if txt=='True': 
+        return True
+    elif txt=='False': 
+        return False
+    else: 
+        raise Exception('Connot convert %s to bool' %(txt))
```
JuliaSprenger: There's a typo in 'Can not'

sin-mike:
> 
> 
> OK for merging for me.

yay!

Issue #568, opened 2018-09-13
BlackrockIO tests fail with NumPy 1.15


jpgill86:
Here we go again...

Just like [last time](https://github.com/NeuralEnsemble/python-neo/issues/556), some tests are failing depending on the version of NumPy installed for the CI builds. This time, the failures are isolated to CircleCI and do not occur on Travis CI. The broken tests, found in `test_blackrockio.py`, are already merged into master; nevertheless, master's latest CircleCI tests passed a couple days ago when master was last committed to (CircleCI report: [3.6](https://circleci.com/gh/NeuralEnsemble/python-neo/687) | [2.7](https://circleci.com/gh/NeuralEnsemble/python-neo/686)). The failures are instead occurring for new builds, and they would likely occur if master was tested again.

Déjà vu...

The problem is that, in NumPy 1.15.0 and 1.15.1, [multidimensional indexing with anything but a tuple is deprecated](https://docs.scipy.org/doc/numpy/release.html#deprecations) and raises a `FutureWarning`. Build logs on CircleCI are filled with this new warning, cluttering them up. The failing tests in `test_blackrockio.py`, called `test_segment_detection_pause` and `test_segment_detection_reset`, both check that a `UserWarning` is raised under specific circumstances, but instead the `FutureWarning` is detected, causing the tests to fail.

On a [test branch](https://github.com/jpgill86/python-neo/commits/numpy-1.15-failing-tests), I've tried [forcing the NumPy version to 1.15.0](https://github.com/jpgill86/python-neo/commit/dc79c6bf35ce000d5835e8dc9132690b492e5802), and verified that the two tests failed on CircleCI (reports [3.6](https://circleci.com/gh/jpgill86/python-neo/14) | [2.7](https://circleci.com/gh/jpgill86/python-neo/13)). I then tried [forcing the NumPy version to 1.14.5](https://github.com/jpgill86/python-neo/commit/cf82dbbb7fc4358713040f41aff51a235cb8c114), and verified that all tests passed on CircleCI (reports [3.6](https://circleci.com/gh/jpgill86/python-neo/15) | [2.7](https://circleci.com/gh/jpgill86/python-neo/16)).

Strangely, all tests pass on TravisCI, even if the NumPy version is forced to 1.15.0. The logs are conspicuously absent of the `FutureWarning`s, so I can only speculate that TravisCI suppresses this warning, preventing it from interfering with the tests.

[...]

PR #569, opened 2018-09-14
Fixed detection of warnings in test_blackrockio.py (#568)

bjoern1001001:
See #568 
Now the position of the warning in the list of warnings is not assumed anymore. 
Instead the test checks if the warning is in the list of warnings.

Apart from that the warnings.simplefilter filter should only be changed within
the context of catch_warnings here in order to ensure it being set to its original
state afterwards. Alternatively it needs to be reset manually. This was done as well.


JuliaSprenger:
The other failing tests are due to some more changes in numpy and will be incorporated here if PR https://github.com/bjoern1001001/python-neo/pull/1 is accepted by @bjoern1001001.

bjoern1001001:
Thank you @JuliaSprenger, I had not noticed that. 
Seems like it's working now.

PR #570, opened 2018-09-15
Blackrock - separate dig input and serial input.

cboulay:
Prior to this merge, all serial input events are also showing up as digital input events. But maybe that's desired?

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -1772,7 +1772,8 @@ class BlackrockRawIO(BaseRawIO):
             'digital_input_port': {
                 'name': 'digital_input_port',
                 'field': 'digital_input',
-                'mask': self.__is_set(data['packet_insertion_reason'], 0),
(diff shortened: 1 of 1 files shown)


cboulay:
Ah, I'm going crazy. Too many rebases and merges. I think this was already changed in another commit that was not in my fork but was upstream. Hold off for now please...

cboulay:
I've partially figured out why I was so confused.

`__get_nonneural_evtypes_variant_b` was fixed very recently.
`__get_nonneural_evtypes_variant_a` still had the bug, if it is indeed a bug in that file format.

I'm testing a file recorded on an NSP with firmware 7.0.4 (most recent), filespec 2.3, yet for some reason it was triggering variant_a and thus this bug. However, after a recent rebase, I'm no longer triggering variant_a and it is correctly using variant_b, which already has this bug fixed.

Unfortunately, I don't have the ability to record filespec 2.1 or 2.2, and the file used for the unittests does not have digital or serial events, so I can't possibly test this.

PR #574, opened 2018-09-20
Blackrock comments

bjoern1001001:
After having a close look at @cboulay 's code from #561 I could not find a solution that I think fits better at the moment. I tried a few things, like wrapping __nonneural_evtypes and __comment_evtypes into a common method that works for both (see commit c18475d), but I noticed this creates too much unnecessary code. Thus for now I would prefer not to do this. Once somebody implements another event type, however, I think a bit of restructuring will be justified.

The original code has not changed much but I introduced a few changes as discussed in #561.

- Strings are decoded according to the charset and reencoded to bytes that are human readable.
- Comment colors are parsed to 8-digit hex values as read from the file. If anybody (@cboulay ?) has more information on how they are encoded, please let me know. These hex strings are then placed into an array that is presented to the user as an annotation to the comment event object.
- I also renamed __nonneural_evtypes to __nonneural_evdicts aand __comment_evtypes to __comment_evdict.

I might, of course, have missed something, so if anybody sees a better way of implementing this, please also let me know.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -553,12 +553,15 @@ class BlackrockRawIO(BaseRawIO):
                 ev_dict = self.__nonneural_evtypes[self.__nev_spec](events_data)
                 if 'Comments' in self.nev_data:
                     ev_dict.update(self.__comment_evtypes[self.__nev_spec](comments_data))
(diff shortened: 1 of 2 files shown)


samuelgarcia:
THis looks good to me.
thanks.

Issue #575, opened 2018-09-21
add more tests for raw NixIO


JuliaSprenger:
The RawIO implementation of the NixIO (from #554) could use some write-read tests to test for consistent neo structure and data content using the non-raw NixIO for writing. I suggest to use a neo structure generated by one of the other rawios (e.g. ```BlackRockIO```) in the unittest. This would imply that the NixIO test would have a direct dependency on one of the other IOs.

JuliaSprenger:
here is a some code as starting point for writing the tests

```python
   def test_check_same_neo_structure(self):
        self.assertEqual(len(self.bl1.segments), len(self.bl2.segments))
        self.assertEqual(len(self.bl1.channel_indexes), len(self.bl2.channel_indexes))
        for seg1, seg2 in zip(self.bl1.segments, self.bl2.segments):
            self.assertEqual(len(seg1.analogsignals), len(seg2.analogsignals))
            self.assertEqual(len(seg1.spiketrains), len(seg2.spiketrains))

        for chid1, chid2 in zip(self.bl1.channel_indexes, self.bl2.channel_indexes):
            self.assertEqual(len(chid1.units), len(chid2.units))
            self.assertEqual(len(chid1.analogsignals), len(chid2.analogsignals))
            for unit1, unit2 in zip(chid1.units, chid2.units):
                self.assertEqual(len(unit1.spiketrains), len(unit2.spiketrains))

    def test_check_same_data_content(self):
        for seg1, seg2 in zip(self.bl1.segments, self.bl2.segments):
            for asig1, asig2 in zip(seg1.analogsignals, seg2.analogsignals):
                np.testing.assert_array_equal(asig1.magnitude, asig2.magnitude)
            for st1, st2 in zip(seg1.spiketrains, seg2.spiketrains):
                np.testing.assert_array_equal(st1.magnitude, st2.times)

        for chid1, chid2 in zip(self.bl1.channel_indexes, self.bl2.channel_indexes):
            for asig1, asig2 in zip(chid1.analogsignals, chid2.analogsignals):
                np.testing.assert_array_equal(asig1.magnitude, asig2.magnitude)

            for st1, st2 in zip(seg1.spiketrains, seg2.spiketrains):
                np.testing.assert_array_equal(st1.magnitude, st2.times)
```

PR #585, opened 2018-10-19
create basic and compatability test for nixio_fr

hkchekc:
in Reference to Issue #575 , some basic test for nixio_fr and compatability test of nixio and nixio_fr is added.

### DIFF
--- a/neo/test/iotest/test_nixio_fr.py
+++ b/neo/test/iotest/test_nixio_fr.py
@@ -28,7 +28,7 @@ class TestNixfr(BaseTestIO, unittest.TestCase, ):
 
     def setUp(self):
         super(TestNixfr, self).setUp()
-        self.testfilename = 'nixio_fr.nix'
+        self.testfilename = self.get_filename_path('nixio_fr.nix')
         self.reader_fr = NixIOfr(filename=self.testfilename)
         self.reader_norm = NixIO(filename=self.testfilename, mode='ro')
         self.blk = self.reader_fr.read_block(block_index=1, load_waveforms=True)
--- a/neo/test/iotest/test_nixio_fr.py
+++ b/neo/test/iotest/test_nixio_fr.py
@@ -19,7 +19,7 @@ class TestNixfr(unittest.TestCase):
     files_to_download = ['nixio_fr.nix']
 
     def setUp(self):
-        self.testfilename = 'nixio_fr.nix'
+        self.testfilename = self.get_filename_path('nixio_fr.nix')
         self.reader_fr = NixIOfr(filename=self.testfilename)
         self.reader_norm = NixIO(filename=self.testfilename, mode='ro')
         self.blk = self.reader_fr.read_block(block_index=1, load_waveforms=True)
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
Hi @hkchekc! Thanks for adding some more tests!
Currently the new tests are failing because of an incomplete path to the test files. You can get the correct path by using `self.get_filename_path(filename)`.

JuliaSprenger:
And you need to inherit from `BaseTestIO` to have this function available. You can compare to https://github.com/NeuralEnsemble/python-neo/blob/master/neo/test/iotest/test_blackrockio.py as an example implementation.

hkchekc:
One test will fail now due to a bug which will be fixed in the PR #581

neo/test/iotest/test_nixio_fr.py
```diff
@@ -0,0 +1,106 @@
...
+class TestNixfr(BaseTestIO, unittest.TestCase, ):
+    ioclass = NixIOfr
+
+    files_to_test = ['nixio_fr.nix']
+
+    files_to_download = ['nixio_fr.nix']
+
+    def setUp(self):
+        super(TestNixfr, self).setUp()
+        self.testfilename = 'nixio_fr.nix'
```
JuliaSprenger: this line needs to be changed to get the correct file path:
```suggestion
        self.testfilename = 'self.get_filename_path('nixio_fr.nix')
```

JuliaSprenger:
@hkchekc: You will also need to adjust the conversion of nix attributes to regular python objects, because nix objects are not supported by neo eg. as annotations.

hkchekc:
We upload a new nix file for test on Gin.

JuliaSprenger:
I merged the new file and rerun the tests. However, they are still failing somewhere within nix when parsing the header. Do you have an idea how to fix this?

JuliaSprenger:
I merged the new file and rerun the tests. However, they are still failing somewhere within nix when parsing the header. Do you have an idea how to fix this?

JuliaSprenger:
@hkchekc: Thanks for fixing the CI tests.
I still get an error when running the tests locally with numpy 1.15.1 and nixio 1.5.0b2

```python
Failure
  File .../python-neo/neo/rawio/tests/common_rawio_test.py", line 151, in test_read_all
    compliance.read_analogsignals(reader)
  File ".../python-neo/neo/rawio/tests/rawio_compliance.py", line 144, in read_analogsignals
    assert raw_chunk.ndim == 2
AssertionError
```

It seems that `raw_chunk.ndim ==1` in at least one case here. Do you have an idea why this might be the case?

hkchekc:
@JuliaSprenger  I try to use numpy 1.15.1 and nixio 1.5.0b2 to run the test_nixio_fr.py and test_nixrawio.py for both 'nixrawio-1.5.nix' and ' nixio_fr.nix' and could not reproduce the error. 

Can you tell me more details?

JuliaSprenger:
Hi @hkchekc,
I had a closer look and think that the error was somehow related to an improper setup of my environment. If I create a new environment everything works smoothly. Sorry for the confusion.
I will merge this now.

hkchekc:
Thanks for all the checkings! @JuliaSprenger

Issue #125, opened 2014-02-03
Codejam #6 : Automatic relationship handling


samuelgarcia:
The actual implementation use list for relation ship.
This is easy to code but not efficient when the number of object is high in models (PyNN, mozaik)

Todd propose NeoContainer that would handle automatic relation ship.

Some are thinking of some kind of indexes but nothing clear at the moment.

here the place to discuss.

samuelgarcia:
done

PR #588, opened 2018-10-23
[WIP] Automatic handling of relationships between objects

apdavison:
As discussed in #125, the parent-child relationships between objects are currently handled manually, e.g. when you add a `SpikeTrain` to the `segment.spiketrains` list, the reverse relationship `spiketrain.segment` is not automatically created, leading to potential inconsistencies.

The aim of this PR, when complete, is to automatically create these relationships by replacing the various child lists (`Segment.spiketrains`, `Segment.analogsignals`, `Block.segments`, etc) by pseudo-list objects, and the reverse relationships `Segment.block`, etc., by properties).

The pseudo-list classes also provide the opportunity for additional logic and optimisations, for example, a multiplexed representation of spike trains (all times in a single array, with a second array indicating which neuron/channel the spike is from) which will be much more efficient for data coming from simulators.

- [x] SpikeTrainList class
- [ ] AnalogSignalList class
- [ ] EventList and EpochList
- [ ] `child.parent` properties
- [ ] ChannelIndexList (or whatever ChannelIndex turns into)

Comments welcome!

### DIFF
--- a/neo/core/spiketrainlist.py
+++ b/neo/core/spiketrainlist.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+"""
+This module implements :class:`SpikeTrainList`, a pseudo-list
+which takes care of relationships between Neo parent-child objects.
+
+In addition, it supports a multiplexed representation of spike trains
+(all times in a single array, with a second array indicating which
+neuron/channel the spike is from).
+"""
+
+import numpy as np
+from .spiketrain import SpikeTrain
+
+
+class SpikeTrainList(object):
+    """
+    docstring needed
+    """
+
+    def __init__(self, items=None, segment=None):
+        """Initialize self"""
+        self._items = items
+        self._spike_time_array = None
+        self._channel_id_array = None
+        self._all_channel_ids = None
+        self._spiketrain_metadata = None
+        self.segment = segment
+
+    def __iter__(self):
+        """Implement iter(self)"""
+        if self._items is None:
+            self._spiketrains_from_array()
+        for item in self._items:
+            yield item
+
+    def __getitem__(self, i):
+        """x.__getitem__(y) <==> x[y]"""
+        if self._items is None:
+            self._spiketrains_from_array()
+        return self._items[i]
+
+    def __str__(self):
+        """Return str(self)"""
+        if self._items is None:
+            if self._spike_time_array is None:
+                return str([])
+            else:
+                return "SpikeTrainList containing {} spikes from {} neurons".format(
+                    self._spike_time_array.size,
+                    self._channel_id_array.size)
+        else:
+            return str(self._items)
+
+    def __len__(self):
+        """Return len(self)"""
+        if self._items is None:
+            if self._all_channel_ids is not None:
+                return len(self._all_channel_ids)
+            elif self._channel_id_array is not None:
+                return np.unique(self._channel_id_array).size
+            else:
+                return 0
+        else:
+            return len(self._items)
+
+    def __add__(self, other):
+        """Return self + other"""
+        if isinstance(other, self.__class__):
+            if self._items is None or other._items is None:
+                # todo: update self._spike_time_array, etc.
+                raise NotImplementedError
+            else:
+                self._items.extend(other._items)
+            return self
+        elif other and isinstance(other[0], SpikeTrain):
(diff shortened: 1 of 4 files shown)


samuelgarcia:
Hi @apdavison ,
What I understand here is that you need a list that internaly interleave or not spiketrain depending the situation.

My first impression is that we should first implement in a generic class that do the bidirectional relationship for all object (at least the one to many) and then hinerits from that generic class.

Then this SpikeTrainList would inherits this generic class.

Naively, I would inherits this generic class from the python list but I don't known the consequence.

Cheers,

Samuel

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,150 @@
...
+        obj._channel_id_array = channel_id_array
+        obj._all_channel_ids = all_channel_ids
+        obj._spiketrain_metadata = {
+            "units": units,
+            "t_start": t_start,
+            "t_stop": t_stop
+        }
+        return obj
+
+    def _spiketrains_from_array(self):
```
JuliaSprenger, line 124: This function is quite expensive, as it generates all Spiketrain objects explicitly. As it is used in many of Spiketrainlist methods, this would cause quite a bit of overhead. Maybe it would be good to use it in only as few places as possible?

apdavison: It's only ever used once, because the generated SpikeTrain objects are cached in `self._items`

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,150 @@
...
+        obj.segment = self.segment
+        self._items.append(obj)
+
+    def extend(self, iterable):
+        """L.extend(iterable) -> None -- extend list by appending elements from the iterable"""
+        if self._items is None:
+            self._spiketrains_from_array()
+        for obj in iterable:
+            obj.segment = self.segment
+        self._items.extend(iterable)
```
JuliaSprenger, line 105: maybe it's necessary to also verify that the items of `iterable` are instances of `SpikeTrain`? Otherwise arbitrary objects might be in the SpikeTrainList...

neo/core/spiketrainlist.py
JuliaSprenger: Yes, you are right. So in this PR the `SpikeTrain` representation is still the reference all data will be converted to eventually. Even when merging two `SpikeTrainList`s coming from a spike_time_array representation. Is there any way we can avoid that? Maybe having a default operation mode that can switch between `SpikeTrain` and `spike_time_list`? Or would it be worth to have the `_spike_time_array` and the corresponding masks for individual channels / units as base representation and only generate SpikeTrains when required (without) caching / duplicating the data? This would rely even stronger on numpy for inserting / reordering of the spike_times lists and masks. I am not sure which method would be more performant in the end. @apdavison  Do you have a better intuition there?

apdavison: you're right that merging two array-representation SpikeTrainLists should avoid creating `SpikeTrain`s. I think that's easy enough to do. 

The general idea is to keep data in the representation they arrive in as long as possible, to avoid unecessary transformations, i.e. the "reference/base" representation depends on how the object was initialized. 

Note that the `multiplexed` property (not yet implemented in this branch, but you can see it in the [spiketrainlist branch](https://github.com/apdavison/python-neo/blob/spiketrainlist/neo/core/spiketrainlist.py#L170) which will replace this PR) allows users to access the array representation.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,150 @@
...
+    def __add__(self, other):
+        """Return self + other"""
+        if isinstance(other, self.__class__):
+            if self._items is None or other._items is None:
+                # todo: update self._spike_time_array, etc.
+                raise NotImplementedError
+            else:
+                self._items.extend(other._items)
+            return self
+        elif other and isinstance(other[0], SpikeTrain):
```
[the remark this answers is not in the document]
apdavison: yes, this was a shortcut, but it would be more robust to do a full check

neo/core/spiketrainlist.py
apdavison: yes, more robust checks are needed

apdavison:
SpikeTrainList now implemented in #1000.

Closing this now.

PR #589, opened 2018-10-24
account for empty nse in rawio.neuralynxrawio

sin-mike:
Guys, I've made an update for a workaround with the empty nse files
There still will be info in self._spike_memmap, but it will be just empty numpy arrays

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -148,11 +148,11 @@ class NeuralynxRawIO(BaseRawIO):
 
                     if (os.path.getsize(filename)<=HEADER_SIZE):
                         self._empty_nse_ntt.append(filename)
-                        data = dict(unit_id    = np.array([]), 
-                                    channel_id = np.array([]), 
-                                    timestamp  = np.array([]), 
-                                    features   = np.array([[]]), 
-                                    samples    = np.array([[]]))
+                        data = dict(unit_id    = np.zeros((0,), dtype=dtype), 
+                                    channel_id = np.zeros((0,), dtype=dtype), 
+                                    timestamp  = np.zeros((0,), dtype=dtype), 
+                                    features   = np.zeros((0,0), dtype=dtype), 
+                                    samples    = np.zeros((0,0), dtype=dtype))
                     else:
                         data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
 
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -74,6 +74,7 @@ class NeuralynxRawIO(BaseRawIO):
         self.internal_unit_ids = []  # channel_index > (channel_id, unit_id)
         self.internal_event_ids = []
         self._empty_ncs = [] # this list contains filenames of empty records
+        self._empty_nse_ntt = []
 
         # explore the directory looking for ncs, nev, nse and ntt
         # And construct channels headers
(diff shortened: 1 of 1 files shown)


neo/rawio/neuralynxrawio.py
```diff
@@ -144,7 +145,17 @@ def _parse_header(self):
...
                     dtype = get_nse_or_ntt_dtype(info, ext)
-                    data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
+
+                    if (os.path.getsize(filename)<=HEADER_SIZE):
+                        self._empty_nse_ntt.append(filename)
+                        data = dict(unit_id    = np.array([]), 
+                                    channel_id = np.array([]), 
+                                    timestamp  = np.array([]), 
+                                    features   = np.array([[]]), 
+                                    samples    = np.array([[]]))
```
samuelgarcia: Here I would prefer :
```python 
data = np.zeros((0,), dtype=dtype)
```

neo/rawio/neuralynxrawio.py
```diff
@@ -89,7 +90,7 @@ def _parse_header(self):
             if ext not in self.extensions:
                 continue
 
-            if (os.path.getsize(filename)<=HEADER_SIZE) and (ext == 'ncs'):
+            if (os.path.getsize(filename)<=HEADER_SIZE) and (ext in ['ncs']):
```
samuelgarcia: Do you mean
```python
(ext in ['ncs', 'ntt']):
```

samuelgarcia:
Thank you for the patch.
I have made 2 comments.

Unfortunatly there are a lot of there weird pep8 **invalid escape sequence '\S'** but it was already in the original code.

neo/rawio/neuralynxrawio.py
sin-mike: At first, there were 
`ext in ['ncs', 'nse', 'ntt']`
but then we had a talk in #553, and then I just put 'ncs'.  
But this version of mine lived in my local verison. Now I just wanted to sync with the master. 

So afaik, there were 2 reasons:
1. Empty NCS are not important
2. Empty spikes (NSE, NTT) are somehow important. However, we didn't ever used this info.

Therefore, I just skipped empty ncs and put empty info to nse and ntt

samuelgarcia: Oups.
I reply to quickly.
I forgot the last patch and mix nse and nse.
I am sorry, this is OK.

neo/rawio/neuralynxrawio.py
```diff
@@ -148,11 +148,11 @@ def _parse_header(self):
...
-                        data = dict(unit_id    = np.array([]), 
-                                    channel_id = np.array([]), 
-                                    timestamp  = np.array([]), 
-                                    features   = np.array([[]]), 
-                                    samples    = np.array([[]]))
+                        data = dict(unit_id    = np.zeros((0,), dtype=dtype), 
+                                    channel_id = np.zeros((0,), dtype=dtype), 
+                                    timestamp  = np.zeros((0,), dtype=dtype), 
+                                    features   = np.zeros((0,0), dtype=dtype), 
+                                    samples    = np.zeros((0,0), dtype=dtype))
```
samuelgarcia: Oups.
My proposal was:
```data = np.zeros((0,), dtype=dtype)```
would replace all lines 151 - 155
data must be numpy .array 1D  (memmap or not) with a structured dtype not a dict.

sin-mike: done)

samuelgarcia:
OK for me.
I think we can merge.

Issue #590, opened 2018-10-25
Set channel list for reading explicitly 


sin-mike:
@samuelgarcia, I did an inspection of neuralynxrawio.py up to the baserawio.py, and I couldn't find a way to set a channel list for reading. Is there a way? Could I implement it in neuralynxrawio explicitly? 

e.g. it will just filter by channel id in the following structures
```
        self.ncs_filenames = OrderedDict()  # chan_id: filename
        self.nse_ntt_filenames = OrderedDict()  # chan_id: filename
        self.nev_filenames = OrderedDict()  # chan_id: filename

        self._nev_memmap = {} #
        self._spike_memmap = {}
        self.internal_unit_ids = []  # channel_index > (channel_id, unit_id)
        self.internal_event_ids = []
```

as well as in `self.header`

samuelgarcia:
Are you at neo.io or neo.rawio level ?

For neo.io level, channel selection when reading is not yet implemented but the futur lazy/proxy object will do this.
See #461.
It should be merged in December.

For neo.rawio level you can do
```
reader = neo.NeuralynxRawIO(dirname='/ath/to/dataset')
reader.parse_header()

# read signal by channel index (0 based)
raw_sig = reader.get_analogsignal_chunk(channel_indexes=[0,5,24, 36])

# read signal by channel name
raw_sig = reader.get_analogsignal_chunk(channel_names=['ch0', 'ch5', 'ch24', 'ch36'])

# read signal by channel id.
# the notion of IO vary from IOs
raw_sig = reader.get_analogsignal_chunk(channel_ids=[...])
```


This is briefly documented here:
https://neo.readthedocs.io/en/0.6/rawio.html#basic-usage
please add any thing in the doc that would help.

sin-mike:
@samuelgarcia Thank you! It makes sense now. However, in our lab, we have some experiments at which the chan_id is not unique. I want people to switch to python+neo from Matlab, but in this exact case, they get a mess in their data. Their use case is that, when you record in DC mode, you can assign two different channels for the same AD channel, each with different preprocessing. It becomes much handier for their purposes. A simple workaround is to split files in the working directory to ensure channel uniqueness. But my suggestion was simply to list explicitly, which channels to read, e.g. passing a subset of files to the reader. Because by default the reader just scans the whole dir at a time of init. When the chan_id is not unique, the ordered dicts are all broken, as in pic:

[...]

PR #591, opened 2018-10-27
Nlx unique chan id fix

sin-mike:
@samuelgarcia 

This PR includes a fix for non-unique chan_id when there are several files for the same AD channel

see  #590,  #589

Now I can read mixed records without splitting folders, like this
image

These two channels are reading from one AD channel, but with different preprocessings:
image

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -148,11 +148,11 @@ class NeuralynxRawIO(BaseRawIO):
 
                     if (os.path.getsize(filename)<=HEADER_SIZE):
                         self._empty_nse_ntt.append(filename)
-                        data = dict(unit_id    = np.array([]), 
-                                    channel_id = np.array([]), 
-                                    timestamp  = np.array([]), 
-                                    features   = np.array([[]]), 
-                                    samples    = np.array([[]]))
+                        data = dict(unit_id    = np.zeros((0,), dtype=dtype), 
+                                    channel_id = np.zeros((0,), dtype=dtype), 
+                                    timestamp  = np.zeros((0,), dtype=dtype), 
+                                    features   = np.zeros((0,0), dtype=dtype), 
+                                    samples    = np.zeros((0,0), dtype=dtype))
                     else:
                         data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
 
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -74,6 +74,7 @@ class NeuralynxRawIO(BaseRawIO):
         self.internal_unit_ids = []  # channel_index > (channel_id, unit_id)
         self.internal_event_ids = []
         self._empty_ncs = [] # this list contains filenames of empty records
+        self._empty_nse_ntt = []
 
         # explore the directory looking for ncs, nev, nse and ntt
         # And construct channels headers
@@ -89,7 +90,7 @@ class NeuralynxRawIO(BaseRawIO):
             if ext not in self.extensions:
                 continue
 
-            if (os.path.getsize(filename)<=HEADER_SIZE) and (ext == 'ncs'):
+            if (os.path.getsize(filename)<=HEADER_SIZE) and (ext in ['ncs']):
                 self._empty_ncs.append(filename)
                 continue
 
@@ -144,7 +145,17 @@ class NeuralynxRawIO(BaseRawIO):
                     self.nse_ntt_filenames[chan_id] = filename
 
                     dtype = get_nse_or_ntt_dtype(info, ext)
-                    data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
+
+                    if (os.path.getsize(filename)<=HEADER_SIZE):
+                        self._empty_nse_ntt.append(filename)
+                        data = dict(unit_id    = np.array([]), 
+                                    channel_id = np.array([]), 
+                                    timestamp  = np.array([]), 
+                                    features   = np.array([[]]), 
+                                    samples    = np.array([[]]))
+                    else:
+                        data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
+
                     self._spike_memmap[chan_id] = data
 
                     unit_ids = np.unique(data['unit_id'])
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -148,11 +148,7 @@ class NeuralynxRawIO(BaseRawIO):
 
                     if (os.path.getsize(filename)<=HEADER_SIZE):
                         self._empty_nse_ntt.append(filename)
-                        data = dict(unit_id    = np.zeros((0,), dtype=dtype), 
-                                    channel_id = np.zeros((0,), dtype=dtype), 
-                                    timestamp  = np.zeros((0,), dtype=dtype), 
-                                    features   = np.zeros((0,0), dtype=dtype), 
-                                    samples    = np.zeros((0,0), dtype=dtype))
+                        data = np.zeros((0,), dtype=dtype)
                     else:
                         data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
 
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -65,7 +65,7 @@ class NeuralynxRawIO(BaseRawIO):
         unit_channels = []
         event_channels = []
 
-        self.ncs_filenames = OrderedDict()  # chan_id: filename
+        self.ncs_filenames = OrderedDict()  # uid: filename
         self.nse_ntt_filenames = OrderedDict()  # chan_id: filename
         self.nev_filenames = OrderedDict()  # chan_id: filename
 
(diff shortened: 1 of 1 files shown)


sin-mike:
Hi, @samuelgarcia, any updates on this PR?

sin-mike:
> 
> 
> @sin-mike
> Thank you very much for this. This is cool to fix.
> I was am a bit busy and still be until wensday.
> I will make comment soon.
> 
> For short, I am a bit discomfort with this uuid, this make the logic harder to understand.
> In my mind the double key (name, chan_id) was already unique for the dict.

@samuelgarcia Indeed, but here I was worried about the consistency, as, in the header, you're using just integer ids. And that using tuple ids could easily break something further. As I discovered, the header is a key structure for the data IO operations, e.g. to restrict IOs to certain channels, I just modified the header and removed rows partially. Therefore I used just integer IDs.

samuelgarcia:
@sin-mike I am back.
thank you for this new comit but I had in mind the posite solution : remove this generated uid.
I think we can go wihtout it. 

In short in the header['signal_channels'] the chan_id must totally consistent with what the user known about the file. So we must put the real chan_id even it is not unique.
Some IOs do not have unique id or unique names for sig channels. it is not problem because we can always  ```acces then read_signal_chunk(channel_indexes=[...])```. In your special case ```read_signal_chunk(channel_names=[...])``` and ```read_signal_chunk(channel_ids=[...])``` will raise an error because it is tested at the BaseRawIO class level.

Other remark: why did you change the id for spike channel ?

I will make comment in the code.

If we use a generated id it should be internal but in our case I think it is useless.

I feel really sorry to add more work.

neo/rawio/neuralynxrawio.py
```diff
@@ -65,31 +65,41 @@ def _parse_header(self):
         unit_channels = []
         event_channels = []
 
-        self.ncs_filenames = OrderedDict()  # chan_id: filename
-        self.nse_ntt_filenames = OrderedDict()  # chan_id: filename
+        self.ncs_filenames = OrderedDict()  # uid: filename
```
samuelgarcia: We can do this with a tupple for the key:
self.ncs_filenames = OrderedDict()  # (chan_name, chan_id): filename

neo/rawio/neuralynxrawio.py
```diff
@@ -65,31 +65,41 @@ def _parse_header(self):
...
+        self.nse_ntt_filenames = OrderedDict()  # uid: filename
```
samuelgarcia: I prefer also :
 # (chan_name, chan_id): filename

neo/rawio/neuralynxrawio.py
```diff
@@ -65,31 +65,41 @@ def _parse_header(self):
...
         self.nev_filenames = OrderedDict()  # chan_id: filename
+        self.uid2chid = OrderedDict() # uid : (file_enum, chan_enum, chan_id)
```
samuelgarcia: I prefer also :
 (chan_name, chan_id): filename

neo/rawio/neuralynxrawio.py
```diff
@@ -65,31 +65,41 @@ def _parse_header(self):
...
         # And construct channels headers
         signal_annotations = []
         unit_annotations = []
         event_annotations = []
 
-        for filename in sorted(os.listdir(self.dirname)):
+        # filter all filenames to match IO extensions,
+        # so that fid indicates only the necessary files
+        all_filenames = filter(lambda f: os.path.splitext(f)[1][1:] in self.extensions, 
+                               sorted(os.listdir(self.dirname)))
```
samuelgarcia: Super good.

neo/rawio/neuralynxrawio.py
```diff
@@ -65,31 +65,41 @@ def _parse_header(self):
...
+
+        # I'm generating a unique id for a channel.
+        # chan_id is not a unique id, as several
+        # CSCs could reference the same chan_id
+        # // Mike
+        gen_uid = lambda x,y,z: x + 1000*y
+
+        for fid, filename in enumerate(all_filenames):
```
samuelgarcia: I am quite unhappy with this id.

neo/rawio/neuralynxrawio.py
```diff
@@ -98,19 +108,27 @@ def _parse_header(self):
...
             chan_ids = info['channel_ids']
 
-            for idx, chan_id in enumerate(chan_ids):
-                chan_name = chan_names[idx]
+            for idx, (chan_id, chan_name) in enumerate(zip(chan_ids, chan_names)):
+                # ^^^ I had never seen cases when idx is non-zero // Mike
+                
+                uid = gen_uid(fid, idx, chan_id)
+                self.uid2chid[uid] = (fid, idx, chan_id)
+
```
samuelgarcia: This could be remove so.

neo/rawio/neuralynxrawio.py
```diff
@@ -98,19 +108,27 @@ def _parse_header(self):
...
                     units = 'uV'
                     gain = info['bit_to_microVolt'][idx]
                     if info['input_inverted']:
                         gain *= -1
                     offset = 0.
                     group_id = 0
-                    sig_channels.append((chan_name, chan_id, info['sampling_rate'],
+
+                    ### Note: vvvv: this structure could contain non-unique chan_name, chan_id
+                    sig_channels.append((chan_name, uid, info['sampling_rate'],
```
samuelgarcia: Let s go back to
sig_channels.append((chan_name, chan_id, info['sampling_rate'],
so the use can have something that match the real id in the file and the software.

neo/rawio/neuralynxrawio.py
```diff
@@ -98,19 +108,27 @@ def _parse_header(self):
...
                                          'int16', units, gain, offset, group_id))
-                    self.ncs_filenames[chan_id] = filename
+                    ### // Mike
+
+                    self.ncs_filenames[uid] = filename
```
samuelgarcia: self.ncs_filenames[( chan_name, chan_id)] = filename

neo/rawio/neuralynxrawio.py
```diff
@@ -133,26 +151,36 @@ def _parse_header(self):
...
                     # a file can contain several unit_id (so several unit channel)
-                    assert chan_id not in self.nse_ntt_filenames, \
-                        'Several nse or ntt files have the same unit_id!!!'
-                    self.nse_ntt_filenames[chan_id] = filename
+
+                    # vvv this assert is not informative anymore
+                    # assert chan_id not in self.nse_ntt_filenames, \
+                    #     'Several nse or ntt files have the same unit_id!!!'
+
+                    self.nse_ntt_filenames[uid] = filename
```
samuelgarcia: self.nse_ntt_filenames[( chan_name, chan_id)] = filename

neo/rawio/neuralynxrawio.py
```diff
@@ -133,26 +151,36 @@ def _parse_header(self):
...
-                    data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
-                    self._spike_memmap[chan_id] = data
+
+                    if (os.path.getsize(filename)<=HEADER_SIZE):
+                        self._empty_nse_ntt.append(filename)
+                        data = np.zeros((0,), dtype=dtype)
+                    else:
+                        data = np.memmap(filename, dtype=dtype, mode='r', offset=HEADER_SIZE)
+
+                    self._spike_memmap[uid] = data
```
samuelgarcia: OK

neo/rawio/neuralynxrawio.py
```diff
@@ -133,26 +151,36 @@ def _parse_header(self):
...
 
                     unit_ids = np.unique(data['unit_id'])
                     for unit_id in unit_ids:
                         # a spike channel for each (chan_id, unit_id)
-                        self.internal_unit_ids.append((chan_id, unit_id))
+                        self.internal_unit_ids.append((uid, unit_id))
```
samuelgarcia: This could disappear.

neo/rawio/neuralynxrawio.py
```diff
@@ -133,26 +151,36 @@ def _parse_header(self):
...
 
-                        unit_name = "ch{}#{}".format(chan_id, unit_id)
+                        unit_name = "ch{}#{}".format(uid, unit_id)
```
samuelgarcia: Let s go back to unit_name = "ch{}#{}".format(chan_id, unit_id) or to
unit_name = "{}ch{}#{}".format(chan_name, chan_id, unit_id)
so the user can make the link with the sofwtare

samuelgarcia:
I made some comment in the code.

Best.

sin-mike:
@samuelgarcia 

Ok, I tried with tuple ids (chan

but...

it breaks at header info formatting:
`206         sig_channels = np.array(sig_channels, dtype=_signal_channel_dtype)`

The signal channel dtypes come from:
`from .baserawio import (BaseRawIO, _signal_channel_dtype,   _unit_channel_dtype, _event_channel_dtype)`

So, in my version, it was just:
`sig_channels.append((chan_name, uid, info['sampling_rate'], 'int16', units, gain, offset, group_id))`

Then, with tuple ids, I must write: 
`sig_channels.append((chan_name, (chan_name, chan_id), info['sampling_rate'], 'int16', units, gain, offset, group_id))`

with tuples, it breaks the header dtypes, and I tough the header is something built in stone, so I'm not allowed to change its fields.

If I'm allowed to change the header structure, then, no problem, tuple ids will work, otherwise, `id` field must be ints

samuelgarcia:
(chan_name, chan_id) would not be for the header['signal_channels'] but for self.ncs_filenames[ (chan_name, chan_id)]

PR #600, opened 2018-11-15
rtu to fix this overflow.

### LINKED ISSUE #528: overflow encountered in long_scalars
Got a RuntimeWarning: overflow encountered in long_scalars on spike2rawio.py:89 : 
chan_info['t_start'] = block_info['start_time'] * \
                        info['us_per_time'] * info['dtime_base']

while loading a 10500s / 850Mo smr file

I'll try with several files to check whereas the issue is repeatable

samuelgarcia:
This should fix #528


samuelgarcia:
Hi @FGabrielli.
The PR include a patch that should fix the overflow problem in spike2io.
We can not test it because we do not have theses big files.
If you could test this and validate the patch it would be great and we could merge this before the release 0.7.

Thanks a lot.

Samuel

samuelgarcia:
If pass test.
This is ready to merge.

Issue #613, opened 2018-11-27
Spiketrain behavior on rescale changed


mdenker:
The behavior when rescaling a SpikeTrain changed in 0.7.0 proposed, and is now very counter-intuitive. The change surfaces in a number of tests in Elephant.

Behavior 0.5.2 (`t_start` and `t_stop` are rescaled accordingly):
```
In [1]: import quantities as pq
In [2]: import neo
In [3]: a=neo.SpikeTrain([1,2,3]*pq.ms,t_start=0*pq.s,t_stop=10*pq.h)
In [4]: b=a.rescale(pq.s)
In [5]: print b
[0.001 0.002 0.003] s
In [6]: print b.t_start
0.0 s
In [7]: print b.t_stop
36000.0 s
```

Behavior 0.7.0 (`t_start` and `t_stop` are rescaled to millisecond):
```
In [1]: import quantities as pq
In [2]: import neo
In [3]: a=neo.SpikeTrain([1,2,3]*pq.ms,t_start=0*pq.s,t_stop=10*pq.h)
In [4]: b=a.rescale(pq.s)
In [5]: print b
[0.001 0.002 0.003] s
In [6]: print b.t_start
0.0 ms
In [7]: print b.t_stop
36000000.0 ms
```

mdenker:
Related to https://github.com/NeuralEnsemble/elephant/pull/189

mdenker:
Further investigation yields that in 0.7 when rescaling the spiketrain, this calls the `__new__` of a new Spiketrain instance, which in turn uses the dimensionality of the `times` property of the original object to set that of `t_start` and `t_stop`, i.e., if the spike train `a` above would have times in hourse, `t_start` in the new object would be in hours.

In 0.5.2, the `rescale()` method of Spiketrains has this code
```
if self.dimensionality == pq.quantity.validate_dimensionality(units):
    return self.copy()
spikes = self.view(pq.Quantity)
obj = SpikeTrain(times=spikes, t_stop=self.t_stop, units=units,
                         sampling_rate=self.sampling_rate,
                         t_start=self.t_start, waveforms=self.waveforms,
                         left_sweep=self.left_sweep, name=self.name,
                         file_origin=self.file_origin,
                         description=self.description, **self.annotations)
```
that is replaced in 0.7.0 by 
```
obj = super(SpikeTrain, self).rescale(units)
```

[...]

PR #614, opened 2018-11-27
Fix and associated test for #613

apdavison:


mdenker:
The fix solves the original issue and also associated unit test breaking in the Elephant. Good to merge from my side.

Issue #621, opened 2018-12-12
duplicate_with_new_array changed to duplicate_with_new_data


aguimera:
The basesignal class method  "duplicate_with_new_array" has been changed by "duplicate_with_new_data" in version 0.7, it can produce back compatibility errors.

jpgill86:
Indeed, this is affecting elephant, and a simple fix is proposed in NeuralEnsemble/elephant#189.

In my opinion, the old name should have been retained as an alias for at least one development cycle, eventually deprecated, and later removed, but, hey, hindsight is 20/20. In the words of Michael Scott, we should have had hindsight.

apdavison:
> the old name should have been retained as an alias for at least one development cycle, eventually deprecated, and later removed

I completely agree. It was more a case of oversight than lack of foresight - the change was in a PR that was merged a short time before the release, and we didn't notice the change! 

I actually think we should reintroduce an alias in the next release.

apdavison:
Done. Released in v 0.7.1

PR #622, opened 2018-12-13
Add an alias for `duplicate_with_new_data`, for backwards compatibility.

apdavison:
See #621 

I propose to make a release with version 0.7.1 as soon as this is merged, so that we can reduce the impact of updating Neo on downstream projects. What do you think @samuelgarcia ?

### DIFF
--- a/neo/core/dataobject.py
+++ b/neo/core/dataobject.py
@@ -331,6 +331,12 @@ class DataObject(BaseNeo, pq.Quantity):
             length = 1
         return length
 
+    def duplicate_with_new_array(self, signal, units=None):
+        warnings.warn("Use of the `duplicate_with_new_array function is deprecated. "
+                      "Please use `duplicate_with_new_data` instead.",
+                      DeprecationWarning)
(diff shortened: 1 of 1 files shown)


samuelgarcia:
I agree with this and making a release.
But we have already some commit since the the release.
Do you propose a release with only this in a branch or with the current master ?

In any case I am OK.

apdavison:
As far as I can see, all the commits since 0.7.0 are minor fixes, so I plan to release from master.

samuelgarcia:
OK.

jpgill86:
Since `duplicate_with_new_array` is deprecated, should an issue be created to eventually remove it, tagged with milestone 0.8.0 (or 0.9.0)?

apdavison:
@jpgill86 Go for it! (I suggest 0.9.0 to give people enough time to adapt, as 0.8.0 will be out quite soon).

jpgill86:
Done, though I don't have the ability to tag the new issue with a milestone.

Issue #506, opened 2018-03-22
Sort out coveralls


apdavison:
We are now getting 4 coveralls reports with every commit to a PR. I don't know why.

In addition, the coverage numbers come only from the core tests. I suggest deactivating coveralls on Travis, and taking the numbers only from CircleCI.

samuelgarcia:
OK.

apdavison:
Implemented in #623

PR #623, opened 2018-12-13
Stop pushing coverage information to Coveralls from Travis CI

apdavison:
as we don't test the `io` modules there.

Now Coveralls gets coverage information only from CircleCI

cf #506


samuelgarcia:
OK.

PR #626, opened 2018-12-17
Make NixIO raw implementation default selection for nix format

JuliaSprenger:


samuelgarcia:
Can I merge this ?

achilleas-k:
I don't know if this is a good idea.  The original NIX IO is much more general with respect to the Neo object hierarchy.  The NIX Raw IO on the other hand requires certain assumptions to hold for the file to be readable, such as consistent time dimensions across all signals in a Segment.  I think there are more limitations in the Raw IO that I can't remember now (perhaps @hkchekc can comment on this).

I've been a bit out of the loop on any new developments for the Raw IO API.  Is it accurate to say that there are *valid* Neo files that the Raw IO API doesn't support?  By valid Neo files, I mean files that the Neo API (and the IOs that support writing) will create and save without errors or warnings.  If this is the case, then I think having the NIX Raw IO as the default is going to cause issues.

hkchekc:
@samuelgarcia 

The main issue is that the NIXRawio (or in general the rawios) support only a single block structure or multiple blocks with same within-block structure (i.e. all blocks having same number of segments, ansig, spiketrains etc.) , which is not always the case for NIX files. And as @achilleas-k suggest, (if I am not wrong), these files with difference across blocks are valid for both NEO and NIX. 

I am not sure with the current annotation structure in the RawIO API this issue is resolved.

samuelgarcia:
OK. So lets keep the old order for IOs.

Note that rawio force to have the same channel count across block but do not force to have the same segment ocunt accross blocks.

@JuliaSprenger : Do we close this PR without merge?

samuelgarcia:
Is it OK if I change the "by_property" to "grouping_property" ?

samuelgarcia:
Is it OK to add this behavior:
  * when there is an explicit probe_file with several group AND grouping_property is None then the
    grouping_property is automatically set to "group" ?

samuelgarcia:
Oups.
sorry for the 2 last comment, it is on the wrong project...
I am tired.

JuliaSprenger:
@hkchekc: Would it make sense to extend the functionality of the NixRawIO implentation to cover all cases accepted by the RawIO structure (as @samuelgarcia mentioned above, allow different segment counts per block).

hkchekc:
@JuliaSprenger : wrote something wrong on my previous comment, the NIXrawio allows different seg_count. Sorry

JuliaSprenger:
@hkchekc: Does this mean there are no additional limitations introduced by the RawIO implementation? And if no, what are the limitations?

hkchekc:
@samuelgarcia : does the rawio force same channel count on the segment level or on the block level?

I think we can have for example 2 blocks with segment_count [2,3] but in the channel_count in each segment have to be the same right?

samuelgarcia:
Yes channel count is the same for all blocks with neo.rawio contrary to neo.io whihch is more flexible for that point.

In practice, I never see the channel map changing from one block to another in a file.

JuliaSprenger:
@samuelgarcia: So having different numbers of channels in different blocks does not occur in data coming from a recording system, but might  occur during postprocessing of signals or coming from simulations? Since nix is one of the few formats Neo can write to, should we then restrict the writing to nix files, which are compatible with the limitations of the rawio reader implementation to avoid inconsistency?

samuelgarcia:
Do you mean restrict the writting at neo.io level or only NixIO ?

The best could be have a flag in NixIO.write_block(..., force_channel_consistency_across_block=True).

And in NixRawIO have check of this consistency. If yes, then the IO works if not it raise an error to the user and tell him to user the other NixIO (not raw based).

But it imply soe works.

achilleas-k:
I think the idea of having this kind of strict write mode is interesting but I wonder if it would be better made as a bigger decision.  Is this a special case for the NixIO?  If not, it implies there's a general *strict* Neo structure.  I'm not a heavy Neo user so maybe my opinion doesn't count for much, but if this is going to exist, I think this strict format should be decided on carefully and documented.  I'd even go further and suggest that a *validator* of some sort be written in the library that checks for inconsistencies between object counts and dimensions.

samuelgarcia:
Agree that this strict writting should be done at API level.

JuliaSprenger:
Sounds like a good idea! @samuelgarcia: Do you want to create a separate issue for this and keep this one open until we have implemented the validator?

apdavison:
validator implemented in #828, so closing

Issue #625, opened 2018-12-14
NixIOFr is ignoring annotations


JuliaSprenger:
Using the NixIOFr annotations of neo objects are not loaded, eg. the code below raises an `AssertionError`.
```python
from neo import NixIO, NixIOFr, Block

annotations = {'my_custom_annotation': 'hello block'}

with NixIO(filename) as io:
    bl = Block(**annotations)
    io.write_block(bl)

with NixIOFr(filename) as io:
    bl = io.read_block()
    assert len(bl.annotations) == len(annotations)
```

This leads to the unintentional merge of multiple AnalogSignals, when being loaded as they are considered to be the same object and merged into a single AnalogSignal object

hkchekc:
@JuliaSprenger Hi, thanks for telling. I will see how to fix it under the baserawio annotation generating framework.

On the other hand, I don't really understand how it leads to merge of multiple AnalogSignals. Can you show be some code? I think it may be due to our default signal-grouping-mode is group all signals in same channel?

hkchekc:
Some fixes is in #629

samuelgarcia:
## Note

**signal_group_mode** change the annotations behavior in read_block for all IOs depending on **'split-all'** or **'group-by-same-units'**.

This is because in rawio annotations are by channels.

So, this is mono channel signal with all annotations:
```python
anasig = reader.read_block(signal_group_mode='split-all').segments[0].analosignals[0]
```

and this is a multi channel signal without by-channel annotations because it is a channel bundle:
```python
anasig = reader.read_block(signal_group_mode='group-by-same-units').segments[0].analosignals[0]
```



Since the new array annotation features, annotations by channels will be back with **signal_group_mode='group-by-same-units'** in next version with new lazy mode, I am working on it here #461.
And so the default will be **signal_group_mode='group-by-same-units'** for all IOs without the lost of by-channels annotations.

JuliaSprenger:
The example above still fails, but due to additional `nix_name` annotations being introduced when writing data to nix and not due to the original issue. An updated fair comparison would be

```python
with NixIO(filename,'ow') as io:
    bl0 = Block(**annotations)
    io.write_block(bl0)

with NixIOFr(filename) as io:
    bl1 = io.read_block()
    assert len(bl0.annotations) == len(bl1.annotations)
```

The original issue is therefore solved. A follow up issue is #740.

PR #629, opened 2019-01-04
[NixIOFr] adding support for customize annotations 

hkchekc:
Dealing with issue #625 

now the NixIOFr will read customize annotations

### DIFF
--- a/neo/rawio/nixrawio.py
+++ b/neo/rawio/nixrawio.py
@@ -166,6 +166,36 @@ class NIXRawIO(BaseRawIO):
         self.header['event_channels'] = event_channels
 
         self._generate_minimal_annotations()
+        for blk_idx, blk in enumerate(self.file.blocks):
+            bl_ann = self.raw_annotations['blocks'][blk_idx]
+            for props in blk.metadata.inherited_properties():
+                self._add_annotate(bl_ann, props)
+            for seg_index, seg in enumerate(blk.groups):
+                seg_ann = bl_ann['segments'][seg_index]
+                for props in seg.metadata.inherited_properties():
+                    self._add_annotate(seg_ann, props)
+                ansig_idx = 0
+                for da in seg.data_arrays:
+                    if da.type == 'neo.analogsignal':
+                        ana_an = seg_ann['signals'][ansig_idx]
+                        for props in da.metadata.inherited_properties():
+                            self._add_annotate(ana_an, props)
+                        ansig_idx += 1
+                sp_idx = 0
+                for mt in seg.multi_tags:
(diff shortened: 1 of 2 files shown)


neo/rawio/nixrawio.py
```diff
@@ -324,3 +355,24 @@ def _rescale_epoch_duration(self, raw_duration, dtype='float64'):
         durations = raw_duration.astype(dtype)
         # supposing unit is second, other possibilities maybe mS microS...
         return durations  # return in seconds
+
+    def _add_annotate(self, tar_ann, props, otype):
```
JuliaSprenger: This function could be more general if you would use Neo object types as `otype`.

JuliaSprenger:
Thanks for fixing this @hkchekc!
I think it would make sense to also include a couple of unittests to cover this.

hkchekc:
@JuliaSprenger true, make sense to use the otype for some possible change later. Thanks

I will make the unitttests in another pull request.

samuelgarcia:
Hi @JuliaSprenger and @hkchekc.

Sorry to be late on this subject.

In my mind, at rawio level there is no dependency on neo.core at all (unlike neo.io).

This is important for me.
A use case could be build an API that read ephy file directly intto pandas object (or xarrays) without neo object.
Introducing neo.core break a bit the rule.


What I anderstand here is tat ```_add_annotate``` is some kind of internal method so calling with string or real object do not matter so much.

No ?

hkchekc:
> Hi @JuliaSprenger and @hkchekc.
> 
> Sorry to be late on this subject.
> 
> In my mind, at rawio level there is no dependency on neo.core at all (unlike neo.io).
> 
> This is important for me.
> A use case could be build an API that read ephy file directly intto pandas object (or xarrays) without neo object.
> Introducing neo.core break a bit the rule.
> 
> What I anderstand here is tat `_add_annotate` is some kind of internal method so calling with string or real object do not matter so much.
> 
> No ?

Yes, you are right, now it do not make any differences. I just thought possibly this can be used on further changes. But as you say if that contradicts with the idea of rawio, I revert the commit. Thanks for reminding

samuelgarcia:
### Other note

For rawio developement you can use 

this internal helpers:
  * BaserawIO._raw_annotate(...)  
  * BaserawIO._repr_annotations()

And this public function:
  * BaserawIO.print_annotations()

See neo/rawio/baserawio.py.

hkchekc:
Updates: I add some test and also fix some issues related to the events annotations in this PR. Also, I find #643 is seemingly making the cusomtized annotations for Analogsingals not feasible

PR #636, opened 2019-01-19
[NIXIO] version checks

achilleas-k:
This PR adds a version compatibility check function to the NIX IO.

Currently the IO requires NIXPy (nixio) version 1.5 or greater (betas included). The version checker only checks the first two components of the version string.

The same function is used in the constructor of the NIXRawIO.

Some codestyle cleanup for the NIXRawIO is included.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -1112,13 +1112,13 @@ class NixIO(BaseIO):
         for chx in neoblock.channel_indexes:
             signames = []
             for asig in chx.analogsignals:
-                if not ("nix_name" in asig.annotations and
-                        asig.annotations["nix_name"] in self._signal_map):
+                if not ("nix_name" in asig.annotations
+                        and asig.annotations["nix_name"] in self._signal_map):
                     self._write_analogsignal(asig, nixblock, None)
                 signames.append(asig.annotations["nix_name"])
             for isig in chx.irregularlysampledsignals:
-                if not ("nix_name" in isig.annotations and
-                        isig.annotations["nix_name"] in self._signal_map):
+                if not ("nix_name" in isig.annotations
+                        and isig.annotations["nix_name"] in self._signal_map):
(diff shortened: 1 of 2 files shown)


neo/io/nixio.py
```diff
@@ -80,6 +82,38 @@ def calculate_timestamp(dt):
...
+            "The NixIO requires the Python package for NIX "
+            "(nixio on PyPi). Try `pip install nixio`."
+
+        )
+
+    minverstr = ".".join(map(str, MIN_NIX_VER))
+    nixverstr = nix.__version__
+    nixver = nixverstr.split(".")
+
+    try:
```
JuliaSprenger: Would it makes sense to use `packaging.version` for the version handling here instead of doing explicit string handling?

achilleas-k: I considered it and since the `packaging` module isn't in the standard library, I didn't want to add an extra dependency. I only now realise that it's part of the setuptools and pip group of modules so it's pretty close to standard.

I'll amend.

achilleas-k:
Regarding the errors, I noticed them when preparing this PR and opened an issue on our fork: https://github.com/G-Node/python-neo/issues/21

At the time it looked like it was only occurring on macOS but it looks like it's happening on circle as well. Haven't had the chance to investigate yet.

achilleas-k:
Amended to use `packaging.version` for comparisons.

To allow betas and dev releases of 1.5, the version comparison is made against the base_version of the installed nixio module.

If the version string is not understood for any reason, a warning is presented, but usage is allowed.

achilleas-k:
Looks like the travis environments don't include the `packaging` module.

JuliaSprenger:
Except for Py3.4? That's strange.

achilleas-k:
Quite strange. I also assumed circle was failing for the old reason, but it looks like those envs don't have the module either.

achilleas-k:
Changed to distutils.version.

PR #637, opened 2019-01-21
Update docs to reflect correct URL for test files

rgerkin:
The URL for test files changed at some point but the docs did not reflect that in the example for AxonIO.  That example fails with the old URL and works with the new one.


samuelgarcia:
Thank you.
I don't anderstand why there is so much error in circle ci.

PR #639, opened 2019-01-21
fix logger statement

apdavison:


apdavison:
The CircleCI failures seem to have nothing to do with this PR.

samuelgarcia:
OK.
Maybe we should fix cicle before merging this.
No ?

apdavison:
I'm just going to merge this as it is a trivial change, CircleCI is fixed, but will require messing around with caches to get the tests to pass for the PR.

Issue #579, opened 2018-10-01
Proposal: extend AsciiSignalIO to allow extra information in a JSON file or in file header


apdavison:
At present, `AsciiSignalIO.read_segment` takes several optional arguments describing how the CSV file is structured.

I propose to add support for having this information either:
(i) in a JSON file that would accompany the text file with the data, or
(ii) in a "header" block in the file, with a simple "# name = value" format

The name of the JSON file could either be specified alongside the filename, or could follow a standard pattern, e.g. if the datafile was named "foo.txt" then the IO would automatically look for a file called "foo_about.json" or "foo_map.json"

JuliaSprenger:
I agree, this would be a good extension. I talked to a number of people at the Bernstein Conference, who would appreciate this feature. In my opinion json sounds like the better approach since it is more flexible and keeps the actual csv file clean.

apdavison:
Yes, we seem to have a lot of data in HBP using plain text files for ephys data.

I was thinking we could support both options (external JSON file and internal comments). The advantage of JSON is that you don't need to modify the original data file, but the advantage of internal comments is that everything is in one file.

PR #641, opened 2019-01-28
Some improvements to AsciiSignalIO

apdavison:
in preparation for working on #579. Main changes:

(1) moved most arguments from `read_segment()` method to the IO class constructor, and added a `read_block()` method. This makes AsciiSignalIO more consistent with other Neo IOs.
(2) added an argument "multichannel", to control whether to create a multi-channel AnalogSignal or a list of single-channel signals. The default is the latter, for backwards compatibility.
(3) more consistency checks on argument values
(4) fixed a bug with "genfromtxt" where usecols would be applied twice.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
           keys:
           - v1-dependencies-{{ checksum "requirements.txt" }}
           # fallback to using the latest cache if no exact match is found
-          - v1-dependencies-
+          #- v1-dependencies-
 
       - restore-cache:
           keys:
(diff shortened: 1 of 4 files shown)


samuelgarcia:
Hi @apdavison 

Some comments:
  * The ci cache is also for the test file download. This mean that each time there is a problem with files
    and we want to download all of then again we also need to hack this cache key
  * In BaseFromRaw.read_segment/read_block there is a karg ```signal_group_mode= 'split-all'``` or
      ```signal_group_mode= 'all-in-one'```. Than have the same meaning as ```multichannel=True/False```

Otherwise this is OK for me.

apdavison:
There is a separate cache for the test file download; it is only the "installed packages" cache which is involved here.

I will change `multichannel` to `signal_group_mode` for consistency.

samuelgarcia:
> There is a separate cache for the test file download; it is only the "installed packages" cache which
> is involved here.

Sorry for this bad information.

apdavison:
Anyone have time to merge this?

PR #647, opened 2019-02-14
Changed Datatype of Labels of Events and Epochs to Unicode String

Kleinjohann:
Labels of events and epochs are currently null-terminated strings, but using those is not recommended and users probably convert them to unicode strings anyway for further usage. I thought this was weird, so I changed the labels datatype to unicode strings as suggested in https://github.com/NeuralEnsemble/python-neo/blob/b4c5950d85ed4cedf27caf7dd36aa41067ab01ab/neo/rawio/blackrockrawio.py#L754


samuelgarcia:
Hi Alexander,
Thank you for this.
Good idea of course. the dtype='S' is very old now.

The thing is that we are planning this year some changes in the object model.
This kind of bytes>str (ex str>unicode) is of course one of theses changes.
For API stability reason this have been kept for a long time.

In my mind, the big question is **when** merging this.
The actual matser 0.8.dev, the "big" change is the lazy load with proxy.
The following release 0.9 should be more focus about changing object a bit (some attributes, uselees behavior, and relationship). So I think this PR could be incroporated just after the 0.8 release at the beginning of 0.9.dev.

@apdavison @JuliaSprenger @mdenker : what do you thing ?


best


Samuel

apdavison:
Yes, I think at the start of 0.9 development is a good time.

JuliaSprenger:
> Yes, I think at the start of 0.9 development is a good time.

That is now. @samuelgarcia @Kleinjohann Do you want to continue here and fix the conflicts or start a new PR?

Kleinjohann:
I fixed the conflicts. 
Travis-CI somehow could not download the Python 3.5 archive, not sure what's going on there.

JuliaSprenger:
@samuelgarcia So, do you think now is the right time point for introducing this change? If you are happy with it you can go ahead and merge. Looks good to me.

samuelgarcia:
Yes.
Go go go.
And then #727 ?

PR #648, opened 2019-02-15
Implement JSON metadata support for AsciiSignalIO (cf #579)

apdavison:


appukuttan-shailesh:
The JSON file currently has a parameter named `filename`. @apdavison  mentioned that:
> The idea was to allow cross-checking, to make sure that the correct data file is paired with the correct metadata file. This cross-check is not implemented at the moment, however.

On further discussion, it was identified that there might be occasions (quite commonly; e.g. in a batch of data files) when we want to use the same metadata JSON to handle all the files in that batch. @apdavison  therefore proposed to make this an optional parameter.

samuelgarcia:
This is OK for me.
tell me when you want I merge it.

apdavison:
It's ready to merge, please go ahead.

PR #651, opened 2019-02-26
BlackrockRawIO: Fixed thresholds not being multiplied by the digitisation factor

Kleinjohann:
The thresholds are applied to the digitised signal in the Blackrock software, to use them with the converted analog signal we need to multiply them by the digitisation factor.


JuliaSprenger:
Hi @Kleinjohann. Thanks for the fix. Would you mind also fixing the two pep8 issues on the way?

JuliaSprenger:
Thanks, I will merge this now.

Issue #632, opened 2019-01-15
Merge multiple spiketrains


JuliaSprenger:
It would be nice to have a utility function, which makes merging of multiple `SpikeTrain` objects easier.

JuliaSprenger:
This is related to #526

JuliaSprenger:
Does it make sense to extend the `spiketrain.merge` method to take more than one other spiketrain? This sounds like the most simple solution to me, but maybe there are better ones. Comments?

apdavison:
I think that makes sense. You could change the interface to `def merge(self, *others):`, so that it would work the same as before when merging just a pair.

apdavison:
implemented in #652

PR #652, opened 2019-02-28
Merge Multiple Spiketrains

Kleinjohann:
Addresses #632 

I extended the `spiketrain.merge` function to an arbitrary number of arguments with the interface `def merge(self, *others)` in order not to change anything for the case of just merging two spiketrains as suggested in https://github.com/NeuralEnsemble/python-neo/issues/632#issuecomment-454427484

I similarly extended `BaseNeo.merge` and `BaseNeo.merge_annotations` in order to be able to merge the annotations correctly.


Kleinjohann:
I'm not too sure how merging SpikeTrainProxies is supposed to work, should it just not work and raise an error? Or should it load the SpikeTrains automatically? Or should it return a ProxyObject which will return the merged SpikeTrain when one calls its load function?

samuelgarcia:
Intuitivelly I would say "load automatically" then merge.
With this we could merge a list that mix SpikeTrain and SpikeTrainProxy.

neo/core/spiketrain.py
```diff
@@ -618,77 +620,118 @@ def time_slice(self, t_start, t_stop):
...
-            if hasattr(other, "lazy_shape"):
-                merged_lazy_shape = (self.lazy_shape[0] + other.lazy_shape[0])
-            else:
-                raise MergeError("Cannot merge a lazy object with a real"
-                                 " object.")
-        if other.units != self.units:
-            other = other.rescale(self.units)
-        wfs = [self.waveforms is not None, other.waveforms is not None]
+        for other in others:
+            if type(other) not in [SpikeTrain, neo.io.proxyobjects.SpikeTrainProxy]:
```
apdavison: It is considered more Pythonic (I'm not sure why), to write:
```
if not isinstance(other, (SpikeTrain, neo.io.proxyobjects.SpikeTrainProxy))
```

(by the way, @samuelgarcia, this code makes me think that `proxyobjects` should be moved to `core`, since code in `neo.core` should not be referring to code in `neo.io`)

neo/core/spiketrain.py
```diff
@@ -618,77 +620,118 @@ def time_slice(self, t_start, t_stop):
...
+            if self.t_stop != other.t_stop:
+                raise MergeError("Cannot merge, different t_stop")
+            if self.left_sweep != other.left_sweep:
+                raise MergeError("Cannot merge, different left_sweep")
+            if self.segment != other.segment:
+                raise MergeError("Cannot merge these signals as they belong to"
+                                 " different segments.")
+
+        all_spiketrains = [self]
+        all_spiketrains.extend([st.rescale(self.units) if type(st) is SpikeTrain else
```
apdavison: I'm not convinced by the argument for automatic loading. Since this could be a "heavy" procedure, I think the user should first load the data, then perform the merge.

neo/core/spiketrain.py
samuelgarcia: I agree, at the beginning proxyobject.py and basefromrawio.py was in a separate directory.
I finally put every thing in neo.io.
We can move proxyobject.p > neo.core and keep basefromrawio.py in neo.io.

Kleinjohann: I changed the check to `isinstance`. I agree that `proxyobject.py` should probably not be in `neo.io`, I actually have to do `import neo` in `spiketrain.py` at the moment because importing from `neo.io` causes an import cycle.

samuelgarcia: I will do this in a separate PR.

neo/core/spiketrain.py
Kleinjohann: I changed the behaviour from automatic loading to raising an error.

Kleinjohann:
I incorporated the changes @apdavison requested and resolved the conflicts. 
Merging multiple `SpikeTrain` objects could also be interesting in the context of the new `SpikeTrainList`  in #588. Something like `SpikeTrainList.as_single_spiketrain` would cover most of the use cases I am using this for.

JuliaSprenger:
Hi @Kleinjohann. Thank for the fixes. To me the PR looks fine. The only issue not solved is if SpiketrainProxyObjects should be automagically loaded for merging as suggested by @samuelgarcia.

Kleinjohann:
@JuliaSprenger I went with @apdavison 's remark in https://github.com/NeuralEnsemble/python-neo/pull/652#discussion_r273216832 and changed the behaviour to returning an error instead of loading automagically.

JuliaSprenger:
Ok, then we can leave the `automagic loading of data` discussion for another issue, since it's a rather fundamental one.

PR #655, opened 2019-03-01
Change of the logic for version handling in neuralynx because of "pag…

samuelgarcia:
…asus" new file.

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -547,7 +547,7 @@ txt_header_keys = [
     ('OriginalFileName', '', None),
     ('TimeCreated', '', None),
     ('TimeClosed', '', None),
-    ('ApplicationName', '', None),  # also include version number 
+    ('ApplicationName', '', None),  # also include version number
     ('AcquisitionSystem', '', None),
     ('ReferenceChannel', '', None),
 ]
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @samuelgarcia!
Is the file you got small enough to use this also for a unittest? Then we should add it to the collection of test files.

samuelgarcia:
Unfortunatly not (100Mo).

This issue was declared on tridesclous here : https://github.com/tridesclous/tridesclous/issues/109
have a look. several problem with the file. you can download it.

This PR fix the easy thing with date formating, version number, and app name.

There are some other problem in this file : the gaps...

I ask @beastgit to also open an issue here an to provide a small file.

Maybe we can wait this file before the merge.

JuliaSprenger:
Ok, let's wait for a testfile.

samuelgarcia:
@beastgit : could you send us a test file ?

beastgit:
I am attaching the following file. Please do not make the contents of this
file public.

samuelgarcia:
@beastgit : this is a problem because we need a public file to integrate this in our unitest.

Could you have a file recording with the same device even if it is not connected to any patient.
Noise is perfect. We just need this new kind of header for neo.


Best

JuliaSprenger:
@beastgit: Have you had a chance to get your hands on a testfile? Since it's only about the header we could also try to manually tweak the header of another file, but this would be a dirty solution.

Issue #526, opened 2018-04-20
set up collection of / extend utility functions


JuliaSprenger:
Using the neo structure can be quite cumbersome in some cases, eg like slicing objects based on an epoch array (#525). To improve on this we could A) extend the methods neo objects offer and B) start an object-independent collection of utility functions in a separate file.
Most likely a combination of both is necessary to achieve best usability of the neo structure.

[Here](https://web.gin.g-node.org/INT/multielectrode_grasp/src/master/code/neo_utils.py) are some suggestions for helpful utility functions:
- [ ] create a `neo.Epoch` based on a  `neo.Event` object (or maybe two `neo.Event` objects)
- [ ] implement a `time_slice` function for `neo.Segment` that also slices child objects accordingly
- [ ] slicing a `neo.Segment` based on a `neo.Epoch`
- [ ] add `shift` methods for neo Data objects, which move timescale of the object (including `t_start` and `t_stop` adjustment, etc)
- [ ] support inplace slicing to avoid data duplication and keep original neo structure.

Any thoughts on this? Additional suggestions?

apdavison:
I support a combination of (A) and (B). In general I prefer adding functionality as new methods on the Neo classes, but by making some of this functionality also accessible as utility functions we improve the discoverability (NumPy does this a lot (e.g. `arr.min() and np.min(arr)`))

samuelgarcia:
For this utils functions/methods, I think it could be good to take proxy object in account.

Here are two possible slicing without/with proxy object.
Having the same function/method that autodetect proxyobject and do the slicing could be good.
No ?

**With Standard neo objects**
```
lim0, lim1 = -500*pq.ms, +1500*pq.ms
seg = reader.read_segment(lazy=False)
triggers = seg.events[0]
anasig = seg.analogsignals[0]  # here anasig contain the whole recording in memory
all_sig_chunks = []
for t in triggers.times:
    t0, t1 = (t + lim0), (t + lim1)
    anasig_chunk = anasig.time_slice(t0, t1)
    all_sig_chunks.append(anasig_chunk)
apply_my_fancy_average(all_sig_chunks)
```

**With proxy neo object**
```
lim0, lim1 = -500*pq.ms, +1500*pq.ms
seg = reader.read_segment(lazy=True)
triggers = seg.events[0].load(time_slice=None)  # this load all trigers in memory
anasigproxy = seg.analogsignals[0]  # this is a proxy
all_sig_chunks = []
[...]

PR #657, opened 2019-03-13
Added utility module neo/utils.py

Kleinjohann:
Addresses #526. I added [neo_utils.py](https://web.gin.g-node.org/INT/multielectrode_grasp/src/master/code/neo_utils.py) as `neo/utils.py` as suggested in https://github.com/NeuralEnsemble/python-neo/issues/526#issuecomment-431916318. I ensured compatibility with the current version of neo, especially with proxy objects to get the functionality outlined in https://github.com/NeuralEnsemble/python-neo/issues/526#issuecomment-383529750, and added unittests.

The original had a header with copyright information, a disclaimer, etc. I guess this is not needed since you have corresponding info for neo as a whole, right?

### DIFF
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -368,6 +368,10 @@ def add_epoch(
     --------
     epoch: neo.Epoch
         An Epoch object with the calculated epochs (one per entry in event1).
+
+    See also:
+    ---------
+    Event.to_epoch()
     """
     if event2 is None:
         event2 = event1
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -692,6 +692,7 @@ def seg_time_slice(seg, t_start=None, t_stop=None, reset_time=False, **kwargs):
     # cut analogsignals
     for ana_id in range(len(seg.analogsignals)):
         ana_time_slice = seg.analogsignals[ana_id].time_slice(t_start, t_stop)
+        # TODO: is this still an issue?
         # explicitly copying parents as this is not yet fixed in neo (
         # NeuralEnsemble/python-neo issue #220)
         ana_time_slice.segment = subseg
@@ -728,6 +729,31 @@ def seg_time_slice(seg, t_start=None, t_stop=None, reset_time=False, **kwargs):
     # TODO: Improve
     # seg.create_relationship(force=True)
     return subseg
+
+
+def shift_spiketrain(spiketrain, t_shift):
+    """
+    Shifts a spike train to start at a new time.
+
+    Parameters:
+    -----------
+    spiketrain: Neo SpikeTrain
+        Spiketrain of which a copy will be generated with shifted spikes and
+        starting and stopping times
+    t_shift: Quantity (time)
+        Amount of time by which to shift the SpikeTrain.
+
+    Returns:
+    --------
+    spiketrain: Neo SpikeTrain
+        New instance of a SpikeTrain object starting at t_start (the original
+        SpikeTrain is not modified).
+    """
+    new_st = spiketrain.duplicate_with_new_data(
+        signal=spiketrain.times.view(pq.Quantity) + t_shift,
+        t_start=spiketrain.t_start + t_shift,
+        t_stop=spiketrain.t_stop + t_shift)
+    return new_st
 
 
 def event_time_slice(event, t_start=None, t_stop=None):
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -249,15 +249,12 @@ def _event_epoch_slice_by_valid_ids(obj, valid_ids):
     sparse_annotations = _get_valid_annotations(obj, valid_ids)
 
     # modify array annotations
-    sparse_array_annotations = {key: (value[valid_ids] if len(value) else value) for key, value in obj.array_annotations.items()}
-
-    # modify labels
-    sparse_labels = _get_valid_labels(obj, valid_ids)
+    sparse_array_annotations = {key: value[valid_ids]
+                                for key, value in obj.array_annotations.items() if len(value)}
 
     if type(obj) is neo.Event:
         sparse_obj = neo.Event(
             times=copy.deepcopy(obj.times[valid_ids]),
-            labels=sparse_labels,
             units=copy.deepcopy(obj.units),
             name=copy.deepcopy(obj.name),
             description=copy.deepcopy(obj.description),
@@ -268,7 +265,6 @@ def _event_epoch_slice_by_valid_ids(obj, valid_ids):
         sparse_obj = neo.Epoch(
             times=copy.deepcopy(obj.times[valid_ids]),
             durations=copy.deepcopy(obj.durations[valid_ids]),
-            labels=sparse_labels,
             units=copy.deepcopy(obj.units),
             name=copy.deepcopy(obj.name),
             description=copy.deepcopy(obj.description),
@@ -347,24 +343,6 @@ def _get_valid_annotations(obj, valid_ids):
             sparse_annotations[key] = list(np.array(sparse_annotations[key])[
                 valid_ids])
     return sparse_annotations
-
-
-def _get_valid_labels(obj, valid_ids):
-    """
-    Internal function
-    """
-    labels = obj.labels
-    selected_labels = []
-    if len(labels) > 0:
-        if _is_annotation_list(labels, len(obj)):
-            for vid in valid_ids:
-                selected_labels.append(labels[vid])
-            # sparse_labels = sparse_labels[valid_ids]
-        else:
-            warnings.warn('Can not filter object labels. Shape (%s) does not '
-                          'fit object shape (%s)'
-                          '' % (labels.shape, obj.shape))
-    return np.array(selected_labels)
 
 
 def _is_annotation_list(value, exp_length):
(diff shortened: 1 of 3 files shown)


neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+
+import neo
+import copy
+import warnings
+import inspect
+import numpy as np
+import quantities as pq
+
+
+def get_events(container, properties=None):
```
apdavison: I think it would be more user-friendly to make this `get_events(container, **properties)`, so that it could be used like:
```python
get_events(seg, event_type='trial start')
```

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+        If True, the resulting Neo Epoch object is added to segment.
+
+    Keyword Arguments:
+    ------------------
+    Passed to the Neo Epoch object.
+
+    Returns:
+    --------
+    epoch: neo.Epoch
+        An Epoch object with the calculated epochs (one per entry in event1).
```
apdavison: I suggest adding a section "See also", referring to the `Event.to_epoch()` method.

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+def _is_annotation_list(value, exp_length):
+    """
+    Internal function
+    """
+    return (
+        (isinstance(value, list) or (
+            isinstance(value, np.ndarray) and value.ndim > 0)) and (len(value) == exp_length))
+
+
+def add_epoch(
```
apdavison: This function does two separate things: create an Epoch from one or two Events, and then add the resulting Epoch to a Segment. I think the functionality should be simplified so it only does the first part (and hence renamed to `epoch_from_events()` or something similar).

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+        segment, event1, event2=None, pre=0 * pq.s, post=0 * pq.s,
+        attach_result=True, **kwargs):
+    """
+    Create epochs around a single event, or between pairs of events. Starting
```
apdavison: "Epoch" and "Event" should be capitalized throughout the docstring.

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+        obj=event1, valid_ids=match_ev1)
+    event2_matched = _event_epoch_slice_by_valid_ids(
+        obj=event2, valid_ids=match_ev2)
+
+    return event1_matched, event2_matched
+
+
+def cut_block_by_epochs(block, properties=None, reset_time=False):
+    """
+    This function cuts Neo Segments in a Neo Block according to multiple Neo
```
apdavison: There is no need to prefix the object names by "Neo" in this context!

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+                seg, epoch=epoch, reset_time=reset_time)
+            block.segments += new_segments
+
+        block.segments.remove(seg)
+    block.create_many_to_one_relationship(force=True)
+
+
+def cut_segment_by_epoch(seg, epoch, reset_time=False):
+    """
+    Cuts a Neo Segment according to a neo Epoch object
```
apdavison: Same comment as above about "Neo"

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+
+    event1_matched = _event_epoch_slice_by_valid_ids(
+        obj=event1, valid_ids=match_ev1)
+    event2_matched = _event_epoch_slice_by_valid_ids(
+        obj=event2, valid_ids=match_ev2)
+
+    return event1_matched, event2_matched
+
+
+def cut_block_by_epochs(block, properties=None, reset_time=False):
```
apdavison: This function seems to be doing two things: (1) getting epochs from a block (same as is done by `get_epochs()` above, no?), (2) cutting the block based on the epoch.

For maintainability, testability and consistency with `cut_segment_by_epoch()` below, I think it would be better to simplify this to only (2), so instead of 
```python
result = cut_block_by_epochs(block, some_properties)
```
you would write
```python
epochs = get_epochs(block, some_properties)
result = cut_block_by_epochs(block, epochs)

File: neo/utils.py [hunk 8]
```diff
@@ -0,0 +1,856 @@
...
+        for key, val in epoch.array_annotations.items():
+            if len(val):
+                subseg.annotations[key] = copy.copy(val[ep_id])
+
+        segments.append(subseg)
+
+    return segments
+
+
+def seg_time_slice(seg, t_start=None, t_stop=None, reset_time=False, **kwargs):
```
apdavison: I think this would be more natural/consistent as a method of the `Segment` class (since `AnalogSignal` already has a `time_slice()` method).

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+    reset_time: bool
+        If True the times stamps of all sliced objects are set to fall
+        in the range from 0 to the duration of the epoch.
+        If False, original time stamps are retained.
+        Default is False.
+
+    Keyword Arguments:
+    ------------------
+        Additional keyword arguments used for initialization of the sliced
+        Neo Segment object.
```
apdavison: Same comment as above re: "Neo"

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+        if reset_time:
+            ep_time_slice = shift_epoch(ep_time_slice, t_shift)
+        # appending only non-empty epochs
+        if len(ep_time_slice):
+            subseg.epochs.append(ep_time_slice)
+
+    return subseg
+
+
+def shift_spiketrain(spiketrain, t_shift):
```
apdavison: This should be a method of the `SpikeTrain` class

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+        SpikeTrain is not modified).
+    """
+    new_st = spiketrain.duplicate_with_new_data(
+        signal=spiketrain.times.view(pq.Quantity) + t_shift,
+        t_start=spiketrain.t_start + t_shift,
+        t_stop=spiketrain.t_stop + t_shift)
+    return new_st
+
+
+def event_time_slice(event, t_start=None, t_stop=None):
```
apdavison: Should be a method `time_slice` of the Event class

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+
+    valid_ids = np.where(np.logical_and(
+        event.times >= t_start, event.times < t_stop))[0]
+
+    new_event = _event_epoch_slice_by_valid_ids(event, valid_ids=valid_ids)
+
+    return new_event
+
+
+def epoch_time_slice(epoch, t_start=None, t_stop=None):
```
apdavison: Should be a method time_slice of the Epoch class

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+
+    valid_ids = np.where(np.logical_and(
+        epoch.times >= t_start, epoch.times + epoch.durations < t_stop))[0]
+
+    new_epoch = _event_epoch_slice_by_valid_ids(epoch, valid_ids=valid_ids)
+
+    return new_epoch
+
+
+def shift_event(ev, t_shift):
```
apdavison: Should be a method `time_shift` of the Event class

neo/utils.py
```diff
@@ -0,0 +1,856 @@
...
+    Returns:
+    --------
+    epoch: Neo Event
+        New instance of an Event object starting at t_shift later than the
+        original Event (the original Event is not modified).
+    """
+    return _shift_time_signal(ev, t_shift)
+
+
+def shift_epoch(epoch, t_shift):
```
apdavison: Should be a method `time_shift` of the Epoch class

neo/utils.py
Kleinjohann: Yes, I agree. I changed get_epochs as well.

neo/utils.py
Kleinjohann: Done.

neo/utils.py
Kleinjohann: I went through the whole file and made sure that I consistently call the actual objects Epochs and Events while referring to single events/epochs they contain by the lower-case names.

neo/utils.py
Kleinjohann: I went through the whole file and removed all unnecessary prefixes.

neo/utils.py
Kleinjohann: I agree, all the shift/slice functions should be moved to the respective classes. Should I do that in this PR or in a separate one?

neo/utils.py
Kleinjohann: The idea behind this is that a Segment is only cut by Epochs which belong to that Segment. For each Segment, we have to go through all of its Epochs and filter and slice them according to the requested properties. Only these selected Epochs are used to cut that same Segment. If the Epoch extraction is a separate step we have to somehow keep track of which Segment each Epoch comes from. Since the extracted Epochs are duplicates or slices of existing ones they are not attached to the neo structure and this metadata is lost.

neo/utils.py
Kleinjohann: Is there a utility function to attach dataobjects to a Segment? I know it does not require much code but I always found it pretty convenient to have the resulting Epoch properly integrated into the neo structure and I'm not sure how easy it is for users to discover how to do this themselves.

neo/utils.py
apdavison: If you'd prefer to do that in a separate PR, that's fine with me.

PR #662, opened 2019-04-05
Fix Nix Ignoring Array Annotations

Kleinjohann:
The NixIO currently just ignores array annotations. I went through it with @achilleas-k and added them analogously to annotations. For writing this is straightforward, when reading I check the shape of all annotations and turn them into array annotations if it matches.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -406,23 +406,15 @@ class NixIO(BaseIO):
             sig_length = 1
 
         array_annotations = {}
-
         if sig_length > 1:
             for attr_key, attr_val in neo_attrs.items():
                 if isinstance(attr_val, (list, np.ndarray, pq.Quantity)):
                     if isinstance(attr_val, (np.ndarray, pq.Quantity)) and attr_val.shape == ():
                         attr_val = attr_val.flatten()
                     if len(attr_val) == sig_length:
-                        if isinstance(attr_val, list) or (isinstance(attr_val, np.ndarray) and not (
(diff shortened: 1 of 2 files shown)


neo/io/nixio.py
```diff
@@ -397,9 +397,28 @@ def _nix_to_neo_analogsignal(self, nix_da_group):
         else:
             t_start = create_quantity(timedim.offset, timedim.unit)
 
+        # find annotations with length equal to signal length and turn them into array-annotations
```
apdavison: This block of code (until `del neo_attrs[key]`) is repeated many times. For maintainability, it would make sense to pull it out as a function.

JuliaSprenger:
Instead of guessing the type of the `annotation` by it's length it would be safer to introduce an additional flag for this in the nix file to be sure to recover saved `array_annotations` properly.

neo/io/nixio.py
Kleinjohann: I went with Julia's suggestion (https://github.com/NeuralEnsemble/python-neo/pull/662#issuecomment-481376527) which makes this entire block unnecessary.

JuliaSprenger:
I had a look at it and the updated version also works for me. Depending on the result of #659, the implementation of the writing of `array_annotations` can be generalized (e.g. not checking for `labels` and `durations` explicitly when writing objects), but for now this is a good solution.

JuliaSprenger:
@Kleinjohann: I would like to merge this one now. Could you resolve the 4 conflicting lines?

PR #663, opened 2019-04-08
Fix for #661

apdavison:
#661

PR #664, opened 2019-04-08
Fix for #661

### LINKED ISSUE #661: segments[0].filter recursion if String key not found.
example:script  (taken from doc except for last line)
```python
from neo.core import (Block, Segment, ChannelIndex, AnalogSignal)
from quantities import nA, kHz
import numpy as np

blk = Block()
for ind in range(3):
     seg = Segment(name='segment %d' % ind, index=ind)
     blk.segments.append(seg)

for ind in range(2):
     chx = ChannelIndex(name='Array probe %d' % ind,
                        index=np.arange(64))
     blk.channel_indexes.append(chx)

for seg in blk.segments:
     for chx in blk.channel_indexes:
         a = AnalogSignal(np.random.randn(10000, 64)*nA,
                          sampling_rate=10*kHz)
         chx.analogsignals.append(a)
         seg.analogsignals.append(a)

print(blk)
blk.segments[0].filter("foo")
```
-
Clearly the filter("foo") should fail but NOT with the error 
RecursionError: maximum recursion depth exceeded while calling a Python object

-
We suspect this is called because the filter checks if the key is an iterable and then check each element separately.  Which is a great idea until you remember that a string is an iterable so on not finding "foo" it tries "f"  which is still and iterable with one element "f" ............

apdavison:
Raise a meaningful error if `filter()` is called with a single string rather than dict/kwargs

Fixes #661

### DIFF
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -8,6 +8,10 @@ object inherit from.  It provides shared methods for all container types.
 
 # needed for python 3 compatibility
 from __future__ import absolute_import, division, print_function
+try:
+    basestring
+except NameError:
+    basestring = str
 
 from neo.core.baseneo import BaseNeo, _reference_name, _container_name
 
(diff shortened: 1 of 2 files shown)


apdavison:
Replaces #663

JuliaSprenger:
I tested this also including code from #661 and it works nicely. I will merge this.

Issue #666, opened 2019-04-16
Can't write coordinates of channel indexes with NixIO


rgutzen:
Neo version = 0.7.1
NixIO version = 1.5.0b3
Quantities version = 0.12.1

```
import neo
import numpy as np
import quantities as pq

bl = neo.Block()
bl.channel_indexes.append(neo.ChannelIndex(np.array([1]), name='testchx'))
bl.channel_indexes[0].coordinates = (10, 20)*pq.mm

with neo.io.NixIO('tmp.nix', 'ow') as nixio:
    nixio.write_block(bl)
```

causes the following error message:

```
------------------------------------------
IndexErrorTraceback (most recent call last)
<ipython-input-25-a1dd10c3586c> in <module>
      8 
      9 with neo.io.NixIO('tmp.nix', 'ow') as nixio:
---> 10     nixio.write_block(bl)

~/anaconda3/envs/py36/lib/python3.6/site-packages/neo/io/nixio.py in write_block(self, block, use_obj_names)
    561         # descend into ChannelIndexes
    562         for chx in block.channel_indexes:
--> 563             self._write_channelindex(chx, nixblock)
    564 
    565         self._create_source_links(block, nixblock)

~/anaconda3/envs/py36/lib/python3.6/site-packages/neo/io/nixio.py in _write_channelindex(self, chx, nixblock)
    611             if chx.coordinates is not None and len(chx.coordinates):
    612                 coords = chx.coordinates[idx]
--> 613                 coordunits = stringify(coords[0].dimensionality)
    614                 nixcoords = tuple(c.magnitude.item() for c in coords)
    615                 chanprop = chanmd.create_property("coordinates", nixcoords)

~/anaconda3/envs/py36/lib/python3.6/site-packages/quantities/quantity.py in __getitem__(self, key)
    362     @with_doc(np.ndarray.__getitem__)
    363     def __getitem__(self, key):
--> 364         ret = super(Quantity, self).__getitem__(key)
    365         if isinstance(ret, Quantity):
    366             return ret

IndexError: too many indices for array
```

achilleas-k:
I see the problem.  While getting the dimensionality/unit of the coordinates, the IO assumes the `coordinates` property is 2-dimensional.

```python
    coords = chx.coordinates[idx]
    coordunits = stringify(coords[0].dimensionality)
```

[...]

PR #667, opened 2019-04-23
[nixio] Support 1D coordinates for single ChannelIndex

achilleas-k:
This PR adds support 1-dimensional ChannelIndex coordinates.  1D coordinates in ChannelIndex can occur when there is only one channel and the coordinates are specified as a vector (see issue #666).  When 1D coordinates are found they are wrapped in a list and treated the same as the general case.


apdavison:
The failing tests are unrelated to this PR, and are fixed in master.

PR #538, opened 2018-05-16
[WIP] Deep copy annotations when using time_slice

JuliaSprenger:
This aims at issue #536 by ensuring a deep copy of the annotation dictionary when using time_slice

PR #668, opened 2019-05-13
Deep copy annotations when using time_slice

bjoern1001001:
When time slicing an object, so far annotations and array_annotations have not been deepcopied. 
#538 aimed to fix this by deepcopying all attributes when using time_slice. 
This pull request implements the latest changes into #538 and fixes merge conflicts.

The current implementation does work, so it is in theory ready to be merged (except that tests are missing).
However I tried using deepcopy to achieve this task in a simpler way. What I noticed is that this does not work, because deepcopy also copies e.g. the segment that an analogsignal is linked to. 
So my question is whether this is desired behavior or deepcopy should only copy the object itself and leave references intact?
I was rather expecting the latter behavior, i.e. that parent objects will not be deepcopied.


JuliaSprenger:
Hi @bjoern1001001. Thanks for taking care of this.
In my opinion the references should not be kept when deep copying, since this would lead to unidirectional linking between objects. To avoid this it would be better to remove all references for a deep copy.
Do you have time to also add some tests to this PR?

bjoern1001001:
That's a good point. So what you're saying is that after time_slice and after deepcopy any attributes of the result referencing a parent, e.g. analogsignal.segment should simply be None? 

Yes, I'll add some tests.

samuelgarcia:
I think I agree with Julia.

bjoern1001001:
Thank you both for your feedback.

I just implemented the behavior you suggested and noticed that, at least for analogsignal, there are explicit tests requiring the opposite, e.g. `test__deepcopy_should_let_access_to_parent_objects` and `test__time_slice_should_let_access_to_parent_objects`. So it seems like there is a reason it is implemented in this way. Does anybody know more about this?

The current version that I pushed should have precisely the behavior that I described in my last comment. So if a data object is deepcopied, references to parents will be set to None. And if a container object is deepcopied, this copy will be recursive and all children will be deepcopied as well and all references set correctly within the deepcopied container.

This makes the implementation for time_slice a lot easier and more consistent with less duplicate code.

Tests for the deepcopy behavior are not yet implemented, but you can check out the code and try for yourself if you want to.

apdavison:
I think we've never clearly agreed on what should happen to parent references when copying and slicing objects. I think experience has shown, however, that we should not copy such references, and we should require users to explicitly re-add objects to the relevant container if that's what they want. We should therefore invert and rename the tests you mentioned in your last comment.

We should also ensure that this behaviour is clearly documented.

bjoern1001001:
Thanks for the input @apdavison.

I have now finished implementing tests and added docstrings to the __deepcopy__ methods that clearly state the current behavior.

In my opinion this pull request is thus finished, so you can have a look and see whether the implementation is what you were expecting or I should change anything.


On a somewhat related topic: I'm assuming a shallow copy should not keep the references to parents either then? If so, shallow copy would probably have to be overridden for all objects.
Should this be done or is the current behavior of shallow copies fine?

apdavison:
The test failure with Python3 seems to be unrelated to this PR (see https://stackoverflow.com/questions/55293711/what-to-use-instead-of-csv2rec)

PR #675, opened 2019-06-05
Turn time-slicing and -shifting utilities into methods of the corresponding classes

Kleinjohann:
As suggested by @apdavison in #657 I turned utility functions that operate on single objects into methods of the respective classes. Time-shifting and -slicing deep-copies the object and thus destroys links to parents as discussed in #668 .

I circumnavigated import-cycles by identifying proxy-objects by their `_rawio` attribute instead of an explicit `isinstance`. I'm open to suggestions on how to do this more elegantly.

While adding `IrregularlySampledSignals` to `Segment.time_slice` I noticed that there is no proxy-object for them. The code should still work if and when they are added.

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -483,6 +483,26 @@ class AnalogSignal(BaseSignal):
 
         return obj
 
+    def time_shift(self, t_shift):
+        """
+        Shifts a :class:`AnalogSignal` to start at a new time.
+
+        Parameters:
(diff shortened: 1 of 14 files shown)


apdavison:
hi @Kleinjohann the changes look great, but after merging an older PR there are now conflicts. Please could you take a look at them?

Kleinjohann:
@apdavison should be fine now

PR #676, opened 2019-06-10
Fix merge behaviour

apdavison:
When merging two container objects, the children merged from the other container into the target container should have the appropriate parent attribute set to reference the target.

Example:

After `seg1.merge(seg2)`, we expect that all the AnalogSignal objects in `seg1.analogsignals` should have `sig.segment == seg1`, whether `sig` was originally in `seg1.analogsignals` or `seg2.analogsignals`.

After the operation, `seg2` will be in an inconsistent state, but I think this is ok, as I wouldn't normally expect to continue using an object after merging it into another. The alternative is to take a copy of `seg2` before merging it, but this would likely have a bad effect on the performance of `merge()` for little benefit.

### DIFF
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -562,6 +562,9 @@ class Container(BaseNeo):
 
         Annotations are merged such that only items not present in the current
         annotations are added.
+
+        Note that the `other` object will be in an inconsistent state after the
+        merge operation and should not be used further.
         """
         # merge containers with the same name
         for container in (self._container_child_containers +
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -585,8 +585,8 @@ class Container(BaseNeo):
             ids = [id(obj) for obj in objs]
             for obj in getattr(other, container):
                 if id(obj) in ids:
-                    continue
-                if hasattr(obj, 'merge') and obj.name is not None and obj.name in lookup:
+                    pass
+                elif hasattr(obj, 'merge') and obj.name is not None and obj.name in lookup:
                     ind = lookup[obj.name]
                     try:
                         newobj = getattr(self, container)[ind].merge(obj)
@@ -598,6 +598,7 @@ class Container(BaseNeo):
                     lookup[obj.name] = obj
                     ids.append(id(obj))
                     getattr(self, container).append(obj)
+                obj.set_parent(self)
 
         # use the BaseNeo merge as well
         super(Container, self).merge(other)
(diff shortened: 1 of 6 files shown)


samuelgarcia:
The solution is OK for me.
If a user really want a copy then it have to be done before the merge.
I merge it.

samuelgarcia:
No I wait for @JuliaSprenger comments.

neo/core/container.py
```diff
@@ -562,6 +562,9 @@ def merge(self, other):
 
         Annotations are merged such that only items not present in the current
         annotations are added.
+
+        Note that the `other` object will be in an inconsistent state after the
+        merge operation and should not be used further.
```
JuliaSprenger: It's not clear to me in which cases the `other` objects in corrupted after the merge. Is there a way to prevent this? Otherwise I think many users will miss this comment and run into confusion.

apdavison: if we take the example of two `Segments` A and B, each containing some `AnalogSignals`, and we perform
```
A.merge(B)
```
we add all the signals in `B.analogsignals` to `A.analogsignals`, and we change the `signal.segment` attribute of all the signals in `B.analogsignals` to point to `A`.

The question is then what to do with `B.analogsignals`. We could:

1. empty it. This means that `A.merge(B)` has a side-effect on `B`.
2. leave it unchanged. This means that `B.analogsignals` references signals which do not point back to `B`. This is what I mean by "inconsistent".
3. copy `B` before merging. This has a potentially large memory and performance overhead.

I went with option 2. Option 1 would also be ok, but also has its problems. To minimize confusion, we could add an option like `preserve_other` or `copy`, which would give the user the choice of doing either 2 or 3. In that case, what should the default be: copy or not?

JuliaSprenger: Thanks for the clarification. I was thinking the object itself  or child data objects would be inconsistent, but yes, for the links this makes more sense. Maybe change the comment to *Note that the `other` object will be linked inconsistently to other Neo objects after the  merge operation and should not be used further.*
I like your suggestion of adding an option for this. I think it would be most user friendly. Intuitively I would expect option 1) to be the default or would even go beyond, as merging of 2 Segments should only result in a single remaining Segment, i.e. Segment B will be removed.

apdavison: ok. I've changed the comment as you suggest.

If you can merge this PR I'll create a new issue for the remaining changes.

JuliaSprenger:
Ok, then we merge this now and continue discussing in another issue.

PR #678, opened 2019-06-12
read all .continuous files

rishidhingra:
fixes issue where openephysrawio reader ignores aux and adc .continuous files specified by the current/old open ephys format.

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -484,7 +484,7 @@ def explore_folder(dirname):
         order = [list.tolist() for list in order]
         for i, sublist in enumerate(order):
             if i > 0:
-                order[i] = [x+max(order[i-1])+1 for x in order[i]]
+                order[i] = [x+max(order[i-1]) + 1 for x in order[i]]
         order = [item for sublist in order for item in sublist]
         continuous_filenames = [continuous_filenames[i] for i in order]
         info['continuous'][seg_index] = continuous_filenames
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -288,7 +288,7 @@ class OpenEphysRawIO(BaseRawIO):
         block_stop = i_stop // RECORD_SIZE + 1
         sl0 = i_start % RECORD_SIZE
         sl1 = sl0 + (i_stop - i_start)
-        
+
         if channel_indexes is None:
             channel_indexes = slice(None)
         channel_indexes = np.arange(self.header['signal_channels'].size)[channel_indexes]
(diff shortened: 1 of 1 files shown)


neo/rawio/openephysrawio.py
```diff
@@ -87,13 +88,12 @@ def _parse_header(self):
...
             all_last_timestamps = []
             all_samplerate = []
-            for continuous_filename in info['continuous'][seg_index]:
+            for chan_id, continuous_filename in enumerate(info['continuous'][seg_index]):
                 fullname = os.path.join(self.dirname, continuous_filename)
                 chan_info = read_file_header(fullname)
 
                 s = continuous_filename.replace('.continuous', '').split('_')
                 processor_id, ch_name = s[0], s[1]
-                chan_id = int(ch_name.replace('CH', ''))
```
samuelgarcia: I don't anderstand this.

In neo rawio, the chan_id is something that match the original file format. Sometimes is is one based, sometimes can contains gaps.

Here, with enumerate it is more a chan_index.

In rawio the chan_index is the position, inside the "sig_channels" vector.

samuelgarcia:
Hi @rishidhingra.
Thank you for the patch.
I think I don't anderstand it. Sorry.
Could you send me trhe file that trigegr the bug ?
And make a description of the problem with more detail please ?

rishidhingra:
Hi Samuel,

Let me try explaining better. The bottom line is that there are more .continuous files produced by the openephys format than just the ones from the probes (which are named like 100_CH1.continuous, etc.), in particular the 100_ADC1.continuous files are produced when an experimenter uses the breakout board to record signals from other devices in parallel to the probe data. There are also files with an AUX prefix that can come from the accelerometers on some headstages (for recording animal position in awake recordings).

My patch accepts any of these prefixes, rather than only the CH (MEA) channel prefix. You can reproduce the bug by simply renaming a few .continuous files with the other prefixes (ADC or AUX). 

Apologies for not giving a more detailed explanation initially.

Best,
Rishi

________________________________

neo/rawio/openephysrawio.py
samuelgarcia: Could we go back to chan_id even if chan_id can in some case have multiple entries.
 ```
            chan_str = re.split(r'(\d+)', s[1])[0]
            chan_id = int(ch_name.replace(chan_str, ''))
 ```
Here it is a channel index which is redundant to the channel position.

samuelgarcia:
Hi @rishidhingra.
Ok I have read the PR.
It is Ok for me.

It would be would very good to have a dataset (very light)  [here](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data).

Could you:
  * fix pep8 
  * go back to 
       ```
       chan_str = re.split(r'(\d+)', s[1])[0]
       chan_id = int(ch_name.replace(chan_str, ''))
       ```

     Instead  of chan_id becoming a global index.
     It is cool for the end user to retrieve the same number as they have on the device.
     Even if we can have 2 chan_id=0 (CH0 and AUX0)

So we could include, in the l release 0.8.

samuelgarcia:
Hi @rishidhingra.
I have made some change directly on your branch could you have a look ?

There was a mix between chan_id and chan_index.
Now chan_id is clearly the XX CHXX so it is not unique when there is AUX.

And the chan_index is the global count index.

is it OK for you ?

samuelgarcia:
@rishidhingra : for me this PR is ready to merge.
Could you make a last check on your dataset ?

PR #681, opened 2019-06-29
switch to tuples in chan_id in neuralynxrawio.py

sin-mike:
@samuelgarcia 

As we discussed in #591, I've switched to tuples in memmaps, because channel IDs are not unique for the acquisition, but the channel names are. The thing is that NLX allows to set different preprocessing parameters for each channel ID and create new files, e.g. one file contains highpassed data for spikes, second - true DC recording, third - common channel with LFP 0.1Hz-100Hz, etc. And they all could be collected from the same channel ID!

Note, CSC1 and CSC33 have the same ID, but different gains (also different highpass parameters)
image

An example of simultaneous recording of DC and AC from the same channel ID
image

Now, memmaps and dicts have the following structure:
image

### DIFF
--- a/doc/source/authors.rst
+++ b/doc/source/authors.rst
@@ -43,8 +43,8 @@ and may not be the current affiliation of a contributor.
 * Jeffrey Gill [18]
 * Lucas (lkoelman@github)
 * Mark Histed
-* Mike Sintsov
-* Scott W Harden [19]
+* Mike Sintsov [19]
+* Scott W Harden [20]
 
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Hi @sin-mike.
This is totally perfect.
It was exatly the solution I hd in mind.
I am really sorry of being so few active on this issue.

Could you fix the unitest that I guess still access with id in the dict ?
See travis.

best.

sin-mike:
@samuelgarcia all done!

PR #689, opened 2019-07-14
Fix for duration of signals in `generate_one_simple_segment`

jpgill86:
The default `sampling_rate` has units kHz, and the default `duration` has units s. By dropping units from their product without first simplifying, the number of data points was off by a factor of 1000. Consequently, the default 6 second signal was instead 6 milliseconds long.


samuelgarcia:
Good catch!!
OK for me.

PR #693, opened 2019-07-16
fix nlx segment search in case of faulty blocks

sin-mike:
Guys, I found another bug

So, I have a record in which there are some blocks which are logged partially (faulty block). When you searched for a segment with diff in timestamps, you assumed, that all blocks are correct, but not in this case:

image

It results in a faulty detection of segment boundaries, where you can see the next segment starts before the end of the previous segment due to faulty blocks:
image

If we look at the exact signals, we will see that a part of the faulty block repeats ADC values stored from the previous block:
image

The same blocks, but reconstructed in absolute times:
image

My solution was to include faulty blocks to the detection procedure of segments

Now the timing is correct:
image

PS. I must admit though, that the solution is not perfect, as, in cases of faulty blocks, I should have fixed the reader itself, rather than to oversegment the record. However, this is a quick and dirty hack allows me to process the data correctly now.

PS2. Looking back, I now understand that the faulty block doesn't affect the processing, but the incorrect timing of segments just made me feel suspicious of what inside the hood of the segment detection. Now I'm absolutely sure the origin of the faulty timing is not dangerous


samuelgarcia:
Hi @sin-mike 
Thank you very much.
This investigation is totally amazing.
Do you have a quite small file that could be included in gin.gnode so test with CI could be lauched on it ?

I am not a user of this file format.
Maybe @JuliaSprenger or @mdenker could have a look on this PR ?

sin-mike:
> Hi @sin-mike
> Thank you very much.
> This investigation is totally amazing.
> Do you have a quite small file that could be included in gin.gnode so test with CI could be lauched on it ?
> 
> I am not a user of this file format.
> Maybe @JuliaSprenger or @mdenker could have a look on this PR ?

The file I worked with during this discovery is 150Mb, I don't have any other at the moment

samuelgarcia:
Outch. this is too big.

sin-mike:
@samuelgarcia 
I'm looking for publishing another PR above this one. Have you decided, if you're merging this one, or are there other updates required?

JuliaSprenger:
@sin-mike: Do you think it would be possible to cut the file you have been working with to get a smaller file for the tests? I think this would help a lot in verifying the correct handling of the wrong time stamps.

samuelgarcia:
@sin-mike next Tuesday and Wendsay we will review and merge many PR.
Normally this one should be merge.
I have to admit that I did not anderstood yet thepatch I would like to understand it before merge.

If you have another PR you can make it. Just branch on that one and clearly writte on the PR that it depend on that one. So you can work independently that us.
Sorry to be slow for reviewing and merging.

sin-mike:
> @sin-mike: Do you think it would be possible to cut the file you have been working with to get a smaller file for the tests? I think this would help a lot in verifying the correct handling of the wrong time stamps.

Tough one. I did never made this. I'll try. Could you suggest the tools?

sin-mike:
@JuliaSprenger 
I have it, about 8Mb, which exhibits 2 failed segments + events

image

image

image

where to store the file?

samuelgarcia:
Hi Mike.
we have a repo for test files here https://gin.g-node.org/NeuralEnsemble/ephy_testing_data
It is like github you can submit file there.

If you can put the file I could read more easily this PR and merge it for the release 0.8 in preparation.

neo/rawio/neuralynxrawio.py
```diff
@@ -440,46 +440,67 @@ def read_ncs_files(self, ncs_filenames):
...
+            # but for a file I have found many deltas0==15999, 16000, 16001 (for sampling at 32000)
             # I guess this is a round problem
             # So this is the same with a tolerance of 1 or 2 ticks
-            mask = deltas0 != good_delta
-            for tolerance in (1, 2):
-                mask &= (deltas0 != good_delta - tolerance)
-                mask &= (deltas0 != good_delta + tolerance)
+            max_tolerance = 2
+            mask = np.abs((deltas0 - good_delta).astype('int64'))>max_tolerance
+
```
samuelgarcia: More easy to read. The previous implementation was a bit faster in my memory but no sure.
In any case, it is better to have something readable.

apdavison:
Decided we will merge this without a test file if it is blocking the release, but it would be much better to have the test file beforehand.

neo/rawio/neuralynxrawio.py
sin-mike: > More easy to read. The previous implementation was a bit faster in my memory but no sure.
> In any case, it is better to have something readable.

@samuelgarcia  do you mean, I should've left the original version?

sin-mike:
> Decided we will merge this without a test file if it is blocking the release, but it would be much better to have the test file beforehand.

@apdavison I've uploaded a file https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/11/files

samuelgarcia:
Hi Mike.
I have merged files at gin gnode.
Could you add this files in test_neuralynrawio.py and test_neuralynxio.py ?

neo/rawio/neuralynxrawio.py
samuelgarcia: No I mean it is perfect because more readable.

sin-mike:
@samuelgarcia looks like the tests failed, but in nixio, rather than in neuralynx, after I pulled from master

```ERROR: test_write_proxyobjects (neo.test.iotest.test_nixio.NixIOWriteTest)```

samuelgarcia:
@sin-mike : there is a bug in master that block all PR. We have wait and rerun circle when will be fixed.
Thank you for being very reactive on this.
I guess this will be included in the relase 0.8/

samuelgarcia:
@sin-mike : the bug in the master is normaly resolved.
Could you rebase to master that would rerun the ci ?

samuelgarcia:
Close by #726

PR #695, opened 2019-07-18
[nixio] Discard 'neo_name' property on read

### LINKED ISSUE #694: Duplicate neo_name and nix_name when saving object to new nix file
Neo objects are assigned unique identifiers (_neo_name_, _nix_name_) when saved with the NixIO (see #331). However, because these identifiers are required to be unique, this leads to an error when a loaded neo object is saved to a different file.

**Minimal example:**
```
import neo

with neo.NixIO('/path/to/file.nix', 'ro') as io:
    img_block = io.read_block()

# Perform some edit which keeps the annotations 'neo_name' and 'nix_name'
# of the block and/or its child objects.

with neo.NixIO('/path/to/new_file.nix', 'ow') as io:
    io.write(img_block)
```
neo: 0.7.2
nixio: 1.5.0b1

**Error message:**
`DuplicateName: Duplicate name - names have to be unique for a given entity type & parent. (create_property)`

A workaround is to manually delete the respective annotations before saving, but since loading an object and saving it to another file is a rather common operation, this probably should be handled by the NixIO itself. A solution, for example, could be to have a boolean argument _ignore_duplicate_names_ in the write function or a check which tests whether the object is written to an existing or a new file.

achilleas-k:
Small modification of the naming scheme described in PR #331:
In step 4, upon reading the 'neo_name' property and restoring the original Neo object's name, the 'neo_name' attribute is now discarded.  Since we already restore this to the name of the object, there's no need to keep it around in the object's annotations.  More importantly, it would create a conflict on rewrite of an object, since the creation of the property is handled explicitly, outside of the general annotation conversion function.

Fixes #694

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -1160,7 +1160,9 @@ class NixIO(BaseIO):
                 else:
                     values = list(values)
                 neo_attrs[prop.name] = values
-        neo_attrs["name"] = stringify(neo_attrs.get("neo_name"))
+        # since the 'neo_name' NIX property becomes the actual object's name,
+        # there's no reason to keep it in the annotations
+        neo_attrs["name"] = stringify(neo_attrs.pop("neo_name", None))
 
         if "file_datetime" in neo_attrs:
             neo_attrs["file_datetime"] = datetime.fromtimestamp(


JuliaSprenger:
Thanks for the quick fix. I will merge this now.

Issue #530, opened 2018-05-03
AnalogSignal.time_slice() does not always return AnalogSignals with the same shape


grg2rsr:
I am trying to extract spike waveforms by slicing an AnalogSignal at time points from a SpikeTrain
the Problem: the returned slices are not always of the same shape. It seems like there is some rounding problem.

A test snipped to reproduce the problem
```python
import scipy as sp
from scipy import signal
import neo
import quantities as pq

# generate some random AnalogSignal
AnalogSignal = neo.core.AnalogSignal(sp.randn(25000)*pq.uV,t_start=0*pq.s,sampling_rate=25*pq.kHz)

# get the indices of all spikes
spikes_inds = signal.argrelmax(AnalogSignal > 3.25*pq.uV)[0]

# get the spike time points by slicing into AnalogSignal.times and make a SpikeTrain
spike_times = AnalogSignal.times[spikes_inds]
SpikeTrain = neo.core.SpikeTrain(spike_times,t_start=AnalogSignal.t_start,t_stop=AnalogSignal.t_stop,sampling_rate=AnalogSignal.sampling_rate)

# extracting spike shapes by AnalogSignal.time_slice
wsize = 3*pq.ms
expected_shape = int((wsize * AnalogSignal.sampling_rate).simplified.magnitude)

for t in SpikeTrain:
    Asig = AnalogSignal.time_slice(t-wsize/2, t+wsize/2)
    print(expected_shape,Asig.shape[0],Asig.duration.rescale(pq.s))
```
output:
```
75 75 0.003 s
75 74 0.00296 s
75 74 0.00296 s
75 76 0.00304 s
75 75 0.003 s
75 75 0.003 s
75 74 0.00296 s
75 76 0.00304 s
75 74 0.00296 s
75 76 0.00304 s
75 76 0.00304 s
75 76 0.00304 s
75 74 0.00296 s
75 75 0.003 s
75 76 0.00304 s
75 75 0.003 s
75 74 0.00296 s
75 75 0.003 s
75 74 0.00296 s
75 75 0.003 s
```

`expected_shape` is sometimes not equal `Asig.shape[0]` if for odd numbers of `wsize`.

note that this code example produces an unrelated error if the spikes fall too close to the borders of the time segment, but no handling of this is included for brevity reason. Just rerun it in that case.

[...]

PR #703, opened 2019-07-23
Time slice consistency

apdavison:
cf #530

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -446,9 +446,8 @@ class AnalogSignal(BaseSignal):
 
     def time_index(self, t):
         """Return the array index corresponding to the time `t`"""
-        t = t.rescale(self.sampling_period.units)
         i = (t - self.t_start) * self.sampling_rate
-        i = int(np.rint(i.magnitude))
+        i = int(np.rint(i.simplified.magnitude))
         return i
 
     def time_slice(self, t_start, t_stop):
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -466,6 +466,7 @@ class AnalogSignal(BaseSignal):
         # checking start time and transforming to start index
         if t_start is None:
             i = 0
+            t_start = 0 * pq.s
         else:
             i = self.time_index(t_start)
 
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -473,8 +473,8 @@ class AnalogSignal(BaseSignal):
         if t_stop is None:
             j = len(self)
         else:
-            duration = (t_stop - t_start).rescale(self.sampling_period.units)
-            j = i + int(np.rint(duration * self.sampling_rate))
+            delta = (t_stop - t_start) * self.sampling_rate
+            j = i + int(np.rint(delta.simplified.magnitude))
 
         if (i < 0) or (j > len(self)):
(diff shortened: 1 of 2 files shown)


neo/test/coretest/test_analogsignal.py
```diff
@@ -466,6 +466,31 @@ def test__time_slice_should_set_parents_to_None(self):
...
+                           sampling_rate=25 * pq.kHz)
+
+        window_size = 3.0 * pq.ms
+
+        expected_shape = int(np.rint((window_size * sig.sampling_rate).simplified.magnitude))
+
+        # test with random times
+        t_start = (window_size / 2).magnitude
+        t_stop = (sig.t_stop.rescale(pq.ms) - window_size / 2).magnitude
+        for t in np.random.uniform(t_start, t_stop, size=1000):
```
samuelgarcia: Maybe we could put a seed here or even better voltontary loop over a ambiguous float at sub sample resolution.

apdavison: I think random values without a seed are better in this case, because we'll get different values every time the test is run, and hence over time the tests will cover a much wider range of values.

JuliaSprenger: I think a seed makes more sense, because with a true random component in the tests (and potentially randomly failing tests) unittests get quite hard to debug.

neo/core/analogsignal.py
```diff
@@ -447,33 +447,38 @@ def _pp(line):
     def time_index(self, t):
         """Return the array index corresponding to the time `t`"""
         t = t.rescale(self.sampling_period.units)
```
JuliaSprenger: does t need to be rescaled here? and if so then the invers of the sampling_rate should be used, because this is what this is compared to later.
I think it would be best not to rescale here but do a `rescale('dimensionless') in the line below.

neo/test/coretest/test_analogsignal.py
apdavison: I agree with using a seed in cases where there is a small but finite chance of the test failing for certain random values. In this case, however, if the code is correct then the probability of failure  should be zero.

neo/core/analogsignal.py
apdavison: good idea. I've pushed a new commit

neo/test/coretest/test_analogsignal.py
JuliaSprenger: Ok, I see your point. Let's use the current implementation for now.

samuelgarcia: I am OK for no seed.

neo/core/analogsignal.py
```diff
@@ -446,34 +446,38 @@ def _pp(line):
...
         else:
             i = self.time_index(t_start)
 
         # checking stop time and transforming to stop index
         if t_stop is None:
             j = len(self)
         else:
-            j = self.time_index(t_stop)
+            duration = (t_stop - t_start).rescale(self.sampling_period.units)
+            j = i + int(np.rint(duration * self.sampling_rate))
```
samuelgarcia: Maybe here you can also use simplified, no ?
```j = i + int(np.rint((duration * self.sampling_rate)).simplified)```

apdavison: good idea. I've pushed a new commit. I think this can be merged now.

Issue #490, opened 2018-03-06
__getitem__ and scalar for neo objects


samuelgarcia:
here an issue from #454 by @bjoern1001001.

I have a question about this, because I have been working on `SpikeTrain.__getitem__` for the last days. It's rather general, but I noticed this here.

I noticed here that what you wrote works in Epoch for things like epoch[1], while it would raise an error in SpikeTrain, because when you create the new Epoch in `__getitem__`, times is a quantity scalar, NOT an array or Epoch. Creating a SpikeTrain like this in `__getitem__` via `SpikeTrain(times=super(...).__getitem, ...)` would raise an error here saying it's testing the length of an unsized object. Is it supposed to be okay that epochs are created with a scalar quantity instead of a quantity array? And if that is the case, shouldn't it be converted to an array then in order to ensure consistency?

A little example of what happens:

`epoch = neo.Epoch([1,2,3]*pq.s)`
`ep1 = epoch[1] # ep1 is now an Epoch with times=2*pq.s, not [2]*pq.s or array([2]*pq.s)`

This is because the following happens in `__getitem__`:
`times = super(...).__getitem__(1)` 
Times now becomes a scalar quantity, because numpy `__getitem__` returns a scalar instead of an array when called with int index. From this pq.Quantity.__getitem__ creates a quantity, because it is not yet a quantity.

`obj = Epoch(times=times, ...)`
This is not a problem, because there is no test about whether times is an array or not
 `...`
`return obj # This returns obj, with times=2*pq.s`

Should it be like this or should a test for this be introduced?

Apart from that I'd like to know if any of you has deeper knowledge of `__getitem__` because I noticed together with @JuliaSprenger that there sometimes a scalar quantity is returned instead of a SpikeTrain object. I've been trying to change this but it fails. Trying to copy what you did, I noticed that this works for Epoch, but not for SpikeTrain.

Again some code to illustrate what I mean:

What currently happens:
`st1 = new SpikeTrain(times=[1,2,3]*pq.s, ...)`
`testobj = st1[1]`
testobj is now a (scalar) quantity object, not a SpikeTrain
`print(testobject) # This prints 2 s`

`new_st = st1[0:1]`
new_st is a SpikeTrain, because numpy returns a SpikeTrain array
`print(new_st) # This prints [1] s` 

If one would do the same as you did in Epoch for SpikeTrain, it would raise an error
`st2 = st1[1] # This would raise an error`

What happens SpikeTrain `__getitem__`:

[...]

PR #706, opened 2019-07-23
Improve consistency between Epoch, Event and SpikeTrain

apdavison:
* all can be created with lists (provided the units parameter is provided), not just with arrays
* `obj.__getitem__` with a scalar argument returns a Quantity
* `obj.__getitem__` with a slice or array argument returns an object of the same type

cf #490

### DIFF
--- a/neo/core/epoch.py
+++ b/neo/core/epoch.py
@@ -58,10 +58,12 @@ class Epoch(DataObject):
               dtype='|S4')
 
     *Required attributes/properties*:
-        :times: (quantity array 1D) The start times of each time period.
-        :durations: (quantity array 1D or quantity scalar) The length(s) of each time period.
+        :times: (quantity array 1D, numpy array 1D or list) The start times
+           of each time period.
+        :durations: (quantity array 1D, numpy array 1D, list, or quantity scalar)
+           The length(s) of each time period.
            If a scalar, the same value is used for all time periods.
-        :labels: (numpy.array 1D dtype='S') Names or labels for the time periods.
+        :labels: (numpy.array 1D dtype='S' or list) Names or labels for the time periods.
 
     *Recommended attributes/properties*:
         :name: (str) A label for the dataset,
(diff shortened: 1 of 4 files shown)


samuelgarcia:
OK for me.

A question unrelated to this PR:
  * obj._durations is not removed when \_\_getitem\_\_ ?.

apdavison:
> * obj._durations is not removed when __getitem__ ?.

Probably it should be, I was just being lazy.

samuelgarcia:
Can we merge this ?

JuliaSprenger:
@bjoern1001001: Is this compatible with your ideas from a couple of month ago?

bjoern1001001:
@JuliaSprenger Yes, the described behavior is what I had in mind. 

I just had a quick look on the code, and I noticed a few things I'm not sure about:

- Why is there no _copy_data_complement after event.py line 227? Currently if a slice is provided, labels will not be copied in Event, only a reference is given to the new object. I think the expected behavior is creating a copy.
- Are labels and durations correctly sliced even after doing _copy_data_complement? (epoch.py line 209)

I did not go into depth here, so maybe my points are not valid, I just wanted to raise them and make sure you are aware.

apdavison:
@bjoern1001001 thanks for the feedback, and for spotting the missing `_copy_data_complement` in Event.

Note that labels and durations are no longer handled by `_copy_data_complement` (since #659 was merged), they are handled explicitly.

bjoern1001001:
Ah, thank you, I was not aware of that.

This means, however, that if `obj.__getitem__` is called with a slice or array argument, none of the array attributes (`times`, `labels`, `durations`) are copied in the current implementation. Instead, numpy will create views into these arrays. So, if an Event is sliced, and labels of the returned object are altered, the original Event will be changed, too. 
I think general consensus was to always copy when slicing, that's why the `array_annotations` are explicitly deepcopied here, and #668 does this explicitly for time_slicing.

apdavison:
hmmm, we seem to be in a mess with the copy vs view thing. One of the guiding principles of Neo is that data objects should behave the same way as plain NumPy arrays unless there is a good reason not to. Slicing an array gives a view; the same should be true (and is at present) for AnalogSignal, SpikeTrain, Epoch, Event, etc.

I think I wasn't thinking broadly enough when I merged #668. We decided to do a thorough review of objects lifecycles for the 0.9 release, so I suggest we merge this PR (which does increase the consistency of Event and Epoch wrt SpikeTrain and AnalogSignal, even if there are still inconsistencies within Neo more generally), and continue this discussion in a new ticket. For 0.9, we need to add a page to the documentation which explicitly describes the expected behaviour when slicing, shallow or deep copying, merging, concatenating, etc.

bjoern1001001:
I agree. This PR should be merged like this and the whole issue of copying or not should be discussed in depth on its own.

Issue #643, opened 2019-02-06
Annotation problem for Analogsignal in BaseFromRawIO


hkchekc:
If I am not mistaken, in the baserawio.py and proxyobjects.py, the annotations of the Analogsingal Object cannot be inherited from the rawio.raw_annotations (Unless the length of signal_channel is 1) :

https://github.com/NeuralEnsemble/python-neo/blob/b4c5950d85ed4cedf27caf7dd36aa41067ab01ab/neo/io/proxyobjects.py#L115

I want to know is that intentional and is there anyway for getting customized annotations through?

Thank you.

samuelgarcia:
In #461 (proxy and lazy) recently merge, what I did is to add array_annotations in IO.
In short, channel_names and channel_ids are now in array_annotations.
By nature, all annotation that concern all channels should be in array_annotations.

At rawio level, channel annotations are a dict for each channel.
So mapping the list of dict to of one dict (annotation) is obvious only when there is one channel.

For particular annotations that concern all channel, it is easy to create an array_annotation.
I did something similar for  it for ```ChannelIndex``` here https://github.com/NeuralEnsemble/python-neo/blob/b4c5950d85ed4cedf27caf7dd36aa41067ab01ab/neo/io/basefromrawio.py#L135
It should be easy to do the same in AnalogSignalProxy.
I can do it.

hkchekc:
will be great

> In #461 (proxy and lazy) recently merge, what I did is to add array_annotations in IO.
> In short, channel_names and channel_ids are now in array_annotations.
> By nature, all annotation that concern all channels should be in array_annotations.
> 
> At rawio level, channel annotations are a dict for each channel.
> So mapping the list of dict to of one dict (annotation) is obvious only when there is one channel.
> 
> For particular annotations that concern all channel, it is easy to create an array_annotation.
> I did something similar for it for `ChannelIndex` here
> 
> [python-neo/neo/io/basefromrawio.py](https://github.com/NeuralEnsemble/python-neo/blob/b4c5950d85ed4cedf27caf7dd36aa41067ab01ab/neo/io/basefromrawio.py#L135)
> 
> Line 135 in [b4c5950](/NeuralEnsemble/python-neo/commit/b4c5950d85ed4cedf27caf7dd36aa41067ab01ab)
>  keys = list(self.raw_annotations['signal_channels'][ind_abs[0]].keys()) 
> 
> It should be easy to do the same in AnalogSignalProxy.
> I can do it.

[...]

PR #709, opened 2019-07-24
Create array_annotations for AnalogSignalProxy

samuelgarcia:
See #643


neo/io/proxyobjects.py
```diff
@@ -588,3 +597,33 @@ def consolidate_time_slice(time_slice, seg_t_start, seg_t_stop, strict_slicing):
...
+    common_keys = None
+    for ind in global_channel_indexes:
+        keys = [k for k, v in sig_annotations[ind].items() if not \
+                    isinstance(v, (list, tuple, np.ndarray))]
+        if common_keys is None:
+            common_keys = keys
+        else:
+            common_keys = [k for k in common_keys if k in keys]
+
+    # this is redundant and done at other level
```
JuliaSprenger: this comment is either outdated or code needs to be changed here. Which of the two is correct?

JuliaSprenger: Or do you want to keep `name`, `channel_id` and `file_origin` here and remove the more specific code at the other location?

samuelgarcia: channel_names and channel_ids are done separatly directly from the header['signal_channel'].
So it is redundant here, this is why they are not included here.

file_origin in array_annoations have no sens. But they are push automatically for all IOs.
It is already in annotations.

samuelgarcia: Maybe I need to reformulate the comments.

JuliaSprenger: why does 'file_origin` not make sense? For some IOs all individual traces in an AnalogSignal might come from different files, e.g. NeuralynxIO.

samuelgarcia: because an array per channel of file origin is not util.
file_origin is more a scalar annoataions.
It repeat the same file_origin over channel.
No ?

JuliaSprenger: Not in the case of Neuralynx, where you have one file per channel.

samuelgarcia: OK.
I can put it if you want but for the main case it can add a quite heavy array (full X nb_channel) that can be some kb.

JuliaSprenger: But it's just an array of strings with length = nb_channels, so it shouldn't be heavy, no?

samuelgarcia: If full path is 150bytes > 1024 channel then it is 150kb.
Which is quite light compare to signal but quqite heavy for a metadata.... :)
As you want.
I not not sure that this metadatat need to be per channel.

JuliaSprenger: Can we implement it for sake of completeness now as array annotation and if this turns out to be too heavy in the future we can still revert?

samuelgarcia: OK done

Issue #710, opened 2019-07-24
grouping mode default should now be 'group-by-same-units'


samuelgarcia:
For smooth trnasition with very behavrio of neo (one channel = one AnalogSignal)
when have the option ```signal_group_mode = 'split-all'/'group-by-same-units'```

For each io the default mode can be diffrent because of this ```_prefered_signal_group_mode```
 
I think we should change this
```_prefered_signal_group_mode = 'split-all'```

to  this

```_prefered_signal_group_mode = 'group-by-same-units'```

for all IOs.

samuelgarcia:
Lets put a warning in 0.8 and change in 0.9

apdavison:
This changes the default behaviour of IO classes, so I think we should add a DeprecationWarning for now, and change the behaviour in 0.9

samuelgarcia:
Duplicate of #832

PR #715, opened 2019-07-24
Warning for signal group mode.

samuelgarcia:
#712

### DIFF
--- a/neo/io/basefromrawio.py
+++ b/neo/io/basefromrawio.py
@@ -108,7 +108,8 @@ class BaseFromRaw(BaseIO):
         if signal_group_mode is None:
             signal_group_mode = self._prefered_signal_group_mode
             if self._prefered_signal_group_mode == 'split-all':
-                self.logger.warning("the default signal_group_mode will be 'split-all' in next release")
+                self.logger.warning("the default signal_group_mode will'\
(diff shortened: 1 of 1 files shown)


neo/io/basefromrawio.py
```diff
@@ -107,6 +107,9 @@ def read_block(self, block_index=0, lazy=False, signal_group_mode=None,
 
         if signal_group_mode is None:
             signal_group_mode = self._prefered_signal_group_mode
+            if self._prefered_signal_group_mode == 'split-all':
+                self.logger.warning("the default signal_group_mode will'\
+                        'be 'split-all' in next release")
```
jpgill86: In #710, you said you want to change the default to ``group-by-same-units``, right? Isn't this warning wrong then?

JuliaSprenger: shouldn't this read `he default signal_group_mode will be 'group_by_same_units' in next release`?

jpgill86: Maybe you should also say "will change from 'split-all' to 'group-by-same-units' in the next release". Otherwise it isn't clear to non-expert users how they will be affected.

samuelgarcia: yes of course.
I totally stupid I wrote the contrary.

PR #722, opened 2019-07-29
Fixed handling of empty .nev files in NeuralynxRawIO

alafuzof:
Hi guys, I noticed a small bug in how empty or header only files are handled in NeuralynxRawIO. While empty .ncs files are skipped entirely and empty .nse and .ntt files are mapped to empty Numpy arrays, the current code attempts to memmap empty .nev files, which results in an error since the memmap offset is larger than the file size.

In contrast to .nse and .ntt files (see discussion on #553 ), I'm not convinced that empty .nev files are interesting enough to convey to the user, so I chose to handle them like the .ncs files, i.e. ignore them.

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -183,7 +183,7 @@ class NeuralynxRawIO(BaseRawIO):
 
                     if (os.path.getsize(filename) <= HEADER_SIZE):
                         self._empty_nev.append(filename)
-                        data = np.zeros((0,), dtype=dtype)
+                        data = np.zeros((0,), dtype=nev_dtype)
                         internal_ids = []
                     else:
                         data = np.memmap(filename, dtype=nev_dtype, mode='r', offset=HEADER_SIZE)
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -73,7 +73,7 @@ class NeuralynxRawIO(BaseRawIO):
         self._spike_memmap = {}
         self.internal_unit_ids = []  # channel_index > ((channel_name, channel_id), unit_id)
         self.internal_event_ids = []
-        self._empty_ncs = []  # this list contains filenames of empty records
+        self._empty_ncs_nev = []  # this list contains filenames of empty records
         self._empty_nse_ntt = []
 
         # explore the directory looking for ncs, nev, nse and ntt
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @alafuzof!
Thanks for noticing and fixing this!
I am still not sure if generating an empty Neo event object for an empty nev file wouldn't be better, because then the generated Neo structure would be independent of the actual data content of the files. @samuelgarcia: What is your current opinion on the topic?

samuelgarcia:
I think empty file should generate zeros size object. No ?
This patch remove the object I think. No ?

JuliaSprenger:
@samuelgarcia: I agree.
@alafuzof: Could you change the behaviour such that there are still empty neo events generated for empty nev files?

alafuzof:
It's certainly possible to place an empty array in the __nev_memmap_ array, but I'm not sure how an empty nev file should be handled in terms of the _event_channels_ and _internal_events_ids_ arrays. Right now those arrays are populated based on the events found in the event file, but obviously there are no events in an empty file...

samuelgarcia:
Could we postpone this for 0.9 ?
I will out for 3 weeks.

sin-mike:
@alafuzof I'm routinely working with nlx files
I've thought about this thing too, but I came to the conclusion, that when nev is empty, no records are available at all. If there is at least a block, cheetah puts "Record Start" and "Record End" events to nev. At least "Record End", because otherwise, you won't be able to properly close files with cheetah.
Could you please provide an example when Cheetah could normally generate empty nev file?

apdavison:
Hi @alafuzof: a lot of time has passed, and I guess everyone has forgotten about this PR. Please could you reply to @sin-mike's comment above, so we can move this towards a resolution?

apdavison:
We would like to fix this problem soon, but we think it is important to create an empty object rather than ignore it entirely (or at least have the option to ignore it). Would you like to make that change, then we can merge this?

samuelgarcia:
@alafuzof : I think for "nev" files you should do the same as  "nse" or  "ntt"


see here https://github.com/NeuralEnsemble/python-neo/blob/3ed56f84efe8e0b37abbecab3d9b09350c11b02b/neo/rawio/neuralynxrawio.py#L150

neo/rawio/neuralynxrawio.py
```diff
@@ -179,10 +180,14 @@ def _parse_header(self):
...
                     # each ('event_id',  'ttl_input') give a new event channel
                     self.nev_filenames[chan_id] = filename
-                    data = np.memmap(
-                        filename, dtype=nev_dtype, mode='r', offset=HEADER_SIZE)
-                    internal_ids = np.unique(
-                        data[['event_id', 'ttl_input']]).tolist()
+
+                    if (os.path.getsize(filename) <= HEADER_SIZE):
+                        self._empty_nev.append(filename)
+                        data = np.zeros((0,), dtype=dtype)
```
samuelgarcia, line 186: I don't the dtype variable.
I would do:
`dtype=dtype` > `dtype=nev_dtype`

JuliaSprenger: Yes, you are right. Fixed it in the latest commit.

samuelgarcia:
I think this is OK.
Lets merge.

PR #726, opened 2019-07-31
REBASED Nlx lost segments with lost samples 

sin-mike:
This is just a rebased version of #693 

The previous PR became eventually broken when I had reset it and used rebase instead of merge.

PR #728, opened 2019-07-31
Nlx pick channels

sin-mike:
Dear @samuelgarcia , I've added a function to facilitate the selection of channels, which we lack in our lab.

It is on top of the #726 as you suggested.

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -55,6 +55,7 @@ class NeuralynxRawIO(BaseRawIO):
     def __init__(self, dirname='', **kargs):
         self.dirname = dirname
         BaseRawIO.__init__(self, **kargs)
+        self._header = self.header
 
     def _source_name(self):
         return self.dirname
@@ -407,6 +408,26 @@ class NeuralynxRawIO(BaseRawIO):
         event_times /= 1e6
         event_times -= self.global_t_start
         return event_times
+
+
+    def pick_channels(self, channels):
+        """ 
+        Given a list of channels construct a header to read only selected channels
+        channels:
+            list of channel names
+            'all': all original channel names 
+        """
+
+        if (type(channels) == str) and channels == 'all':
+            self.header = self._header
+
+        sigs = np.array(list(filter(lambda x: x[0] in channels,
+                                    self._header['signal_channels'])))
+
+        self.header['signal_channels'] = sigs
+
+
+
 
     def read_ncs_files(self, ncs_filenames):
         """
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -55,7 +55,6 @@ class NeuralynxRawIO(BaseRawIO):
     def __init__(self, dirname='', **kargs):
         self.dirname = dirname
         BaseRawIO.__init__(self, **kargs)
-        self._header = self.header
 
     def _source_name(self):
         return self.dirname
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Why don't you use the new lazy object for channel selection in neo.io ?
https://neo.readthedocs.io/en/latest/io.html#lazy-option-and-proxy-object

Note that could can also load channel in rawio level with index, ids and names:
https://neo.readthedocs.io/en/latest/rawio.html#basic-usage

From what I anderstand, in this patch you hide some channel after parse.
I am not sure how it will be compatible for neo.ioNeuralynxIO.

Am I missing something ?

sin-mike:
> Why don't you use the new lazy object for channel selection in neo.io ?
> https://neo.readthedocs.io/en/latest/io.html#lazy-option-and-proxy-object
> 
> Note that could can also load channel in rawio level with index, ids and names:
> https://neo.readthedocs.io/en/latest/rawio.html#basic-usage
> 
> From what I anderstand, in this patch you hide some channel after parse.
> I am not sure how it will be compatible for neo.ioNeuralynxIO.
> 
> Am I missing something ?

@samuelgarcia my motivation was to be able to keep only the necessary channels for further manipulation. In addition, it keeps the order of the channels, if you want to use the order other than the lexicographical one

sin-mike:
@samuelgarcia 
So, I took a dive into the code and found some inconvenience, that people in my lab won't handle
E.g. the code from `analogsignal.load`
```
raw_signal = self._rawio.get_analogsignal_chunk(block_index=self._block_index,
                    seg_index=self._seg_index, i_start=i_start, i_stop=i_stop,
                    channel_indexes=self._global_channel_indexes[channel_indexes])
```
It uses `channel_indexes` from the global range, rather than actual signal names
How do you think, if I expand this to handle the actual channel names, would it work for you?

samuelgarcia:
I think this patch could be generalized to all rawio with an external function like this:
```python
full_channel_reader = MyRawIO(...)
# by index
some_channel_reader = reduce_channel_on_io(full_channel_reader, channel_indexes=[0,5,12])

# by name if unique
some_channel_reader = reduce_channel_on_io(full_channel_reader,
                                                         channel_names=['CH0', 'CH5'])
# by id if unique
some_channel_reader = reduce_channel_on_io(full_channel_reader, 
                                                 channel_ids=[1, 6, 12])

```
(the name of the function must be better :) )

And also, we could propagate this into neo.io with BaseFromRaw.

What do you think ?

samuelgarcia:
get_analogsignal_chunk already support index/name/ids.

sin-mike:
> ```python
> 
>                                                          channel_
> ```

@samuelgarcia  this is a bit bigger than I thought my patch would be. I need a timeout for a tour to the big code

sin-mike:
@samuelgarcia 
No, it was pretty fast, I understood your idea
I can start fixing with this one:
```
def channel_name_to_index(self, channel_names):
        """
        Transform channel_names to channel_indexes.
        Based on self.header['signal_channels']
        """
        ch = self.header['signal_channels']
        channel_indexes, = np.nonzero(np.in1d(ch['name'], channel_names))
        assert len(channel_indexes) == len(channel_names), 'not match'
        return channel_indexes
```
2 problems:
1. When you select with channel names, the returned order becomes broken, therefore the user must additionally bother about sorting stuff.
2. When you provide list of channel names bigger than the actual list, the API could unsafely return everything it finds (e.g. in strict_slicing=False), rather than asserting.

samuelgarcia:
@sin-mike : can I close ethis ?

samuelgarcia:
I colse this and leave #730 open.

PR #730, opened 2019-08-01
add pick_signal_channels to baserawio.py

sin-mike:
@samuelgarcia 
following #728 I've added a function to the `baserawio.py`, keeping the options for ordering and strict selecting the channels in other functions

### DIFF
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -480,9 +480,6 @@ class BaseRawIO(object):
         depending which is not None
         """
 
-        check = (channel_indexes is None) + (channel_names is None) + (channel_ids is None)
-        assert(check == 2), 'multiple options for channel selection'
-
         if channel_indexes is None and channel_names is not None:
             channel_indexes = self.channel_name_to_index(channel_names, user_order, strict_selection)
 
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -421,38 +421,115 @@ class BaseRawIO(object):
         else:
             return [None]
 
-    def channel_name_to_index(self, channel_names):
+    def channel_name_to_index(self, channel_names, user_order=False, strict_selection=True):
         """
         Transform channel_names to channel_indexes.
         Based on self.header['signal_channels']
+        Args:
+             channel_names: list, requested channel names
+             user_order: bool (False), keep the output index order
+                the same as requested
+             strict_selection: bool (True), check whether all the requested channel names
+                were present in the header
         """
         ch = self.header['signal_channels']
-        channel_indexes, = np.nonzero(np.in1d(ch['name'], channel_names))
-        assert len(channel_indexes) == len(channel_names), 'not match'
+        chnames = ch['name']
+        channel_indexes, = np.nonzero(np.in1d(chnames, channel_names))
+
+        if user_order and (len(channel_indexes) > 0):
+            lch = list(channel_names)
+            channel_indexes = sorted(channel_indexes, key=lambda x: lch.index(chnames[x]))
+            channel_indexes = np.array(channel_indexes)
+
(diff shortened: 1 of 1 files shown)


neo/rawio/baserawio.py
```diff
@@ -421,39 +421,113 @@ def get_group_channel_indexes(self):
         else:
             return [None]
 
-    def channel_name_to_index(self, channel_names):
+    def channel_name_to_index(self, channel_names, user_order=False, strict_selection=True):
```
samuelgarcia, line 424: very good

neo/rawio/baserawio.py
```diff
@@ -421,39 +421,113 @@ def get_group_channel_indexes(self):
...
+            channel_indexes = np.array(channel_indexes)
+
+        if strict_selection:
+            assert len(channel_indexes) == len(channel_names), \
+                'selection {} doesn\'t match header names {}'.format(channel_names, chnames)
+
         return channel_indexes
 
-    def channel_id_to_index(self, channel_ids):
+    def channel_id_to_index(self, channel_ids, user_order=False, strict_selection=True):
```
samuelgarcia, line 450: very good

samuelgarcia:
Hi @sin-mike.
Sorry for the super super long delay.
I prefer a lot this idea becaise it is generic.
However, I am wondering how this ```pick_signal_channels``` could return another reader and not modify inplace the actual reader.
I think the inplace modification can be dangerous.
What do you think ?

sin-mike:
> Hi @sin-mike.
> Sorry for the super super long delay.
> I prefer a lot this idea becaise it is generic.
> However, I am wondering how this `pick_signal_channels` could return another reader and not modify inplace the actual reader.
> I think the inplace modification can be dangerous.
> What do you think ?

Hi Samuel,

I've been using this feature since then in this exact version, and I didn't observe any drawbacks in our processing pipelines. However, I understand your concern. My suggestion then is to make a protected field `_header`, so that the function can change only the public `header`, while the `_header` is kept untouched. And make a `self.reset_header` function

Mike

samuelgarcia:
Hi @sin-mike 
sorry again for this super long delay.

I like a lot the  `user_order=False, strict_selection=True` but I think it is not possible to include the `pick_signal_channels` (inplace modification) in the neo master/

I think this will lead to some border effect for some rawios even if if works for you.

Maybe you you remove remove this `pick_signal_channels` right now so I could merge the `user_order=False, strict_selection=True` and then you could make another PR that mimic the  `pick_signal_channels(..)` but instead of inplace modification it could just duplicate the the rawio instance and mofify the header. It would be less dangerous.

What do you think ?

samuelgarcia:
Hi @sin-mike.
Sorry for the discontinuous discussion about this old PR.

We will release neo 0.9.0 in september.
Do you think you could a bit more work to test my last idea or would you prefer to prefer to postpone this PR in september.


My last idea is basically to have something like this:

```python
import neo

rawreader = neo.rawio.MyFormetIO(filename=...)  # this have all channel
# this is a copy of the rawreader but with less channels
subrawreader = neo.rawio.tools.pick_signal_channels(rawreader)
```

This would avoid an inplace modification which is dangerous in my mind.

What do ou think ?

samuelgarcia:
Hi Mike,
as the relase date is approaching I propose to postpone this to 0.10.
Cheers

samuelgarcia:
Dear Mike.
I have work a bit rawio API here #949 

The first change you propose here is not possible (in place channel modification).
As I said having another class doing it would be totally ok.

The second change ` strict_selection=True` is no more relevant in the rawio refactoring becaise ids have to be unique inside a stream and this is tested always.

About `user_order=False` I think this is safer to always keep order.

So I close this PR.

Thanks again for you participation in neo.

PR #737, opened 2019-09-23
Can't use exact comparison for floating point.

KyrusMama:
We found a bug where an `assert` [here](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/spike2rawio.py#L153) was hit because of an incorrect floating point comparison. 

Specifically, without the fix, it did an exact comparison to a floating point value, but for floating points, you need to use `np.isclose()`. Hence the assert failed, because this comparison was incorrect.

If you try to run with the [attached ](https://drive.google.com/file/d/15WhILxaz-OwURLjcKWpvdYvjDvQQ0qOS/view?usp=sharing) file it will raise this assert.

This seems to fix the issue, but we are not sure why the comparison is performed in the first place, so we don't know if it is a correct fix.

### DIFF
--- a/neo/rawio/spike2rawio.py
+++ b/neo/rawio/spike2rawio.py
@@ -127,7 +127,12 @@ class Spike2RawIO(BaseRawIO):
                     # detect gaps
                     inter_block_sizes = data_blocks['start_time'][1:] - \
                         data_blocks['end_time'][:-1]
-                    gaps_block_ind, = np.nonzero(inter_block_sizes > interval)
+                    # Due to rounding errors, have to check if the value is larger
+                    # which means its not close and larger
+                    gaps_block_ind, = np.nonzero(np.logical_and(
+                        inter_block_sizes > interval,
+                        np.logical_not(np.isclose(inter_block_sizes, interval))
+                    ))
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi.
Thank you for the catch.
I anderstand the patch but why not forcing all_nb_seg to be int64 ?
Would it work ?

KyrusMama:
I think it should, so long as you round/ceil it to an integer. Just forcing it to an integer would floor it, and the same problem would arise. 
As I said though, I'm not sure what this calculation and comparison is really for, so I may be wrong.
Thanks.

samuelgarcia:
Something annoye me with this patch because it means that the round precision is bellow the "interval".

In this part of of code, **inter_block_sizes** are supposed to be int32. interval is float be should be integer this is why the **>** can fail. Could you force interval to be int after (after a rounding) instead of your patch ?

samuelgarcia:
This is PR is not relevant anymore because now the master have this:

https://github.com/NeuralEnsemble/python-neo/blob/054394e869d794c41d27d3db7481bdab8682f938/neo/rawio/spike2rawio.py#L124


So I close the PR

PR #739, opened 2019-09-24
[WIP] Nest3io

JuliaSprenger:
This PR adds support for the sion format, the new output format of NEST. The IO depends on the nestio-tools (https://github.com/apeyser/nestio-tools) and the sionlib library (https://www.fz-juelich.de/ias/jsc/EN/Expertise/Support/Software/SIONlib/sionlib-download_node.html).

Current status: Work in progress.

### DIFF
--- a/neo/io/__init__.py
+++ b/neo/io/__init__.py
@@ -133,7 +133,11 @@ Classes:
 
     .. autoattribute:: extensions
 
-.. autoclass:: neo.io.NestIO
+.. autoclass:: neo.io.Nest2IO
+
+    .. autoattribute:: extensions
+
+.. autoclass:: neo.io.Nest3IO
 
     .. autoattribute:: extensions
 
@@ -243,7 +247,8 @@ from neo.io.kwikio import KwikIO
 from neo.io.micromedio import MicromedIO
 from neo.io.hdf5io import NeoHdf5IO
 from neo.io.neomatlabio import NeoMatlabIO
-from neo.io.nestio import NestIO
+from neo.io.nest2io import Nest2IO
+from neo.io.nest3io import Nest3IO
 from neo.io.neuralynxio import NeuralynxIO
 from neo.io.neuralynxio_v1 import NeuralynxIO as OldNeuralynxIO
 from neo.io.neuroexplorerio import NeuroExplorerIO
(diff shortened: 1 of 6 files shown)


essink:
Hi @JuliaSprenger,

in order to load data generated with NEST 3 (no `.gdfs` anymore), I created a [hotfix branch](https://github.com/essink/python-neo/tree/fix/nestio_nest3.0). It uses the existing `nestio` from the master branch and only makes small adjustments to allow loading the resulting `.dat` files. 

Disclaimer: It is not properly tested, but seems to work at least for the spiking data. At the time of creating the hotfix, I was not aware of this pull request. The hotfix does not make use of any knowledge of nestio-tools or the sionlib library and thus does not generalize the `nestio` satisfyingly. Nevertheless, I wanted to let you know in the hope that it may help somehow.

Best,
Simon

JuliaSprenger:
Hi @essink Thanks for linking your solution here. Unfortunately this topic is quite low in my queue, so if you want to make a full PR out of this feel free to go ahead.

morales-gregorio:
Hi @rgutzen @jasperalbers and I just also stumbled upon the NestIO 3.0 problem. Any chance this is merged anytime soon? 

@rgutzen created a simple hotfix that requires only minor changes in the `__init__()` of the class `ColumnIO` such that:
```
def __init__(self, filename):
    """
    filename: string, path to ASCII file to read.
    """

    self.filename = filename
    
    # read the first line to check the data type (int or float) of the data
    f = open(self.filename)
    line = f.readline()
    header_size = 0
    
    while line:
        if line[0].isdigit():
            break
        else:
            header_size += 1
            line = f.readline()

    # Warn user
    if header_size > 0:
        warnings.warn(f'Ignoring {str(header_size)} header lines.')
    
    additional_parameters = {}
    if '.' not in line:
        additional_parameters['dtype'] = np.int32
    
    self.data = np.loadtxt(self.filename, skiprows=header_size,
                           **additional_parameters)
``` 
In principle this can work for both nest 2.x and 3.x, since the only thing it does is ignoring the header lines in the file and proceeding as usual. We can create a different PR to include this in the nestio if you want.

Best,
Aitor, Robin and Jasper

PR #741, opened 2019-09-26
Add support for image sequence data (e.g. calcium imaging of neuronal activity)

apdavison:
provides the classes `ImageSequence` and three `RegionOfInterest` subclasses, as well as three IO modules for imaging data.

The principal author of this PR was @teogale


samuelgarcia:
Hi Andrew
it is a big one.
I will have a look.

neo/core/imagesequence.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __array_finalize__spec(self, obj):
+
+        self.sampling_rate = getattr(obj, 'sampling_rate', None)
+        self.spatial_scale = getattr(obj, 'spatial_scale', None)
+        self.units = getattr(obj, 'units', None)
+
+        return obj
+
+    def signal_from_region(self, *region):
```
samuelgarcia: I think we would a **reduce_mode='average' / 'median'/'max'** kargs to control the method of the reduce.
No ?

neo/core/regionofinterest.py
```diff
@@ -0,0 +1,172 @@
+# -*- coding: utf-8 -*-
+
+from math import floor, ceil
+
```
samuelgarcia: A general and stupid question : this RegionOfInterest is binary by design (in or out) but we could also have a weighted region of interest. No ?

neo/core/regionofinterest.py
```diff
@@ -0,0 +1,172 @@
...
+                max_y = i[1]
+
+        list_coord = []
+        for y in range(int(floor(min_y)), int(ceil(max_y))):
+            for x in range(int(floor(min_x)), int(ceil(max_x))):
+                list_coord.append((x, y))
+
+        pixel_list = self.polygon_ray_casting(self.vertices, list_coord)
+
+        return pixel_list
```
samuelgarcia: We could also have a BooleanMaskRegionOfInterest, no ?

neo/core/imagesequence.py
apdavison: yes, that would be a good improvement. I suggest leaving it for a future update, however.

neo/core/regionofinterest.py
apdavison: Yes, that would make sense for circular regions of interest (e.g. circular gaussian weighting). I'm not sure how you would specify it for a more complex shape, however.

neo/core/regionofinterest.py
apdavison: Yes, we should add that in a future version.

apdavison:
ok to merge?

apdavison:
> I am not super happy how the raltionship between ROI and block is done.

I'm not happy either, but, as you say, we need to refactor relationships anyway!

Thanks for the merge.

PR #747, opened 2019-10-14
Clean gain for some files because it can be numerical or string.

### LINKED ISSUE #744: Trouble reading some BCI2000 files
Hi everyone

Thanks to an update in reading BCI2000 data (#736, #734) I can now import my *.dat files into neo. However, I've just figured out that an issue remains for some of my files, and it looks to be due to how the gain is specified in the header.

could you help me on this ?

attached: 
- snapchot of the error I get using neo reader
- 
Capture d’écran 2019-10-04 à 14 09 14
Capture d’écran 2019-10-03 à 17 29 36
Capture d’écran 2019-10-03 à 17 29 47

and two snapshots of the SourceChGain value (observed using Matlab load_bcidat.m) in one working  file (index 78), and in a failing one (85)

Thank you for your help on this!
Françoise

samuelgarcia:
try to fix #744

### DIFF
--- a/neo/rawio/bci2000rawio.py
+++ b/neo/rawio/bci2000rawio.py
@@ -47,7 +47,19 @@ class BCI2000RawIO(BaseRawIO):
             dtype = file_info['DataFormat']
             units = 'uV'
             gain = param_defs['SourceChGain']['value'][chan_ix]
+            
+            if isinstance(gain, str):
+                r = re.findall('(\d+)(\D+)', gain)
+                # some files have strange units attached to gain
+                # in that case it is ignored
+                if len(r) ==1:
+                    gain = r[0][0]
+                gain = float(gain)
+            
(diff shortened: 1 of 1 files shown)


neo/rawio/bci2000rawio.py
```diff
@@ -47,7 +47,19 @@ def _parse_header(self):
             dtype = file_info['DataFormat']
             units = 'uV'
             gain = param_defs['SourceChGain']['value'][chan_ix]
+            
+            if isinstance(gain, str):
+                r = re.findall('(\d+)(\D+)', gain)
```
apdavison, line 52: I think you need to use a raw string here: `r = re.findall(r'(\d+)(\D+)', gain)`

samuelgarcia:
@apdavison : can you merge this ?

Issue #748, opened 2019-10-15
[NixIO] duplicate naming issue with array_annotations


rgutzen:
When reading a neo object with array annotations and writing it back with the NixIO, the array annotation causes the error:
`DuplicateName: Duplicate name - names have to be unique for a given entity type & parent. (create_property)`

Here a minimal code example:

```
import neo
import quantities as pq
import numpy as np

asig = neo.AnalogSignal(signal=np.array([[1,2,3,4,5], [6,7,8,9,10]]),
                        units='dimensionless',
                        sampling_rate=1*pq.Hz,
                        array_annotations={'labels': ['a','b','c','d','e']})

block = neo.Block()
seg = neo.Segment()
block.segments.append(seg)
block.segments[0].analogsignals.append(asig)

with neo.NixIO('temp.nix') as io:
    io.write(block)
with neo.NixIO('temp.nix') as io:
    loaded_block = io.read_block()
with neo.NixIO('temp2.nix') as io:
    io.write(loaded_block)
```

neo v0.8.0
nixio v1.5.0b1

rgutzen:
The issue arises from this part of the loading routine which adds properties which are an array_annotation to a corresponding nested dict within _neo_attrs_, but also adds them as a regular annotation, therefore creating a duplicate.

nixio.py line 1237-1244 (in method __nix_attr_to_neo_):
```
                elif prop.definition == ARRAYANNOTATION:
                    if 'array_annotations' in neo_attrs:
                        neo_attrs['array_annotations'][prop.name] = values
                    else:
                        neo_attrs['array_annotations'] = {prop.name: values}
                else:
                    values = list(values)
                neo_attrs[prop.name] = values
```

Is showing array_annotations also additionally in annotations a desired feature?

It seems that PR #662 intended not to duplicate the array_annotations, but the duplication was then introduced with the merge commit 0a4ccfae4aca7d854d3f1065bd8782136981f1ab
Please comment @Kleinjohann ?

PR #750, opened 2019-10-16
[Update to PR #662] fix duplication of array_annotations into annotations

rgutzen:
Handling Issue #748 
The issue arises from this part of the loading routine which adds properties which are an array_annotation to a corresponding nested dict within _neo_attrs_, but also adds them as a regular annotation, therefore creating a duplicate.

nixio.py line 1237-1244 (in method __nix_attr_to_neo_):
```
                elif prop.definition == ARRAYANNOTATION:
                    if 'array_annotations' in neo_attrs:
                        neo_attrs['array_annotations'][prop.name] = values
                    else:
                        neo_attrs['array_annotations'] = {prop.name: values}
                else:
                    values = list(values)
                neo_attrs[prop.name] = values
```

It seems that PR #662 intended not to duplicate the array_annotations, but the duplication was then introduced with the merge commit 0a4ccfae4aca7d854d3f1065bd8782136981f1ab
Please comment @Kleinjohann ?

The PR separates the properties again, to be either stored directly in _neo_attrs_, or in _neo_attrs['array_annotations']_ but not both.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -1190,8 +1190,6 @@ class NixIO(BaseIO):
                         values = ""
                 elif len(values) == 1:
                     values = values[0]
-                else:
-                    values = list(values)
                 if prop.definition == ARRAYANNOTATION:
                     if 'array_annotations' in neo_attrs:
                         neo_attrs['array_annotations'][prop.name] = values
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -1190,14 +1190,15 @@ class NixIO(BaseIO):
                         values = ""
                 elif len(values) == 1:
                     values = values[0]
-                elif prop.definition == ARRAYANNOTATION:
+                else:
+                    values = list(values)
+                if prop.definition == ARRAYANNOTATION:
                     if 'array_annotations' in neo_attrs:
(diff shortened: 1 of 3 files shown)


neo/io/nixio.py
```diff
@@ -1190,14 +1190,15 @@ def _nix_attr_to_neo(nix_obj):
                         values = ""
                 elif len(values) == 1:
                     values = values[0]
-                elif prop.definition == ARRAYANNOTATION:
+                else:
+                    values = list(values)
```
achilleas-k: This is probably the reason the tests are failing.  The repetition of the `list()` conversion here is undoing the quantity creation from earlier in the method and the unit information is lost.  I think this `else` block is unnecessary.

Kleinjohann:
> It seems that PR #662 intended not to duplicate the array_annotations, but the duplication was then introduced with the merge commit [0a4ccfa](https://github.com/NeuralEnsemble/python-neo/commit/0a4ccfae4aca7d854d3f1065bd8782136981f1ab)
> Please comment @Kleinjohann ?

You are right, this is my fault, I remember something weird happening to that code block in the merge and it seems I did not resolve it correctly.

PR #753, opened 2019-10-21
ImageSequence support in NixIO

teogale:


JuliaSprenger:
Hi @teogale!
Thanks for extending the NixIO functionality! Could you also add a test to cover this new feature?

teogale:
Hi @JuliaSprenger !
I will take a look at the test, but i cannot ensure an update as i am quite limited in time.

achilleas-k:
If you want me to have a look at this PR (since it's NIXIO related), it might have to wait a few days since I'm currently at SfN and can't really dig into it with full attention.

I did have a quick look and I think it might be possible to save each ImageSequence as a single DataArray instead of splitting it into multiple ones like we do with Signal objects.  I'm not sure though.  I'll have to think about this a bit.  If it's possible, it would be a lot more efficient since it would save all the object overhead (in both space and time).

achilleas-k:
I added my comments to PR #760 since it includes the commits from this PR as well.

JuliaSprenger:
The changes were merged via #804.

PR #754, opened 2019-10-24
add micro Volt conversion and import missing package

rgutzen:
When loading a new set of Spike2 files (from 2018), the IO couldn't handle the units 'µV'.
The changes add a corresponding conversion, fix a wrong conversion and a missing import, which was discovered while fixing this bug.

### DIFF
--- a/neo/io/proxyobjects.py
+++ b/neo/io/proxyobjects.py
@@ -12,7 +12,7 @@ ineherits neo.rawio.
 
 import numpy as np
 import quantities as pq
-import loggging
+import logging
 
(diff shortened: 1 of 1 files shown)


neo/io/proxyobjects.py
```diff
@@ -12,6 +12,7 @@
 
 import numpy as np
 import quantities as pq
+import loggging
```
apdavison: too many 'gggg's (-:

rgutzen: that, of course, explains the failing tests ^^

PR #764, opened 2019-11-07
Fix for SpikeTrain consistency check encountering CompundUnits

### LINKED ISSUE #763: Regression in the `check_has_dimensions_time` method of the SpikeTrain class for Compound Units
Regression in the `check_has_dimensions_time` method of the SpikeTrain class.

Calling
`neo.SpikeTrain([1,2,3]*pq.CompoundUnit("1/10*s"), t_start=0*pq.ms, t_stop=10000*pq.s)-5*pq.CompoundUnit("1/100*s")`
used to issue
`ValueError: value 5.0 has dimensions s, not [time]`
due to an incorrect handling of CompoundUnits -- these need to be simplified first in order to check consistency.

This fixes the issue by an additional simplified statement, making the above statement return
`<SpikeTrain(array([0.5, 1.5, 2.5]) * (1/10*s), [-0.49999999999999994 (1/10*s), 99999.5 (1/10*s)])>`

mdenker:
This fixes #763 

An additional simplified statement makes sure that CompoundUnits are first simplified such that the "true" unit can be reliably extracted. Additional, a regression test based on #763 is implemented, which failed on the pervious build.

### DIFF
--- a/neo/core/spiketrain.py
+++ b/neo/core/spiketrain.py
@@ -38,10 +38,13 @@ def check_has_dimensions_time(*values):
     '''
     errmsgs = []
     for value in values:
-        dim = value.dimensionality
-        if (len(dim) != 1 or list(dim.values())[0] != 1 or not isinstance(list(dim.keys())[0],
-                                                                          pq.UnitTime)):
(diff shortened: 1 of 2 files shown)


apdavison:
The test failures are nothing to do with this PR, as far as I can see.

Issue #766, opened 2019-11-14
[NixIO] saving/loading looses the links between analogsignals and channel_indexes


rgutzen:
Minimal example:
neo v0.8.0
nixio v1.5.0b1

```
import neo
import quantities as pq
import numpy as np
import itertools

# create dummy objects
asig = neo.AnalogSignal(np.arange(3),
                        units='mV',
                        sampling_rate=1000*pq.Hz)
chidx = neo.ChannelIndex(name='Electrode array', 
                         index=np.arange(3))

# create links
asig.channel_index = chidx
chidx.analogsignals = [asig]

# build neo structure
blk = neo.Block()
seg = neo.Segment()
seg.analogsignals.append(asig)
blk.channel_indexes.append(chidx)
blk.segments.append(seg)

# check links
print('Pre-nixify:')
print(blk.channel_indexes[0].analogsignals)
print(blk.segments[0].analogsignals[0].channel_index)

# write/read
with neo.NixIO('temp.nix') as io:
    io.write(blk)
with neo.NixIO('temp.nix') as io:
    loaded_blk = io.read_block()

# check links
print('Post-nixify:')
print(loaded_blk.channel_indexes[0].analogsignals)
print(loaded_blk.segments[0].analogsignals[0].channel_index)
```

Output:
```
>> Pre-nixify:
>> [<AnalogSignal(array([[0], [1], [2]]) * mV, [0.0 s, 0.003 s], sampling rate: 1000.0 Hz)>]
>> <neo.core.channelindex.ChannelIndex object at 0x7fcd1fe17cc0>
>> Post-nixify:
>> []
>> None
```

achilleas-k:
I can partially confirm this.  I get the following output, with the same lib versions you specified:
```
Neo version: 0.8.0
NIX version: v1.5.0b1
Pre-nixify:
[<AnalogSignal(array([[0],
       [1],
       [2]]) * mV, [0.0 s, 0.003 s], sampling rate: 1000.0 Hz)>]
<neo.core.channelindex.ChannelIndex object at 0x7f845129f3d0>
Post-nixify:
[<AnalogSignal(array([[0],
       [1],
       [2]]) * mV, [0.0 s, 0.003 s], sampling rate: 1000.0 Hz)>]
None
```

ChannelIndex -> AnalogSignal link is preserved, but the inverse isn't.

achilleas-k:
[...]

PR #767, opened 2019-11-14
[nixio] Create many to one and many to many relationships

achilleas-k:
Runs the two utility methods before returning loaded Neo Blocks.

See #766.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -291,6 +291,10 @@ class NixIO(BaseIO):
             # parent reference
             newchx.block = neo_block
 
+        # create object links
+        neo_block.create_many_to_one_relationship()
+        neo_block.create_many_to_many_relationship()
+
         # reset maps
         self._neo_map = dict()
         self._ref_map = dict()
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -292,8 +292,7 @@ class NixIO(BaseIO):
             newchx.block = neo_block
 
         # create object links
-        neo_block.create_many_to_one_relationship()
-        neo_block.create_many_to_many_relationship()
+        neo_block.create_relationship()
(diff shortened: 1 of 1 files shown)


achilleas-k:
I believe the circleci tests for NIX are failing because circle isn't pulling down the latest test files but is pulling down the latest nixio beta.  The circleci log shows that it's not restoring the test file cache though, so I'm not sure that's the issue, but the error I'm seeing is consistent with what I get when I run the test with nixio-1.5.0b4 and the file [before this PR was merged](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/13) last week in the test file repository.

neo/io/nixio.py
```diff
@@ -291,6 +291,10 @@ def _nix_to_neo_block(self, nix_block):
             # parent reference
             newchx.block = neo_block
 
+        # create object links
```
JuliaSprenger: why not using `create_relationship` instead of calling these two functions explicitely?

JuliaSprenger:
@achilleas-k: In your PR on gin you only updated one of the 3 nix files being present. I tried to run the tests locally and already get the failing tests with one of the non-updated test files. I gues these files might also have an incorrect header?

neo/io/nixio.py
achilleas-k: Wasn't aware of that function (I was going off [the IO dev docs](https://neo.readthedocs.io/en/stable/io_developers_guide.html)).  Will update now.

achilleas-k:
Oh, I see it now.  Yes, it was my fault.  I wasn't getting the errors locally which I why I assumed it was a caching issue for the CI.  Not sure why I wasn't getting the error before, but I get it now as well.  Will update the test file on GIN.

Issue #771, opened 2019-11-18
NixIO ChannelIndex Reading channel_names as empty


jmbudd:
**Minimal Example**
Create ChannelIndex for single Block and save as .h5.

```python
from neo.core import (Block, Segment, ChannelIndex)
from neo.io import NixIO

bl = Block()

chx = ChannelIndex(name="Array", index=[0,1,2], channel_names=['ch0', 'ch1', 'ch2'])
bl.channel_indexes.append(chx)

with NixIO(filename="/tmp/chx.h5", mode='rw') as writer:
    writer.write_block(bl)
```
```python
>>> bl.channel_indexes[0].channel_names
array(['ch0', 'ch1', 'ch2'], dtype='<U3')
```

But when .h5 file was loaded the channel names array was empty  (note: array originally encoded as unicode but read as str):

```python
>>> with NixIO(filename="/tmp/chx.h5", mode='ro') as reader:
...     blk = reader.read_block()
... 
>>> blk.channel_indexes[0].channel_names
array([], dtype='|S1')
>>> blk.channel_indexes[0].index
array([0, 1, 2])
```

HDF Viewer suggests channel names were stored in .h5 file:
![chx](https://user-images.githubusercontent.com/40672011/69092188-243d2280-0a4c-11ea-800f-3b02f031f8b6.png)

**Setup**
Python 3.5.2
h5py==2.10.0
neo==0.8.0
nixio==1.5.0b3
numpy==1.17.3
ubuntu 16.04

achilleas-k:
Thanks.  I found the issue.  It was introduced in #695 (as a fix for #694).
Fix coming in a bit.

PR #772, opened 2019-11-19
[nixio] Fix channel name reading

achilleas-k:
Bug introduced in #695 (as a fix for #694).

See #771.


JuliaSprenger:
I can confirm this fixes issue #771

PR #775, opened 2019-11-22
Fix round problem for spike2 smr.

samuelgarcia:


apdavison:
Hi @samuelgarcia can you explain a little bit what the problem was that led to this PR? It is possible to provide a test?

samuelgarcia:
I must admit that I don't remember but it is important.
The way I did indexing was wrong in some case.
It is how we detect the boundary blocks.
While I was fixing this I discover that spike2 and plexon have the same problem so I fixed this at the two place.

And the title is wring it is not a "round problem" as I though at the begining.

samuelgarcia:
Oups. Now I remember.
spike2 when reading chunk tat is not multiple of internal chunksize was causing misalignement bugs.
There is no specific tests for that.

Could we merge this quite soon ? Some people I collborate with need this.

JuliaSprenger:
Couldn't we add a test that is trying to read one of the test files with an incompatible chunk size?

samuelgarcia:
It was not raising errors but mixing chunks.
I could try to write a general test for for all IOs that read signals with differents chunksize and check If the final output remain the same.

I hope it will not trigger any bugs for others IOs.

samuelgarcia:
I have added a test. Lets see.

samuelgarcia:
This is ready to merge.

samuelgarcia:
could we merge this before #824 ?

PR #776, opened 2019-12-04
Add resample and decimate utility methods for Analog and IrregularlysampledSignals

JuliaSprenger:
In the context of #526 it might be useful to add utility functions for adjusting the time sampling of `AnalogSignals` and `IrregularlySampledSignals`. This PR is wrapping the two scipy functions `decimate` and `resample` to provide a mechanism for decimating and resampling within Neo and adds some tests with it.

There is one issue coming with this PR: Scipy itself is not a Neo dependency yet, even though it is used in a couple of places for specific IOs. Here, I the new utility methods can only be used in case of scipy being available and will raise an ImportError if this is not the case.

Note: As scipy is not available in the Travis test setup, the new tests fail. Should we include scipy in the test environment / requirements for Neo or is this introducing too strong dependencies? Alternatively, these utility functions might be included in [elephant](https://github.com/neuralensemble/elephant), with the drawback ob being less visible to the users.

### DIFF
--- a/neo/core/irregularlysampledsignal.py
+++ b/neo/core/irregularlysampledsignal.py
@@ -24,11 +24,13 @@ the old object.
 from __future__ import absolute_import, division, print_function
 
 from copy import deepcopy, copy
+
 try:
-    SCIPY_AVAIL = True
     import scipy.signal
-except ImportError:
-    SCIPY_AVAIL = False
+except ImportError as err:
+    HAVE_SCIPY = False
+else:
+    HAVE_SCIPY = True
 
 import numpy as np
 import quantities as pq
@@ -353,7 +355,7 @@ class IrregularlySampledSignal(BaseSignal):
         else:
             raise NotImplementedError
 
-    def resample(self, at=None, interpolation=None):
+    def resample_irregularly(self, at=None, interpolation=None):
         '''
         Resample the signal, returning either an :class:`AnalogSignal` object
         or another :class:`IrregularlySampledSignal` object.
@@ -369,7 +371,7 @@ class IrregularlySampledSignal(BaseSignal):
         # further interpolation methods could be added
         raise NotImplementedError
 
-    def resample_regularly(self, sample_count, **kwargs):
+    def resample(self, sample_count, **kwargs):
         """
         Resample the data points of the signal.
         This function is a wrapper of scipy.signal.resample and accepts the same set of keyword arguments,
@@ -388,7 +390,7 @@ class IrregularlySampledSignal(BaseSignal):
             The original :class:`AnalogSignal` is not modified.
         """
 
-        if not SCIPY_AVAIL:
+        if not HAVE_SCIPY:
             raise ImportError('Resampling requires availability of scipy.signal')
 
         # Resampling is only permitted along the time axis (axis=0)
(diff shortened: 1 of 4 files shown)


JuliaSprenger:
@samuelgarcia @apdavison Do you have an opinion on the Scipy issue here?

apdavison:
For a long time I was against having scipy as a dependency, because it was often very slow and/or painful to install. Since the rise of  wheels and conda, this is no longer really a  problem, so I am ok with adding it as a dependency.

I think the name "decimate" is terrible, and "resample_regularly" could be just "resample". What do you think about having a single function "resample" for both AnalogSignal and IrrSampSig, which would use either scipy decimate or scipy resample under the hood, depending on the arguments?

samuelgarcia:
Hi julia.
This is a good idea.

I think we still use scipy as a weak dependency with the
```python
try:
   import scipy 
    HAVE_SCIPY= True
except importError
    HAVE_SCIPY= False
```
neo could be used in a web leigth weioght environement this could be usefull to not have it.
I don't known.
For desktop anybody have scipy.

neo/core/irregularlysampledsignal.py
```diff
@@ -353,7 +355,7 @@ def mean(self, interpolation=None):
         else:
             raise NotImplementedError
 
-    def resample(self, at=None, interpolation=None):
+    def resample_irregularly(self, at=None, interpolation=None):
```
apdavison: What is the thinking behind providing a not-implemented method? Wouldn't it be better just to remove this?

JuliaSprenger: I guess the idea was to make a suggestion for the future interface, so I kept it. But I am happy to remove it if you agree.

apdavison: I think it would be better to remove it to avoid confusing users - but perhaps you could open an issue so as to remind us of the idea?

dizcza:
I was thinking to add resampling in Elephant, but I'm glad to see it's being implemented in Neo.
I agree with Andrew regarding (1) I also don't experience issues anymore with slow pip install \<whatever\> and (2) the same API - `resample` for both types of signal.
In the docs, it'd be nice to distinguish when to use `decimate` and when `resample`.

However, what I wanted to check and what I don't see here is some kind of tests that check for the artifacts of the resample/downsampled signal. It seems we, at our lab, are very cautious about adding spurious spikes as well as removing ones due to imperfect data processing. So I wonder, under what circumstances, if any, these methods can
* add spurious spikes;
* remove original spikes.

I was thinking of simulating raw analog signals from spikes, applying these methods, and thresholding back to see if the original spiketrain is preserved. Or testing on a chunk of real data.

JuliaSprenger:
@apdavison I removed the old resample interface and renamed the methods to `resample` and `downsample` and added more description and crossreferences in the docstring.
@samuelgarcia I updated the usage of scipy following your `HAVE_SCIPY` schema.
@dizcza I added two sentences in the docstrings making the difference between the two functions more clear. However, I think that the tests you are suggesting should be another PR, since the problems you are describing highly depend on the type of data (sources) you are looking at and will probably lead to longer discussion, especially since the process of spike sorting you are hinting at is not part of Neo.

PR #782, opened 2019-12-10
STring bug in openephysio.

### LINKED ISSUE #781: Wrong endianness when reading OpenEphys data from the legacy format.
Reading OpenEphys data using `neo.io.OpenEphysIO` does not allow to recover proper data (plotting the signal makes it clear it can not be a continuous recording). I think I tracked down the issue to the definition of the `continuous_dtype` line [374](https://github.com/NeuralEnsemble/python-neo/blob/9c3cadd839d25cf14b18074a80b479999a3a720d/neo/rawio/openephysrawio.py#L374) of `openephysrawio` which is:

```
continuous_dtype = [('timestamp', 'int64'), ('nb_sample', 'uint16'),
    ('rec_num', 'uint16'), ('samples', 'int16', RECORD_SIZE),
    ('markers', 'uint8', 10)]
```

though the [documentation](https://open-ephys.atlassian.net/wiki/spaces/OEW/pages/65667092/Open+Ephys+format) indicates that the samples are written as big-endian int16. At least on my linux the system is little-endian. This is easily fixed by updating the data type definition: 
```
continuous_dtype = [('timestamp', 'int64'), ('nb_sample', 'uint16'),
    ('rec_num', 'uint16'), ('samples', '>i2', RECORD_SIZE),
    ('markers', 'uint8', 10)]
```
It could also be worth having the little-endianness of the other fields made explicit.

I hope I am not wrong and that this helps other users.

samuelgarcia:
This should fix #781

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -372,7 +372,7 @@ class OpenEphysRawIO(BaseRawIO):
 
 
 continuous_dtype = [('timestamp', 'int64'), ('nb_sample', 'uint16'),
-    ('rec_num', 'uint16'), ('samples', 'int16', RECORD_SIZE),
+    ('rec_num', 'uint16'), ('samples', '>i2', RECORD_SIZE),
     ('markers', 'uint8', 10)]
 
 events_dtype = [('timestamp', 'int64'), ('sample_pos', 'int16'),


samuelgarcia:
thanks.

PR #784, opened 2019-12-11
[NIXIO] Save and load date, time, and datetime annotations

### LINKED ISSUE #783: [NixIO] Error for datetime annotations
Saving a neo block structure that contains an annotation of type datetime (date/time) results in a `ValueError`. E.g.
```python
from neo import NixIO, Block
import datetime

b = Block(custon_date=datetime.date.today())
with NixIO('test.nix','ow') as io:
    io.write_block(b)
```
causes `ValueError: Unknown type for value 2019-12-10`.
However, datetime objects are a supported annotation object type within neo, see 
https://github.com/NeuralEnsemble/python-neo/blob/a9ac68db3a2bfdc6b8f0d96237e091e7772610ed/neo/core/baseneo.py#L20
and are frequently generated when loading datasets with other IOs.

achilleas-k:
This PR adds support for saving and loading date, time, and datetime objects to NIX without losing the original data type.
On write, it marks annotations by setting property.definition to DATE, TIME, or DATETIME.  On load, it uses the definition to recreate the original object.

Fixes #783.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -63,6 +63,10 @@ DATEANNOTATION = "DATE"
 TIMEANNOTATION = "TIME"
 MIN_NIX_VER = Version("1.5.0")
 
+datefmt = "%Y-%m-%d"
+timefmt = "%H:%M:%S.%f"
+datetimefmt = datefmt + "T" + timefmt
+
 
 def stringify(value):
     if value is None:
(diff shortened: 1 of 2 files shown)


achilleas-k:
Looks like I used some datetime methods that are only available in newer versions.  I'll fix it tomorrow.

JuliaSprenger:
Works nicely! Thanks :)

PR #789, opened 2020-01-28
Improve utilities

JuliaSprenger:
This is a first change in a list of adjustments that need to be implemented in the utility functions to fully make use of the new array_annotation feature.

[edit] This PR also removes the notion of treating regular annotations that are lists of the shape of the data like array annotations. For this purpose array annotations should be used.

### DIFF
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -421,6 +421,7 @@ def add_epoch(
     ep = neo.Epoch(times=times, durations=durations, **kwargs)
 
     ep.annotate(**event1.annotations)
+    ep.array_annotate(**event1.array_annotations)
 
     if attach_result:
         segment.epochs.append(ep)
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -50,8 +50,8 @@ def get_events(container, **properties):
     Example:
     --------
         >>> event = neo.Event(times=[0.5, 10.0, 25.2] * pq.s)
-        >>> event.annotate(event_type='trial start',
-                           trial_id=[1, 2, 3])
+        >>> event.annotate(event_type='trial start')
+        >>> event.array_annotate(trial_id=[1, 2, 3])
         >>> seg = neo.Segment()
         >>> seg.events = [event]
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
of course.

JuliaSprenger:
@samuelgarcia @apdavison The first commit of the PR is a small, but important change I would like to implement rather sooner than later. Would you like me to open a separate PR with only this change, so we can merge it today?

apdavison:
> This PR also removes the notion of treating regular annotations that are lists of the shape of the data like array annotations. For this purpose array annotations should be used.

what is the motivation for removing that? It seems like a nice feature

JuliaSprenger:
To me this feature was useful while neo did not have array_annotations. The current mechanism automagically modifies some of the regular annotations in case the shape fits to the data dimensions. However, this might also occur by chance or might not be intended as regular annotations are not considered to relate to individual data points of the object. In case such a relation is intended, array_annotations should be used instead as there the relation to the data is clearly defined.

apdavison:
My estimation is that the likelihood of the annotation being of the same size as the data by chance is very, very small, and it would be nicer from a usability perspective to have a single method `annotate()` for both "array" and "regular" annotations - i.e. I would argue for removing the separate `array_annotate()` method.

(I don't feel at all strongly about this, however, and I agree it would be better to separate them cleanly than to keep the current half-and-half situation).

JuliaSprenger:
That's an interesting suggestion and would definitely simplify the user interface. However, I don't think it's a good solution as 1) even if the case sounds unlikely, having a regular annotation with the same length might occur in some cases and lead to confusing behaviour if automagic slicing is applied to annotations. 2) What behaviour would you expect in the case of a spiketrain only containing a single spike time. In this case basically all annotations would match the length of the data. Slicing this spiketrain would then potentially remove also all annotations.
I think that annotations and array_annotations should be kept separate and well documented to make them as user friendly as possible.

For the current situation however, I would argue, that at least the first commit should be merged to make the functionality of utils consistent - independent of the long term plans for annotations and array_annotations.

apdavison:
You are right. I think point (2) is the killer argument. Spiketrains, Events and Epochs containing only a few elements are very common, I guess I was thinking only of AnalogSignals.

I'm happy to merge this PR as is. Do you have any more commits to add, or shall I go ahead and merge?

JuliaSprenger:
I will do some more checks begining of next week. I maybe there will be some more small changes then. Let's see.

JuliaSprenger:
No more changes required here. Ready to merge from my side :)

PR #790, opened 2020-01-29
Fix reading RealWave Spike2 channels and channels with unrecognized quantities

SimonDanner:
RealWave channels were read as int32 instead of float32. 

Furthermore, logging wasn't imported in proxzobjects.py. This caused an error when the units were not recognized.


samuelgarcia:
Oups.
That was a big bug. Thank you for catching.

PR #791, opened 2020-02-06
Improvements to support for .pxp files by IgorIO

apdavison:
Add support for IgorIO.read(), read_block(), read_segment() for .pxp files (currently only read_analogsignal() works for such files)

Unfortunately, the only test file I have to hand is 480 MB, which I think is too big to add to the test file repository. Does anyone have any smaller .pxp files we could use for automated testing?

### DIFF
--- a/neo/io/igorproio.py
+++ b/neo/io/igorproio.py
@@ -128,7 +128,7 @@ class IgorIO(BaseIO):
                         location = location[element.encode('utf8')]
             data = location.wave
 
-        return self._wave_to_analogsignal(data['wave'])
+        return self._wave_to_analogsignal(data['wave'], [])
 
     def _wave_to_analogsignal(self, content, dirpath):
         if "padding" in content:
--- a/neo/io/igorproio.py
+++ b/neo/io/igorproio.py
@@ -76,7 +76,7 @@ class IgorIO(BaseIO):
         self._filesystem = None
 
     def read_block(self, lazy=False):
-        assert not lazy, 'Do not support lazy'
+        assert not lazy, 'This IO does not support lazy mode'
 
         block = Block(file_origin=self.filename)
         block.segments.append(self.read_segment(lazy=lazy))
@@ -84,7 +84,7 @@ class IgorIO(BaseIO):
         return block
 
     def read_segment(self, lazy=False):
-        assert not lazy, 'Do not support lazy'
+        assert not lazy, 'This IO does not support lazy mode'
         segment = Segment(file_origin=self.filename)
 
         if self.extension == 'pxp':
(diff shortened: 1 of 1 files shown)


apdavison:
@rgerkin would you mind taking a look?

a1eko:
Andrew, there is a small 64k test file 'polar-graphs-demo.pxp' here
https://github.com/wking/igor/blob/master/test/data/

rgerkin:
@apdavison I have lots of files that I can test this on.  My understanding of what you want to do here is:
-- read_segment will recursively load every wave or variable from the file, flattening the directory structure completely
-- read_block will create a block with one segment, equal to the result of read_segment

I tested it on one file and it works.  However, do you think it might be better to preserve the folder hierarchy in the Igor file (folder is just the Igor name for a directory within the .pxp file, not a directory on disk)?  Every folder could be a block (many people organize different recordings or stimulus conditions by folder).  It should be possible to do the same kind of recursive walk to implement that.  I'm not sure how to deal with subfolders (does Neo do blocks within blocks?).

apdavison:
@rgerkin if you have files that have a hierarchy, I'd be happy to improve the PR - the only files I have all have a flat folder structure.

One option is to have the bottom level of hierarchy be a Segment containing multiple signals, then the second level would be a Block containing multiple segments, finally we can return a list of blocks (this is what `IO.read()` returns). For hierarchies beyond three levels, we would need a special IO method to return lists of lists...

The semantics of the Neo Segment is that all the data objects inside have a common time reference. Would this fit with how .pxp files are usually used? Are there any metadata inside the files that indicate this, do you know?

An alternative (could be an option) would be to preserve the folder hierarchy through an annotation holding the original path of each signal.

apdavison:
@a1eko many thanks for that link.

rgerkin:
@apdavison I've sent you a link with some wild Igor file to use as test cases.  I can produce some more carefully structured new ones at some point to use for the resulting unit tests.  

Generally there is no guarantee about any metadata in an Igor .pxp file.  The acquisition package that I wrote a long time ago [here](https://github.com/rgerkin/recording-artist) has some, but lots of other people use Igor in different ways.  I think instead we should wait for something like [this](https://github.com/oruebel/ndx-icephys-meta) to be discussed and developed, and then Igor users can make sure to put the appropriate metadata into their files.  

Hierarchies could go arbitrarily deep.  In my acquisition package, each cell has a folder, and each folder has many objects corresponding to sequentially collected sweeps (AnalogSignals).  Sweeps with the same name across folders are recorded concurrently.  But for other acquisition packages in Igor, such as what Allen Institute uses, or what BBP has used, or other smaller labs, completely different conventions are used.  Even in my package, it is theoretically possible for the sampling rate to change between sweeps, so one cannot even rely on the same time base across the experiment.  

To account for all of this variability, your final suggestion of just keeping an annotation with the original path is probably best.

apdavison:
I finally got back to this. @rgerkin all the test files you sent load successfully, except "2018.06.02 Cell1.pxp" where I get an error from the igor package (https://pypi.python.org/pypi/igor/):
```
igor - ERROR - could not reshape data from [2] to b''
buffer is too small for requested array
```
I don't think there's any way to fix this from within Neo, so I think I will merge this now.

rgerkin:
@apdavison I haven't seen that before, and the `igor` package on PyPI hasn't been updated since 2016 so I'm not sure what happened.

PR #793, opened 2020-02-18
Add function to take the absolute value of analogsignals

morales-gregorio:
Dear all,

I have implemented an addtional function for the analogsignal class that should allow taking the absolute value of an analogsignal object without losing the annotations.

I have gone for a PR directly because I am indeed using this function already, it is working as expected. I thought it would be useful to have it available as a method.

Let me know your thoughts on it, or if something needs to be corrected.

Best,
Aitor

### DIFF
--- a/doc/source/authors.rst
+++ b/doc/source/authors.rst
@@ -51,6 +51,7 @@ and may not be the current affiliation of a contributor.
 * Christian Kothe
 * rishidhingra@github
 * Hugo van Kemenade
+* Aitor Morales-Gregorio [13]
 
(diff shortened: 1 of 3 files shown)


JuliaSprenger:
Hi @morales-gregorio. Thanks for adding new functionality. Could you also provide a test with it?

morales-gregorio:
I have now added the corresponding test, please have a look and let me know if anything else is needed.

apdavison:
This looks good to me. As I believe this is your first contribution, please could you add your name (and affiliation, if appropriate) to doc/source/authors.rst?

Issue #221, opened 2015-11-19
Add support for the NWB format


apdavison:
NWB = "Neurodata Without Borders" https://neurodatawithoutborders.github.io
- specification - https://github.com/NeurodataWithoutBorders/specification 
- Python APIs - (1) https://github.com/AllenInstitute/nwb-api/tree/master/ainwb (2) https://github.com/AllenInstitute/AllenSDK/blob/master/allensdk/core/nwb_data_set.py (3) https://github.com/NeurodataWithoutBorders/api-python 
- sample datasets @ CRCNS - https://crcns.org/NWB 
- sample datasets @ Allen Institute - http://alleninstitute.github.io/AllenSDK/cell_types.html#neurodata-without-borders

neuromusic:
+1

wvangeit:
+1

neuromusic:
as a heads-up @apdavison et al, NWB is going through a major refactor, including a new read/write API. @ajtritt & @oruebel are leading up the effort. the coming months might be a ripe time to get NWB support in Neo

apdavison:
Thanks, @neuromusic, I had gathered something was going on from some of your recent tweets. Is the specification itself being revised much, or just the Python API? Where should I look to follow what's going on?

neuromusic:
everything should settle down by ~ SfN. there's a lot of work being done right now, including on documentation and continuous integration

importantly, the NWB stack has been split up into an API layer, a schema specification layer, and a backend layer.

**pynwb** is the new API
- docs: http://pynwb.readthedocs.io/en/latest/
- repo: https://github.com/NeurodataWithoutBorders/pynwb

**nwb-schema** specifies the schema in a bunch of yml files
- docs: http://nwb-schema.readthedocs.io/en/latest/
- repo: https://github.com/NeurodataWithoutBorders/nwb-schema

There are some backward-incompatible changes to the schema (hence the 2.0 designation) but it's not an overhaul. Maybe we can try to get the NEO schema to become NWB 3.0 😄 

There's a google group setup, but there's nothing there beyond the announcement of the overhaul. You'l probably get more out of watching the issues on those two repos.

apdavison:
Great, thanks for the links!

rgerkin:
@apdavison @neuromusic Has anyone taken this up?  This seems pretty timely.  I could try to get a student on it if no one else has made any progress.

oruebel:
[...]

PR #796, opened 2020-03-06
Add NWBIO

apdavison:
This PR adds NWBIO, an IO module for reading and writing the Neurodata Without Borders format (cf #221).

This is a work in progress, which doesn't support all possible contents of an NWB file, but I think it is good enough for people to try it, and let us know (by commenting here) what doesn't work, or could be done better. Many thanks to @legouee for her work on this.

**TODO**

- [x] write Block, Segment, Epoch, Event, AnalogSignal, IrregularlySampledSignal, SpikeTrain using base TimeSeries
- [ ] write Unit, ChannelIndex (or whatever replaces them)
- [ ] write ImageSequence and related
- [x] allow user to choose where to write data (e.g. "acquisition", "stimulus", "analysis") by providing guiding annotations
- [x] allow user to write subclasses of TimeSeries by providing guiding annotations
- [x] read "acquisition" and "stimulus" groups
- [ ] read metadata associated with processing modules and make it available through annotations
- [ ] support array annotations

Usage example: https://github.com/apdavison/python-neo/blob/nwb2/examples/NWB-Allen-Institute-Example.ipynb


bendichter:
I tried this and got an error. Am I doing something wrong?

```python
url = "http://download.alleninstitute.org/informatics-archive/prerelease/H19.28.012.11.05-2.nwb"
local_filename = url.split("/")[-1]
if not exists(local_filename):
    local_filename, headers = urlretrieve(url)
    
ior = NWBIO(local_filename)
blocks = ior.read(lazy=True)

iow = NWBIO(filename='neo_test.nwb', mode='w')
iow.write_all_blocks(blocks)
```


```
TypeError                                 Traceback (most recent call last)
<ipython-input-16-c1c2eb01f5f2> in <module>
      1 test_file_name = 'neo_test.nwb'
      2 iow = NWBIO(filename=test_file_name, mode='w')
----> 3 iow.write_all_blocks(blocks)

~/dev/python-neo/neo/io/nwbio.py in write_all_blocks(self, blocks, **kwargs)
    290                 for block in blocks:
    291                     if annotation_name in block.annotations:
--> 292                         annotations[annotation_name].add(block.annotations[annotation_name])
    293                 if annotation_name in annotations:
    294                     if len(annotations[annotation_name]) > 1:

TypeError: unhashable type: 'dict'
```

bendichter:
Also, more broadly, let's think about the scope. It looks like NEO handles both intracellular electrophysiology and extracellular electrophysiology. In NWB, these different types of experiments have very different structures of metadata. The demo and tests seem to be focused on intracellular electrophysiology. Is that the goal for this, or do you also want to include extracellular?

samuelgarcia:
@bendichter : neo object are totaly agnostic of intra/extra/imaging electrophysology.

At writting time if it matter for NWB maybe we could use some annotation as a hint for the writter to use the good structure in the file.
I did not work at all on this IO so I can't tell.

samuelgarcia:
The two I see for this "routing" of AnalogSignal inside NWB is or/and:
  * some extra kwargs :  in write_block(block, sig_type='electrical_series')
  * use neo annotations at object level (aka metadata) anasig.annotate(sig_type='electrical_series')

bendichter:
Yes, that would put it in the right place, but it would still be missing some of the essential metadata. For instance, an `ElectricalSeries` requires a vector reference to rows of the electrodes table. The electrodes table stores information about the electrodes like position, device, filtering, impedance, etc. For a `TwoPhotonSeries` you need input what the excitations and emission lambdas are, the spacing of the image, information about the acquisition device, etc.

This could all be done with annotations, but they would need to be pretty sophisticated and have a specific structure. I think we should have a convo about this

bendichter:
Thanks for working on this, @apdavison. After your recent changes, I am running into this new error:

```python
ior = NWBIO(local_filename)
blocks = ior.read(lazy=True)

iow = NWBIO(filename='neo_test.nwb', mode='w')
iow.write_all_blocks(blocks)
```

```
TypeError                                 Traceback (most recent call last)
<ipython-input-2-3a8932303c81> in <module>
      5 
      6 iow = NWBIO(filename='neo_test.nwb', mode='w')
----> 7 iow.write_all_blocks(blocks)

~/dev/python-neo/neo/io/nwbio.py in write_all_blocks(self, blocks, **kwargs)
    337 
    338         for i, block in enumerate(blocks):
--> 339             self.write_block(nwbfile, block)
    340         io_nwb.write(nwbfile)
    341         io_nwb.close()

~/dev/python-neo/neo/io/nwbio.py in write_block(self, nwbfile, block, **kwargs)
    352             if not segment.name:
    353                 segment.name = "%s : segment%d" % (block.name, i)
--> 354             self._write_segment(nwbfile, segment)
    355         self.blocks_written += 1
    356 

~/dev/python-neo/neo/io/nwbio.py in _write_segment(self, nwbfile, segment)
    362             if not signal.name:
    363                 signal.name = "%s : analogsignal%d" % (segment.name, i)
--> 364             self._write_signal(nwbfile, signal)
    365 
    366         for i, train in enumerate(segment.spiketrains):

~/dev/python-neo/neo/io/nwbio.py in _write_signal(self, nwbfile, signal)
    400         else:
    401             raise TypeError("signal has type {0}, should be AnalogSignal or IrregularlySampledSignal".format(
--> 402                 signal.__class__.__name__))
    403         nwbfile.add_acquisition(tS)
    404         return tS

TypeError: signal has type AnalogSignalProxy, should be AnalogSignal or IrregularlySampledSignal
```

apdavison:
Hi @bendichter, we haven't yet implemented support for lazy objects. If you run 
`blocks = ior.read(lazy=False)`
it should now work (with the latest version of the nwb2 branch).

apdavison:
> This could all be done with annotations, but they would need to be pretty sophisticated and have a specific structure.

This is exactly the plan for properly supporting NWB (see the TODO list at the top of the PR). I don't really see any other way of doing it, although I suspect we will end up adding some utility functions to simplify the process of providing the correct annotations when creating the Neo structure.

In general terms, my proposed roadmap, which Elodie and I will be working on (together with anyone else who would like to help out!) is:
1. continue trying to round-trip (read and then write again) a wide range of NWB files; we will add annotations at the read stage, and then use those annotations at the write stage. The same annotations could be used by someone creating a Neo structure from scratch. It would be a great help if @bendichter could suggest some suitable NWB files, with a range of sophistication in their content, for such round-trip testing.
2. once we've proved the concept works with test files, we'll then systematically complete the annotation handling with reference to the NWB spec (hopefully we'll also be able to handle NWB extensions).
3. document the annotations, and possibly implement helper/utility functions to make adding these annotations easier for common use cases.

bendichter:
@apdavison OK, but handling the full range of NWB files will be quite a lot of work with this approach. Let's focus on intracellular electrophysiology for now, because I think that will be the easiest and you've already gotten a good start.

With your latest changes the code runs! Some discrepancies I'm finding in the round trip:

* the original has a structure `/general/intracellular_ephys/sweep_table` which matches stimuli to response. This is essential, because it's the only way we know which stimuli go with which responses. In this particular case, you could use the names of the stimuli and the responses, but that's not generally true.

* `/general/intracellular_ephys` also contains `IntracellularElectrode` objects named "Electrode 0" and "Electrode 1".

minor:
*  it appears the round trip creates a `/units` group that doesn't appear to contain any information.

* an empty "epochs" object was created as well.

apdavison:
Thanks for the feedback @bendichter.

@legouee is going to pick this up now, starting with the the specialized time series ("acquisition" should contain VoltageClampSeries, not plain TimeSeries), and then moving on to the other discrepancies until the output file is identical to the input.

I agree it makes sense to focus on perfecting intracellular ephys before moving on to more complex use cases.

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+
+IO class for reading data from a Neurodata Without Borders (NWB) dataset
+
+Documentation : https://neurodatawithoutborders.github.io
+Depends on: h5py, nwb, dateutil
+Supported: Read, Write
+Specification - https://github.com/NeurodataWithoutBorders/specification
+Python APIs - (1) https://github.com/AllenInstitute/nwb-api/tree/master/ainwb
+	          (2) https://github.com/AllenInstitute/AllenSDK/blob/master/allensdk/core/nwb_data_set.py
+              (3) https://github.com/NeurodataWithoutBorders/api-python
```
bendichter, line 13: PyNWB is not mentioned?

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+
+    def __init__(self, filename, mode='r'):
+        """
+        Arguments:
+            filename : the filename
+        """
+        if not have_pynwb:
+            raise Exception("Please install the pynwb package to use NWBIO")
+        if not have_hdmf:
+            raise Exception("Please install the hdmf package to use NWBIO")
```
bendichter, line 236: hdmf has been a requirement of pynwb for a long time, so I think having pynwb installed but not having hdmf installed would be unlikely.

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+            if not segment.name:
+                segment.name = "%s : segment%d" % (block.name, i)
+            self._write_segment(nwbfile, segment, electrodes)
+        self.blocks_written += 1
+
+    def _write_electrodes(self, nwbfile, block):
+        # this handles only icephys_electrode for now
+        electrodes = {}
+        devices = {}
+        for segment in block.segments:
```
bendichter, line 474: Currently in NWB it is a requirement that all times are with respect to the same session start time. I don't know how we would properly handle multiple segments.

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+
+        for i, epoch in enumerate(segment.epochs):
+            if not epoch.name:
+                epoch.name = "%s : epoch%d" % (segment.name, i)
+            self._write_epoch(nwbfile, epoch)
+
+    def _write_signal(self, nwbfile, signal, electrodes):
+        hierarchy = {'block': signal.segment.block.name, 'segment': signal.segment.name}
+        if "nwb_type" in signal.annotations:
+            timeseries_class = get_class(*signal.annotations["nwb_type"])
```
bendichter, line 519: We call this "neurodata_type". I understand why you might want to call it "nwb_type" here, but just letting you know in case you wanted to be consistent with our naming convention

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+        else:
+            timeseries_class = TimeSeries  # default
+        additional_metadata = {name[4:]: value
+                               for name, value in signal.annotations.items()
+                               if name.startswith("nwb:")}
+        if "nwb_electrode" in signal.annotations:
+            electrode_name = signal.annotations["nwb_electrode"]["name"]
+            additional_metadata["electrode"] = electrodes[electrode_name]
```
bendichter, line 527: This would only be relevant for descendants of `PatchClampSeries`. Extracellular electrophysiology analog signals form a map to electrodes in a different way. They use an `electrodes` arg that is a `DynamicTableRegion`, pointing to a region of the `electrodes` `DynamicTable`.

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+        if timeseries_class != TimeSeries:
+            conversion, units = get_units_conversion(signal, timeseries_class)
+            additional_metadata["conversion"] = conversion
```
bendichter, line 530: Why only get conversion when the neurodata_type is not `TimeSeries`? [`TimeSeries`](https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=TimeSeries#timeseries) and all its descendants contain a `conversion` field

bendichter:
@apdavison I'd like to work on this for extracellular electrophysiology, starting to think about writing.

ecephys requires an electrodes table that has information about each electrode, including:
* position (x,y,z) in the brain
* imp (impedance)
* location (brain region)
* filtering
* group (which is a `pynwb.ecephys.ElectrodeGroup` object).

The signal is in an `ElectricalSeries` object. `ElectricalSeries.electrodes` is a`DynamicTableRegion` object that points to a region of the electrodes table. Most of the time this region is the entire table, and we can probably assume that by default. That is generally broken when 1) some electrodes aren't recorded from or 2) when there are multiple groups of electrodes that have different metadata, like starting time.

I think we'd need some way to specify these electrodes. The only way I can think to do this is to specify them in the annotations of the `AnalogSignal` that will become the `ElectricalSeries`. My preference would be to just hold within `AnalogSignal.annotations` the indices that form the view of the electrodes table, and to specify the data that goes into this table elsewhere, but I am not sure where that would go, or how it would be expressed.

neo/io/nwbio.py
```diff
@@ -0,0 +1,810 @@
...
+
+    def _read_stimulus_group(self, lazy):
+        self._read_timeseries_group("stimulus", lazy)
+
+    def write_all_blocks(self, blocks, **kwargs):
+        """
+        Write list of blocks to the file
+        """
+        # todo: allow metadata in NWBFile constructor to be taken from kwargs
+        start_time = datetime.now()
```
bendichter, line 397: I don't like setting `now` as a default for `sessions_start_time`. It's almost always wrong.

apdavison: `session_start_time` is required by NWB, so would you suggest we raise an Exception here if a value is not provided?

neo/io/nwbio.py
apdavison: because in the general case of TimeSeries, there are no "expected" units, and so the conversion factor will always be 1.

neo/io/nwbio.py
apdavison: thanks for the pointer.

That reminds me, at the moment the `pynwb.icephys` module has `PatchClampSeries` but no `SharpElectrodeSeries`. Should I open a ticket for this?

neo/io/nwbio.py
apdavison: Perhaps "nwb_neurodata_type"? That preserves the clarity of the "nwb_" prefix in a Neo context, but makes the NWB equivalency more evident.

bendichter: yeah I think that would work

neo/io/nwbio.py
bendichter: sure! That would be a ticket in the nwb-schema repo

neo/io/nwbio.py
bendichter: Hmm, well TimeSeries is required to have a unit, and although it is not specified, `conversion` still takes you from the values of the data to the physical values. This is particularly useful when the original data is stored as `int16` values.

neo/io/nwbio.py
apdavison: What about raising an Exception when writing to NWB if any segments have overlapping (t_start, t_stop)?

neo/io/nwbio.py
bendichter: I think that's probably better. I know it's not ideal, but I don't like the idea of automatically writing data which we know to be incorrect.

neo/io/nwbio.py
bendichter: Maybe I am not fully understanding the concept of a segment. Do all segments have a common time reference? If so, I don't think there is an issue with writing multiple of them to NWB.

neo/io/nwbio.py
apdavison: the scenario I have in mind here is where "generic" Neo data is being saved to NWB format without the annotations needed to determine a more suitable subclass of `TimeSeries`. In such a scenario, the Neo `AnalogSignal` definitely has a unit, and the conversion from the on-disk data to the physical values would have already taken place in the Neo IO class.

neo/io/nwbio.py
apdavison: Segments do not necessarily have a common time reference, but they may have. If the segments' time values do not overlap, I think it is a reasonable assumption that they _do_ share a common reference.

neo/io/nwbio.py
legouee: Thanks for the pointer. You are right. PyNWB is missing. I add it.

neo/io/nwbio.py
legouee: I had some issues with the hdmf package a long time ago. It is for this reason that I added this additional condition. If now hdmf is a requirement of pynwb, these two lines are no longer needed.

neo/io/nwbio.py
legouee: You are right. "nwb neurodata_type" seems to be clear to both NWB and Neo.

examples/NWB-Allen-Institute-Example.ipynb
```diff
@@ -0,0 +1,1464 @@
+{
```
samuelgarcia, line 1: I would not push NB in git.
I propose instead when possible to use script + sphinx-gallery that generate NB

neo/io/proxyobjects.py
```diff
@@ -224,6 +205,34 @@ def load(self, time_slice=None, strict_slicing=True,
...
+            :strict_slicing: True by default.
+                Control if an error is raise or not when one of  time_slice member
+                (t_start or t_stop) is outside the real time range of the segment.
+        '''
+
+        if channel_indexes is None:
+            channel_indexes = slice(None)
+
+        i_start, i_stop, sig_t_start = self._time_slice_indices(time_slice,
+                                                                strict_slicing=strict_slicing)
```
samuelgarcia, line 235: I don't anderstand the patch here.
Git make strange stuff.

neo/io/nwbio.py
```diff
@@ -0,0 +1,811 @@
...
+
+import numpy as np
+import quantities as pq
+from neo.io.baseio import BaseIO
+from neo.io.proxyobjects import (
+    AnalogSignalProxy as BaseAnalogSignalProxy,
+    EventProxy as BaseEventProxy,
+    EpochProxy as BaseEpochProxy,
+    SpikeTrainProxy as BaseSpikeTrainProxy
+)
```
samuelgarcia, line 39: I plan to make a BaseAnalogSignalProxy very soon.
Maybe we should wait for this.

neo/io/nwbio.py
```diff
@@ -0,0 +1,811 @@
...
+            raise NotImplementedError("Compound units not yet supported")  # e.g. volt^2
+        uq_def = uq.definition
+        return float(uq_def.magnitude), uq_def
+    conv, unit2 = _decompose(unit)
+    while conv != 1:
+        conversion *= conv
+        unit = unit2
+        conv, unit2 = _decompose(unit)
+    return list(unit.dimensionality.keys())[0].name, conversion
+
```
samuelgarcia, line 188: Maybe this coulld live in a units_tools.py file. No ?

neo/io/nwbio.py
```diff
@@ -0,0 +1,811 @@
...
+    UnitCurrent('nanoampere', 0.001 * uA, 'nA')
+
+    """
+    if conversion not in prefix_map:
+        raise ValueError(f"Can't handle this conversion factor: {conversion}")
+    unit_name = prefix_map[conversion] + base_unit_name
+    if unit_name[-1] == "s":  # strip trailing 's', e.g. "volts" --> "volt"
+        unit_name = unit_name[:-1]
+    return getattr(pq, unit_name)
+
```
samuelgarcia, line 206: Maybe this coulld live in a units_tools.py file. No ?

neo/io/nwbio.py
```diff
@@ -0,0 +1,811 @@
...
+        self._read_units(lazy=lazy)
+        self._read_epochs_group(lazy)
+
+        return list(self._blocks.values())
+
+    def read_block(self, lazy=False, **kargs):
+        """
+        Load the first block in the file.
+        """
+        return self.read_all_blocks(lazy=lazy)[0]
```
samuelgarcia, line 278: This implementation is quite strange.
generaly, the read_all_blocks() loop over read_block()

I also would add a block_index here instead of always use 0

apdavison:
There are many things still to improve, but I think NWBIO is now suitable for intracellular electrophysiology data. I think this can be merged now, and future improvements can be made in separate PRs.

examples/NWB-Allen-Institute-Example.ipynb
JuliaSprenger: I agree, NB can become quite big if they include images and are updated each time they are executed.

neo/io/proxyobjects.py
JuliaSprenger: The code was just moved to a dedicated `_time_slice_indices` method. @samuelgarcia Maybe your confusion fixed in the latest version?

examples/NWB-Allen-Institute-Example.ipynb
apdavison: removed

neo/io/nwbio.py
JuliaSprenger: @samuelgarcia  This PR has priority over the new BaseAnalogSignalProxy implementation, no?

neo/io/nwbio.py
samuelgarcia: I agree I think we can remove this have_hdmf, no ?

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        self._read_units(lazy=lazy)
+        self._read_epochs_group(lazy)
+
+        return list(self._blocks.values())
+
+    def read_block(self, lazy=False, block_index=0, **kargs):
+        """
+        Load the first block in the file.
+        """
+        return self.read_all_blocks(lazy=lazy)[block_index]
```
samuelgarcia, line 283: Is it efficient to do this ? read all of then and only render one ?

neo/io/nwbio.py
apdavison: We could, but I don't think it causes any harm, so I suggest we leave it for now so we can focus on getting the release finished.

neo/io/nwbio.py
apdavison: no, this could be improved in a future PR

neo/test/iotest/test_nwbio.py
```diff
@@ -0,0 +1,262 @@
...
+except (ImportError, SyntaxError):
+    NWBIO = None
+    HAVE_PYNWB = False
+import quantities as pq
+import numpy as np
+from numpy.testing import assert_array_equal, assert_allclose
+
+
+@unittest.skipUnless(HAVE_PYNWB, "requires pynwb")
+class TestNWBIO(unittest.TestCase):
```
samuelgarcia, line 31: All our IO must have at least class test that inherits
`BaseTestIO` because it run standard test for all IOs.

Here we will run only specific test.


We also need to push some nwb data in gin.

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+            conversion, units = get_units_conversion(signal, timeseries_class)
+            additional_metadata["conversion"] = conversion
+        else:
+            units = signal.units
+        if isinstance(signal, AnalogSignal):
+            sampling_rate = signal.sampling_rate.rescale("Hz")
+            tS = timeseries_class(
+                name=signal.name,
+                starting_time=time_in_seconds(signal.t_start),
+                data=signal,
```
samuelgarcia, line 549: With implementation we handle the signal in memmory.
We also should have later a mechanism that write a AnalogSignal in a chunk mode to avoid a full load in memory otherwise it will be impossible to write big files to nwb.

Maybe we can open issue for this to not forget it.

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        if not have_hdmf:
+            raise Exception("Please install the hdmf package to use NWBIO")
+        BaseIO.__init__(self, filename=filename)
+        self.filename = filename
+        self.blocks_written = 0
+        self.nwb_file_mode = mode
+
+    def read_all_blocks(self, lazy=False, **kwargs):
+        """
+
```
JuliaSprenger, line 250: there seems to be some documentation intended here...

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+                    annotations[annotation_name], = annotations[annotation_name]
+        if "identifier" not in annotations:
+            annotations["identifier"] = self.filename
+        if "session_description" not in annotations:
+            annotations["session_description"] = blocks[0].description or self.filename
+            # todo: concatenate descriptions of multiple blocks if different
+        if "session_start_time" not in annotations:
+            raise Exception("Writing to NWB requires an annotation 'session_start_time'")
+        # todo: handle subject
+        # todo: store additional Neo annotations somewhere in NWB file
```
JuliaSprenger, line 432: are these todos for this or the next PR? If they are going to be postponed, would it make sense to collect all of those options for improvements in a single issue?

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        io_validate = pynwb.NWBHDF5IO(self.filename, "r")
+        errors = pynwb.validate(io_validate, namespace="core")
+        if errors:
+            raise Exception(f"Errors found when validating {self.filename}")
+        io_validate.close()
+
+    def write_block(self, nwbfile, block, **kwargs):
+        """
+        Write a Block to the file
+            :param block: Block to be written
```
JuliaSprenger, line 469: this is incompletely documented

neo/io/nwbio.py
JuliaSprenger: this should go into the collective improvements issue for nwb

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        nwbfile.add_unit(spike_times=spiketrain.rescale('s').magnitude,
+                         obs_intervals=[[float(spiketrain.t_start.rescale('s')),
+                                         float(spiketrain.t_stop.rescale('s'))]],
+                         _name=spiketrain.name,
+                         # _description=spiketrain.description,
+                         segment=spiketrain.segment.name,
+                         block=spiketrain.segment.block.name)
+        # todo: handle annotations (using add_unit_column()?)
+        # todo: handle Neo Units
+        # todo: handle spike waveforms, if any (see SpikeEventSeries)
```
JuliaSprenger, line 588: same here?

neo/io/nwbio.py
JuliaSprenger: In fact, the BaseAnalogSignalproxy could be later moved from here to a central location...

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+                    electrode_metadata[field_name] = value
+            for field_name in timeseries.electrode.device.__class__.__nwbfields__:
+                value = getattr(timeseries.electrode.device, field_name)
+                if value is not None:
+                    electrode_metadata["device"][field_name] = value
+            self.annotations["nwb_electrode"] = electrode_metadata
+
+    def load(self, time_slice=None, strict_slicing=True):
+        """
+        *Args*:
```
JuliaSprenger, line 677: I don't think this style of docstrings is used in many other classes of neo.. Maybe it would be good to use a more consistent style and check also that is is rendered properly in the documentation on readthedocs?

neo/test/iotest/test_nwbio.py
```diff
@@ -0,0 +1,262 @@
...
+            local_filename = os.path.join(self.local_test_dir, url.split("/")[-1])
+            if not os.path.exists(local_filename):
+                try:
+                    urlretrieve(url, local_filename)
+                except IOError as exc:
+                    raise unittest.TestCase.failureException(exc)
+            io = NWBIO(local_filename, 'r')
+            blocks = io.read()
+
+    def test_roundtrip(self):
```
JuliaSprenger, line 52: nice! I think having a roundtrip test is essential!

JuliaSprenger:
@apdavison @legouee I think we can merge this PR soon to have a first version of the NWBIO in the next release. The most important change requests for this would be the incorporation of some testfiles also on gin (if you have some, as we can then tag the release to a fixed version of the test files), the incorporation of the BaseTests and some completion that is still lacking on the docstrings. Also I am not sure about some of the discussions in this thread form the end of last year, if there are outdated, could you resolve them?

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+                     name=self.name)
+
+
+class SpikeTrainProxy(BaseSpikeTrainProxy):
+
+    def __init__(self,  units_table, id):
+        self._units_table = units_table
+        self.id = id
+        self.units = pq.s
+        t_start, t_stop = units_table.get_unit_obs_intervals(id)[0]
```
bendichter, line 787: If I understand correctly, this will only work if there is a single observation interval. It will not work if there are no observation intervals, and if there are multiple, it will only output the first.

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+class EpochProxy(BaseEpochProxy):
+
+    def __init__(self, epochs_table, epoch_name=None, index=None):
+        self._epochs_table = epochs_table
+        if index is not None:
+            self._index = index
+            self.shape = (index.sum(),)
+        else:
+            self._index = slice(None)
+            self.shape = epochs_table.n_rows  # untested, just guessed that n_rows exists
```
bendichter, line 758:
```suggestion
            self.shape = (len(epochs_table),)
```

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        self.name = epoch_name
+
+    def load(self, time_slice=None, strict_slicing=True):
```
bendichter, line 761: time_slice appears to do nothing. How is this supposed to interact with `self._index`?

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        """
+        interval = None
+        if time_slice:
+            interval = (float(t) for t in time_slice)  # convert from quantities
+        spike_times = self._units_table.get_unit_spike_times(self.id, in_interval=interval)
+        return SpikeTrain(
+                    spike_times * self.units,
+                    self.t_stop,
+                    units=self.units,
+                    #sampling_rate=array(1.) * Hz,
```
bendichter, line 815: https://pynwb.readthedocs.io/en/stable/pynwb.misc.html?highlight=Units#pynwb.misc.Units
```suggestion
                    waveform_rate,
```

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        self.filename = filename
+        self.blocks_written = 0
+        self.nwb_file_mode = mode
+
+    def read_all_blocks(self, lazy=False, **kwargs):
+        """
+
+        """
+        assert self.nwb_file_mode in ('r',)
+        io = pynwb.NWBHDF5IO(self.filename, mode=self.nwb_file_mode)  # Open a file with NWBHDF5IO
```
bendichter, line 253:
```suggestion
        io = pynwb.NWBHDF5IO(self.filename, mode=self.nwb_file_mode, load_namespaces=True)  # Open a file with NWBHDF5IO
```

This will automatically generate the API for any extensions in the file.

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+            else:  # IrregularlySampledSignal
+                signal = AnalogSignalProxy(timeseries, group_name)
+                if not lazy:
+                    signal = signal.load()
+                segment.irregularlysampledsignals.append(signal)
+                signal.segment = segment
+
+    def _read_units(self, lazy):
+        if self._file.units:
+            for id in self._file.units.id[:]:
```
bendichter, line 375:
```suggestion
            for id in range(len(self._file.units)):
```

`id` is by default 0, 1, 2, 3, ... but that can be changed to any unique set of integers. I think this would be safer if you are going to use it to index a list

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        if "session_start_time" not in annotations:
+            raise Exception("Writing to NWB requires an annotation 'session_start_time'")
+        # todo: handle subject
+        # todo: store additional Neo annotations somewhere in NWB file
+        nwbfile = NWBFile(**annotations)
+
+        assert self.nwb_file_mode in ('w',)  # possibly expand to 'a'ppend later
+        if self.nwb_file_mode == "w" and os.path.exists(self.filename):
+            os.remove(self.filename)
+        io_nwb = pynwb.NWBHDF5IO(self.filename, manager=get_manager(), mode=self.nwb_file_mode)
```
bendichter, line 438: What is the purpose of using `manager=get_manager()` here?

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+        for i, block in enumerate(blocks):
+            self.write_block(nwbfile, block)
+        io_nwb.write(nwbfile)
+        io_nwb.close()
+
+        io_validate = pynwb.NWBHDF5IO(self.filename, "r")
+        errors = pynwb.validate(io_validate, namespace="core")
+        if errors:
+            raise Exception(f"Errors found when validating {self.filename}")
+        io_validate.close()
```
bendichter, line 464:
```suggestion
        with pynwb.NWBHDF5IO(self.filename, "r") as  io_validate:
            errors = pynwb.validate(io_validate, namespace="core")
            if errors:
                raise Exception(f"Errors found when validating {self.filename}")
```

Using a context is preferred here because it automatically closes the file in case an error prevents the full execution of the inner code.

neo/io/nwbio.py
bendichter, line 588: You should be able to just add a `waveforms` field here.

bendichter: https://pynwb.readthedocs.io/en/stable/pynwb.misc.html?highlight=Units#pynwb.misc.Units.add_unit

neo/io/nwbio.py
apdavison: added to #1013

neo/io/nwbio.py
apdavison: created https://github.com/NeurodataWithoutBorders/nwb-schema/issues/479

neo/io/nwbio.py
apdavison: this is a "todo"

JuliaSprenger:
@legouee @apdavison The test file is now on gin. Can you include it in the set of files used by the unittests and also include the common rawio tests by inheriting from `BaseTestIO`?

neo/io/nwbio.py
JuliaSprenger: is this on the todo list? #1013

samuelgarcia:
@JuliaSprenger : since it now use `BaseTestIO` for tests, I think we can merge no ?

samuelgarcia:
There are some warning in test_nwb.py
I guess we can live with then for a while.

neo/io/nwbio.py
JuliaSprenger: seems to be implemented.

neo/io/nwbio.py
JuliaSprenger: added to #1013

neo/io/nwbio.py
JuliaSprenger: `now` has been replaced.

neo/test/iotest/test_nwbio.py
JuliaSprenger: done

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+from neo.core import baseneo
+
+import logging
+import os
+from itertools import chain
+from datetime import datetime
+import json
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:  # Python 2
```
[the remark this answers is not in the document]
JuliaSprenger: has been removed

neo/io/nwbio.py
```diff
@@ -0,0 +1,823 @@
...
+    from pynwb import image
+    from pynwb.image import ImageSeries
+    from pynwb.spec import NWBAttributeSpec, NWBDatasetSpec, NWBGroupSpec, NWBNamespace, NWBNamespaceBuilder
+    from pynwb.device import Device
+    # For calcium imaging data
+    from pynwb.ophys import TwoPhotonSeries, OpticalChannel, ImageSegmentation, Fluorescence
+    have_pynwb = True
+except ImportError:
+    have_pynwb = False
+except SyntaxError:  # pynwb doesn't support Python 2.7
```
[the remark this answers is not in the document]
JuliaSprenger: has been removed

neo/io/nwbio.py
JuliaSprenger: added to #1013

neo/io/nwbio.py
JuliaSprenger: this is part of #1013

neo/io/nwbio.py
JuliaSprenger: added to #1013

neo/io/nwbio.py
JuliaSprenger: get_manager is not used any more

JuliaSprenger (approved):

JuliaSprenger:
@apdavison @legouee I cleaned all open comments, moved unresolved ones to #1013 for the future. I fixed the layout to be pep8 compatible and fixed some docstrings / typos. Under the assumption that you don't plan to include https://github.com/apdavison/python-neo/pull/6 here any more I would be ok to merge the current version and fix the remaining issues listed in #1013 in the next release.

Issue #786, opened 2019-12-17
[NixIOFr] ChannelIndex and AnalogSignal mixup when loading


morales-gregorio:
Dear all,

I am working with relatively big data files (>10GB) and the lazy mode for the BlackrockIO is currently my best friend. I think it is an amazing feature :)

I wanted to save the processed data as `.nix` files, but unfortunately no lazy mode is available and it becomes quite complicated to work with the big datafiles. Is there a plan to implement it? Is there already a developer/beta version of this?

I am using neo `0.8.0` and nixio `v1.5.0b4`.

All best,

Aitor

achilleas-k:
The are two NIX readers in Neo, the old IO that also supports writing (NixIO) and the new Raw-based class that only supports reading (NixIOFr).  The latter, NixIOFr, uses the raw IO API and supports lazy loading.

The following should work with your package versions.

```python
io = neo.NixIOFr("stuff.nix")
blk = io.read_block(lazy=lazy)

signalproxy = blk.segments[0].analogsignals[0]
print(signalproxy)  # AnalogSignalProxy object
signal = signalproxy.load()
print(signal)  # AnalogSignal data
```

rgutzen:
Thank you for that the info. I'm trying NixIOFr now as you suggested, but run into a ValueError `Not a location (invalid object ID)`

Here a minimal code:
```
asig = neo.AnalogSignal(np.arange(100).reshape((10,10)),
                        units='mV',
                        sampling_rate=1000*pq.Hz)
blk = neo.Block()
blk.segments.append(neo.Segment())
blk.segments[0].analogsignals.append(asig)

with neo.NixIO('temp.nix') as io:
    io.write_block(blk)

with neo.NixIOFr('temp.nix') as io:
    blk = io.read_block(lazy=True)

blk.segments[0].analogsignals[0].load()
```

neo: 0.9.0dev
nixio : v1.5.0b4

achilleas-k:
The file is closed when exiting the context manager, so there's no way to load the data from the file outside the context block.
This should work:
```python
import neo
import numpy as np
import quantities as pq


asig = neo.AnalogSignal(np.arange(100).reshape((10, 10)),
                        units='mV',
                        sampling_rate=1000*pq.Hz)
blk = neo.Block()
blk.segments.append(neo.Segment())
blk.segments[0].analogsignals.append(asig)

with neo.NixIO('temp.nix') as io:
    io.write_block(blk)

[...]

PR #797, opened 2020-03-12
make ChannelIndex creation in RawIO more robust

rgutzen:
This fixes the issue of getting a TypeError when loading a file with a RawIO, and an (array_)annotation is named like a ChannelIndex keyword argument, e.g., 'channel_names'. 
See the related Issue #786


JuliaSprenger:
@rgutzen Thanks for the quick fix! Looks functional.

PR #798, opened 2020-03-27
 Fix: In openephysrawio.py get segment indices from folder exploration…

remi-pr:
… instead of assuming continuous numbering

In the IO for OpenEPhys data, under certain conditions (eg. users record into the same directory, but later remove some segments) opening the folder becomes impossible.
This is because the code assumes the segments are continuously numbered, and they are loop through using` range(nb_segments)`..
This does not work when the user removes, after the fact, some recordings. In the proposed fix one would loop using the keys of the `info['continuous']` dictionary which are obtained by parsing the file names. This should be more robust.
I am happy to provide some data to test.

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -257,14 +257,14 @@ class OpenEphysRawIO(BaseRawIO):
         self._generate_minimal_annotations()
         bl_ann = self.raw_annotations['blocks'][0]
         for seg_index, oe_index in enumerate(oe_indices):
-            seg_ann = bl_ann['segments'][oe_index]
+            seg_ann = bl_ann['segments'][seg_index]
             if len(info['continuous']) > 0:
                 fullname = os.path.join(self.dirname, info['continuous'][oe_index][0])
                 chan_info = read_file_header(fullname)
                 seg_ann['openephys_version'] = chan_info['version']
                 bl_ann['openephys_version'] = chan_info['version']
                 seg_ann['date_created'] = chan_info['date_created']
-                seg_ann['segment_index'] = oe_index + 1
+                seg_ann['openephys_segment_index'] = oe_index + 1
 
     def _segment_t_start(self, block_index, seg_index):
         # segment start/stop are difine by  continuous channels
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -81,7 +81,8 @@ class OpenEphysRawIO(BaseRawIO):
         self._sig_length = {}
         self._sig_timestamp0 = {}
         sig_channels = []
-        for seg_index in range(nb_segment):
+        seg_indices = sorted(list(info['continuous'].keys()))
+        for ix, seg_index in enumerate(seg_indices):
             self._sigs_memmap[seg_index] = {}
 
             all_sigs_length = []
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Remi.
Thank you for patch.
This is unfortuantly not good.
In neo rawio seg_index have to be continuous.
In short seg_index is your ix.

So this ```for seg_index in range(nb_segment):``` is always true and important for annoatations steps.

You could go for something like this
```python
oe_indexes = sorted(list(info['continuous'].keys()))
for seg_index, oe_index in enumerate(oe_indexes):
   ...
```

Note that you could also you could insert the oe_index inside annoattions for index so that the end user could retrieve it.

remi-pr:
OK, I think I understand and I changed the code to follow your advice the best I could.
If that fits better with the rest of the code I will have a question on how to count the number of segments.

neo/rawio/openephysrawio.py
```diff
@@ -255,14 +256,15 @@ def _parse_header(self):
         # Annotate some objects from coninuous files
         self._generate_minimal_annotations()
         bl_ann = self.raw_annotations['blocks'][0]
-        for seg_index in range(nb_segment):
-            seg_ann = bl_ann['segments'][seg_index]
+        for seg_index, oe_index in enumerate(oe_indices):
+            seg_ann = bl_ann['segments'][oe_index]
```
samuelgarcia: No here it is seg_index, because neo.io route annoatations with this seg_index

neo/rawio/openephysrawio.py
```diff
@@ -255,14 +256,15 @@ def _parse_header(self):
...
             if len(info['continuous']) > 0:
-                fullname = os.path.join(self.dirname, info['continuous'][seg_index][0])
+                fullname = os.path.join(self.dirname, info['continuous'][oe_index][0])
                 chan_info = read_file_header(fullname)
                 seg_ann['openephys_version'] = chan_info['version']
                 bl_ann['openephys_version'] = chan_info['version']
                 seg_ann['date_created'] = chan_info['date_created']
+                seg_ann['segment_index'] = oe_index + 1
```
samuelgarcia: I would not use 'segment_index' because it is ambiguous but maybe
'origin_segment_index' or 'openephys_segment_index' or ...
Does openephyse is "one based" ?
Is the +1 relevent ?

neo/rawio/openephysrawio.py
remi-pr: Oh yes absolutely, my mistake.

neo/rawio/openephysrawio.py
remi-pr: `openephys_segment_index` sounds better indeed. 
I do not really have an opinion on the +1. I added it because there is a subtraction happening when exploring the folder: see for example in the [docstring](https://github.com/NeuralEnsemble/python-neo/blob/62d8e3ccda1d379d5eb2eea8cc64504b74ba434e/neo/rawio/openephysrawio.py#L437) of `explore_folder`. I'll wait for your opinion before making another commit.

samuelgarcia: OK. I forgot about this.
OE is quite inconsistent about one (seg) or zero (chans) based.
It make sense to keep thos +1 so.

samuelgarcia:
Sorry for long delay.
I think this is ready to merge no ?

remi-pr:
If you are happy with it then yes, sure.

Issue #801, opened 2020-04-01
Invalid length of array annotations for Spiketrain, Event and Epoch


JuliaSprenger:
When creating a Spiketrain, Event or Epoch is is not ensured that the data are a one dimensional array. Since array_annotations are assumed to match the last dimension, this can lead to unexpected behaviour, e.g.

```python
e = neo.SpikeTrain(np.array([1,2,3]).reshape((3,-1))*pq.s, t_stop=10*pq.s)
print(e.array_annotations.length)
```

Here the array annotations are expected to be of length 1, even though the length of the data is 3.

A potential solution to this is to ensure that data arrays for Spiketrains, Events and Epochs are always 1-dimensional.

PR #802, opened 2020-04-02
Enforce times of Events, Epochs, Spiketrains to be 1d

JuliaSprenger:
The dimensions of time arrays of Event, Epoch and Spiketrains are frequently assumed to be 1-dimensional, e.g. see #801, but this is never enforced.
This lead to inconsistent behaviour with the introduction of array_annotations, which are always matching the last dimension of a data object. In cases when one of the objects above was initialised with a 2-dimensional array containing data only along the first dimension, this leads to array_annotations always being forced to be of length 1 even though there is more than one sample contained.

This PR enforces the time arrays of Events, Epochs and Spiketrains to be 1-dimensional via reshaping and adds corresponding tests.

### DIFF
--- a/neo/core/epoch.py
+++ b/neo/core/epoch.py
@@ -92,7 +92,7 @@ class Epoch(DataObject):
         elif isinstance(times, (list, tuple)):
             times = np.array(times)
         if len(times.shape) > 1:
-            times = times.reshape(-1)
+            raise ValueError("Times array has more than 1 dimension")
         if isinstance(durations, (list, tuple)):
             durations = np.array(durations)
         if durations is None:
--- a/neo/core/epoch.py
+++ b/neo/core/epoch.py
@@ -91,6 +91,8 @@ class Epoch(DataObject):
             times = np.array([]) * pq.s
         elif isinstance(times, (list, tuple)):
             times = np.array(times)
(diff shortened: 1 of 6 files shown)


neo/core/epoch.py
```diff
@@ -91,6 +91,8 @@ def __new__(cls, times=None, durations=None, labels=None, units=None, name=None,
             times = np.array([]) * pq.s
         elif isinstance(times, (list, tuple)):
             times = np.array(times)
+        if len(times.shape) > 1:
```
apdavison, line 94: I think this could lead to unexpected behaviour. Wouldn't it be better to raise an exception if the array is not 1D?

JuliaSprenger: Yes, you are right, that this is a 'magical' conversion not visible to the user and might therefore be unexpected. However, I think that raising an exception here instead of doing the conversion might break existing code if there are additional, empty dimensions.
For now I will add an exception here.

apdavison: I see your point. Since we're not yet at version 1.0 I think it's ok to just add the exception now. In future, we should probably add a deprecation warning in one major release then add the exception in the next one.

JuliaSprenger:
@apdavison Ready to merge from my side.

PR #760, opened 2019-10-31
[W.i.p] test_nixio.py

teogale:
End of my internship. Here are the latest modification i made to test_nixio.py not everything is working.
Probably some part or also missed.

PR #804, opened 2020-04-06
NIXIO ImageSequence support

achilleas-k:
Rebased and cleaned the branch for PR #760.

### DIFF
--- a/neo/test/iotest/test_nixio.py
+++ b/neo/test/iotest/test_nixio.py
@@ -199,7 +199,7 @@ class NixIOTest(unittest.TestCase):
             for da in data_arrays:
                 if da.metadata.name == nixname:
                     dalist.append(da)
-                nsig = np.shape(sig)[-1]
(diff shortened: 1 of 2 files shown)


neo/test/iotest/test_nixio.py
```diff
@@ -195,7 +199,7 @@ def compare_signals_das(self, neosignals, data_arrays):
             for da in data_arrays:
                 if da.metadata.name == nixname:
                     dalist.append(da)
-            _, nsig = np.shape(sig)
+                nsig = np.shape(sig)[-1]
```
JuliaSprenger, line 202: I think this is indented one level too deep. Would make more sense outside of the foor loop, no?

achilleas-k: Yes, good catch.

JuliaSprenger:
The travis tests succeeded but this was not displayed properly by github.

PR #806, opened 2020-04-17
Support Empty Array Annotations in NixIO

Kleinjohann:
Currently, the `property.definition` attribute is used to tag both empty annotations and array annotations. This breaks down in the case of empty array annotations. 

I moved the array annotation tag to `property.type` which appeared to be unused as well, and I added a test with an empty array annotation. 

@achilleas-k is using the `property.type` attribute feasible / does that cause any other issues down the line? It only seems to work with `nixio<=1.5.0b3`.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -854,7 +854,7 @@ class NixIO(BaseIO):
         if anasig.array_annotations:
             for k, v in anasig.array_annotations.items():
                 p = self._write_property(metadata, k, v)
-                p.definition = ARRAYANNOTATION
+                p.type = ARRAYANNOTATION
 
         self._signal_map[nix_name] = nixdas
 
@@ -992,7 +992,7 @@ class NixIO(BaseIO):
         if irsig.array_annotations:
             for k, v in irsig.array_annotations.items():
                 p = self._write_property(metadata, k, v)
-                p.definition = ARRAYANNOTATION
+                p.type = ARRAYANNOTATION
 
         self._signal_map[nix_name] = nixdas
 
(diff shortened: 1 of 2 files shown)


achilleas-k:
There shouldn't be any issue using the property.type since we don't use it anywhere else in the Neo NIXIO.  Not sure exactly what bug this is triggering.  It seems there's an issue with the `_parent` reference for sections.  I can see it breaking locally, but when trying to reproduce with a simpler script, it's not occuring, so I'm going to have to figure out exactly what's going on.

achilleas-k:
Didn't get the change to look at this closely during the last few days but I narrowed down the issue:
G-Node/nixpy#460

Should be able to fix it in the next day or so and get back to you.

achilleas-k:
G-Node/nixpy#462 fixes the issue (tested locally with your branch).  Should be merged soon and I'll schedule a release.

JuliaSprenger:
@achilleas-k Any news here? Are you planning another beta release soon?

achilleas-k:
Sorry for not getting back to you sooner, I was off last week.

I'm working on something now that should be the last bit before we can move forward with 1.5 proper.  I'll probably release another beta when it's ready before finalising everything.  I'm hoping I can get it out sometime next week.

JuliaSprenger:
@achilleas-k Any news on the nixpy side? Let's us know when you are ready ;)

achilleas-k:
Thanks for the nudge, Julia.  The latest beta should work.  I had forgotten about this issue.

JuliaSprenger:
The failing tests are not related to the NixIO and are fixed in master already.

Issue #808, opened 2020-04-17
quantities breaks/fails on rescaling


TRuikes:
We encoutered two issues:
```python
import quantities as pq
import numpy as np
import neo
times = np.arange(0, 3) * pq.s
ev = neo.Event(
    times=times
)
new_time = ev.rescale('us')
```

giving: 
```

---------------------------------------------------------------------------

TypeError                                 Traceback (most recent call last)

<ipython-input-14-2b5a13ab6d4a> in <module>
      6     times=times
      7 )
----> 8 new_time = ev.rescale('us')

~\Miniconda3\envs\sp_git\lib\site-packages\neo\core\event.py in rescale(self, units)
    165         units
    166         '''
--> 167         obj = super(Event, self).rescale(units)
    168         obj.segment = self.segment
    169         return obj

~\Miniconda3\envs\sp_git\lib\site-packages\neo\core\dataobject.py in rescale(self, units)
    280 
    281         # Rescale the object into a new object
--> 282         obj = self.duplicate_with_new_data(signal=self.view(pq.Quantity).rescale(dim), units=units)
    283 
    284         # Expected behavior is deepcopy, so deepcopying array_annotations

TypeError: duplicate_with_new_data() got an unexpected keyword argument 'signal'

```

and:
```python

import quantities as pq
import numpy as np
import neo
times = np.arange(0, 3) * pq.s
ev = neo.Event(
    times=times
)

a = 1*pq.us
b = 2*pq.us

ev_a = ev - a
ev_b = ev + b

for t0, t1 in zip(ev_a, ev_b):
    print(t1-t0)
```
Yielding not exact results:
```

3e-06 s
3.000000000086267e-06 s
2.9999999997532e-06 s
```

which we deal with by rescaling `ev` to us (not trough ev.rescale('us'), see 1st issue), then it does return exact values.

apdavison:
The first issue is a bug, it should be fixedd by #810.

The second issue is fundamentally a problem of the floating point representation, there's nothing we can do at the level of Neo, I don't think.

PR #810, opened 2020-04-25
Fix and tests for #808

apdavison:


JuliaSprenger:
Looks good to me, thanks for fixing it!

PR #815, opened 2020-05-05
Add 'keep_original_times' keyword for NeuralynxIO

### LINKED ISSUE #814: Neuralynx loader output for a given file differs depending on what other files are present in folder
When loading a Neuralynx data file, the timestamps assigned by neo depend on what other files are present in the same folder. 

For instance, loading a .ntt file with no other files present gives a first timestamp of 0.0 s, but loading that same file with some .ncs files also present gives a first timestamp of ~0.12 s.

Minimal working example [here](https://github.com/vandermeerlab/data-formats/blob/master/neo-nept-loading.ipynb).

I would consider this behavior undesirable because of the considerable potential for inconsistencies. The same code can now give different results depending on what files are present at the time of loading, and it is unclear how any timestamped data from different sources (e.g. spike sorters) could be aligned with the neo timestamps.

I would prefer an option, maybe even the default, to have the loader output the timestamps actually reported by the Neuralynx recording system.

JuliaSprenger:
This PR fixes #814 by introducing a new keyword argument for the NeuralynxIO.
For tests to succeed first https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/16 needs to be merged.

### DIFF
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -26,7 +26,26 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
     _prefered_signal_group_mode = 'group-by-same-units'
     mode = 'dir'
 
-    def __init__(self, dirname, use_cache=False, cache_path='same_as_resource'):
-        NeuralynxRawIO.__init__(self, dirname=dirname,
-                                use_cache=use_cache, cache_path=cache_path)
+    def __init__(self, dirname, use_cache=False, cache_path='same_as_resource',
+                 keep_original_times=False):
+        """
+        Initialise IO instance
+
+        Parameters
+        ----------
+        dirname : str
+            Directory containing data files
+        use_cache : bool, optional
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Waoooh
Super efficient.

JuliaSprenger:
I am surprised the new test to be not failing as the test file is still broken. I think the test allows too much variance in the data. I will investigate this.

JuliaSprenger:
I found and fixed more errors in the test files. They are fixed in https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/16.
Also, the test is now more strict and should succeed after the update of the test files on gin.

samuelgarcia:
Hi Julia.
This looks good.
Tell me when I can merge.

JuliaSprenger:
Ready from my side.

JuliaSprenger:
Ok, tests succeed now. Ready to merge from my side.

samuelgarcia:
bien joué!

Issue #456, opened 2018-01-10
Proposal: Split ChannelIndex into View, Group and Multiannotations


JuliaSprenger:
**Current State**
The current implementation of ChannelIndex is trying to provide a number of functionalities within the neo structure:
- providing information about individual traces of AnalogSignals like the channel id and the channel name (labelling)
- grouping a subset of traces within an Analogsignal via the index attribute (masking)
- linking between / grouping AnalogSignals (grouping)

_Problems_
However the implementation of all of these features in a single object lead to a number of inconsistencies and limitations (see also Issue #443). One of the limitation is, that if multiple AnalogSignals are attached to a ChannelIndex, they have to have the same number of traces. This does not allow to group AnalogSignals occurring within a single neo Segment. Another limitation is introduced by AnalogSignals only having a single parent ChannelIndex, which provides a general (global) channel ID and a general channel name. Additional / alternative channel numbering / naming is not possible this way.

**The Proposal**
The proposal is to split the different aspects of the ChannelIndex into different neo objects and features.

_Labelling_
A simple mechanism for providing channel ids and channel names is already proposed in Issue #390. Multiannotations (array-valued annotations) would allow each trace in the AnalogSignal to be labelled with a general channel ID as well as additional alternative ids using different keys of the multiannotaion dictionary. The actual keys to use are a matter of convention and up to discussion.

_Masking_
For selecting data (traces) of a data object we propose to implement a new neo object named 'View'. This object should have a single object of type AnalogSignal/IrregularSampledSignal as attribute and provide a mask for the data contained in the data object (eg. only selecting a subset of traces of the AnalogSignal/IrregularSampledSignal). It should also provide a method to actually return the corresponding slice of the linked object. For simplicity reasons the view object is linking to the data object, but not in the opposite direction. This simplifies the neo structure, but also introduces new limitations, which a more powerful implementation of the filter functionality should be able to overcome.

_Grouping_
Finally, the grouping aspect will be taken over by a new neo group object. This object should be able to link to multiple object of types AnalogSignal, IrregularSampledSignal, SpikeTrain and View. The group object can also take over the role of the neo Unit object, by only having SpikeTrains as children. Groups have a single neo Block as parent, but a neo Block can have multiple Groups attached.

Ideas / comments / feedback appreciated!

samuelgarcia:
[...]

PR #817, opened 2020-05-19
Replace ChannelIndex and Unit by View and Group

apdavison:
The goal of this is to complete the implementation of #456. Please take a look, and let me know if this fits your understanding of what was agreed, and if you see any difficulties with this approach.

Most of the necessary changes to `core` are done, I think, `io` has not been touched yet.

An important decision is whether we keep ChannelIndex and Unit for one major release, but deprecate them, or whether we remove them in the same release where we add View/Group. The advantages of keeping them are that (i) people can test the old and new approaches side-by-side, (ii) existing files in NIX format can be read, converted by the user, and rewritten with the new objects (otherwise NIXIO will have to be convert things automatically, which may not be easy). The disadvantages are the increased complexity and possible confusion of having two ways to do the same thing in Neo at the same time.


mdenker:
Thanks, Andrew, I will go over this PR over the next days. I don't have an ad-hoc opinion on whether to keep or drop the old classes yet.

samuelgarcia:
Hi Andrew.
Waooohh. You did it. You are our heroe.

I also try to have a look before the end of the week end.
I think it is important to discuss/validate this important PR during a short period.

After everyone in core team have read this, maybe we could plan a very short call to accelerate discussion, what do you think ?

samuelgarcia:
I will start to read your PR.

Some feeling right now (before reading) :
  * I can help to make change at IO level
  * I would vote for hard break (no ChannelIndex in next release) even if it is painfull
  * I don't think it is more diffulct to make NIX compatible with previous format than having still 
    ChannelIndex
  *  @achilleas-k should have a strong weight in this choice (slow/hard transition)
  * I can have the opposite opinion tomorow

achilleas-k:
I haven't had a good look at the PR but I skimmed through it yesterday.  I agree with Samuel that it might not be that difficult to read old NIX files into the new format.  I think an important part of the decision is whether the user or the IO should control the old-to-new conversion.
- If the ChannelIndex is kept for a transition period, it's up to the user to convert those ChannelIndex objects to one of the new ones when they read a file.
- If the ChannelIndex is removed, the IO would have to decide how to read those objects from old files.  I don't know how straightforward this would be.

I suppose in the latter case it could be accompanied by a warning informing the user that "N ChannelIndex objects have been converted to View".  Then it's up to the user to create a different kind of object if they need to (assuming View isn't the only replacement type).

That's just some initial thoughts I had on the decision.  I understand why you would want to avoid having the complexity of keeping around the old object types alongside the new ones, especially if it might lead to users mixing old with new classes while they work with the transition version.

doc/source/grouping.rst
```diff
@@ -0,0 +1,145 @@
...
+        SpikeTrain(np.arange(7, 100) * ms, t_stop=100 * ms)
+    ]
+    segment.analogsignals.append(signal)
+    segment.spiketrains.extend(spiketrains)
+    # assign each spiketrain to a neuron (now using Group)
+    units = []
+    for i, spiketrain in enumerate(spiketrains):
+        unit = Group(spiketrain, name=f"Neuron #{i + 1}")
+        units.append(unit)
+
```
samuelgarcia, line 133: I think this example would more clear with sevral segment.
a unit is more easy to understand across segment. No ?

doc/source/scripts/multi_tetrode_example.py
```diff
@@ -0,0 +1,103 @@
...
+    # create spike trains with dummy data
+    # we will pretend the spikes have been extracted from the dummy signal
+    for tetrode_id in ("Tetrode #1", "Tetrode #2"):
+        for i in range(n_units[tetrode_id]):
+            spiketrain = SpikeTrain(np.random.uniform(0, 100, size=30) * ms, t_stop=100 * ms)
+            # assign each spiketrain to the appropriate segment
+            segment.spiketrains.append(spiketrain)
+            spiketrain.segment = segment
+            # assign each spiketrain to a given neuron
+            current_group = next(iter_group)
```
samuelgarcia, line 59: I like this example.
Maybe the use of next.cycle is a bit confusion for a python beginner.
Maybe an old school and less elegant indexing would be more explicit.
No ?

neo/core/group.py
```diff
@@ -0,0 +1,58 @@
...
+        :class:`AnalogSignal`, :class:`IrregularlySampledSignal`, :class:`SpikeTrain`,
+        :class:`Event`, :class:`Epoch`, :class:`View`, :class:`Group
+    """
+    _data_child_objects = (
+        'AnalogSignal', 'IrregularlySampledSignal', 'SpikeTrain', 'Event', 'Epoch', 'View'
+    )
+    _container_child_objects = ('Segment', 'Group')
+    _single_parent_objects = ('Block',)
+
+    def __init__(self, *objects, name=None, description=None, file_origin=None, **annotations):
```
samuelgarcia, line 42: Maybe, I am old school but I would expect ```objects``` an iterable and not ```*objects``` here.

A standard use case will be
```python
a_list = []
for i in myloop:
    a_list.append(an_obbect)   

# This is more intuitive a biginer
g = Group(a_list)

# if I understand correctly you propose this
g = Group(*a_list)
```

neo/core/view.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # check type and dtype of index and convert index to a common form
+        # (accept list or array of bool or int, convert to int array)
+        self.index = np.array(index)
+        if len(self.index.shape) != 1:
+            raise ValueError("index must be a 1D array")
+        if self.index.dtype == np.bool: # convert boolean mask to integer index
+            if self.index.size != self.obj.shape[-1]:
+                raise ValueError("index size does not match number of channels in signal")
+            self.index = np.arange(self.obj.shape[-1])[self.index]
```
samuelgarcia, line 58: or `self.index, = np.nonzero(self.index)`

samuelgarcia:
Hi andrew.
I made some comments in the code directly.

My main feelings:
 
**Group**
super cool and totally fit the need. I totally validate.
The signature have to be discuss maybe.

**View**
I am more dibutative with the view object.
I am not sure it will be so usefull for long term usage.
I known we already discuss with Julia and Michael.
One object attached to the tree only to handle a subset of channel would make user code more heavy than dealing directly with the index vector in my opignon.
And about the semantic, in numpy a view is an array iself that is a lazy representation of a subarray, it is a bit different concept finally.

Let Julia, Michael, Achilleas have a look and we could maybe plan a short meeting, no ?

JuliaSprenger:
@apdavison Thanks for going ahead and starting this!

Regarding the deprecation of `ChannelIndex` objects I think it would be better to have a consistent solution across IOs. Currently this affects `KlustaKwikIO` `NixIO` `NeoMatlabIO` `NSDFIO`  and `PickleIO` as all of these can read and write Blocks. Also we should consider that neo structures are not only generated by the IOs, but might also be generated by custom user scripts that might currently rely on the `ChannelIndex` object. Updating such preprocessing to the new `Group` and `View` objects might imply a lot of tideous work for users. When abruptly removing `ChannelIndex` objects, the output of such preprocessing could only be automatically updated using two different neo versions, one for saving the preprocessing result using the old `ChannelIndex` mechansim and one for loading it with a later version of Neo. This is quite cumbersome and potentially resource consuming.

Therefore I think we should provide a function to convert any given neo structure based on `ChannelIndex`es to one based on `Group`s and `View`s. This way we have a well defined mapping between the old structures and the new ones. However, this would require the `ChannelIndex` object to be present in the next version(s) of neo.

Besides these thoughts regarding `ChannelIndex`es, I did not have the time yet to test the proposed implementation yet, but will do so soon!

neo/core/group.py
```diff
@@ -0,0 +1,58 @@
...
+    Can contain any of the data objects, views, or other groups,
+    outside the hierarchy of the segment and block containers.
+    A common use is to link the :class:`SpikeTrain` objects within a :class:`Block`,
+    possibly across multiple Segments, that were emitted by the same neuron.
+
+    *Required attributes/properties*:
+        None
+
+    *Recommended attributes/properties*:
+        :objects: (Neo object) Objects with which to pre-populate the :class:`Group`
```
dizcza, line 24: need to specify of which type: only AnalogSignal and SpikeTrain, right?

doc/source/grouping.rst
apdavison: This example is just a direct translation of the previous one, it's probably not the best example for the new features.

We should write new examples when we rewrite the documentation.

neo/core/view.py
apdavison: +1

neo/core/group.py
apdavison: I agree

doc/source/scripts/multi_tetrode_example.py
apdavison: yes, but I propose to leave it for now since we plan to rewrite the documentation (and examples) anyway.

neo/core/group.py
apdavison: no, a Group can contain any Neo object.

apdavison:
I've made the agreed changes.

What do you think about "ChannelView" instead of "View"?

samuelgarcia:
@apdavison :  I think that I am able to push some change into your branch. Is it OK for you ?

apdavison:
todo: make a list of which IOs need to be updated

apdavison:
todo: document migration, or add a function to automatically migrate existing code, from old to new structure

apdavison:
IO modules that need to be updated:

- [ ] basefromrawio.py
- [ ] blackrockio_v4.py
- [ ] brainwaredamio.py
- [ ] brainwaref32io.py
- [ ] brainwaresrcio.py
- [ ] hdf5io.py
- [ ] kwikio.py
- [ ] neuralynxio_v1.py
-  [ ] nixio.py
- [ ] nsdfio.py

samuelgarcia:
I will try  basefromrawio.py soon.

samuelgarcia:
@apdavison @JuliaSprenger @mdenker : I have made a proff of concept for integrating group into BaseFromRawIO.
Could you have a look ?
The change is small but important to be OK of the new use of theses groups.
Here Group is used only for "Group across segment".

samuelgarcia:
Any news here for porting IOs ?

I am trying to port the kwikio
We are in the case of multiple channel group of spike sorting.
In short
ChannelIndex was used for gouping Units family when there are on the same group AND also for the channel index.
What is the strategy we should do:
  * one group per spiketrain acrros segment ? only one segment here
  * use metagroup (=ChannelIndex) for grouping other group (=Units) ? Isn't it too complicated for end user ?
  * What about channel indexing. Here there is not link to analosignal at all. Do we put channel_index in annotation of Group ?

apdavison:
Hi @samuelgarcia - if I'm understanding the use case correctly, there should be one Group per unit, containing one spiketrain per segment + a ChannelView (assuming all the channels are in the same AnalogSignal).

samuelgarcia:
In knwkio there is not link between AnalogSignal and Spiketrain so it is not the ChannelView case the ChannelIndex used to be more for a group of group.

dizcza:
I've just tested this branch with the current Elephant master and it works without any changes - all tests passed.
Great job! :+1:

apdavison:
Thanks @dizcza, that's good to hear!

neo/io/nixio.py
```diff
@@ -676,6 +706,10 @@ def write_block(self, block, use_obj_names=False):
         for chx in block.channel_indexes:
```
samuelgarcia, line 706: Maybe we could put a warning here to tell that it won't be supported soon.

samuelgarcia:
@apdavison @achilleas-k 
Hi Andrew.
Cool that you have strated this Group in Nix.
A conceptual question : how we handle that fact that a data object is both attached to Segment and Group ?
Is the object is written 2 times ? Or is there an internal pointer ?

achilleas-k:
> A conceptual question : how we handle that fact that a data object is both attached to Segment and Group ?

In NIX a Segment is represented by a nix.Group.  Now a neo.Group is also represented by a nix.Group.

Groups in NIX hold only links to objects.  The canonical location for data objects in NIX is on the Block (e.g., `block.data_arrays`).  An object can be linked to multiple Groups since they hold only (hard) links to objects, so you can have a situation where, for example `blk.groups[0].data_arrays[0] == blk.groups[1].data_arrays[3]` is valid.

apdavison:
@achilleas-k if you have any more comments or suggestions about the way I've implemented support for Neo Groups in NixIO, I'd be very happy to hear them! 

@samuelgarcia conceptually, I tend to think of the Segment as the principal container for a signal, while the Group contains only a reference to it. As @achilleas-k says, they are both implemented as links in NIX, but in NixIO the signals are actually written to file while writing the Segment, and the code for writing Neo Groups assumes that the signal has already been written.

By the way, NixIO allows signals to be written to file that aren't in any Segment (they're only linked to a ChannelIndex). I propose to be stricter than this for Groups, and require that all data objects _must_ be within a Segment. What do you think?

achilleas-k:
> @achilleas-k if you have any more comments or suggestions about the way I've implemented support for Neo Groups in NixIO, I'd be very happy to hear them!

So far, the ChannelIndex was represented by a Source object.  I never liked the way it was represented because the `Source -> DataObject` relationships were reversed (DataObjects in NIX link to Sources).  With the new objects coming in to replace ChannelIndex (or break it into different types), I think it's a good idea to reevaluate the structure and I like your implementation of using Group instead.  We might want to plan this out and make the mapping more consistent, use Group to replace all these and the `type` field to differentiate between them in NIX.  It would also be good to have an upgrade path for older files so that as soon as someone loads an old file with the new NixIO, they get notified somehow that they should save the data (maybe to a new file) to get it up to date with the new mapping and structure.

This all got me thinking as well about the documentation we keep on the G-Node fork of Neo: https://github.com/G-Node/python-neo/wiki/Mapping
Do you think it would be a good idea to include it in the Neo docs?  Perhaps in the developer docs.  It would be easier to keep it up to date with changes to the IO if it's not in a separate repository/wiki.

JuliaSprenger:
@apdavison: Some suggestions from my side for an automatic conversion feature and small fixes are here: https://github.com/apdavison/python-neo/pull/3 and  https://github.com/apdavison/python-neo/pull/4

JuliaSprenger:
I suggest to start a list of practical todos here, so we keep an overview of what is missing before this PR can be merged:
- [ ] NixIO can not yet write nested Group objects.
- [ ] Unittests are still failing, for first fixes, see PRs linked above.

Feel free to extend this list.

samuelgarcia:
Hi all.
The 3 brainware IO are easy to fix for the ChannelIndex stuff.
I will do it now.

apdavison:
A question about "Represent units as groups (BasefromrawIO)":

In the master branch, in `BaseFromRaw`, SpikeTrains are put into Units, and these Units are attached to ChannelIndexes, but those ChannelIndexes do not seem to link to any AnalogSignals.

i.e. raw IOs do not support [use-case 2 for ChannelIndex](https://neo.readthedocs.io/en/stable/core.html#grouping-linking-objects)

> 2. for spike sorting of extracellular signals, where spikes may be recorded on more than one recording channel, and the ChannelIndex can be used to associate each Unit with the group of recording channels from which it was obtained.

Is that correct? Don't any of the file formats contain the information necessary for this association?

JuliaSprenger:
> Is that correct? Don't any of the file formats contain the information necessary for this association?

I would hope that this association is possible via the channel ids. Or are there systems that use different channel_ids for continuously sampled data and spiketrain data coming from the same electrode? 
If this is association is not possible in some cases, then I would suggest to still implement it for the other cases, as this is a key information needed for working with spiketrains and LFP data.

samuelgarcia:
@apdavison :
in some old school file format spiketrain belong to a channel.
The rawio at the moment to not handle this and so the ChannelIndex was used only for grouping Unit and not linking to channels. And it have never been handle by an IO for experimental format (with higer level level format like nix, yes)
In short the like ChannelIndex were never used in io even before rawio layer for those file format.

ChannelIndex was more an object to reconstruct this with spike sorting tools but quasi impossible to handle at file reader level (or too much work). In my opinion this relationship between a SpikeTrain and some channel is less and less usefull in the spike sorting work with dense probe.
A SpikeTrain have a "template" waveform on all (at least many) channel. Then if you put a threhold on the amplitude of this waveform then you have a sparse "template" and in that case you have indeed a relationship between a spiketrain and some channels but this relationship itself is not important. What is important is the template waveform itself (sparse or not) and the position channel wise of the extremum amplitude or even better the soma location on the probe.

In short this object can be usefull but it is not a raw input from the data.
It is some thing we infer from data with processing but do not come from raw experimental.

apdavison:
thanks @samuelgarcia and @JuliaSprenger 

so it seems that rawio has never had full support for Unit. In that case, I think the new basefromrawio that Samuel wrote is essentially feature-equivalent with the previous version. We can try to add more functionality, but perhaps after the release.

I've now fixed the conflicts with master, and the tests seem to be passing, so I think this is ready to merge. I suggest we merge this now, and then make any remaining changes that are needed before the release with PRs onto master.

JuliaSprenger:
Hi @samuelgarcia & @apdavison. Ok, then we postpone the unit-channel linking enhancement discussion to another release and I will merge this PR now. Thanks for the amount of work put into the new concept. I hope this is the last structural change we need for representing channel relations.

JuliaSprenger (approved):

samuelgarcia:
Champagne!!!

Issue #819, opened 2020-05-27
[NeuralynxIO] Improve features


TRuikes:
Hi! While working with NeuralynxIO in different datasets I came across a few bug/missing features, which I think will be useful (necessary) additions:

* [Input is automatically inverted](https://github.com/NeuralEnsemble/python-neo/blob/7409f47b5debd4d2a75bbf0e77ac10562446c97a/neo/rawio/neuralynxrawio.py#L109-L110), I was not aware this was happening, but is important to know.
* [Using regexp breaks](https://github.com/NeuralEnsemble/python-neo/blob/7409f47b5debd4d2a75bbf0e77ac10562446c97a/neo/rawio/neuralynxrawio.py#L685-L691), in two cases: 1) somehow the header string type does not match the application version (happens if I save old NL data using a new version of Neuraview) 2) if we have this in the header: `## Time Closed File was not closed properly`; we don't know how it got there, but the data itself seems intact
* [NCS files within one session can have different data lengths, causing this to break](https://github.com/NeuralEnsemble/python-neo/blob/7409f47b5debd4d2a75bbf0e77ac10562446c97a/neo/rawio/neuralynxrawio.py#L504). When `DspFilterDelay_µs > 0` and `DspDelayCompensation == Enabled`, there seems to be padding of the data by Cheetah (im not 100% confident about how). For a tetrode, someone would have 3 channels saved with `DspFilterDelay==0`, but one is being bandpassed (`DspFilterDelay >0`), yielding patches before and after the signal if the Dsp Delay compensation is on.
* If `DspDelayCompensation == Disabled` and `DspFilterDelay > 0`, this is not dealt with, and one would have channels with improper alignment
[...]

PR #820, opened 2020-05-28
(WIP) update neuralynx io

TRuikes:
Add info about inversion
Commented not-used lines in `read_txt_header`
In `read_txt_header` writes None to 'recording_opened' or 'recording_closed' if either breaks

TODO:
deal with ring buffer errors
deal with data of different lenghts (dsp)

follow up on #819 
Data which cannot be read by current IO (or explaining the issues) are [here](https://gin.g-node.org/TRuikes/ephy_testing_data/src/master/neuralynx/examples_breaking_data)

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -32,6 +32,7 @@ from collections import OrderedDict
 BLOCK_SIZE = 512  # nb sample per signal block
 HEADER_SIZE = 2 ** 14  # file have a txt header of 16kB
 
+
 class NeuralynxRawIO(BaseRawIO):
     """"
     Class for reading dataset recorded by Neuralynx.
@@ -66,7 +67,6 @@ class NeuralynxRawIO(BaseRawIO):
         event_channels = []
 
         self.ncs_filenames = OrderedDict()  # (chan_name, chan_id): filename
-        self._ncs_info = OrderedDict()  # (chan_name, chan_id): info
         self.nse_ntt_filenames = OrderedDict()  # (chan_name, chan_id): filename
         self.nev_filenames = OrderedDict()  # chan_id: filename
 
@@ -115,7 +115,30 @@ class NeuralynxRawIO(BaseRawIO):
                     sig_channels.append((chan_name, chan_id, info['sampling_rate'],
                                          'int16', units, gain, offset, group_id))
                     self.ncs_filenames[chan_uid] = filename
-                    signal_annotations.append(info)  # is there a reason to not append all information
+                    keys = [
+                        'DspFilterDelay_µs',
+                        'recording_opened',
+                        'FileType',
+                        'DspDelayCompensation',
+                        'recording_closed',
+                        'DspLowCutFilterType',
+                        'HardwareSubSystemName',
+                        'DspLowCutNumTaps',
+                        'DSPLowCutFilterEnabled',
+                        'HardwareSubSystemType',
+                        'DspHighCutNumTaps',
+                        'ADMaxValue',
+                        'DspLowCutFrequency',
+                        'DSPHighCutFilterEnabled',
+                        'RecordSize',
+                        'InputRange',
+                        'DspHighCutFrequency',
+                        'input_inverted',
+                        'NumADChannels',
+                        'DspHighCutFilterType',
+                    ]
+                    d = {k: info[k] for k in keys if k in info}
+                    signal_annotations.append(d)
 
                 elif ext in ('nse', 'ntt'):
                     # nse and ntt are pretty similar except for the wavform shape
@@ -181,7 +204,7 @@ class NeuralynxRawIO(BaseRawIO):
             self._sigs_sampling_rate = sampling_rate[0]
 
         # read ncs files for gaps detection and nb_segment computation
-        self.read_ncs_files(ncs_filenames=self.ncs_filenames)  # no need to pass class attributes to function
+        self.read_ncs_files(self.ncs_filenames)
 
         # timestamp limit in nev, nse
         # so need to scan all spike and event to
@@ -920,8 +943,7 @@ def read_txt_header(filename):
 
 ncs_dtype = [('timestamp', 'uint64'), ('channel_id', 'uint32'), ('sample_rate', 'uint32'),
              ('nb_valid', 'uint32'), ('samples', 'int16', (BLOCK_SIZE,))]
-ncs_dtype2 = [('timestamp', 'uint64', (1,1)), ('channel_id', 'uint32', (1,1)), ('sample_rate', 'uint32', (1,1)),
-             ('nb_valid', 'uint32', (1,1)), ('samples', 'int16', (BLOCK_SIZE, 512))]
+
 nev_dtype = [
     ('reserved', '<i2'),
     ('system_id', '<i2'),
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -603,8 +603,8 @@ class NeuralynxRawIO(BaseRawIO):
             # gap indices are detected by dt > good_delta (thus pointing to rows)
             gap_indices = np.where(np.diff(data['timestamp']) > good_delta)[0]  # row (block) indices
             for g in gap_indices:
-                assert (data['timestamp'][g+1] - data['timestamp'][g]) <= max_gap_duration_to_fill * 1e6, 'found a big gap, what should we do???'
-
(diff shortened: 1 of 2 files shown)


TRuikes:
I've moved all the patching functionality for dealing with misalginment and gaps to a separate function, which we can call 1 time per dataset to generate a patched dataset. This way, if you don't think it should be in rawio, we can move it elsewhere.

TRuikes:
I'm finished with the draft, for these two files it seems to work, I'm going to check on the other data  we have in the future.

At this point the whole patching story is a separate function. I left it as a static method for convenience.

neo/rawio/neuralynxrawio.py
```diff
@@ -216,7 +218,7 @@ def _parse_header(self):
                     ts0 = ts[0]
                     ts1 = ts[-1]
                 ts0 = min(ts0, ts[0])
-                ts1 = max(ts0, ts[-1])
+                ts1 = max(ts1, ts[-1])  # TODO: shouldn't this be ts1
```
samuelgarcia, line 221: Yes I think so.

neo/rawio/neuralynxrawio.py
```diff
@@ -672,23 +910,33 @@ def read_txt_header(filename):
     if old_date_format:
         datetime1_regex = r'## Time Opened \(m/d/y\): (?P<date>\S+)  \(h:m:s\.ms\) (?P<time>\S+)'
         datetime2_regex = r'## Time Closed \(m/d/y\): (?P<date>\S+)  \(h:m:s\.ms\) (?P<time>\S+)'
-        filename_regex = r'## File Name (?P<filename>\S+)'
+        # filename_regex = r'## File Name (?P<filename>\S+)'  # not used
```
samuelgarcia, line 913: ok

neo/rawio/neuralynxrawio.py
```diff
@@ -672,23 +910,33 @@ def read_txt_header(filename):
...
         datetimeformat = '%m/%d/%Y %H:%M:%S.%f'
     else:
         datetime1_regex = r'-TimeCreated (?P<date>\S+) (?P<time>\S+)'
         datetime2_regex = r'-TimeClosed (?P<date>\S+) (?P<time>\S+)'
-        filename_regex = r'-OriginalFileName "?(?P<filename>\S+)"?'
+        # filename_regex = r'-OriginalFileName "?(?P<filename>\S+)"?'  # not used
```
samuelgarcia, line 918: ok

neo/rawio/neuralynxrawio.py
```diff
@@ -672,23 +910,33 @@ def read_txt_header(filename):
...
 
-    dt1 = re.search(datetime1_regex, txt_header).groupdict()
-    dt2 = re.search(datetime2_regex, txt_header).groupdict()
+    # Using this at least the code won't break if the regexp fails
+    if dt1 is not None:
+        dt1 = dt1.groupdict()
+        info['recording_opened'] = datetime.datetime.strptime(
+            dt1['date'] + ' ' + dt1['time'], datetimeformat)
+    else:
+        info['recording_opened'] = None
```
samuelgarcia, line 932: good idea

samuelgarcia:
@TRRuikes : if I understand your PR you want be able to rewrite some some neuralynx files that contains gaps into mono-segment (without gap) new neuralynx folder.

I think that this would be very usefull in a generic (io inpedant) function.

For instance taking any rawio instance that have multi-segment signals and writte then into a raw binary IO files.

Having this functionality inside a rawio is a bit dangerous for long term maintenance because it is too specific.


@JuliaSprenger : what do you think ?

neo/rawio/neuralynxrawio.py
```diff
@@ -411,6 +413,242 @@ def _rescale_event_timestamp(self, event_timestamps, dtype):
...
+            patched_data = np.core.records.fromarrays(
+                [ts_interpolated + t0_all, chan_id, sample_rate, nb_valid, records], dtype=np.dtype(ncs_dtype)
+            )
+
+            header = read_ncs_header(ncs_in_name)
+            with open(ncs_out_name, 'wb') as f:
+                f.write(header)
+                f.write(patched_data.tobytes())
+
+        print('Done!')
```
JuliaSprenger, line 650: Please remove the print statement. This will clutter the output when running a full analyis and not only the patching.

JuliaSprenger:
@samuelgarcia I agree, the patching functionality should rather go into some sort of utility function, but in a more generic fashion, such that it can be used across IOs.
@TRuikes: Also, if you have files exhibiting these irregularities, would you be willing to contribute one to the set of [neo testfiles](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/) so we can cover this in the unittests?

neo/rawio/neuralynxrawio.py
JuliaSprenger: This is an issue that could be addressed in a more general fashion in #851. @PeterNSteinmetz, what do you think?

PeterNSteinmetz: Yes, @JuliaSprenger, I think this is dealt with in #851, with a different pattern in the NlxHeader parsing code.

TRuikes: @PeterNSteinmetz nice update on the io, I think we had a few recording sessions where the recording software was closed too early. In this case there is no closing time in the header (but the keyword is there [example file](https://gin.g-node.org/TRuikes/ephy_testing_data/src/master/neuralynx/examples_breaking_data/dsp_compensation_and_ringbuff_errors))

I fixed it with [these lines](https://github.com/TRuikes/python-neo/blob/3a9d537bb38510bf8863d75b3c08f5fcc29a549f/neo/rawio/neuralynxrawio/nlxheader.py#L217-L224), for our data.

TRuikes:
hi @JuliaSprenger @samuelgarcia  @PeterNSteinmetz sorry for the no reply.

I think I'll close this PR since it's on an old version of neo now, and you've implemented some nice features. I had uploaded the breaking data on gin, but something went wrong with a PR, [so I tried again just now](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/33). It should contain a few files which were/are not readable by the io (the reason specified in the description)

There is still a header parsing bug for me, I commented the solution for my data above.

Your concatenate function works nice, the padding keyword makes it even better. The only problem now is that 'ringbuffer errors' (ie losing datapoints during recording) occurs per channel. So in the case of my example data, for 1 ncs file it might detect 8 segments, and for another it might detect 10 segments, which the current [neo does not allow to read](https://github.com/NeuralEnsemble/python-neo/blob/7695048371b15b57beaf546b9982f19856c658ea/neo/rawio/neuralynxrawio/neuralynxrawio.py#L519-L521) (this is a good thing, in the sense that the user is aware something is up). Since I won't be able to read the data, I also cannot use the concatenate function for individual signals to fix them. One solution would be to allow the Neuralynx Io to read a single file, and extend the Io with a 'writer'. This way I can loop over all individual channels, use concatenate and write to store a fixed version.

Finally [this section](https://github.com/TRuikes/python-neo/blob/3a9d537bb38510bf8863d75b3c08f5fcc29a549f/neo/rawio/neuralynxrawio/neuralynxrawio.py#L505-L513) might be a neat addition. It will notify the user if during recordings a lag in one of the channels arises due to digital filtering, and cheetah not correcting for it (a problem we had, an example of this in the data uploaded on gin).

PeterNSteinmetz:
@TRuikes I think the longer term solution for this lies in handling the multiple sections (which is a name I introduced in the source code changes for contiguous groups of blocks of records) in some standardized way. This could be either creating multiple analog signals for a given channel which has multiple sections, or splitting up the Neo segments. I lean toward the former solution presently as the Neo definition is simply a set of data sharing the same common clock, which is the case for these different sections. Analog signals could be named with their normal name and a suffix, like _sect0, _sect1, etc. How would that strike you for your use case?

neo/rawio/neuralynxrawio.py
PeterNSteinmetz: @TRuikes I agree that is the right way to handle it and it is reasonable to allow the closing time to be optional generally. Would you create a pull request from that change?

TRuikes:
It sounds good, I think the most important thing is that all the data is able to load and we are warned if something suspiscious happens.

JuliaSprenger:
Hi @TRuikes I think the main aim of this PR is already covered in the master branch now via #836, correct? I suggest we continue the discussion in #819 and close this PR. What do you think?

PR #823, opened 2020-06-11
AxographRawIO: Handle recoverable errors, support files written by axographio package

jpgill86:
``AxographRawIO`` should now be able to recover from unexpected end-of-file errors and text decoding errors if they occur when parsing sections of the header containing non-essential metadata. Warnings are logged in these cases, but the process should not fatally stall.

Previously, support for some AxoGraph files written by the independent [axographio](https://pypi.org/project/axographio/) Python package was missing. This has now been added. Namely, files written in the latest format version with a time vector (rather than a ``t_start, sampling_period`` pair using ``axographio.linearsequence``; these were already supported) can now be read.

Additionally, the reader will now check that time vectors are uniformly sampled, rather than just assuming they are.

Tests were added for all new functionality.

### DIFF
--- a/neo/rawio/axographrawio.py
+++ b/neo/rawio/axographrawio.py
@@ -686,49 +686,6 @@ class AxographRawIO(BaseRawIO):
                     dtype = 'f'
                     gain, offset = 1, 0  # data is neither scaled nor off-set
 
-                    if i == 0:
-
-                        # there is no guarantee that this time column is
-                        # regularly sampled, and in fact the test file for
-                        # version 1 has slight variations in the intervals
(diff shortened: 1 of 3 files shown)


jpgill86:
This PR is ready to be merged!

samuelgarcia:
Hi Jeffrey.
This looks good.
As I don't known this format and you have put test files. I guess everything is OK.
I merge.

jpgill86:
Thanks @samuelgarcia!

jpgill86:
Hi @samuelgarcia, could you add this PR to the 0.9.0 milestone so that it shows up in the list of pull requests that were part of the new release?

PR #824, opened 2020-06-13
fix for v9 SON files

MarinManuel:
In the CED documentation for the SON library, it says:

> In version 9 and above files, if pos is of type TDOF, the disk offset is pos*DISKBLOCK. In version 8 and previous files, the disk offset is pos. 

this PR fixes the offsets for v9 files, and fixes #818

PR #826, opened 2020-06-16
ChannelIndex.channel_names is now 'U' instead of 'S'

samuelgarcia:


samuelgarcia:
This need to merged before  #824

samuelgarcia:
@MarinManuel @JuliaSprenger : I prefer this fix for channel names.
The 'U' is now official.

JuliaSprenger:
I think there are a still a number of places in the code, where 'S' is used instead of 'U', e.g. https://github.com/NeuralEnsemble/python-neo/blob/fce4227410249660f62ce0f3c3b596fc2cde4f68/neo/io/brainwaredamio.py#L141
These are not affecting the functionality, since the type is anyway converted later, but maybe it would make sense to systematically replace 'S' by 'U' also in all non-relevant places?

samuelgarcia:
I think I already fix the brainwaredamio no ?
Test are passing so I think there is no channelindex.channel_names with 'S' anymore.
Each object read by IOs are tested for there types and dtypes normally.
What do you think of this solution ?

JuliaSprenger:
Yes, you are right, I missed that one. Looks good to me!
Thanks for the fix.

Issue #832, opened 2020-07-07
Force group channel in IOs by default for all IOS


samuelgarcia:
```python
    _prefered_signal_group_mode = 'split-all'
    _prefered_units_group_mode = 'split-all'
```
must now be
```python
    _prefered_signal_group_mode = 'group-by-same-units'
    _prefered_units_group_mode = 'all-in-one'
```

PR #837, opened 2020-07-23
Default group mode to 'group-by-same-units'

samuelgarcia:
'split-all' > 'group-by-same-units'

#715
#832 
#710

### DIFF
--- a/neo/io/axographio.py
+++ b/neo/io/axographio.py
@@ -48,7 +48,7 @@ class AxographIO(AxographRawIO, BaseFromRaw):
     name = 'AxographIO'
     description = 'This IO reads .axgd/.axgx files created with AxoGraph'
 
-    _prefered_signal_group_mode = 'split-all'
+    _prefered_signal_group_mode = 'group-by-same-units'
 
     def __init__(self, filename='', force_single_segment=False):
         AxographRawIO.__init__(self, filename, force_single_segment)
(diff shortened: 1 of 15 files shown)


samuelgarcia:
@jpgill86 
Hi Jeffrey.
I have changiing the default behavior of grouping in basefromrawio.py for signal group mode.
This has broken all your tests in axongraph so I put `blk = reader.read_block(signal_group_mode='split-all')` evrywhere.
Maybe you may want to refactor you axograph tests.
Tell me.

jpgill86:
Hi @samuelgarcia, I'll take a look when I can, soon I hope!

samuelgarcia:
I ppropose to merge this PR soon.
And if you want change tests you will be able to do it.
Absically I force the old befault behavior (split) in your tests.

jpgill86:
That makes sense. No need to wait for me to merge!

apdavison:
todo: add warning when default is used, otherwise a lot of code could break without the user understanding why.

We will keep this warning for one release - drop for 0.10.0

samuelgarcia:
In the code wa already add this:
```
            if self._prefered_signal_group_mode == 'split-all':
                self.logger.warning("the default signal_group_mode will change from "\
                                "'split-all' to 'group-by-same-units' in next release")
```
since 1147637256cb5e473845e895f0d4d67b6ece6938 in July 2019.

So the warning cycle already accur in last release. I think there is no need to do it in the one.
@apdavison @JuliaSprenger @mdenker :  what do you think ?

apdavison:
Decision on meeting on 07/09/2020:
  - change the default for 0.9.0
  - for affected IOs emit a warning, to be removed in 0.10.0

"affected IOs" means IOs where the default has changed and where the flag is None.

samuelgarcia:
@apdavison @JuliaSprenger : is this warning OK ?

Issue #828, opened 2020-07-03
Add validator to ensure neo structure is rawio compatible


JuliaSprenger:
This idea was developed in #626.

The validator can be used to ensure a saved neo structure can be loaded with a rawio.
It should be part of `neo.utils`. Besides providing a boolean feedback, it might also indicate to the user the structural parts that are not consistent with the rawio requirements.

JuliaSprenger:
Aspects that should be covered are
- consistent channels across segments
- array annotations can only be basic (default) annotations like channel_id, that are recognized on rawio level as array annotation type
- consistent linking between objects (bidirectional where possible)
- any more ideas?

apdavison:
Implemented in #839

PR #839, opened 2020-07-23
Add is_block_rawio_compatible() in neo.utils

samuelgarcia:
This implement proof of concept for #828.

@JuliaSprenger @achilleas-k : please make coments when you have time.

### DIFF
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -572,7 +572,7 @@ def is_block_rawio_compatible(block, return_problems=False):
     block: Block
         A blck
     return_problems: bool (False by default)
-        Control if a list a str that desribe problems is also provided as return value
+        Control if a list a str that describe problems is also provided as return value
 
     Returns:
     --------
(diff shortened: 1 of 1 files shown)


samuelgarcia:
@JuliaSprenger : I don't understand what you mean about : "consistent linking between objects".

In rawio there is no link between objects except of course for block>segment>data_object
And also automatically generated ChannelIndex (and soon Group) to group data object across segment.

achilleas-k:
Thanks.  This will be very useful for working with the RawIO and keeping things consistent.  I think once this is finalised and merged, I could simplify some things in the NIXRawIO by running a Block through this checker and then reading everything under the assumption that everything is Raw compatible.

The docstring mentions array annotations but they aren't checked.  What does "only basic" mean exactly?  No compound objects, only basic types?

samuelgarcia:
For array annotations, the best would to add annotation at rawio level.
It is on my todo list.
Now only 'name' and 'id' in self.header['signal_channels'] are converted to array annotation.
This means that except this two ones all array_annotaion will be lost in write/read cycle.

I can one some lines to check this in this `is_block_rawio_compatible()` and change it later when arrayèannoation will be better handled in rawio level.

samuelgarcia:
@achilleas-k : Finally I wont have time to improve array_annotations checking here.
I prefer to merge this soon and then we will be able to enhence this function.
Strictly speaking arrau_annations is not a blocker for rawio. They are just ignored. Asymetry in channel acrosssegments is a real blocker for instanc (and tested here).

samuelgarcia:
I think this is ready to merge for now.

This will improved later when:
  * array_annotations will have a better support at rawio level
  * when nixrawio will use it for real.

achilleas-k:
Sounds good!

neo/utils.py
```diff
@@ -555,3 +555,75 @@ def cut_segment_by_epoch(seg, epoch, reset_time=False):
...
 
     return segments
+
+
+def is_block_rawio_compatible(block, return_problems=False):
+    """
+    The neo.rawio layer have some restriction compared to neo.io layer:
+      * consistent channels across segments
+      * no IrregularlySampledSignal
+      * consistent sampling rate accross segments
```
JuliaSprenger, line 565: should read 'across`

neo/utils.py
```diff
@@ -555,3 +555,75 @@ def cut_segment_by_epoch(seg, epoch, reset_time=False):
...
+
+    This function test if a neo.Block that could be written in a nix file could be read
```
JuliaSprenger, line 567: should read `tests`

neo/utils.py
```diff
@@ -555,3 +555,75 @@ def cut_segment_by_epoch(seg, epoch, reset_time=False):
...
+    back with the NIXRawIO.
+
+    Parameters
+    ----------
+    block: Block
+        A blck
```
JuliaSprenger, line 573: `block`

Issue #816, opened 2020-05-13
BlackRockIO lazy=True does not load waveforms


dizcza:
Hello neo devs!
When I use R2G data https://gin.g-node.org/INT/multielectrode_grasp, I can load waveforms only when lazy flag is off. Steps to reproduce:

```python
from neo.io import BlackRockIO

RAW_DATA_DIR = Path('datasets')
session_name = str(RAW_DATA_DIR / "i140703-001")
nev_override = str(RAW_DATA_DIR / "i140703-001-03")
session = BlackrockIO(filename=session_name, nsx_to_load=[],
                      nev_override=nev_override,
                      verbose=False)

segment = session.read_segment(lazy=False, load_waveforms=True)
for spiketrain in segment.spiketrains:
    spiketrain = spiketrain.time_slice(t_start=10 * pq.s, t_stop=30 * pq.s)
    assert spiketrain.waveforms is not None  # works fine

segment_lazy = session.read_segment(lazy=True, load_waveforms=True)
for spiketrain in segment_lazy:
    spiketrain = spiketrain.load(time_slice=(10 * pq.s, 30 * pq.s))
    assert spiketrain.waveforms is not None  # fails
```

I have another question: I pass `nsx_to_load=[]` not to load the continuous data. This behaviour - empty list - is not documented, but it seems to work fine.

neo v0.8.0
python 3.8.2

samuelgarcia:
When ```lazy=True``` the load_waveforms must be set directly in :

```spiketrain.load(time_slice=(10 * pq.s, 30 * pq.s), load_waveforms=True)```

samuelgarcia:
See https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/basefromrawio.py#L271

dizcza:
I see, thanks.
I expected the waveforms to be loaded because I pass `session.read_segment(lazy=True, load_waveforms=True)`. I didn't know that the `load_waveforms` flag is ignored while reading the segments.
Does it make sense to add a warning then? Or, if passed, save it and then automatically pass to `spiketrain.load()`?

samuelgarcia:
Yes, i known : it is not intuitive.

kargs in read_segment/read_block are only when lazy=False (because pass trhourg directly to lazy object).

Theses kargs are ignored when lazy=True.

I think saving kargs for future load is complicated and even less intuitive.

A warning for ```lazy=True and load_waveforms=True``` is a good solution.

[...]

PR #840, opened 2020-07-23
Handling load_waveforms params when lazy load.

samuelgarcia:
See #816


samuelgarcia:
I think this is ready to merge.

Issue #755, opened 2019-10-30
cut_block_by_epochs should return a new block and not do inplace segment replacement


samuelgarcia:
cut_block_by_epochs should return a new block and not do inplace segment replacement

samuelgarcia:
For instance cut_segment_by_epoch return new segments.

PR #841, opened 2020-07-23
cut_block_by_epochs give a new Block now.

samuelgarcia:
Now cut_block_by_epochs give a new block and don't do a inplace replacement of segments.
Similarly to cut_segment_by_epoch.

See #755 

@mdenker @JuliaSprenger : could you check this ?
I think you are originals authors no ?

### DIFF
--- a/neo/utils.py
+++ b/neo/utils.py
@@ -473,8 +473,9 @@ def cut_block_by_epochs(block, properties=None, reset_time=False):
         raise TypeError(
             'block needs to be a Block, not %s' % type(block))
 
-    old_segments = copy.copy(block.segments)
-    for seg in old_segments:
+    new_block = neo.Block()
+
+    for seg in block.segments:
         epochs = _get_from_list(seg.epochs, prop=properties)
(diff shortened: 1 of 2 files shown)


neo/utils.py
```diff
@@ -490,10 +491,11 @@ def cut_block_by_epochs(block, properties=None, reset_time=False):
         for epoch in epochs:
             new_segments = cut_segment_by_epoch(
                 seg, epoch=epoch, reset_time=reset_time)
-            block.segments += new_segments
```
JuliaSprenger, line 493: is there a reason to use `+=` rather than `extend` here?

samuelgarcia: I am anticipating the day wher we will have pseudo list for relationship.
I guess append and extend will be part of methods not sure for +=

JuliaSprenger: then it would be better to stick to `extend`, no?

samuelgarcia:
Ready to merge.
Should close #755

PR #842, opened 2020-07-27
Accept `None` as valid time slice argument for segments

JuliaSprenger:
`time_slice` methods of data objects accept None as `t_start` or `t_stop` when the corresponding samples should not be cut, e.g. [AnalogSignals](https://github.com/NeuralEnsemble/python-neo/blob/c9e5829f3209f65bdd0a3100dbb7666f1ce41ec8/neo/core/analogsignal.py#L469). For segments this yields an error. With this PR `t_start` and `t_stop` for segments are replaced by the respective border time of the segment.

### DIFF
--- a/neo/core/segment.py
+++ b/neo/core/segment.py
@@ -285,6 +285,11 @@ class Segment(Container):
 
         subseg.annotations = deepcopy(self.annotations)
 
+        if t_start is None:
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi Julia. ready to merge ?

Issue #830, opened 2020-07-03
[Feature] Function to merge analogsignals


morales-gregorio:
Hey all,

I wrote the following function to merge analogsignals, because I needed it. The function is thought to be memory efficient, and thus systematically deletes the data that has been duplicated. 

If you think this function can find a home somewhere in the neo codebase, let me know where and I can create a corresponding test and PR it.

```
def merge_anasiglist(anasiglist):
    """
    Merges neo.AnalogSignal objects into a single object.

    Units, sampling_rate, t_start, t_stop and signals shape must be the same
    for all signals. Otherwise a ValueError is raised.

    Parameters
    ----------
    anasiglist: list of neo.AnalogSignal
        list of analogsignals that will be merged

    Returns
    -------
    merged_anasig: neo.AnalogSignal
        merged output signal
    """
    # Check units, sampling_rate, t_start, t_stop and signal shape
    for anasig in anasiglist:
        if not anasiglist[0].units == anasig.units:
            raise ValueError('Units must be the same for all signals')
        if not anasiglist[0].sampling_rate == anasig.sampling_rate:
            raise ValueError('Sampling rate must be the same for all signals')
        if not anasiglist[0].t_start == anasig.t_start:
            raise ValueError('t_start must be the same for all signals')
        if not anasiglist[0].t_stop == anasig.t_stop:
            raise ValueError('t_stop must be the same for all signals')
        if not anasiglist[0].magnitude.shape == anasig.magnitude.shape:
            raise ValueError('All signals must have the same shape')

    # Initialize the arrays
    anasig0 = anasiglist.pop(0)
    data_array = anasig0.magnitude
    sr = anasig0.sampling_rate
    t_start = anasig0.t_start
    t_stop = anasig0.t_stop
    units = anasig0.units

    # Get the full array annotations
    for anasig in anasiglist:
        anasig0.array_annotations = anasig0._merge_array_annotations(anasig)

    array_annot = anasig0.array_annotations
    del anasig0

    while len(anasiglist) != 0:
        anasig = anasiglist.pop(0)
        data_array = np.concatenate((data_array, anasig.magnitude),
                                    axis=np.argmin(anasig.magnitude.shape))
        del anasig

    merged_anasig = neo.AnalogSignal(data_array,
                                     sampling_rate=sr,
                                     t_start=t_start,
                                     t_stop=t_stop,
                                     units=units,
                                     array_annotations=array_annot)
    return merged_anasig
```

Cheers!
[...]

PR #846, opened 2020-08-26
include analog signal merging utility function and the corresponding …

morales-gregorio:
Hey!

I finally managed to create the PR for issue #830

Please have a look if you think something is wrong or should be improved.

Best,
Aitor

### DIFF
--- a/neo/test/test_utils.py
+++ b/neo/test/test_utils.py
@@ -471,7 +471,7 @@ class TestUtilsWithoutProxyObjects(unittest.TestCase):
     def test_merge_anasiglist(self):
         baselist = [AnalogSignal(np.arange(55.0).reshape((11, 5)),
                                  units="mV",
-                                 sampling_rate=1*pq.kHz)]*2
+                                 sampling_rate=1 * pq.kHz)] * 2
 
         # Sanity of inputs
         self.assertRaises(TypeError, merge_anasiglist, baselist[0])
@@ -483,7 +483,7 @@ class TestUtilsWithoutProxyObjects(unittest.TestCase):
         # Different units
         wrongunits = AnalogSignal(np.arange(55.0).reshape((11, 5)),
                                   units="uV",
-                                  sampling_rate=1*pq.kHz)
+                                  sampling_rate=1 * pq.kHz)
         analist = baselist + [wrongunits]
         self.assertRaises(ValueError, merge_anasiglist, analist)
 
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
I think there's some common features to PR #836 and this one. I will have a look, maybe we can unify the functionality somehow.

neo/test/test_utils.py
```diff
@@ -20,7 +20,10 @@
                             assert_same_attributes,
                             assert_same_annotations)
 
-from neo.utils import (get_events, get_epochs, add_epoch, match_events, cut_block_by_epochs)
+from neo.utils import (get_events, get_epochs, add_epoch, match_events,
```
JuliaSprenger, line 23: I don't think this change is needed since Neo allows for up to 99 characters per line (https://github.com/NeuralEnsemble/python-neo/blob/57f8d079deef7533394b4c1e0db103bf963d203c/.pep8speaks.yml#L2) Maybe you are using the default PEP8 settings for autoformatting here?

neo/utils.py
```diff
@@ -10,6 +10,86 @@
...
+    Units, sampling_rate, t_start, t_stop and signals shape must be the same
+    for all signals. Otherwise a ValueError is raised.
+
+    Parameters
+    ----------
+    anasiglist: list of neo.AnalogSignal
+        list of analogsignals that will be merged
+    axis: int
+        axis along which to perform the merging
+        `axis = 1` corresponds to stacking the analogsignals
```
JuliaSprenger, line 26: By implementing the combination of signals along two different axis you are covering two different scenarios, that require different checks on the neo level. Eg. when stacking signals (ie merging multiple channels in one Analogsignal) it is useful to ensure they have the same length and cover the same time range. However for concatenating analogsignals along time, these conditions do not need to be fulfilled. I would argue therefore these different cases should also be implemented in separate functions. This would also make it easier for the users to decide which of the actions they actually want to use.

JuliaSprenger: Another feature that needs to be treated differently depending on the axis are the array_annotations. I am not sure in how far this implementation is taking care of this, but I think we should double check here.

neo/utils.py
```diff
@@ -10,6 +10,86 @@
...
+    data_array = anasig0.magnitude
+    sr = anasig0.sampling_rate
+    t_start = anasig0.t_start
+    units = anasig0.units
+
+    # Get the full array annotations
+    for anasig in anasiglist:
+        anasig0.array_annotations = anasig0._merge_array_annotations(anasig)
+
+    array_annot = anasig0.array_annotations
```
rgutzen, line 75: It seems that there is a distinction missing whether the analogsignals are concatenated or stacked, which would require different handling of the array annotations.
I guess the protocol should be something like:

- _concatenating_: keeping only array_annotations and annotations that are present and equal in all analogsignals in the list, while omitting (?) the rest.
- _stacking_: stacking the array_annotatations; keeping only the annotations that are present and equal in all analogsignals; annotations that are different for the analogsignal should then be used as array_annotations instead.

or what do you think is a good merging strategy?

JuliaSprenger:
Hi @morales-gregorio,
I improved the performance of concatenating Analogsignals in the proposed `concatenate` function at #836. This at least covers the aspect of concatenation of AnalogSignals across time, that is also implemented in this PR. Could you check out the performance of #836 and see if this suits your needs?

apdavison:
The performance issues that motivated this PR should have been addressed in #836. Please re-open, or create a new issue, if the performance issues remain.

PR #851, opened 2020-09-03
Fixes to make tests pass with Pegasus event files only directory.

PeterNSteinmetz:
This set of changes fixes a bug which prevented the Pegasus .Nev event only files from passing the test suite. 

It also allows filename extensions to be in any letter case. For the InputInverted header parameter to be missing, 
meaning not inverted. Checks for 8 hashes at the start of the header as a check on header contents. Allows
any whitespace between a header key and value. 

Internal changes are addition of a large number of explanatory comments and refactoring of the code to handle
the header into a separate class. Many of these in preparation for handling earlier versions of neuralynx file headers
and more flexible handling of segments withins .Ncs files.

### DIFF
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -214,10 +214,16 @@ class NeuralynxRawIO(BaseRawIO):
         unit_channels = np.array(unit_channels, dtype=_unit_channel_dtype)
         event_channels = np.array(event_channels, dtype=_event_channel_dtype)
 
+        # require all sampled signals, ncs files, to have same sampling rate
         if sig_channels.size > 0:
             sampling_rate = np.unique(sig_channels['sampling_rate'])
             assert sampling_rate.size == 1
             self._sigs_sampling_rate = sampling_rate[0]
+
+        # set 2 attributes needed later for header in case there are no ncs files in dataset,
+        #   e.g. Pegasus
+        self._timestamp_limits = None
+        self._nb_segment = 1
 
         # Read ncs files for gaps detection and nb_segment computation.
         # :TODO: current algorithm depends on side-effect of read_ncs_files on
@@ -238,6 +244,7 @@ class NeuralynxRawIO(BaseRawIO):
                 ts0 = min(ts0, ts[0])
                 ts1 = max(ts0, ts[-1]) # :FIXME: possible error, should be ts1
 
+        # decide on segment and global start and stop times based on files available
         if self._timestamp_limits is None:
             # case  NO ncs but HAVE nev or nse
             self._timestamp_limits = [(ts0, ts1)]
--- a/neo/rawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio.py
@@ -244,7 +244,7 @@ class NeuralynxRawIO(BaseRawIO):
                     ts0 = ts[0]
                     ts1 = ts[-1]
                 ts0 = min(ts0, ts[0])
-                ts1 = max(ts0, ts[-1])  # :FIXME: possible error, should be ts1
+                ts1 = max(ts1, ts[-1])
 
         # decide on segment and global start and stop times based on files available
         if self._timestamp_limits is None:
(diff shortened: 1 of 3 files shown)


apdavison:
Many thanks for this PR. Please also add your name and affiliation to `doc/source/authors.rst`

neo/rawio/neuralynxrawio.py
```diff
@@ -297,7 +322,27 @@ def _get_signal_size(self, block_index, seg_index, channel_indexes):
     def _get_signal_t_start(self, block_index, seg_index, channel_indexes):
         return self._sigs_t_start[seg_index] - self.global_t_start
 
-    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, channel_indexes):
+    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, channel_indices):
```
samuelgarcia, line 325: This channel_indexes>channel_indices could be globally a good idea.
But dor consistency we would need to change all IOs.
I propose to go back for this PR to channel_indexes because it is a very scondary change change.

@apdavison : should also make some comment sof this naming choice.

samuelgarcia:
Hi Peter,
thank you very much for this.
Except the comment I have made in the code, I think I am OK with this patch.
Lets see if Andrew or Julia will have time to review this.
Cheers
Samuel

PeterNSteinmetz:
It is definitely easy to revert the change of the local argument name to channel_indices if that is what people want. Since this is a name simply local to the function and makes no difference to the caller, I would suggest just gradually changing this spelling in functions as they are changed for other reasons.

PeterNSteinmetz:
Bumping this. Can I get some advice on use of channel_indices for local argument names in non-public functions, i.e., those beginning with _ ? I can then change this in both of my current pull requests.

neo/rawio/neuralynxrawio.py
JuliaSprenger: I agree with @samuelgarcia, the naming discussion should be solved in another issue.

JuliaSprenger:
Hi @PeterNSteinmetz, thanks for the the PR . Look good to me. Please just revert the name changes for `channel_indices` and I would be happy to merge.

neo/rawio/neuralynxrawio.py
```diff
@@ -221,8 +244,9 @@ def _parse_header(self):
                     ts0 = ts[0]
                     ts1 = ts[-1]
                 ts0 = min(ts0, ts[0])
-                ts1 = max(ts0, ts[-1])
+                ts1 = max(ts0, ts[-1])  # :FIXME: possible error, should be ts1
```
JuliaSprenger, line 247: Yes, I think this is indeed wrong and should read `ts1 = max(ts1, ts[-1])` instead.

neo/rawio/neuralynxrawio.py
```diff
@@ -297,7 +322,27 @@ def _get_signal_size(self, block_index, seg_index, channel_indexes):
...
+        ----------
+        block_index:
+            index of block in dataset, ignored as only 1 block in this implementation
+        seg_index:
+            index of segment to use
+        i_start:
+            sample index of first sample within segment to retrieve
+        i_stop:
+            sample index of last sample within segment to retrieve
+        channel_indices:
```
JuliaSprenger, line 339: This should read `channel_indexes`.

JuliaSprenger (approved):
There are still a number of `TODO` notes in your changes. Are you planning to address these in this PR, i.e. is this PR still work in progress?

neo/rawio/neuralynxrawio.py
PeterNSteinmetz: OK, I just pushed a commit to this branch and pull request which reverts to channel_indexes, per convention.

neo/rawio/neuralynxrawio.py
PeterNSteinmetz: I fixed the this particular bug and removed the :FIXME: comment. The :TODO: comments can remain for now and will be partially addressed in a new PR soon.

neo/rawio/neuralynxrawio.py
PeterNSteinmetz: Also reverted in last commit.

PeterNSteinmetz:
Appears now to past all tests and should be ready to merge.

JuliaSprenger (approved):

PR #863, opened 2020-09-15
Updated install instructions to modern best practices

### LINKED ISSUE #861: Unconventional installation recommendations in the docs
The installation docs start out with 

> For Debian/Ubuntu, you can install these using:
> $ apt-get install python-numpy python-pip
> $ pip install quantities
> **You may need to run these as root**. For other operating systems, you can download installers from the links above, or use a scientific Python distribution such as Anaconda.

These are quite unconventional recommendations:
* You're installing python packages using the distro, which installs the package for the system Python installs. It is absolutely not recommended to use these system Pythons (python2.x and python3.x) yourself. Most users (should) have their own setup such as pyenv or virtual envs which these instructions will ignore (iirc).
* The part in bold condones installing Python packages with root; it's probably aimed at the `apt-get install` but can easily be interpreted that if something goes wrong with the `pip install` they should just sudo it. During `pip install` python packages can execute arbitrary code! Doing `sudo pip install` is a huge security risk, you're giving untrusted arbitrary code sudo rights.

May I propose updating this to:

```
python -m ensurepip
pip install --upgrade pip
pip install numpy quantities
```

This uses the pip bootstrapping module `ensurepip`, the recommended way of installing pip, upgrades it and then installs the python dependencies. All taking virtual environments and other shindigs into account.

I'd even leave off the first 2 lines, as most people with python have pip. 

Ideally, make sure that your own python package lists its dependencies in `setup.py` under the `install_requires` kwarg and then a simple `pip install neo` would install Neo and these instructions can be left off entirely.

### LINKED ISSUE #862: Installation docs claim you can't install alpha or beta versions
> Warning
>
> alpha and beta releases cannot be installed from PyPI.

Sure they can :smiley: ! Just specify the alpha/beta semver:

    pip install neo==0.8.0a5
    pip install neo==0.8.0b0

I see that your project hasn't uploaded any, maybe that's why? https://pypi.org/project/neo/#history

Maybe to avoid making people think PyPI doesn't allow it:

> Note
>
> Our alpha and beta releases are not available on PyPI

Helveg:
I've tweaked the install instructions as they were giving me trouble breathing: ``sudo pip install`` makes me go
</img>

Some other changes include explaining how to install the IO module dependencies and in general updating to more modern and recommended practices when using pip.

closes #862, closes #861

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -5,21 +5,19 @@ Installation
 Neo is a pure Python package, so it should be easy to get it running on any
 system.
 
+Installing from the Python Package Index
+========================================
+
 Dependencies
-============
+------------
 
     * Python_ >= 3.5
     * numpy_ >= 1.11.0
     * quantities_ >= 0.12.1
 
-For Debian/Ubuntu, you can install these using::
+You can install the latest published version of Neo and its dependencies using::
 
-    $ apt-get install python-numpy python-pip
-    $ pip install quantities
-
-You may need to run these as root. For other operating systems, you can
-download installers from the links above, or use a scientific Python distribution
-such as Anaconda_.
+    $ pip install neo
 
 Certain IO modules have additional dependencies. If these are not satisfied,
 Neo will still install but the IO module that uses them will fail on loading:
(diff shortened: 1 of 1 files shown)


doc/source/install.rst
```diff
@@ -60,14 +64,6 @@ Then:
     $ python setup.py install
```
samuelgarcia, line 64: I think today the practice is `pip install .` no ?

samuelgarcia:
Thank you.
Ok for me.

apdavison (approved):
Looks fine to me. Many thanks for the PR.

doc/source/install.rst
Helveg: I don't really know; I've never encountered a use case for manually downloading and installing the package this way, I just left it in as is.

Issue #854, opened 2020-09-09
Error: "Unable to parse units μV"


arnaumanasanch:
I am working with "*.smr" files from Spike2. Sometimes I get the following error: "Unable to parse units μV" so data is not loaded. Can anyone tell me how to solve this? I suppose it is because this "mu" greek character is not included in the type of available units in neo.Spike2IO. Kind regards.

apdavison:
Hi @arnaumanasanch. Please can you post the full error traceback, so we can figure out where in the code this error is occuring?

arnaumanasanch:
Hi @apdavison , here is the traceback. I tried importing "logging" in the env but didn't work.  Thank you so much. 
```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
C:\ProgramData\Anaconda3\lib\site-packages\quantities\registry.py in __getitem__(self, string)
     20             try:
---> 21                 return eval(string, self.__context)
     22             except NameError:

<string> in <module>

NameError: name 'μV' is not defined

During handling of the above exception, another exception occurred:

LookupError                               Traceback (most recent call last)
C:\ProgramData\Anaconda3\lib\site-packages\neo\io\proxyobjects.py in ensure_signal_units(units)
    519     try:
--> 520         units = pq.Quantity(1, units)
    521     except:

C:\ProgramData\Anaconda3\lib\site-packages\quantities\quantity.py in __new__(cls, data, units, dtype, copy)
    126         ret = np.array(data, dtype=dtype, copy=copy).view(cls)
--> 127         ret._dimensionality.update(validate_dimensionality(units))
    128         return ret

C:\ProgramData\Anaconda3\lib\site-packages\quantities\quantity.py in validate_dimensionality(value)
     33         try:
---> 34             return unit_registry[value].dimensionality
     35         except (KeyError, UnicodeDecodeError):

C:\ProgramData\Anaconda3\lib\site-packages\quantities\registry.py in __getitem__(self, label)
     60 
---> 61         return self.__registry[label]
     62 

C:\ProgramData\Anaconda3\lib\site-packages\quantities\registry.py in __getitem__(self, string)
     24                 raise LookupError(
---> 25                     'Unable to parse units: "%s"'%string
     26                 )

LookupError: Unable to parse units: "µV"

During handling of the above exception, another exception occurred:

NameError                                 Traceback (most recent call last)
<ipython-input-3-370b6d6b29ae> in <module>
      3 import logging
[...]

PR #870, opened 2020-09-22
Possible fix for #854

apdavison:


samuelgarcia:
Good catch

apdavison:
Turns out this wasn't really necessary, as #854 was already fixed in master. Probably still worth merging though.

samuelgarcia:
Yes. Merge it. It will help a lot with the error come back.

PR #873, opened 2020-09-24
Fix micromed label encoding.

samuelgarcia:


JuliaSprenger:
Hi @samuelgarcia,
do we have a test file for these changes?

samuelgarcia:
Hi Julia.
Unfortunatly not.
Micromed is a medical system and I am not allow to push data from patient.
I tested it locally.
This kind of change of to be merge without CI warranty.

JuliaSprenger:
Ok, then we have to trust you blindly here...

samuelgarcia:
:)
Thank you.

Issue #857, opened 2020-09-13
Consistent string formatting


apdavison:
As part of the cleanup for v1.0 we should make the string formatting consistent across the code base.

Currently there is a mixture of `%` and `.format()`.

By that time, we should only be supporting versions of Python that support f-strings, so I would suggest we use those throughout.

zm711:
Just to address what I've done in #1396. 

* So docstrings can not be f-strings and as of my reading this is not possible bug report and discussion about why this isn't happening [here](https://bugs.python.org/issue28739). So docstrings will remain %.
* There is one conversion to what I think is hex that still needs to be fixed. There is one if-else that would require some pretty confusing nested f-strings to work (and I'm not sure of minimum version of python) so maybe we hold off and leave that as a format since it will be clearer.

PR #874, opened 2020-09-24
NIXIO Python 3.6 code cleanup

achilleas-k:
Since Python 3.6 is the minimum supported version and in the interest of Consistent string formatting (Issue #857), I've cleaned up some of the code in the NIXIO.

- F-strings are used for all string formatting.
- Removed all Python 2 specific code and exceptions.
- Used `raise ... from` for exception chaining ([PEP 3134](https://www.python.org/dev/peps/pep-3134/)).
- Unfold lines for readability when they fit into 99 chars (the project setting).

I'd also like to add type hints (in another PR).  I think they're very useful, both for the module internally but also for anyone using them outside the library for type linting and they have no effect for anyone not using them.

### DIFF
--- a/neo/io/nixio.py
+++ b/neo/io/nixio.py
@@ -32,9 +32,8 @@ import quantities as pq
 import numpy as np
 
 from .baseio import BaseIO
-from ..core import (Block, Segment, ChannelIndex, AnalogSignal,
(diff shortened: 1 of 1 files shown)


JuliaSprenger (approved):
Thanks for the code beautification. Looks good to me!

PR #884, opened 2020-10-21
Make segment time evaluation compatible with ProxyObjects

JuliaSprenger:
Currently segment `t_start` and `t_stop` evaluation of a Segment fails in case of EventProxy or EpochProxy objects being present as these don't provide `times` values. However, in contrast to Events or Epochs these objects come with `t_start` and `t_stop` attributes, that can be used instead for the time range calculation of the segment.
This PR extends the `t_start` and `t_stop` calculation of Segments to be compatible with EventProxy and EpochProxy objects.

### DIFF
--- a/neo/core/segment.py
+++ b/neo/core/segment.py
@@ -103,7 +103,12 @@ class Segment(Container):
         '''
         t_starts = [sig.t_start for sig in self.analogsignals +
                     self.spiketrains + self.irregularlysampledsignals]
-        t_starts += [e.times[0] for e in self.epochs + self.events if len(e.times) > 0]
+
+        for e in self.epochs + self.events:
+            if hasattr(e, 't_start'):  # in case of proxy objects
+                t_starts += [e.t_start]
+            elif len(e) > 0:
+                t_starts += [e.times[0]]
 
         # t_start is not defined if no children are present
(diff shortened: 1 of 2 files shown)


apdavison:
@JuliaSprenger are you planning to add unit tests? I think it's ok if you don't as we're rushing to finish the release, but I don't want to merge this if it's not ready

apdavison (commented):
The implementation looks ok, although I think it would be better to achieve this by adding `t_start` and `t_stop` properties to `Epoch` and `Event`

JuliaSprenger:
In principle I agree, however practically this is not that easy since I think that the `t_start` and `t_stop` information for event recordings is not always present in recording files. For this release I would suggest to use the proposed fix and we can evaluate if a general introduction of t_start and t_stop is compatible with all IOs for the next release.

JuliaSprenger:
I added a test for covering segment times in combination with ProxyObjects. Ready for merge from my side.

samuelgarcia:
Hi. I also agree that we should add t_start/t_stop to Event/Epoch the same we have for spiketrain.
For IOs base on rawio it is easy we can use segment t_start/t_stop.
For other ones :  I don't known.

PR #886, opened 2020-10-25
Vectorize AnalogSignal.time_index()

jpgill86:
This change allows the `AnalogSignal.time_index()` method to accept arrays of times, not just a scalar. If a scalar time is provided, the method returns a scalar index as before; if an array of times is provided, the method now returns an array of indices.

This is particularly useful with SpikeTrains:
```python
import neo
import quantities as pq

sig = neo.AnalogSignal(range(10)*pq.V, sampling_rate=1*pq.Hz)

st = neo.SpikeTrain([2, 4, 9]*pq.s, t_stop=10*pq.s)

indices = sig.time_index(st)  # previously would have needed to use something like
                              #   np.array([sig.time_index(t) for t in st])
                              # which scales poorly in comparison to the proposed
                              # solution when used on a large number of spikes

amplitudes = sig.as_quantity()[indices]
```

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -448,9 +448,9 @@ class AnalogSignal(BaseSignal):
             _pp(line)
 
     def time_index(self, t):
-        """Return the array index corresponding to the time `t`"""
+        """Return the array index (or indices) corresponding to the time (or times) `t`"""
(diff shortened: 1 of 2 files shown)


jpgill86:
Thanks for merging @apdavison!

PR #892, opened 2020-11-02
fixed a bug in loading hd5py data

dizcza:
The code to reproduce the bug:

```python
import neo

# curl https://web.gin.g-node.org/INM-6/elephant-data/raw/master/dataset-1/dataset-1.h5 --output dataset-1.h5 --location
hdf5neo = neo.io.NeoHdf5IO("dataset-1.h5")
block = hdf5neo.read_block()
```
`dataset.value` threw a deprecation warning in h5py v2.x and became an error in v3.0.0. Now it works for both h5py v2 and v3.

Deprecation warning in v2.x:

```
.../site-packages/h5py/_hl/dataset.py:313: H5pyDeprecationWarning: 
"dataset.value has been deprecated. Use dataset[()] instead. Use dataset[()] instead.", H5pyDeprecationWarning)
```

Fixed error in v3.0.0
```python
      1 hdf5neo = neo.io.NeoHdf5IO("dataset-1.h5")
----> 2 block = hdf5neo.read_block()

~/anaconda3/envs/temp/lib/python3.6/site-packages/neo/io/hdf5io.py in read_block(self, lazy, **kargs)
     87         """
     88         assert not lazy, 'Do not support lazy'
---> 89         return self.read_all_blocks(lazy=lazy)[0]
     90 
     91     def _read_block(self, node):

~/anaconda3/envs/temp/lib/python3.6/site-packages/neo/io/hdf5io.py in read_all_blocks(self, lazy, merge_singles, **kargs)
     79         for name, node in self._data.items():
     80             if "Block" in name:
---> 81                 blocks.append(self._read_block(node))
     82         return blocks
     83 

~/anaconda3/envs/temp/lib/python3.6/site-packages/neo/io/hdf5io.py in _read_block(self, node)
     97         for name, child_node in node['segments'].items():
     98             if "Segment" in name:
---> 99                 block.segments.append(self._read_segment(child_node, parent=block))
    100 
    101         if len(node['recordingchannelgroups']) > 0:

~/anaconda3/envs/temp/lib/python3.6/site-packages/neo/io/hdf5io.py in _read_segment(self, node, parent)
    172         for name, child_node in node['spiketrains'].items():
    173             if "SpikeTrain" in name:
--> 174                 spiketrains.append(self._read_spiketrain(child_node, parent=segment))
    175         segment.spiketrains = spiketrains
    176 

[...]

PR #893, opened 2020-11-03
Adapt to API change of h5py version 3.0

JuliaSprenger:
Currently hdf5io unittests are broken due to an update in the h5py API.

This PR fixes the IO by extending the first fix implemented in #892

### DIFF
--- a/neo/io/hdf5io.py
+++ b/neo/io/hdf5io.py
@@ -5,6 +5,7 @@
 
 
 import logging
+from distutils.version import LooseVersion
 import pickle
 import numpy as np
 import quantities as pq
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Cool.
Shall we also check the version when importing ?

JuliaSprenger:
I added a minimal requirement of h5py version 2.6 which is 4 years old, as at least locally this seemsstill compatible with the new implementation.

JuliaSprenger:
Finally travis tests finished. Ready from my side.

JuliaSprenger:
@samuelgarcia Any more suggestions for improvements?

samuelgarcia:
What I saw everywhere is that this HDF5 version 3.0 broke lot of libs.
Maybe we should have 2 envs in circle-ci one with version 2.6 and one with version 3.0.
What do you think ?

Is it the only faillure we have ?

JuliaSprenger:
I think we should merge this now as this is currently breaking the master branch and add a circleci coverage with h5py==2.6 in another PR.

samuelgarcia:
OK I merge

PR #904, opened 2020-11-13
Modify diagram of object for doc

samuelgarcia:
Modify diagram of object for doc
Add : 
  * Group
  * ImageSequence

Still missing:
  *  ChannelView


samuelgarcia:
@JuliaSprenger tell me if this is OK

samuelgarcia:
I propose to merge this diagram soon.
Add:
  * ImageSequence
  * Group

Still missing:
  * RegionOfInterest
  * ChannelView

Because of our fake object generator ChannelView is too tricky for now.

I propose to merge very very soon on this state to have a valide diagram in doc.

JuliaSprenger:
Hi @samuelgarcia I think for the diagram it would be best to not use the fake object generator any more, but create a neo object 'manually' including all the required objects (also RegionOfInterest, ChannelView). What do you think?

samuelgarcia:
I known, we need to generate from class desscrition itself.
But the way we have done is wrong because to get all attributes we need to instanciate object because thery are not class atributes.
This is bad.
It is a deep modification and I did not have time to do it.
We can open an issue : "make attribute available at class level and not at instance level" then the diagram generator will be ok.
Then having all object generated will super easy.
I don't vote for manual diagram.

JuliaSprenger:
Hi @samuelgarcia 
I am not sure if making the attributes at class level is a good idea, as the actual list of attributes might differ between instances of the same object, e.g. groups can allow different types of child objects, where each type would correspond to it's own attribute to access the child objects. Maybe it is better to stick with the diagram generation based on an instance of a neo structure, then the attributes are better defined.

PR #911, opened 2020-11-27
Accept file and path names to be pathlib objects

JuliaSprenger:
All IOs implicitly assume that the file and directory names provided are strings. Therefore many IOs fail, when a path specified as a [pathlib](https://docs.python.org/3/library/pathlib.html) object is provided instead. This PR introduces the automatic conversion of filenames and directory names, for all IOs and adds a common IO test.

### DIFF
--- a/neo/test/iotest/common_io_test.py
+++ b/neo/test/iotest/common_io_test.py
@@ -21,6 +21,7 @@ import os
 import inspect
 from copy import copy
 import unittest
+import pathlib
 
 from neo.core import Block, Segment
 from neo.io.basefromrawio import BaseFromRaw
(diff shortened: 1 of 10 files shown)


neo/test/iotest/common_io_test.py
```diff
@@ -192,65 +194,6 @@ def generic_io_object(self, filename=None, return_path=False, clean=False):
                                         return_path=return_path,
                                         clean=clean)
 
-    def create_file_reader(self, filename=None, return_path=False,
```
samuelgarcia, line 195: I don't remember why we add this before.
it is not used anymore ?

samuelgarcia:
Ok for me.

neo/test/iotest/common_io_test.py
JuliaSprenger: It is not used within Neo and is also not working, e.g. when trying to create a generic reader as the default filename does not exist.

JuliaSprenger:
I am not sure why circleCI is not receiving output for some time. I think we should investigate this phenomenon in general, since this has been an issue also in other PRs... Maybe it's a matter of adjusting the CI configuration to our current requirements?

JuliaSprenger:
This is rebased and tests are fixed. @samuelgarcia ready for final review.

samuelgarcia:
OK for me.
I merge.

Issue #906, opened 2020-11-18
Random Issue with Neo


ohnopleasehelpme:
Im using Neo in conjunction with SpikeInterface - so far I am unable to use Neo to open Neuralynx (.ncs) files:

import spikeinterface.extractors as se
import spikeinterface.toolkit as st
import spikeinterface.sorters as ss
import spikeinterface.comparison as sc
import spikeinterface.widgets as sw
import matplotlib.pylab as plt
import numpy as np
import neo
reader = neo.NeuralynxIO(dirname='~/PycharmProjects/Spike2020/data')
print(reader)
print(reader.segment_count(0))

I get the error:

runfile('~/PycharmProjects/Spike2020/Spike.py', wdir='~/PycharmProjects/Spike2020')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_bundle/pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "~/PycharmProjects/Spike2020/Spike.py", line 11, in <module>
    reader = neo.NeuralynxIO(dirname='~/PycharmProjects/Spike2020/data')
  File "~/opt/miniconda3/envs/Spike2020/lib/python3.7/site-packages/neo/io/neuralynxio.py", line 51, in __init__
    BaseFromRaw.__init__(self, dirname)
  File "~/opt/miniconda3/envs/Spike2020/lib/python3.7/site-packages/neo/io/basefromrawio.py", line 75, in __init__
    self.parse_header()
  File "~/opt/miniconda3/envs/Spike2020/lib/python3.7/site-packages/neo/rawio/baserawio.py", line 149, in parse_header
    self._parse_header()
  File "~/opt/miniconda3/envs/Spike2020/lib/python3.7/site-packages/neo/rawio/neuralynxrawio.py", line 111, in _parse_header
    info = NlxHeader.build_for_file(filename)
  File "~/opt/miniconda3/envs/Spike2020/lib/python3.7/site-packages/neo/rawio/neuralynxrawio.py", line 801, in build_for_file
[...]

PR #913, opened 2020-11-30
Fixes for cheetah v5.4.0

PeterNSteinmetz:
This request incorporates changes to handle headers from Cheetah v5.4.0. Also better error recovery and messages when new header styles that are not understood are encountered.

This also handles the type of variability in the time gaps between records in a contiguous section which can arise when the sampling rate is quite low, like 1017 Hz. 

This resolves issue #906 

This will not pass tests until PR # 32 in ephy_testing_data is incorporated at gin-g.node.


PeterNSteinmetz:
Two W204 PEP8 warnings to be ignored.

apdavison:
Note that PR # 32 in ephy_testing_data at gin-g.node was replaced by [# 35](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/35)

PeterNSteinmetz:
Looks like the circleci tests are failing because the test data for Cheetah 5.4.0 is not yet in the gin repository.

samuelgarcia:
This is strange the file is in gin:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/neuralynx/Cheetah_v5.4.0

Note that the old [PR](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/32) you did on gin is not deletable.
And have been replace by another one by Julia.

PeterNSteinmetz:
Maybe the CircleCI tests need to be restarted again? That looks like the correct file. It is truncated and was committed simply using git, no annexing needed.

JuliaSprenger:
Yes, I restarted the test.

PeterNSteinmetz:
These changes seems to pass the tests locally, so I assume it has to do with some problem with integrating the files in gin and CircleCI. Looks like it is now hanging on a timeout somehow?

PeterNSteinmetz:
Looks like this is now passing. Will likely need the squash commit as well. It is becoming staler as time goes by and will now need #968 to handle the later changes in #949.

JuliaSprenger:
I don't think this one should be affected by #949 as it does not interact with the baserawio specifications. Now I have to fix it in #968  if there's still incompatiblities that I didn't consider yet.

PR #928, opened 2020-12-17
PhyIO

rjurkus:
This is a draft PR for PhyIO. At the moment it has a POC PhyRawIO implementation.

Needed:
- [x] RawIO test
- [x] PhyIO Class
- [x] PhyIO test

### DIFF
--- a/neo/rawio/phyrawio.py
+++ b/neo/rawio/phyrawio.py
@@ -117,7 +117,7 @@ class PhyRawIO(BaseRawIO):
         self.unit_labels = list(clust_ids)
 
         self._t_start = 0.
-        self._t_stop = max(self._spike_times) / self._sampling_frequency
+        self._t_stop = max(self._spike_times).item() / self._sampling_frequency
 
         sig_channels = []
         sig_channels = np.array(sig_channels, dtype=_signal_channel_dtype)
--- a/neo/rawio/phyrawio.py
+++ b/neo/rawio/phyrawio.py
@@ -121,31 +121,26 @@ class PhyRawIO(BaseRawIO):
         annotation_lists = [self._parse_tsv_or_csv_to_list_of_dict(file)
                             for file in csv_tsv_files]
 
-        for block_index in range(1):
-            bl_ann = self.raw_annotations['blocks'][block_index]
-            bl_ann['name'] = f'Block #{block_index}'
-            bl_ann['block_extra_info'] = f'This is the block {block_index}'
-            for seg_index in range([1][block_index]):
-                seg_ann = bl_ann['segments'][seg_index]
-                seg_ann['name'] = f'Seg #{seg_index} Block #{block_index}'
-                seg_ann['seg_extra_info'] = f'This is the seg {seg_index} ' \
-                                            f'of block {block_index}'
-                for index, clust_id in enumerate(clust_ids):
-                    spiketrain_an = seg_ann['units'][index]
+        bl_ann = self.raw_annotations['blocks'][0]
+        bl_ann['name'] = "Block #0"
+        seg_ann = bl_ann['segments'][0]
+        seg_ann['name'] = 'Seg #0 Block #0'
+        for index, clust_id in enumerate(clust_ids):
+            spiketrain_an = seg_ann['units'][index]
 
-                    # Loop over list of list of dict and annotate each st
-                    for annotation_list in annotation_lists:
-                        clust_key, property_name = tuple(annotation_list[0].
-                                                         keys())
-                        if property_name == 'KSLabel':
-                            annotation_name = 'quality'
-                        else:
-                            annotation_name = property_name.lower()
-                        for annotation_dict in annotation_list:
-                            if int(annotation_dict[clust_key]) == clust_id:
-                                spiketrain_an[annotation_name] = \
-                                    annotation_dict[property_name]
-                                break
+            # Loop over list of list of dict and annotate each st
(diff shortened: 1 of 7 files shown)


samuelgarcia:
Hi Regi,
this is cool.
I will have a look tomorow.
I can help to push file for testing and make also test.
Thanks a lot

Qkley:
@samuelgarcia, I implemented a slightly different way to parse params.py. This method differs from the SpikeInterface approach and the ProbeInterface approach in that we do not use `exec()`.

neo/test/rawiotest/test_phyrawio.py
```diff
@@ -0,0 +1,48 @@
...
+If you have problems, do not hesitate to ask help github (prefered)
+of neuralensemble list.
+
+Note that same mechanism is used a neo.io API so files are tested
+several time with neo.rawio (numpy buffer) and neo.io (neo object tree).
+See neo.test.iotest.*
+
+
+Author: Samuel Garcia
+
```
samuelgarcia, line 22: You can remove this

Qkley: Yes, I will fix/edit docstrings, PEP8 and add unit tests for tsv/csv parser. However, I don't think I'll be able to do it today.

neo/rawio/phyrawio.py
```diff
@@ -0,0 +1,259 @@
...
+            contents = contents.replace('\n', ' ')
+            pattern = re.compile(r'(\S*)[\s]?=[\s]?(\S*)')
+            elements = pattern.findall(contents)
+            for key, value in elements:
+                metadata[key.lower()] = ast.literal_eval(value)
+
+        self._sampling_frequency = metadata['sample_rate']
+
+        clust_ids = np.unique(self._spike_clusters)
+        self.unit_labels = list(clust_ids)
```
samuelgarcia, line 117: I would use `self.unit_ids` everywhere instead of `self.unit_labels`
It will be more consistent across methods.
"label" was more an example for tridesclous than a recommandation.

neo/rawio/phyrawio.py
```diff
@@ -0,0 +1,259 @@
...
+
+        for block_index in range(1):
+            bl_ann = self.raw_annotations['blocks'][block_index]
+            bl_ann['name'] = f'Block #{block_index}'
+            bl_ann['block_extra_info'] = f'This is the block {block_index}'
+            for seg_index in range([1][block_index]):
+                seg_ann = bl_ann['segments'][seg_index]
+                seg_ann['name'] = f'Seg #{seg_index} Block #{block_index}'
+                seg_ann['seg_extra_info'] = f'This is the seg {seg_index} of ' \
+                                            f'block {block_index}'
```
samuelgarcia, line 167: Theses annoatations are taken from example I think. you can remove then.

neo/rawio/phyrawio.py
```diff
@@ -0,0 +1,259 @@
...
+            for key, value in elements:
+                metadata[key.lower()] = ast.literal_eval(value)
+
+        self._sampling_frequency = metadata['sample_rate']
+
+        clust_ids = np.unique(self._spike_clusters)
+        self.unit_labels = list(clust_ids)
+
+        self._t_start = 0.
+        self._t_stop = max(self._spike_times) / self._sampling_frequency
```
samuelgarcia: I guess there is no other solution for this.
It is ideal. Maybe you should add it in the class doc something like : `"t_stop" is not the real stop because the format don't handle it.`

Qkley: This is outdated. I did `max(self._spike_times).item() / self._sampling_frequency` in a later commit

samuelgarcia:
This looks super cool.
I think still need to clean docstrings.
Don't forget to put your name + affiliation in the module and also in `doc/source/authors.rst`.

neo/test/rawiotest/test_phyrawio.py
Qkley: Removed

neo/rawio/phyrawio.py
Qkley: `clust_ids` are not necessarily sequential. In larger data sets some clusters have no spikes and their ids would not appear in `clust_ids` or `self.unit_labels` (e.g. 0, 1, 3, 4, 5, ...). I thought Tridesclous had `self.unit_labels` for this reason, unless I am wrong?

neo/test/iotest/test_phyio.py
```diff
@@ -0,0 +1,94 @@
...
+        'phy_example_0/cluster_group.tsv'
+    ]
+
+
+class SpecificTestPhyIO(unittest.TestCase):
+    def setUp(self):
+        temp_directory = Path(tempfile.gettempdir())
+        self.temp_folder = temp_directory.joinpath(
+            'files_for_testing_neo/phy/phy_example_0')
+
```
samuelgarcia, line 40: To avoid circle ci fails.
You should not use another class but extend the TestPhyIO class.
So you could remove this setUp func and do instaed this:

```python
    def test_read_segment_lazy(self):
        dirname = self.get_filename_path('phy_example_0')
        r = PhyIO(dirname=self.temp_folder)
```

Tell me if it helps.

Qkley: This has helped, thanks! I have extended the TestPhyIO class and removed ```'seg_extra_info' in seg.annotations``` assertion. Now the Circle CI passes.

PR #915, opened 2020-12-04
Tweak raw io docs

PeterNSteinmetz:
I changed the wording about ambiguity of channel_indexes because saying 'not ambiguous' in this sentence threw me for a while. Finally figured out that what is meant is that using the indexes is unambiguous or unique to a channel. 

Also added comments regarding restrictions on sets of channels being used in RawIO functions, both in the BaseRawIO source code and the documentation.

PR #929, opened 2021-01-07
Rev raw io docs

PeterNSteinmetz:
This new branch has only the changes in documentation and comments addressing the need to have groups of channels with the same t_start and signal_length as well as the use of the method to obtain these. 

Replaces PR #915 to have a cleaner history.


samuelgarcia:
Hi Peter.
Thanks a lot for all this clarification and for the new clean PR.

Issue #930, opened 2021-01-14
[Doc] Lists of implemented formats for both IOs and RawIO not compiling


jpgill86:
These section links point to the two lists of implemented formats in the version 0.9.0 docs:

* https://neo.readthedocs.io/en/0.9.0/io.html#module-neo.io
* https://neo.readthedocs.io/en/0.9.0/rawio.html#module-neo.rawio

These appear to be no longer compiling in the latest version of the docs:

* https://neo.readthedocs.io/en/latest/io.html#module-neo.io
* https://neo.readthedocs.io/en/latest/rawio.html#module-neo.rawio

There are just section headers with nothing in the section.

jpgill86:
I see this in the [build log](https://readthedocs.org/api/v2/build/12717669.txt):

```
WARNING: autodoc: failed to import module 'core' from module 'neo'; the following exception was raised:
cannot import name 'NeuralynxRawIO' from 'neo.rawio.neuralynxrawio' (/home/docs/checkouts/readthedocs.org/user_builds/neo/envs/latest/lib/python3.7/site-packages/neo-0.9.0-py3.7.egg/neo/rawio/neuralynxrawio.py)
WARNING: autodoc: failed to import module 'io' from module 'neo'; the following exception was raised:
cannot import name 'NeuralynxRawIO' from 'neo.rawio.neuralynxrawio' (/home/docs/checkouts/readthedocs.org/user_builds/neo/envs/latest/lib/python3.7/site-packages/neo-0.9.0-py3.7.egg/neo/rawio/neuralynxrawio.py)
WARNING: autodoc: failed to import class 'rawio.ExampleRawIO' from module 'neo'; the following exception was raised:
cannot import name 'NeuralynxRawIO' from 'neo.rawio.neuralynxrawio' (/home/docs/checkouts/readthedocs.org/user_builds/neo/envs/latest/lib/python3.7/site-packages/neo-0.9.0-py3.7.egg/neo/rawio/neuralynxrawio.py)
WARNING: autodoc: failed to import class 'io.ExampleIO' from module 'neo'; the following exception was raised:
cannot import name 'NeuralynxRawIO' from 'neo.rawio.neuralynxrawio' (/home/docs/checkouts/readthedocs.org/user_builds/neo/envs/latest/lib/python3.7/site-packages/neo-0.9.0-py3.7.egg/neo/rawio/neuralynxrawio.py)
[...]

PR #917, opened 2020-12-04
Fix neuralynxrawio import, installation and files

JuliaSprenger:
Fix implementation of NeuralynxRawIO in form of multiple files

- Preserve previous import patterns
- Add new folder to regular installation modules
- Remove broken old neuralynxrawio.py file

Fixes https://github.com/NeuralEnsemble/python-neo/pull/881#issuecomment-739036272

PR #934, opened 2021-01-18
Include NeuralynxRawIO directory in non-develop installs, wheels, and source distributions

jpgill86:
This change is needed for `setup.py` to discover the `neo/rawio/neuralynxrawio` directory. Otherwise, it is excluded from non-development installs (`python setup.py install` and `pip install .`), wheels (`python setup.py bdist_wheel`), and source distributions (`python setup.py sdist`).

Note that this was not a problem for development-mode installations (`python setup.py develop` and `pip install -e .`) because when installing from the repo the NeuralynxRawIO directory is, of course, already there. It's only when the other methods are used that the directory fails to be copied to the correct final location (e.g., into `site-packages` or the source tarball).

The automated CircleCI and Travis CI tests use `pip install .`, so one would expect them to catch this problem. However, this slipped past them. I think that's because tests are executed from the repo's top-level directory, so `neo/rawio/neuralynxrawio` can be located in the working directory, even if it isn't in `site-packages`.

The only reason I noticed this is because of #930. ReadTheDocs does not change its working directory to the package's top-level before installing (instead, it installs with a lengthy path: `/home/docs/checkouts/readthedocs.org/user_builds/neo/envs/latest/bin/python /home/docs/checkouts/readthedocs.org/user_builds/neo/checkouts/latest/setup.py install --force`). Consequently, it cannot locate the NeuralynxRawIO directory when it tries to import neo after installation, and autodoc fails to compile the module and class documentation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,8 +25,9 @@ setup(
     name="neo",
     version=neo_version,
     packages=[
-        'neo', 'neo.core', 'neo.io', 'neo.rawio', 'neo.test',
-        'neo.test.coretest', 'neo.test.iotest', 'neo.test.rawiotest'],
+        'neo', 'neo.core', 'neo.io', 'neo.rawio', 'neo.rawio.neuralynxrawio',
+        'neo.test', 'neo.test.coretest', 'neo.test.iotest',
+        'neo.test.rawiotest'],
     install_requires=install_requires,
     extras_require=extras_require,
     author="Neo authors and contributors",
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from setuptools import setup
+from setuptools import setup, find_packages
 import os
 
 long_description = open("README.rst").read()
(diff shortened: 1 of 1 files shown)


jpgill86:
Neo can avoid a problem like this in the future if ever new sub-packages are introduced by using setuptool's `find_packages()`, which will automatically detect all sub-packages. Presently, it returns the same list that is hard-coded into `setup.py`, including `neo.rawio.neuralynxrawio`, so there are none to filter out:
```
$ cd python-neo
$ python -c "from setuptools import find_packages; print(find_packages())"
['neo', 'neo.core', 'neo.io', 'neo.rawio', 'neo.test', 'neo.rawio.neuralynxrawio', 'neo.test.coretest', 'neo.test.iotest', 'neo.test.rawiotest']
```
I will push another commit that uses `find_packages()`.

JuliaSprenger:
Hi @jpgill86,
yes, that's indeed a long-needed fix. Thanks for the implementation!

jpgill86:
Thanks @JuliaSprenger!

jpgill86:
Oh, I see now what you were referring to @JuliaSprenger: #917. I guess that can be closed now?

JuliaSprenger:
Almost, I still think the import pattern is nicer in #917, as it preserves the . @jpgill86  What do you think?

jpgill86:
I agree!

PR #935, opened 2021-01-20
Allow passing full path to ENT and POS files to ElanIO + fix event detection

EtienneCmb:
Hi,

I'm using Neo for reading Elan files (.eeg, .ent, .pos). However, by default the class only support the same base name for all three  files. In my case, they have different names and unfortunately I'm not allowed to change it.

In addition to that, maybe the default event encoding changed, but the events are not correctly detected.

### DIFF
--- a/neo/rawio/elanrawio.py
+++ b/neo/rawio/elanrawio.py
@@ -31,13 +31,21 @@ class ElanRawIO(BaseRawIO):
     extensions = ['eeg']
     rawmode = 'one-file'
 
-    def __init__(self, filename=''):
+    def __init__(self, filename='', entfile='', posfile=''):
         BaseRawIO.__init__(self)
         self.filename = filename
 
+        # check whether ent and pos files are defined
+        if not entfile:
+            entfile = self.filename + '.ent'
+        if not posfile:
+            posfile = self.filename + '.pos'
+        self.entfile = entfile
+        self.posfile = posfile
+
     def _parse_header(self):
 
-        with open(self.filename + '.ent', mode='rt', encoding='ascii', newline=None) as f:
+        with open(self.entfile, mode='rt', encoding='ascii', newline=None) as f:
 
             # version
             version = f.readline()[:-1]
@@ -139,12 +147,12 @@ class ElanRawIO(BaseRawIO):
         self._raw_signals = self._raw_signals[:, :-2]
 
         # triggers
-        with open(self.filename + '.pos', mode='rt', encoding='ascii', newline=None) as f:
+        with open(self.posfile, mode='rt', encoding='ascii', newline=None) as f:
             self._raw_event_timestamps = []
             self._event_labels = []
             self._reject_codes = []
             for l in f.readlines():
-                r = re.findall(r' *(\d+) *(\d+) *(\d+) *', l)
+                r = re.findall(r' *(\d+)\s* *(\d+)\s* *(\d+) *', l)
                 self._raw_event_timestamps.append(int(r[0][0]))
                 self._event_labels.append(str(r[0][1]))
                 self._reject_codes.append(str(r[0][2]))
--- a/neo/rawio/elanrawio.py
+++ b/neo/rawio/elanrawio.py
@@ -36,9 +36,9 @@ class ElanRawIO(BaseRawIO):
         self.filename = filename
 
         # check whether ent and pos files are defined
-        if not entfile:
+        if entfile is None:
             entfile = self.filename + '.ent'
-        if not posfile:
+        if posfile is None:
             posfile = self.filename + '.pos'
         self.entfile = entfile
         self.posfile = posfile
--- a/neo/rawio/elanrawio.py
+++ b/neo/rawio/elanrawio.py
@@ -31,7 +31,7 @@ class ElanRawIO(BaseRawIO):
     extensions = ['eeg']
     rawmode = 'one-file'
 
-    def __init__(self, filename='', entfile='', posfile=''):
+    def __init__(self, filename=None, entfile=None, posfile=None):
         BaseRawIO.__init__(self)
         self.filename = filename
 
--- a/neo/io/elanio.py
+++ b/neo/io/elanio.py
@@ -11,11 +11,23 @@ class ElanIO(ElanRawIO, BaseFromRaw):
     Elan is developed in Lyon, France, at INSERM U821
 
     https://elan.lyon.inserm.fr
+
+    Args:
+        filename (string) :
+            Full path to the .eeg file
+        entfile (string) :
+            Full path to the .ent file (optional). If None, the path to the ent
+            file is inferred from the filename by adding the ".ent" extension
+            to it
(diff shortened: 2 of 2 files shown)


neo/rawio/elanrawio.py
```diff
@@ -139,12 +147,12 @@ def _parse_header(self):
...
 
         # triggers
-        with open(self.filename + '.pos', mode='rt', encoding='ascii', newline=None) as f:
+        with open(self.posfile, mode='rt', encoding='ascii', newline=None) as f:
             self._raw_event_timestamps = []
             self._event_labels = []
             self._reject_codes = []
             for l in f.readlines():
-                r = re.findall(r' *(\d+) *(\d+) *(\d+) *', l)
+                r = re.findall(r' *(\d+)\s* *(\d+)\s* *(\d+) *', l)
```
samuelgarcia, line 155: Does it work with older file or does it break something ?
Do you have a file we could put in https://gin.g-node.org/NeuralEnsemble/ephy_testing_data to ensure that this work in all cases ?

neo/rawio/elanrawio.py
```diff
@@ -31,13 +31,21 @@ class ElanRawIO(BaseRawIO):
     extensions = ['eeg']
     rawmode = 'one-file'
 
-    def __init__(self, filename=''):
+    def __init__(self, filename=None, entfile=None, posfile=None):
         BaseRawIO.__init__(self)
         self.filename = filename
 
+        # check whether ent and pos files are defined
+        if not entfile:
```
JuliaSprenger, line 39: I think it would be better to compare the object identity via `is` here, to avoid puzzling  behavior in rare cases, e.g. when dealing with empty strings and pathlib objects.

neo/rawio/elanrawio.py
```diff
@@ -31,13 +31,21 @@ class ElanRawIO(BaseRawIO):
...
+            entfile = self.filename + '.ent'
+        if not posfile:
```
JuliaSprenger, line 41: same here

JuliaSprenger (commented):
Hi @EtienneCmb,
thanks for the improvements.
I think it would make sense to also cover the two new parameters in a simple test. What do you think?

neo/rawio/elanrawio.py
EtienneCmb: I think it should work with older files. For example, here's how the regexp works on the file you sent me ([link](https://regex101.com/r/07iYqf/1)) . Then, here's a more recent new pos file ([link](https://regex101.com/r/EVwW53/1)). In the second one, you can try to replace with the old regex pattern and you should see that event are not correctly found (maybe because new pos files uses tab...? not sure)

neo/rawio/elanrawio.py
EtienneCmb: Yes, you're right, I also had issues with `is`. I can replace by `isinstance(entfile, string)` if you prefer

JuliaSprenger: Sorry, I think I was not clear enough in the first attempt. What I want to propose is something like this
```suggestion
        if not (entfile is None):
```
That is exactly only evaluated True if the default value is kept and not in any other case (e.g. empty list, string, ...)

EtienneCmb: I think the `is` is going to produce a `SyntaxWarning` in Python >3.8 (see [here](https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior)). In the past I also had issues with the `is` when combined with inputs to scripts (e.g with ArgumentParser). So I would personally replace with `isinstance`

samuelgarcia: I think the `is None` or `is not None` will still be valid for long time. This syntax wrning will is for other types no ?

EtienneCmb: Yes probably. And yes, I think for the moment it's limited to strings and numbers. That said, I had issues with scripts inputs like `python myscript.py --entfile='myfile.ent'`. But probably as you both said, it's not the case here as it's a comparison with a `None`. I let you decide, let me know what fit best with the rest of the package

JuliaSprenger: Ah interesting. I didn't see this warning yet. Ok, but we only want to cover the case when `entfile` realliy is None, so would you then do `isinstance(entfile, type(None))`?

EtienneCmb: @JuliaSprenger actually I would do `if not isinstance(entfile, str)`

apdavison: The change in Python 3.8 is not related to the situation here. `if not entfile` seems to me like the best  approach, as it handles both None and empty strings. If you want to be really strict, you could check if the file actually exists.

apdavison: (if you want to allow empty strings to pass, then `if entfile is not None`

JuliaSprenger: If you force your path to be of type string then you are excluding any approach of intelligent path handling, e.g. via [pathlib](https://docs.python.org/3/library/pathlib.html). For now I think @samuelgarcia is right and `is None` should be valid for long enough for us.

EtienneCmb: Ah ok, didn't know pathlib. Then `entfile is None` is fine to me. I'll do the changes then. Thank you both !

apdavison: A downvote for `entfile is None`, because it means things will break further on if someone puts `entfile=""`, which is the sort of thing that happens when using Neo in scripts and applications. Probably the safest approach is to add a check that the file actually exists.

EtienneCmb: @apdavison good point, I can still move to `os.path.isfile`. I thought it was performed somewhere else, internally, didn't check that

JuliaSprenger: @apdavison I think in the next steps the IO should at any point check the existence of the required files. However, the `is None` question is about how you deal with default values and when to replace these. I think the `if not filename` is not the ideal solution since it can lead to unexpected behavior. E.g. when the users explicitly don't want to load the file, not in the default location, but also not in the custom location, they might just set the filename to an empty string. With the `if not filename` construct this would still load the file from the default location, even though the default value was explicitely changed by the user.

JuliaSprenger: Sorry, this was probably getting too philosophical. I am ok with any of the 2 implementations, then it's up to @samuelgarcia to merge ;)

JuliaSprenger: I created https://github.com/NeuralEnsemble/python-neo/issues/938 so we can discuss this general issue independent of this PR.
@apdavison @samuelgarcia Would it be ok for you to merge the current version and potentially update this later depending on https://github.com/NeuralEnsemble/python-neo/issues/938 ?

apdavison: yes, sure

samuelgarcia: yep. Lets merge

PR #937, opened 2021-01-25
Good channel order for old openephys format.

samuelgarcia:
In old implementation channel order was not the natural one.
"CH" channels where put after the "ADC" channels. This was the total inverse of the sofwtare.
This put the good order.

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -473,27 +473,31 @@ def explore_folder(dirname):
                 info['nb_segment'] += 1
 
     # order continuous file by channel number within segment
+    # order "CH before "ADC"
     for seg_index, continuous_filenames in info['continuous'].items():
-        chan_ids = {}
+        chan_ids_by_type = {}
+        filenames_by_type = {}
         for continuous_filename in continuous_filenames:
             s = continuous_filename.replace('.continuous', '').split('_')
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
Hi @samuelgarcia,
the new implementation is definitely more readable than the old one. However, it's not clear to me why the IO should reorder the channels if they were stored in a particular order in the file. The information which channel the data originated from was also already accessible before, right?

samuelgarcia:
It is a several file data format.
In the previous implementation the order was bad. (I did it).
The order was not corresponding to the sofwtare. it was not a poblem becuase it was 2 distinct AnalogSignal (2 groups).
But this could leave to problem when making the electrode probe location with probeinterface and PRB files.
This new order is the good one.

JuliaSprenger:
@samuelgarcia Do you plan to also include a test for this one or is this PR complete from your side?

samuelgarcia:
Honestly, it was not my plan but if it is too shocking I can try.

JuliaSprenger:
@samuelgarcia this could turn into a showcase PR ;)

samuelgarcia:
héhéhéhé.

samuelgarcia:
I could not leave a PR without tests.
I must be exemplary.
It is done now.

JuliaSprenger:
Thanks for adding a test! I will merge now.

PR #943, opened 2021-02-09
Remove deprecated classes (ChannelIndex, Unit) and IOs; simplify tests

apdavison:
This is a big PR, so please review carefully!

- removes the deprecated core classes (ChannelIndex, Unit)
- removes IO modules which are deprecated or unmaintained
- simplifies the unit tests for core classes


apdavison:
On reflection, I think this PR is doing too much. I'm closing this, and will replace with several individual PRs for the different cleanups.

PR #950, opened 2021-03-02
Remove ElphyIO, as it is not currently maintained

apdavison:


samuelgarcia:
Hi Andrew.
I known this is unmaintain but would be relevant to have a minimal version only to retrieve signal ?

apdavison:
I don't know how feasible that is. The last person to maintain this was @dguarino, who had plans to rewrite and simplify the entire module, but he is working on other things now.

JuliaSprenger:
@samuelgarcia I think we could keep a minimal version of the IO if we would have corresponding tests. There are actually some files on gin, but no io specific unittest based on these. Do you think it would be feasible on adding some tests, i.e. do you know another tool that reads elphy files so we can compare the resulting traces?

JuliaSprenger:
closed via #1011

PR #951, opened 2021-03-02
Remove NeoHdf5IO

apdavison:
NeoHdf5IO has been read-only since Neo v0.5, and will not be updated to handle the new grouping API (Group, etc.)

For HDF5 files created by Neo v0.4 or earlier, install Neo v0.9, load the data then save to NixIO, and then install the latest version of Neo.


samuelgarcia:
Can I merge ?

Issue #850, opened 2020-09-03
Depreciate neuralynxio_v1 and blackrockio_v4


samuelgarcia:
neuralynxio_v1 and blackrockio_v4 use ChannelIndex.
It will be a pain to make the change to go to Group/ChannelView right now.
Furthermore I don't think it will be usefull.

Since neo 0.9 will still have ChannelIndex I propose:
  * depreciate with a warning theses two IOs in 0.9 but keep ChannelIndex in them.
  * remove then when ChannelIndex will also be removed in 0.10

@JuliaSprenger @mdenker : any discord  :stuck_out_tongue_winking_eye: ?

PR #952, opened 2021-03-02
Remove old BlackrockIO

apdavison:
(previously deprecated, cf #850)


apdavison:
For anyone finding this via a search, please note that Blackrock files are still supported - this PR is just removing an older implementation which coexisted with the newer implementation for a while.

samuelgarcia:
ok

PR #953, opened 2021-03-02
Remove old NeuralynxIO

apdavison:
(previously deprecated, cf #850)

For anyone finding this via a search, please note that Neuralynx is still supported - this PR is just removing an older implementation which coexisted with the newer implementation for a while.


samuelgarcia:
ok

PR #949, opened 2021-02-19
Rawio improvement

samuelgarcia:
See #895 

Some important change in rawio API.
  * use now the concept of signals_stream + signal_channels to avoid ambiguity
  * _get_analogsignal_chunk work with local to stream index
  * better raw_annotations 
  *  raw_annotations also handle array_annotations
  * signal 'id' is now a str (like all other 'id')

TODO:
  * create generic lay object
  * implemnent prefer_slice=True/False

For nowonly ExampleIO/Examplerawio is done.

PR #954, opened 2021-03-05
Spikegadgets implementation

samuelgarcia:
Draft for spikegadgets implementation.
Depend on #949.
Do not merge before #949.


neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+                        stream_ids.append(stream_id)
+                        stream_name = stream_id
+                        signal_streams.append((stream_name, stream_id))
+                        self._mask_channels_bytes[stream_id] = []
+                    
+                    name = channel.attrib['id']
+                    chan_id = channel.attrib['id']
+                    dtype = 'int16'
+                    units = 'uV' # TODO check where is the info
+                    gain = 1. # TODO check where is the info
```
JuliaSprenger, line 121: this should be in the header, no?

neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+        signal_channels = []
+        
+        # walk in xml device and keep only "analog" one
+        self._mask_channels_bytes  = {}
+        for device in hconf:
+            stream_id = device.attrib['name']
+            for channel in device:
+
+                if 'interleavedDataIDByte' in channel.attrib:
+                    # TODO deal with "headstageSensor" wich have interleaved
```
JuliaSprenger, line 106: which have interleaved what?
```suggestion
                    # TODO deal with "headstageSensor" which have interleaved
```

neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+            stream_name = stream_id
+            signal_streams.append((stream_name, stream_id))
+            self._mask_channels_bytes[stream_id] = []
+            
+            chan_ind = 0
+            for trode in sconf:
+                for  schan in trode:
+                    name = 'trode' + trode.attrib['id'] + 'chan' + schan.attrib['hwChan']
+                    chan_id = schan.attrib['hwChan']
+                    units = 'uV' # TODO check where is the info
```
JuliaSprenger, line 143: same here...?

neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+                chan_inds = np.arange(num_chan)[channel_indexes]
+            else:
+                chan_inds = channel_indexes
+            stream_mask = np.zeros(raw_unit8.shape[1], dtype='bool')
+            for chan_ind in chan_inds:
+                chan_mask = self._mask_channels_bytes[stream_id][chan_ind]
+                stream_mask |= chan_mask
+            # TODO : make a fix when "channel_indexes" arein wring order.
+
+        # this do a copy from memmap to memory
```
JuliaSprenger, line 240: Is there a way to avoid copying the data here? I thought this method should be fast and ideally not copy the data...

JuliaSprenger, line 240:
```suggestion
        # this copies the data from the memmap into memory
```

neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+        raw_unit8_mask = raw_unit8[:, stream_mask]
+        shape = raw_unit8_mask.shape
+        shape = (shape[0], shape[1] // 2)
+        # reshape the and re type by view
```
JuliaSprenger, line 244:
```suggestion
        # reshape the and retype by view
```

bendichter:
@samuelgarcia what's the status of this PR? Is it still active?

samuelgarcia:
Yes, this is still active. I am still waiting fileswith the correct version that support the scaling to make tests.

bendichter:
Ok, who will give you those files? Can I help with that?

samuelgarcia:
Matthias Karlsson will send me some files.
The one that are in gin already are older version of the format and don't have the gain for scalign included.
I need more recent to make tests.

CodyCBakerPhD:
@samuelgarcia I may have some of this newer version data as well - when I try testing this branch on it acting through `spikeextractors`, the gains all default to `1.` since I assume they're not being set.

The traces look fine otherwise, but I notice that using `get_traces` there does not ultimately return a `np.memmap` (instead returns a `np.array`), which is odd since it seems like the code is reading the file that way - any clue what might be going on there?

samuelgarcia:
Hi Cody.
If you have this new format I would be happy to include it in gin. So I could continue (and maybe finish) the work started here.
Maybe you will need to reduce theses files (knowning the pacquet size).

Yes at the moment gain is 1 just as proof of concept to retrieve buffer.

The output of  of get_traces will never be a memmap. Internall the neo IO handle a memmap but when you ask for trace a "boolean mask" is applied to select channel is in the same stream. This operation imply a copy in numpy ecosystem.
See lines 224.

CodyCBakerPhD:
Data is up on GIN now @samuelgarcia, hope it helps finish this off!

samuelgarcia:
Hi Cody,
To finish this PR I really would need to get at least one file that contain the gain.
The 2 files that we have on gin don't have this gain included, they are too old version of file format.

bendichter:
@samuelgarcia These data files are representative of what we need to read and from what I understand are representative of what a large portion of the community needs. We would really like to use NEO for this if possible. I don't see why you would need newer files in order to read older files. I understand that you want to support the latest format, and I agree that is important, but would it be possible to move forward with reading of an older format (that is still in common use)?

CodyCBakerPhD:
@samuelgarcia I agree that setting the conversion factor to `None` or `nan` is a better way to deal with these other analog signals, since they almost certainly don't share the same electrode properties as the primary electrodes. Please let me know if there is anything else I can help with to get this merged.

samuelgarcia:
Ready to review (and merge maybe)

JuliaSprenger:
@samuelgarcia can you please first address all open comments before we continue with the review?

samuelgarcia:
Putting gain to nan is a bad idea in neo because the neo.io will apply this gain anyway and the end user will have a signals with nan.
Having unit='' and gain=1 is safer and more easy to end user.

JuliaSprenger:
Hi @samuelgarcia I think there are still some comments open to improve the readability of the documentation. Could you merge / comment / refuse those?

neo/rawio/spikegadgetsrawio.py
```diff
@@ -0,0 +1,247 @@
...
+
+class SpikeGadgetsRawIO(BaseRawIO):
+    extensions = ['rec']
+    rawmode = 'one-file'
+
+    def __init__(self, filename='', selected_streams=None):
+        """
+        
+        filename: str
+            filename ".rec"
```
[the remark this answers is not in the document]
samuelgarcia: OK I will remove the .Rec

neo/rawio/spikegadgetsrawio.py
samuelgarcia: No impossible.

neo/rawio/spikegadgetsrawio.py
samuelgarcia: at the moment units won't be handle in this IOs

samuelgarcia:
Hi Julia.
I think it is ok now.

neo/io/spikegadgetsio.py
```diff
@@ -0,0 +1,10 @@
+from neo.io.basefromrawio import BaseFromRaw
+from neo.rawio.spikegadgetsrawio import SpikeGadgetsRawIO
+
+
+class SpikeGadgetsIO(SpikeGadgetsRawIO, BaseFromRaw):
+    """
```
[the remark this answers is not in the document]
JuliaSprenger: I don't see any documentation added here, maybe you didn't commit it?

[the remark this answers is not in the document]
JuliaSprenger: @samuelgarcia ?

Issue #956, opened 2021-03-07
Reading Axona data


sbuergers:
Hi,

### Reading Axona data

I am with CatalystNeuro and we want to build support for reading [Axona](http://space-memory-navigation.org/DacqUSBFileFormats.pdf) data files with python-neo, and then write a wrapper for [spikeextractors](https://github.com/catalystneuro/spikeextractors) using the python-neo implementation. 

I can share some example files if requested. 

I will start working on this here: [https://github.com/catalystneuro/python-neo/tree/axonarawio](https://github.com/catalystneuro/python-neo/tree/axonarawio)

This is a useful resource for reading continuous raw data from Axona .bin files by the Hussaini Lab (Geoff Barrett): [https://github.com/HussainiLab/BinConverter/blob/master/BinConverter/core/readBin.py](https://github.com/HussainiLab/BinConverter/blob/master/BinConverter/core/readBin.py)

For the moment I will focus on the raw continuous data (.bin files containing the raw data and .set files with meta information of the recording setup), but TINT data formats should be readable as well (e.g. .eeg, .eeg1, .1, egf, .egfX, .pos, .cut and .set files).

### Additional considerations

We also want to be able to perform some preprocessing and spike sorting using [SpikeInterface](https://github.com/SpikeInterface/) and subsequently export data to TINT format. I am not sure if it is a good idea to interface with python-neo for this as well and I am happy to discuss options. 

Since I am completely new to python-neo, any tips or hints will be appreciated. :-)

Cheers,
Steffen.

samuelgarcia:
Hi Steffen,
yes make this rawio in neo is a good idea.
Then making the wrapper in spikeextractor will be super easy.
If the file is "raw" based the code should be super easy (if the header is not too complex)

I can help you. Do you have a small test example file that could be public ?
We need to deposite a testing public file [here ](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data)

The branch you mention is a copy/paste of exmple at the moment.
If you send me a file I can help to make the skeleton.

sbuergers:
Hi Samuel,

Thanks for the support. We requested a small example file that could be made public. I will let you know once that is available.

JuliaSprenger:
The files should be available now via https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/41

sbuergers:
[...]

PR #958, opened 2021-03-11
Axonarawio

sbuergers:
Draft for axonarawio.py (which is still very much in development).

Note that the draft still includes a jupyter notebook `read_axona_bin_cont.ipynb` that I use for testing code, which is in this repo for my convenience. But once we actually want to merge to master this has to be deleted first.


samuelgarcia:
Hi steffen,
I will have a look and help quite soon.
I made a PR in gin for testing files. They should be merged soon.

alejoe91 (changes requested):

neo/io/axonaio.py
```diff
@@ -0,0 +1,19 @@
...
+class AxonaIO(AxonaRawIO, BaseFromRaw):
+    name = 'Axona IO'
+    description = "Read raw continuous data (.bin and .set files)"
+
+    # This is an inportant choice when there are several channels.
+    #   'split-all' :  1 AnalogSignal each 1 channel
+    #   'group-by-same-units' : one 2D AnalogSignal for each group of channel with same units
+    _prefered_signal_group_mode = 'group-by-same-units'
+
+    def __init__(self, filename=''):
```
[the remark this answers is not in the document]
sbuergers: Makes sense. Done.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+                            t_start=None, t_stop=None)
+        >>> spike_times = reader.rescale_spike_timestamp(spike_timestamp, 'float64')
+        >>> ev_timestamps, _, ev_labels = reader.event_timestamps(event_channel_index=0)
+
+    """
+    # TODO Why do I need these?
+    extensions = ['bin']
+    rawmode = 'one-file'
+
+    def __init__(self, filename='', sr=48000):
```
alejoe91, line 82: @sbuergers same here:)

neo/io/axonaio.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class AxonaIO(AxonaRawIO, BaseFromRaw):
+    name = 'Axona IO'
+    description = "Read raw continuous data (.bin and .set files)"
+
+    # This is an inportant choice when there are several channels.
+    #   'split-all' :  1 AnalogSignal each 1 channel
+    #   'group-by-same-units' : one 2D AnalogSignal for each group of channel with same units
+    _prefered_signal_group_mode = 'group-by-same-units'
```
samuelgarcia, line 12: It is now the default mode, so you can remove this lines I think.

samuelgarcia:
Hi Steffen. Is it ready to review ?

sbuergers:
@samuelgarcia It's my first draft, but your input would certainly be super useful! 

I am currently debating exactly how I should include the other data formats / files from axona, I will write a more detailed description of this in the issue: https://github.com/NeuralEnsemble/python-neo/issues/956#issue-824015671

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+  2. Step 2: RawIO test:
+
+  3. Step 3 : Create the neo.io class with the wrapper
+    * Create a file in neo/io/ that ends with "io.py"
+    * Create a class that inherits both your RawIO class and BaseFromRaw class
+    * copy/paste from neo/io/exampleio.py
+
+  4.Step 4 : IO test
+    * create a file in neo/test/iotest with the same previous name with "test_" prefix
+    * copy/paste from neo/test/iotest/test_exampleio.py
```
samuelgarcia, line 40:
```suggestion

```

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+                            i_start=0, i_stop=1024,  channel_names=channel_names)
+        >>> float_chunk = reader.rescale_signal_raw_to_float(raw_chunk, dtype='float64',
+                            channel_indexes=[0, 3, 6])
+        >>> spike_timestamp = reader.spike_timestamps(spike_channel_index=0,
+                            t_start=None, t_stop=None)
+        >>> spike_times = reader.rescale_spike_timestamp(spike_timestamp, 'float64')
+        >>> ev_timestamps, _, ev_labels = reader.event_timestamps(event_channel_index=0)
+
+    """
+    # TODO Why do I need these?
```
samuelgarcia, line 78: It can be good for auto guess the file format, even if the "bin" is quite fuzzy.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+    extensions = ['bin']
+    rawmode = 'one-file'
+
+    def __init__(self, filename, sr=48000):
+        BaseRawIO.__init__(self)
+
+        # note that this filename is used in self._source_name
+        self.filename = filename
```
samuelgarcia, line 86: Maybe you can do something in case the filename already ends with 'bin' or 'set'

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # There is no global header for .bin files
+        self.global_header_size = 0
+
+        self.sr = sr
+        self.num_channels = len(self.get_active_tetrode()) * 4
+
+        # How many 432 byte packets does this data contain (<=> num. samples / 3)?
+        self.num_total_packets = int(os.path.getsize(self.bin_file)/self.bytes_packet)
+        self.num_total_samples = self.num_total_packets * 3
```
samuelgarcia, line 105: I would move this in parse_header

neo/rawio/axonarawio.py
alejoe91: In my opinion, the user could pass either a bin or a set file and the class should look for the companion file

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        return self.filename
+
+    def _parse_header(self):
+        # This is the central of a RawIO
+        # we need to collect in the original format all
+        # informations needed for further fast acces
+        # at any place in the file
+        # In short _parse_header can be slow but
+        # _get_analogsignal_chunk need to be as fast as possible
+
```
samuelgarcia, line 121: Please remove all coments given by example.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        # at neo.io level IO are free to add some annoations
+        # to any object. To keep this functionality with the wrapper
+        # BaseFromRaw you can add annoations in a nested dict.
+        self._generate_minimal_annotations()
+
+    def get_signal_streams_header(self):
+        '''
+        create signals stream information (we always expect a single stream)
+        '''
+        return np.array([('stream 0', 0)], dtype=_signal_stream_dtype)
```
samuelgarcia, line 149:
```suggestion
        return np.array([('stream 0', '0')], dtype=_signal_stream_dtype)
```

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        sample3 = sample2 + 64
+
+        sig_ids = np.empty((sample1.size+sample2.size+sample3.size,), dtype=sample1.dtype)
+        sig_ids[0::3] = sample1
+        sig_ids[1::3] = sample2
+        sig_ids[2::3] = sample3
+        sig_ids = sig_ids[rem:(rem+num_samples)]
+
+        # Read one channel at a time
+        raw_signals = np.ndarray(shape=(num_samples, len(channel_indexes)))
```
samuelgarcia, line 206: As the header declare int16, you should force int16 here.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        sample 1: 32b (head) + 2*38b (remappedID) and 2*38b + 1b (second byte of sample)
+        sample 2: 32b (head) + 128 (all channels 1st entry) + 2*38b (remappedID) and ...
+        sample 3: 32b (head) + 128*2 (all channels 1st and 2nd entry) + ...
+
+        NOTE: I believe there is always a single stream (all channels have the same SR)
+        """            
+
+        if channel_indexes is None:
+            channel_indexes = [i+1 for i in range(self.num_channels)]
+
```
samuelgarcia, line 186: i_stop/i_start can be None you should check this.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        return np.arange(1, 5) + 4 * (int(tetrode) - 1)
+
+    def get_channel_offset(self, chan_id):
+        """ 
+        In the .bin file, channels are arranged in a strange order. This method takes
+        a channel index as input and returns the actual offset for the channel in the
+        memory map (self._raw_signals).
+        """
+        channnel_memory_offset = np.array(
```
samuelgarcia, line 277: I would put this variable as global because it is instancieted at each get_analogsignal_chunk and will maybe have some impact of perf.

samuelgarcia:
As a general comments, I would not include the notebook. It is hard to follow change in github.
We will use sphinx-gallery soon with standard "py" file that generate a notebook. Before that better to not have NB.

Did you check performence for reading ?
You have an helper func here
`neo.tests.rawiotest.rawio_compliance.benchmark_speed_read_signals(...)`

neo/rawio/axonarawio.py
samuelgarcia: Yes, I agree. This attr is for auto guess the file format only anf is independant of the init.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        active_tetrode_set = self.get_active_tetrode()
+        num_active_tetrode = len(active_tetrode_set)
+
+        elec_per_tetrode = 4
+        letters = ['a', 'b', 'c', 'd']
+        dtype = 'int16'
+        units = 'uV'
+        gain_list = self.get_channel_gain()
+        offset = 0  # What is the offset?
```
samuelgarcia, line 344: The offset is when you have uint16 for instance and you need to do
`sig_scaled = sig_raw * gain + offset`

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+        offset = 0  # What is the offset? 
+
+        sig_channels = []
+        for itetr in range(num_active_tetrode):
+
+            for ielec in range(elec_per_tetrode):
+                
+                cntr = (itetr*elec_per_tetrode) + ielec
+                ch_name = '{}{}'.format(itetr, letters[ielec])
+                chan_id = cntr + 1
```
samuelgarcia, line 353: Chan_id is no a str

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,374 @@
...
+                gain = gain_list[cntr]
+                stream_id = 0
```
samuelgarcia, line 355: stream_id is also a str

samuelgarcia:
Sorry I accidentaly close/reopen the PR.

neo/rawio/axonarawio.py
sbuergers: Ok, users can input `filename`, `filename.set` or `filename.bin`. 

I do not think `extensions = ['bin']` is used anywhere for auto-guessing, remove?

sbuergers:
Hi Samuel,

thanks a lot for the review! I wasn't planning on including the notebook in the final PR, sorry for including it in the first place, causing confusion.

I will check my changes later this afternoon and test code performance as you suggested, and ping you once I think it's ready for review again.

neo/rawio/axonarawio.py
sbuergers: I added it as class variable. Is that good practice?

sbuergers:
@samuelgarcia can you have a look if this is good now, please? :-)

neo/rawio/axonarawio.py
alejoe91: @sbuergers I think you should add that line. The argument is a `filename`, so it should have an extension. I would check that the extension is either `.set` or `.bin`, then replace the suffix with `''` and create the bin and set filenames.

Note that using `pathlib` makes it muuuuch nicer:

```
from pathlib import Path

# in the __init__
filename = Path(filename)

assert filename.suffix in ['.bin', '.set'], "'filename' should be a .bin or .set file"
self.bin_file = filename.parent / f"{filename.stem}.bin"
self.set_file = filename.parent / f"{filename.stem}.set"

assert self.bin_file.is_file(), "..."
assert self.set_file.is_file(), "..."
```

alejoe91 (approved):
@sbuergers @samuelgarcia 

Looks good to me!

sbuergers:
@samuelgarcia I debugged the Pep8 warnings with Flake8 locally, how do I run the pep8speaks check again?

sbuergers:
@samuelgarcia The tests passed earlier, but I had to do some final Pep8 adjustments (spaces between arithmetic operators, some empty lines to be deleted or added.

samuelgarcia:
Hi Steffen. Thanks very much for this PR. It is OK for me.
I merge.

sbuergers:
Thanks @samuelgarcia 

Is it customary to delete the branch now?

samuelgarcia:
I think you have to delete the branch in your repo by your own.
As it is squash commit, I would advice to do it straigth now to not make futur mistake working on the same branch.

neo/rawio/axonarawio.py
```diff
@@ -0,0 +1,348 @@
...
+        return self.filename
+
+    def _parse_header(self):
+        '''
+        Read important information from .set header file, create memory map
+        to raw data (.bin file) and prepare header dictionary in neo format.
+        '''
+
+        # Get useful parameters from .set file
+        params = ['rawRate']
```
GeoffBarrett, line 85: Just was checking w/ Abid on this project and it popped up in my head. There is this looming bug in the Axona software where it will save the sample rate at 48kHz even if the end-user recorded in their 2ms spike (24kHz) mode. You have to do an additional check for the `Spike2msMode` parameter. If the value is 0, then it is 48k, else 24k.

An additional caveat, I saw that Abid sent you a 2ms mode file that was just converted from the RAW mode. That converter I think will just dump the same `.set` file, so 2ms mode will not be set to 1 (thus the above check isn't even true in the case where the end-user wants to chunk their data from the raw mode to the 24k mode, although I'm not sure why they would want to). 

Could try to just infer sampling rate from the timestamps of the `.N` files. I think a majority of the end-users use the 48k mode though.

sbuergers: Hi Geoff,

Thanks a bunch for pointing this out! We will take it into account!

Cheers,
Steffen.

JuliaSprenger: Hi @GeoffBarrett. Thanks for the feedback. I updated the current AxonaIO PR accordingly: https://github.com/NeuralEnsemble/python-neo/pull/986
Now the `Spike2msMode` is first taken into account to determine the sample rate and only if this fails the `.N` file header `sample_rate` is used. For now I am not trying to extract the sample rate from the timestamp resolution as I think this method also won't be 100% reliable, depending on the number of spikes extracted and the underlying precision of the timestamps.

GeoffBarrett: @JuliaSprenger you're right, the timestamps in the `.N` would just provide the time that the signal went above threshold, so they wouldn't really allow you to infer sampling frequency. I think the best you'd be able to do is read that `Spike2msMode`. I haven't used the pipeline, but potentially the end-user could provide a sample frequency override kwarg, not sure if it's some sort of CLI and the users can provide parameters or something. 

It is definitely an issue that I brought up with Jim Donnett a few years back. If you look through the header info in the 2ms data that Abid has provided you will see the 48k (as well as in the `.set`) so the issue is definitely there. I don't have any 24kHz recorded data, but you could always reach out to Abid and have him record a minute of 24kHz data (not converted from `.bin`), and look at the headers.

PR #963, opened 2021-03-18
Remove NSDFIO

apdavison:
The `nsdf` package is not available on PyPI, and it appears the underlying [nsdf library](https://github.com/nsdf/nsdf) is no longer maintained (no Github activity since February 2018).

@subhacom @mgr0dzicki - do you have any interest in us keeping this module in Neo? If so we would need help in getting the tests to pass.


subhacom:
@apdavison I don't know if anybody else is using `nsdf` any more, even via `neo` . However the version on `github` was usable and stable. Publishing the last master on PyPI is not an issue. But most importantly, if there are not many users, I do not see any reason to support it in `neo`.

apdavison:
Many thanks for the feedback @subhacom. I think to simplify maintenance we'll remove NDSFIO for now; if there is demand from users we can always add it back.

PR #972, opened 2021-04-16
filename needs to have .set or .bin extension

sbuergers:
This is a small update to `axonarawio.py`, enforcing that a filename is passed (as the parameter suggests), with either '.set' or '.bin' extension. A sensible error is thrown if that is not the case. 

Before it was possible to only pass the stem of the filename (e.g. "my_file" instead of "my_file.bin"). 

I updated the docstring of AxonaRawIO accordingly. 

Tests passed on my local machine.

### DIFF
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -33,11 +33,13 @@ class AxonaRawIO(BaseRawIO):
 
     The raw data is saved in .bin binary files with an accompanying
     .set file about the recording setup (see the above manual for details).
+    .bin and .set files have to have the same name. Either can be provided
+    as "filename".
 
     Usage:
         import neo.rawio
         r = neo.rawio.AxonaRawIO(
-            filename=os.path.join(dir_name, base_filename)
+            filename=os.path.join(dir_name, filename)
         )
         r.parse_header()
         print(r)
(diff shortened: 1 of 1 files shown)


sbuergers:
It looks like nixio tests are failing in circleci.

sbuergers:
@samuelgarcia there is another minute change I would like to make, but it is technically unrelated, so I wanted to ask if I should open a new PR for it, or if I may include it here:

In `self.header['signal_channels']` we assign labels like so: 0a, 0b, 0c, 0d, 1a, 1b, ..., 3d; denoting the tetrode (number) and electrode (letter). Yet, electrode and group ids are 1-indexed currently. So it would be nice to instead assign 1a, 1b, 1c, 1d, 2a, 2b, ..., 4d. 

Alternatively we could 0-index everything, which would be more in line with spikeextractors conventions. It's a bit of a funny choice, because even in the .set file some things are 0-indexed, and some are 1-indexed (example snippet below). But having a different indexing scheme for tetrodes and channels is probably not a good idea. 

I think I would actually opt for implementing 0-indexing for everything in neo (and create a new pull request for it). What do you think?

```
gain_ch_0 10000
[...]
gain_ch_1 10000
[...]
collectMask_1 1
collectMask_2 1
collectMask_3 1
collectMask_4 1
collectMask_5 0
collectMask_6 0
collectMask_7 0
collectMask_8 0
collectMask_9 0
collectMask_10 0
collectMask_11 0
collectMask_12 0
collectMask_13 0
collectMask_14 0
collectMask_15 0
collectMask_16 0
```

samuelgarcia:
Hi @sbuergers .
Bout the self.header['signal_channels']['name'] and self.header['signal_channels']['id'].
There is no specific rule about '0' or '1' indexing.
The best it to take the option that best fit the original software, in short the end user should not have to  do convertion +or-1 to go from neo to the software for checking traces.

JuliaSprenger:
Hi @sbuergers This approach is the opposite of what is currently implemented in #986, where the provided file extension is ignored and only the basename is used to search for files with files matching axona extentions. I think the #986 is more flexible for the user as the basename as well as any concrete filename can be provided. What do you think?

JuliaSprenger:
Alternative implementation was merged in #986

Issue #971, opened 2021-04-15
scan_ncs_files() not lazy?


CodyCBakerPhD:
@samuelgarcia @PeterNSteinmetz 

Hello - as a follow up to the fix provided by [PR #957](https://github.com/NeuralEnsemble/python-neo/pull/957) which resolved the basic access issue raised in [Issue #959](https://github.com/NeuralEnsemble/python-neo/issues/959), I am noticing behavior similar to old [Issue #735](https://github.com/NeuralEnsemble/python-neo/issues/735).

Simple summary: I'm trying to utilize the single-file mode for Neuralynx through `spikeextractors` by making a list of NeurlynxRecordingExtractors for each file (safer than using the `dirname` directly) as follows
```
test = [NeuralynxRecordingExtractor(filename=filename) for filename in neuralynx_files]
```
with `neuralynx_files` being a list of several `.ncs` file paths. While this works fine from a purely functional standpoint (all data and metadata is readily reported and appears accurate), I noticed that the more files I included in `neuralynx_files`, the slower the loading was and the more RAM was being used. On the upper end (64 files, up to 1 GB each), my own system would not be able to load in all the files I need to use.

Via a memory profiler I was able to track the memory usage through the `RawNeoIO` in `spikeextractors` to the `parse_header()` command in `neo` itself, and from there down to the call to `scan_ncs_files()`, which appears to be loading in all the data for that `.ncs` file rather than utilizing lazy access through memory maps and such.

Is this intended, or a bug caused by the new `filename` feature?

PeterNSteinmetz:
Not a new bug. This can likely be optimized. When the file does not appear to be one contiguous set of records due to gaps in the time stamps the start of every record in the file must be scanned to determine the appropriate segmenting of those records.

There may be a faster and more memory efficient way though I don’t see why the bits which are read should not be garbage collected. 

This version is correct. Perhaps the next version can be faster or have a better memory footprint.

PeterNSteinmetz:
The new algorithm is more complicated to handle a bunch of different cases that arise with different programs generating these .Ncs files. Likely best to see the code and comments to understand them all, but in terms of the items addressed here by @samuelgarcia, here is a brief summary of the new algorithm. 

First check the first and last record of the file (possible because records are fixed physical length). If it appears all the records are one contiguous segment of records based on the sampling rate in the header and the first and last time stamps, then assume there is one contiguous segment and setup the headers for that. 

[...]

PR #973, opened 2021-04-20
Eliminate CscRecordHeader

PeterNSteinmetz:
Factored out the temporary variables of CscRecordHeader used in Ncs record scanning. Algorithm now reads directly from the memmap which is passed around. No allocation of an object within the loop. Hopefully this will eliminate the curious memory leak noted in issue #971.


PeterNSteinmetz:
Appears to be failing on NixIO. But let's hold off anyway until @JuliaSprenger has weighed in on the memmap performance issue.

JuliaSprenger:
Hi Peter! Thanks for also looking into speed improvements! I rebased your branch onto master to have the latest version of the tests running.

JuliaSprenger (approved):

PR #974, opened 2021-04-21
made channel_names annotation 1 indexed in line with axona file format

sbuergers:
The goal was to match indexing schemes of channels and tetrodes to that used in the .set file (for an example see [here](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/axona/axona_raw.set)).

Within `raw_annotations['blocks'][0]['segments'][0]['signals'][0]['__array_annotations__']`:

`channel_names` were 0-indexed (0a, 0b, 0c, 0d, 1a, 1b, ...), and are now 1-indexed, because the numbers reflect tetrodes, which are 1-indexed in the .set file. 

`channel_ids` were 1-indexed (1,2,3, ...), and are now 0-indexed (0,1,2, ...), because channels are 0-indexed in the .set file. 

Note that in spikeextractors we will use 0-indexing for both channels and tetrodes to keep the convention consistent within spikeextractors between different file formats: https://github.com/SpikeInterface/spikeextractors/pull/605

### DIFF
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -338,7 +338,7 @@ class AxonaRawIO(BaseRawIO):
             for ielec in range(elec_per_tetrode):
 
                 cntr = (itetr * elec_per_tetrode) + ielec
-                ch_name = '{}{}'.format(itetr, letters[ielec])
+                ch_name = '{}{}'.format(itetr+1, letters[ielec])
                 chan_id = str(cntr + 1)
                 gain = gain_list[cntr]
                 stream_id = '0'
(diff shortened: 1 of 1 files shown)


sbuergers:
@JuliaSprenger this is already included in this PR https://github.com/JuliaSprenger/python-neo/pull/1, so it will get merged here https://github.com/NeuralEnsemble/python-neo/pull/986. 

Should this still be merged separately or should I delete this PR? Maybe it's nice for bookkeeping purposes to merge this, too?

JuliaSprenger:
Hi @sbuergers Thanks for the adjustment. I think the failing tests are due to this branch not including the latest changes in master and are related to a different IO.

PR #977, opened 2021-04-22
Proof of concept on github actions + datalad + github cache

samuelgarcia:
This should trigger GH actions


samuelgarcia:
@JuliaSprenger : this is our first trigger on GH actions.
Let's see.

samuelgarcia:
@JuliaSprenger 
No travis is OK.
But gh actions is still super slow.
my guess is that gin propose a limited bandwith...
I don't known what to do.

Here I am only testing neo/test/rawiotest.
On my local machine when file are already downloaded it take me 87s
On GH actions it is more than 50min.
The dataset is more or less 900Mo.
I don't anderstand why it is so slow!!!

PR #978, opened 2021-04-22
Continue github actions.

samuelgarcia:


samuelgarcia:
@JuliaSprenger : test on neo.io and neo.rawio without caching is more than one hour.
When caching the ephy_testing_dataset then it is some minuts.
I still need to fix some bugs but we are almost there.

samuelgarcia:
@JuliaSprenger : all tests pass except 2 test from nix.
COuld you have a look ?

.github/workflows/full-test.yml
```diff
@@ -36,26 +33,27 @@ jobs:
...
-          key: ${{ runner.os }}-datasets-${{ env.HASH_EPHY_DATASET }}
+          key: ${{ runner.os }}-datasets-${{ steps.vars.outputs.HASH_EPHY_DATASET }}
           restore-keys: |
             ${{ runner.os }}-datasets
       - name: Test with pytest
         run: |
-          source ./test_env/bin/activate
-          # onbly rawiotest is tested at the moment
+          source ~/test_env/bin/activate
+          # onbly neo.rawio and neo.io
```
JuliaSprenger, line 57:
```suggestion
          # only neo.rawio and neo.io
```

.github/workflows/full-test.yml
```diff
@@ -18,13 +18,10 @@ jobs:
...
           python-version: 3.8
-      - name: Which python
-        run: |
-          which python
       - uses: actions/cache@v2
         id: cache-venv
         with:
-          path: ./test_env
+          path: ~/test_env
           key: ${{ runner.os }}-venv-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/requirements_testing.txt') }}
```
JuliaSprenger, line 25: It would be good to include some time stamp in the hash, to ensure that the environment is build from scratch at least once per week / month. Maybe it's possible to use the github actions schedule for this, too? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

.github/workflows/full-test.yml
```diff
@@ -36,26 +33,27 @@ jobs:
...
       - name: Test with pytest
         run: |
-          source ./test_env/bin/activate
-          # onbly rawiotest is tested at the moment
+          source ~/test_env/bin/activate
+          # onbly neo.rawio and neo.io
           pytest --cov=neo neo/test/rawiotest
-          # TODO uncomment this soon
-          # pytest --cov=neo neo/test/iotest
+          pytest --cov=neo neo/test/iotest
```
JuliaSprenger, line 59: I think just using `pytest` here would be sufficient since the coverage information is not used later...

samuelgarcia: OK

samuelgarcia:
@JuliaSprenger : I added a clear cache for venv every month.

samuelgarcia:
@JuliaSprenger : can we merge this and then we can open a PR
`NeuralEnsemble:gh_actions_and_datalad > NeuralEnsemble:master`

PR #980, opened 2021-04-23
Axonarawio add pos stream

sbuergers:
In addition to ecephys data the Axona system saves video tracking data in the `.bin` files containing the raw data. 
Each data packet contains a header, footer and 3 samples of ecephys data for each channel. The header includes a
flag, either ADU1 or ADU2, where ADU2 denotes that video tracking data is available in this packet, which is then located
in the header. 

To quote from the [file format manual](http://space-memory-navigation.org/DacqUSBFileFormats.pdf):
```
Each position sample is 20 
bytes long, and consists of a 4-byte frame counter (incremented at around 50 Hz, 
according to the camera sync signal), and then 8 2-byte words. In four-spot mode, the 8 
words are redx, redy, greenx, greeny, bluex, bluey, whitex, whitey. In two-spot mode, they 
are big_spotx, big_spoty, little_spotx, little_spoty, number_of_pixels_in_big_spot, 
number_of_pixels_in_little_spot, total_tracked_pixels, and the 8th word is unused. Each 
word is MSB-first. If a position wasn't tracked (e.g., the light was obscured), then the 
values for x and y will both be 0x3ff (= 1023).
```

In this PR this video tracking position (pos) data is added to `axonarawio.py` as a separate stream.

### DIFF
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -291,7 +291,7 @@ class AxonaRawIO(BaseRawIO):
         ADU2_idx = np.where(flags == b'ADU2')
 
         timestamp = np.ndarray(
-            (self.num_total_packets,), np.int32, self.mmpos, 12, 
+            (self.num_total_packets,), np.int32, self.mmpos, 12,
             self.bytes_packet
         )[ADU2_idx].reshape((-1, 1))
 
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -56,7 +56,7 @@ class AxonaRawIO(BaseRawIO):
 
         float_chunk = r.rescale_signal_raw_to_float(
             raw_chunk, dtype=np.float64,
-            channel_indexes=[0, 3, 6], 
+            channel_indexes=[0, 3, 6],
             stream_index=0
         )
         print('\nRaw acquisition traces in uV:\n', float_chunk)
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Steffen, tell us when it is ready to review.

sbuergers:
Hi Samuel, I think it's ready for review from my end. I am not sure if I dealt with the sampling frequency of the position data correctly. @JuliaSprenger mentioned that there is a way to incorporate non-stationary sampling frequencies, but I wasn't quite sure how. Currently I am simply assuming a fixed rate of 100Hz (which ideally is true, but empirically, at least in our sample data, is only approximately true, because there is a period during the experiment where it starts fluctuating for a while). One of the `.pos` data "channels" contains the actual timestamps of when a position was recorded. 

Tests pass locally.

Interestingly my local linter (flake8) found a Pep8 issue that pep8speaks missed (>79 char line length).

sbuergers:
Julia pointed me toward how to deal with irregularly sampled signals. So converting to draft for now and trying to make appropriate changes.

JuliaSprenger:
Looking at the basefromrawio it does not seem like IrregularySampledSignals are supported on that level yet. @samuelgarcia Do you think it makes sense to find a workaround now to include these for the AxonaIO or do you want to unify AnalogSignals and IrregularlySampledSignals first?

sbuergers:
For the moment we implemented reading the position data in nwb-converion-tools, but once we figure out how to incorporate irregularly sampled signals it shouldn't be too hard to finalize this also for neo!

JuliaSprenger:
Including pos data in the AxonaRawIO will be postponed until IrregularlySampledSignals are supported on the RawIO level. Until then I close this PR. Feel free to reopen to continue working on this.

PR #986, opened 2021-04-30
Enable AxonaIO to read spiketimes and waveforms

JuliaSprenger:
This PR implements the required rawio methods for reading spiketimes and waveforms. It also restructures the IO to be able to support multi-file datasets (.bin & unit files).

This PR depends on the Axona unit test data that will be added to the ephy_testing_data: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/49/files

### DIFF
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -113,7 +113,6 @@ class AxonaRawIO(BaseRawIO):
                   'set': {'filename': self.set_file,
                           'header_encoding': 'cp1252'},
                   'unit': {'data_type': unit_dtype,
-                           'wf_left_sweep_us': 200,
                            'tetrode_ids': [],
                            'header_encoding': 'cp1252'}}
         self.file_parameters = params
@@ -158,9 +157,6 @@ class AxonaRawIO(BaseRawIO):
                 tdict['num_spikes'] = int(tdict['num_spikes'])
                 tdict['header_size'] = len(
                     self.get_header_bstring(tetrode_file))
-                ls = self.file_parameters['unit']['wf_left_sweep_us'] * \
-                     self._to_hz(tdict['timebase']) * 10 ** -6
-                tdict['wf_left_sweep'] = ls
 
                 # memory mapping spiking data
                 spikes = np.memmap(
@@ -175,7 +171,9 @@ class AxonaRawIO(BaseRawIO):
                 wf_units = 'dimensionless'
                 wf_gain = 1
                 wf_offset = 0.
-                wf_left_sweep = tdict['wf_left_sweep']
+                # left sweep information is only stored in set file
+                wf_left_sweep = int(self.file_parameters['set']['file_header']
+                                    ['pretrigSamps'])
                 wf_sampling_rate = self._to_hz(tdict['sample_rate'],
                                                dtype=float)
                 spike_channels.append((unit_name, unit_id, wf_units, wf_gain,
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -380,13 +380,7 @@ class AxonaRawIO(BaseRawIO):
         assert seg_index == 0
 
         tetrode_id = unit_index
-
-        # spike times are repeated for each contact -> use only first contact
-        unit_spikes = self._get_spike_timestamps(block_index, seg_index,
-                                                 unit_index, t_start, t_stop)
         waveforms = self._raw_spikes[tetrode_id]['samples']
-        nb_spikes = len(unit_spikes)
-        nb_samples_per_waveform = waveforms.shape[-1]
 
         # slice timestamps / waveforms only when necessary
         if t_start is None and t_stop is None:
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -97,7 +97,7 @@ class AxonaRawIO(BaseRawIO):
         to raw data (.bin file) and prepare header dictionary in neo format.
         '''
 
-        unit_dtype = np.dtype([('spiketime', '>i4'),
+        unit_dtype = np.dtype([('spiketimes', '>i4'),
                                ('samples', 'int8', (50,))])
 
         # Utility collection of file parameters (general info and header data)
@@ -352,7 +352,7 @@ class AxonaRawIO(BaseRawIO):
         raw_spikes = self._raw_spikes[tetrode_id]
 
         # spike times are repeated for each contact -> use only first contact
-        unit_spikes = raw_spikes['spiketime'][::4]
+        unit_spikes = raw_spikes['spiketimes'][::4]
 
         # slice spike times only if needed
         if t_start is None or t_stop is None:
@@ -389,7 +389,7 @@ class AxonaRawIO(BaseRawIO):
         tetrode_id = unit_index
 
         # spike times are repeated for each contact -> use only first contact
-        unit_spikes = self._raw_spikes[tetrode_id]['spiketime'][::4]
+        unit_spikes = self._raw_spikes[tetrode_id]['spiketimes'][::4]
         waveforms = self._raw_spikes[tetrode_id]['samples']
         nb_spikes = len(unit_spikes)
         nb_samples_per_waveform = waveforms.shape[-1]
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -122,15 +122,16 @@ class AxonaRawIO(BaseRawIO):
         params['set']['file_header'] = set_dict
         params['set']['sampling_rate'] = int(set_dict['rawRate'])
 
+        signal_streams = self._get_signal_streams_header()
+        signal_channels = self._get_signal_chan_header()
+
         # SCAN BIN FILE
-        signal_streams = []
-        signal_channels = []
         if self.bin_file:
             bin_dict = self.file_parameters['bin']
             # add derived parameters from bin file
             bin_dict['num_channels'] = len(self.get_active_tetrode()) * 4
-            num_tot_packets = int(self.bin_file.stat().st_size
-                                  / bin_dict['bytes_packet'])
+            num_tot_packets = int(
+                self.bin_file.stat().st_size / bin_dict['bytes_packet'])
             bin_dict['num_total_packets'] = num_tot_packets
             bin_dict['num_total_samples'] = num_tot_packets * 3
 
@@ -139,9 +140,6 @@ class AxonaRawIO(BaseRawIO):
                 self.bin_file, dtype=self.file_parameters['bin']['data_type'],
                 mode='r', offset=self.file_parameters['bin']['header_size']
             )
-
-            signal_streams = self._get_signal_streams_header()
-            signal_channels = self._get_signal_chan_header()
 
         # SCAN TETRODE FILES
         # In this IO one tetrode corresponds to one unit
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -154,15 +154,12 @@ class AxonaRawIO(BaseRawIO):
                 # collecting tetrode specific parameters and dtype conversions
                 tdict = self.get_header_parameters(tetrode_file, 'unit')
                 tdict['filename'] = tetrode_file
-                tdict['timebase_hz'] = int(tdict['timebase'].replace(' hz', ''))
-                tdict['sample_rate_hz'] = int(
-                    tdict['sample_rate'].replace(' hz', ''))
                 tdict['num_chans'] = int(tdict['num_chans'])
                 tdict['num_spikes'] = int(tdict['num_spikes'])
                 tdict['header_size'] = len(
                     self.get_header_bstring(tetrode_file))
-                ls = self.file_parameters['unit']['wf_left_sweep_us'] * tdict[
-                    'timebase_hz'] * 10 ** -6
+                ls = self.file_parameters['unit']['wf_left_sweep_us'] * \
+                     self._to_hz(tdict['timebase']) * 10 ** -6
                 tdict['wf_left_sweep'] = ls
 
                 # memory mapping spiking data
@@ -179,7 +176,8 @@ class AxonaRawIO(BaseRawIO):
                 wf_gain = 1
                 wf_offset = 0.
                 wf_left_sweep = tdict['wf_left_sweep']
-                wf_sampling_rate = float(tdict['sample_rate_hz'])
+                wf_sampling_rate = self._to_hz(tdict['sample_rate'],
+                                               dtype=float)
                 spike_channels.append((unit_name, unit_id, wf_units, wf_gain,
                                        wf_offset, wf_left_sweep,
                                        wf_sampling_rate))
@@ -333,8 +331,9 @@ class AxonaRawIO(BaseRawIO):
         assert block_index == 0
         assert seg_index == 0
 
-        tetrode_id = unit_index
+        unit_params = self.file_parameters['unit']
 
+        tetrode_id = unit_index
         raw_spikes = self._raw_spikes[tetrode_id]
 
         # spike times are repeated for each contact -> use only first contact
@@ -350,8 +349,8 @@ class AxonaRawIO(BaseRawIO):
 
             # convert to t_start and t_stop to sampling frequency
             # Note: this assumes no time offset!
-            lim0 = t_start * self.file_parameters['unit']['timebase_hz']
-            lim1 = t_stop * self.file_parameters['unit']['timebase_hz']
+            lim0 = t_start * self._to_hz(unit_params['timebase'])
+            lim1 = t_stop * self._to_hz(unit_params['timebase'])
 
             # slice spike times
             mask = (unit_spikes >= lim0) & (unit_spikes <= lim1)
@@ -364,7 +363,7 @@ class AxonaRawIO(BaseRawIO):
 
     def _rescale_spike_timestamp(self, spike_timestamps, dtype):
         spike_times = spike_timestamps.astype(dtype)
-        spike_times /= self.file_parameters['unit']['timebase_hz']
+        spike_times /= self._to_hz(self.file_parameters['unit']['timebase'])
         return spike_times
 
     def _get_spike_raw_waveforms(self, block_index, seg_index, unit_index,
@@ -389,8 +388,8 @@ class AxonaRawIO(BaseRawIO):
 
             # convert to t_start and t_stop to sampling frequency
             # Note: this assumes no time offset!
-            lim0 = t_start * self.file_parameters['unit']['timebase_hz']
-            lim1 = t_stop * self.file_parameters['unit']['timebase_hz']
+            lim0 = t_start * self._to_hz(self.file_parameters['unit']['timebase'])
+            lim1 = t_stop * self._to_hz(self.file_parameters['unit']['timebase'])
 
             # slice spike times
             mask = (unit_spikes >= lim0) & (unit_spikes <= lim1)
@@ -570,3 +569,6 @@ class AxonaRawIO(BaseRawIO):
                                      units, gain, offset, stream_id))
 
         return np.array(sig_channels, dtype=_signal_channel_dtype)
+
+    def _to_hz(self, param, dtype=int):
+        return dtype(param.replace(' hz',''))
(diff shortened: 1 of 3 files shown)


sbuergers (changes requested):
Very nicely done. I mostly have nit-picky comments. I will send you the sample data so we can run proper tests with neo (should then also include the filenames in the rawio and io tests of axona).

neo/rawio/axonarawio.py
```diff
@@ -16,14 +16,14 @@
...
 """
 
-from .baserawio import (BaseRawIO, _signal_channel_dtype, _signal_stream_dtype,
-                        _spike_channel_dtype, _event_channel_dtype,
-                        _common_sig_characteristics)
-import numpy as np
-import os
-import re
-import contextlib
 import datetime
```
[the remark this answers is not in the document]
JuliaSprenger: done

neo/rawio/axonarawio.py
```diff
@@ -49,8 +49,8 @@ class AxonaRawIO(BaseRawIO):
         )
     """
 
-    extensions = ['bin']  # Never used?
-    rawmode = 'one-file'
+    extensions = ['bin', 'set'] + [str(i) for i in range(1, 33)]  # Never used?
```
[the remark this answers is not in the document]
JuliaSprenger: it is part of the IO definition and is used to determine which IO can be used to read a particular file: https://github.com/NeuralEnsemble/python-neo/blob/77455a0661a9088fddb8c05daddd6d377cf3022f/neo/io/__init__.py#L352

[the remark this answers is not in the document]
JuliaSprenger: Therefore I would only add extensions that we currently support and add the others later on when IO functionality is extended.

neo/rawio/axonarawio.py
```diff
@@ -66,11 +66,28 @@ class AxonaRawIO(BaseRawIO):
     def __init__(self, filename):
         BaseRawIO.__init__(self)
 
-        # We accept base filenames, .bin and .set extensions
-        self.filename = filename.replace('.bin', '').replace('.set', '')
-        self.bin_file = os.path.join(self.filename + '.bin')
-        self.set_file = os.path.join(self.filename + '.set')
-        self.set_file_encoding = 'cp1252'
+        # Accepting filename with arbitrary suffix as input
+        self.filename = pathlib.Path(filename).with_suffix('')
```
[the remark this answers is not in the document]
JuliaSprenger: I am not sure if this restriction is necessary, since any filename is accepted from the user side and we only internally use the basename as reference. Artificially restricting this to not accept the basename in the first place feels a bit too artificially restrictive to me...

neo/rawio/axonarawio.py
```diff
@@ -83,55 +100,122 @@ def _parse_header(self):
...
+
+        # SCAN TETRODE FILES
+        # In this IO one tetrode corresponds to one unit
+        self._raw_spikes = []
+        spike_channels = []
+        if self.tetrode_files:
+
+            for i, tetrode_file in enumerate(self.tetrode_files):
+                # collecting tetrode specific parameters and dtype conversions
+                tdict = self.get_header_parameters(tetrode_file, None)
```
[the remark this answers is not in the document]
JuliaSprenger: Yes, that is probably easier to maintain. I added a utility method that is now used whenever a original value needs conversion instead of the precalculation.

neo/rawio/axonarawio.py
```diff
@@ -141,10 +225,24 @@ def _segment_t_start(self, block_index, seg_index):
...
+        if 'num_total_packets' in self.file_parameters['bin']:
+            t_stop = self.file_parameters['bin']['num_total_samples'] / \
+                     self.file_parameters['bin']['sampling_rate']
+
+        if 'unit' in self.file_parameters:
+            # get tetrode recording durations in seconds
+            tetrode_durations = [
+                int(self.file_parameters['unit'][i]['duration'])
+                for i in range(1, 33) if i in
+                self.file_parameters['unit']]
```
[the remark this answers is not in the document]
JuliaSprenger: I simplified this by propagating common tetrode parameters to global unit level. If the tetrode durations are not consistent across files this value won't be propagated and will raise an error. @sbuergers Or are inconsistent tetrode durations to be expected in some datasets?

neo/rawio/axonarawio.py
```diff
@@ -197,50 +298,170 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
+            n_pad = 4 - waveforms.shape[0] % 4
+            pad = np.zeros((n_pad, waveforms.shape[1]))
+            waveforms = np.vstack((waveforms, pad))
+
+        # waveforms must be a 3D numpy array (nb_spike, nb_channel, nb_sample)
+        waveforms = waveforms.reshape(nb_spikes, 4, nb_samples_per_waveform)
+
+        return waveforms
+
+    def get_header_lenth(self, file):
```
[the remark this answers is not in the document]
JuliaSprenger: Thanks good catch! I removed this method in the process of refactoring.

neo/rawio/axonarawio.py
```diff
@@ -250,7 +471,7 @@ def get_active_tetrode(self):
         """
         active_tetrodes = []
 
-        with open(self.set_file, encoding=self.set_file_encoding) as f:
+        with open(self.set_file, encoding='cp1252') as f:
```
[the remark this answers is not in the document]
JuliaSprenger: I also refactored this. So now there is only a single method for reading headers and the header content is available via  `self.file_parameters[<file_type>]['file_header']`

neo/rawio/axonarawio.py
```diff
@@ -83,55 +100,122 @@ def _parse_header(self):
...
+
+            # propagate common tetrode parameters to global unit level
+            units_dict = self.file_parameters['unit']
+            ids = units_dict['tetrode_ids']
+            if ids:
+                for key, value in units_dict[ids[0]].items():
+                    # copy key-value pair if present across all tetrodes
+                    if all([key in units_dict[t] for t in ids]) and \
+                            all([units_dict[t][key] == value for t in ids]):
+                        self.file_parameters['unit'][key] = value
```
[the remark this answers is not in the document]
JuliaSprenger: Yes, that will make it easier to keep an overview.

neo/rawio/axonarawio.py
```diff
@@ -83,55 +100,122 @@ def _parse_header(self):
...
+                          'bytes_head': 32,
+                          'bytes_tail': 16,
+                          'data_type': np.int16,
+                          'header_size': 0,
+                          'header_encoding': 'cp1252',
+                          'sampling_rate': int(global_params['rawRate']),
+                          'num_channels': len(self.get_active_tetrode()) * 4},
+                  'set': {'header_encoding': 'cp1252'},
+                  'unit': {'data_type': unit_dtype,
+                           'wf_left_sweep_us': 200,
```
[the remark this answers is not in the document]
JuliaSprenger: Thanks for the hint. I am now using the information from the `set` file for the `wf_left_sweep`. Regarding the inconsistency in the other dataset there is no other value in the header that we forgot to take into account? Ie. a scaling factor that explains the difference?

JuliaSprenger:
@samuelgarcia Do you have any comments for this PR? I already rebased to include the latest changes in the test suite.

neo/rawio/axonarawio.py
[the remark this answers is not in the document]
alejoe91: Ok for me to keep it as is then 🙂

neo/rawio/axonarawio.py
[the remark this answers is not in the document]
sbuergers: None that I found thus far, but it's something to investigate. When I was last looking at this I abandoned my investigation to work more on milestone 1.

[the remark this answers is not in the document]
JuliaSprenger: Ok, then I would stick to the usage of `pretrigSamps` for this PR and potentially improve this in another PR if we find a more satisfying solution.

neo/rawio/axonarawio.py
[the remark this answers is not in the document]
sbuergers: No, I don't think tetrodes will have different durations, at least not in the conversion code from the Hussaini-lab. So I'd say this is fine.

JuliaSprenger:
I added more testfiles to gin (https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/54). @samuelgarcia can you check these and merge both PRs if you are happy with the current state?

JuliaSprenger:
Hi @samuelgarcia it seems like the new test data files are not automatically downloaded from gin. Can we do a manual reset of the cache?

JuliaSprenger:
I observed the same effect locally. The new files were not automatically downloaded and I had to delete the complete `ephys_testing_data` folder to trigger a new download.

samuelgarcia:
STrange. I will have a look.

sbuergers (changes requested):
I think with this patch it should be good now (besides the gin issues mentioned above):

* [ ] https://github.com/JuliaSprenger/python-neo/pull/1

JuliaSprenger:
@samuelgarcia Any news regarding the download of the new test data?

JuliaSprenger:
@samuelgarcia I cherry-picked your datalad fix also in this branch, so now all files are downloaded properly and tests pass.
Ready for final review / merge from my side.

JuliaSprenger:
Hi @samuelgarcia with the catching of the testfiles fixed and this PR being in a completish state for some time already it's up to your final approval for this to be merged!

neo/rawio/axonarawio.py
```diff
@@ -81,57 +98,149 @@ def _parse_header(self):
...
+            self._raw_signals = np.memmap(
+                self.bin_file, dtype=self.file_parameters['bin']['data_type'],
+                mode='r', offset=self.file_parameters['bin']['header_size']
+            )
+
+            signal_streams = self._get_signal_streams_header()
+            signal_channels = self._get_signal_chan_header()
+
+        # SCAN TETRODE FILES
+        # In this IO one tetrode corresponds to one unit
```
samuelgarcia, line 148: I don't anderstand this.
Do you mean that spike here are unsorted so all spikes per tetrode belong to the same units ?

neo/rawio/axonarawio.py
```diff
@@ -81,57 +98,149 @@ def _parse_header(self):
...
 
         # Annotations
         self._generate_minimal_annotations()
 
+        # Adding custom annotations
         bl_ann = self.raw_annotations['blocks'][0]
         seg_ann = bl_ann['segments'][0]
         seg_ann['rec_datetime'] = self.read_datetime()
-        sig_an = \
+        if seg_ann['signals']:
```
samuelgarcia, line 234: maybe if len(seg_ann['signals']): is more easy to understand

neo/rawio/axonarawio.py
```diff
@@ -81,57 +98,149 @@ def _parse_header(self):
...
             seg_ann['signals'][0]['__array_annotations__']['tetrode_id'] = \
-            [tetr for tetr in self.get_active_tetrode() for _ in range(4)]
+                [tetr for tetr in self.get_active_tetrode() for _ in range(4)]
+
+        if seg_ann['spikes']:
```
samuelgarcia, line 238: Same here

neo/rawio/axonarawio.py
```diff
@@ -81,57 +98,149 @@ def _parse_header(self):
...
+                    self.get_header_bstring(tetrode_file))
+
+                # memory mapping spiking data
+                spikes = np.memmap(
+                    tetrode_file,
+                    dtype=self.file_parameters['unit']['data_type'],
+                    mode='r', offset=tdict['header_size'],
+                    shape=(tdict['num_spikes'] * 4))
+                self._raw_spikes.append(spikes)
+
```
samuelgarcia, line 169: This dtype for spikes is amazing, do you mean that the spike time is duplicated on the 4 channels ? !!!

neo/rawio/axonarawio.py
```diff
@@ -81,57 +98,149 @@ def _parse_header(self):
...
+                tdict['num_spikes'] = int(tdict['num_spikes'])
+                tdict['header_size'] = len(
+                    self.get_header_bstring(tetrode_file))
+
+                # memory mapping spiking data
+                spikes = np.memmap(
+                    tetrode_file,
+                    dtype=self.file_parameters['unit']['data_type'],
+                    mode='r', offset=tdict['header_size'],
+                    shape=(tdict['num_spikes'] * 4))
```
samuelgarcia, line 167: I think i would be more easy to handle this as a 2D array and so use `shape=(tdict['num_spikes'] , 4)`

neo/rawio/axonarawio.py
```diff
@@ -197,51 +327,165 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
 
         return raw_signals
 
+    def _spike_count(self, block_index, seg_index, unit_index):
+        tetrode_id = unit_index
+
+        raw_spikes = self._raw_spikes[tetrode_id]
+        nb_tetrode_spikes = raw_spikes.shape[0]
+        # also take into account last, potentially incomplete set of spikes
+        nb_unit_spikes = int(np.ceil(nb_tetrode_spikes / 4))
```
samuelgarcia, line 345: If `raw_spikes` would be 2d then it would be shape[0] and wowuld avoid this magic //4 number which be a problem when tetrod would become stereotrode for instance.

neo/rawio/axonarawio.py
```diff
@@ -197,51 +327,165 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
+    def _get_spike_timestamps(self, block_index, seg_index, unit_index,
+                              t_start, t_stop):
+        assert block_index == 0
+        assert seg_index == 0
+
+        tetrode_id = unit_index
+        raw_spikes = self._raw_spikes[tetrode_id]
+
+        # spike times are repeated for each contact -> use only first contact
+        unit_spikes = raw_spikes['spiketimes'][::4]
```
samuelgarcia, line 357: Same stuff here about the 2d array of spikes

neo/rawio/axonarawio.py
```diff
@@ -197,51 +327,165 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
+
+    def _get_spike_timestamps(self, block_index, seg_index, unit_index,
+                              t_start, t_stop):
+        assert block_index == 0
+        assert seg_index == 0
+
+        tetrode_id = unit_index
+        raw_spikes = self._raw_spikes[tetrode_id]
+
+        # spike times are repeated for each contact -> use only first contact
```
samuelgarcia, line 356: Are we sure that the 4 times are the same on all channels ?

neo/rawio/axonarawio.py
```diff
@@ -197,51 +327,165 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
+    def _rescale_spike_timestamp(self, spike_timestamps, dtype):
+        spike_times = spike_timestamps.astype(dtype)
+        spike_times /= self._to_hz(self.file_parameters['unit']['timebase'])
+        return spike_times
+
+    def _get_spike_raw_waveforms(self, block_index, seg_index, unit_index,
+                                 t_start, t_stop):
+        assert block_index == 0
+        assert seg_index == 0
+
```
samuelgarcia, line 381: Same idea here self._raw_spike should be a 2D array.

neo/rawio/axonarawio.py
```diff
@@ -336,13 +579,18 @@ def _get_signal_chan_header(self):
...
-                sig_channels.append((ch_name, chan_id, self.sr, dtype,
+                # the sampling rate information is stored in the set header
+                # and not in the bin file
+                sr = self.file_parameters['set']['sampling_rate']
+                sig_channels.append((ch_name, chan_id, sr, dtype,
                                      units, gain, offset, stream_id))
 
         return np.array(sig_channels, dtype=_signal_channel_dtype)
+
+    def _to_hz(self, param, dtype=int):
```
samuelgarcia, line 595: What is the `dtype=int` for ?

samuelgarcia:
Hi Julia and Steffen.
This is looks good.
I proposed a small refactoring using 2D array for spikes.
I think i will be more easy to  maintain with this.
If you don't have time don't worry.

neo/rawio/axonarawio.py
sbuergers: Per tetrode yes. They should always be the same for each channel.

neo/rawio/axonarawio.py
sbuergers: When reading data in `.X` format (tetrode files) that were generated from the raw recording (recommended default case according to the Axona manual), then yes, the units are unsorted. Or rather, the only thing that happened is that a threshold was applied to the raw data to find 'spikes'. Then only data surrounding those spikes was written to files and all data discarded. Each spike timestamp is the same for all channels on a given tetrode. There is also a mechanism for validating these spikes - they should not be too close together, but I am not sure of the details there from the top of my head. Since this is something that the Axona software does as a standard processing step it makes to assume this here.

However, we are now also at the point where we can export spike sorted data (from spikeinterface) to the same `.X` format, so it makes sense to check if that can still be read in with this rawio.

neo/rawio/axonarawio.py
sbuergers: Yes.

neo/rawio/axonarawio.py
sbuergers: I like the idea of making it 2d. In theory stereotrode data should still have 4 channels:

```
Tetrode and stereotrode data is almost identical, in that both store 4-channels worth of 
data per spike. The only difference is that in stereotrode mode, either channels 3&4 or 
1&2 will be all zeroes, corresponding to a spike on stereotrode 1 or 2, respectively (the 
stereotrode format is very wasteful of file space).
 ```

sbuergers:
Hi Sam,

thanks for the review and suggestions! I like the 2d refactor idea, but I am already over my hours for this project, so would prefer to finalize it as is and leave sensible, but not necessarily required improvements for the future.

neo/rawio/axonarawio.py
sbuergers: We should change this to float!

neo/rawio/axonarawio.py
JuliaSprenger: I extended the comment to add the note about the tetrode not being sorted yet at that stage.

neo/rawio/axonarawio.py
JuliaSprenger: I found the original more intuitive so the current waveform handling is based on the other layout. Unless there's performance reasons I would just keep the current implementation...

neo/rawio/axonarawio.py
JuliaSprenger: changed to 2D implementation

neo/rawio/axonarawio.py
JuliaSprenger: changed to 2D implementation

neo/rawio/axonarawio.py
JuliaSprenger: changed to 2D implementation

neo/rawio/axonarawio.py
JuliaSprenger: changed to 2D implementation

neo/rawio/axonarawio.py
JuliaSprenger: changed to 2D implementation

neo/rawio/axonarawio.py
JuliaSprenger: I changed the default dtype and used the keyword int at the places where an integer was used earlier

neo/rawio/axonarawio.py
JuliaSprenger: fixed

neo/rawio/axonarawio.py
JuliaSprenger: fixed

JuliaSprenger:
I had to rebase on master to resolve conflicts with base, but now all comments are addressed. @samuelgarcia You can either provide more feedback or merge.

samuelgarcia:
This looks good.
Do you want I merge now ?
What conflit you want to fix ?

JuliaSprenger:
I am ready to merge. @samuelgarcia go ahead ;)

PR #990, opened 2021-05-22
[neuralynx] first draft for memory and speed improvements

JuliaSprenger:


PeterNSteinmetz:
Please also see PR #973 which eliminated the creation of an object in the inner loop.

PeterNSteinmetz (commented):
The big question is whether this passes all the existing tests of course. 

On line 288 I don't think I would compare to nb_valid for the first record. Rather I would compare to the maximum number of samples for record, which is 512, and I believe is a defined constant. There is no guarantee that the first record is full, though this is usually the case.

On line 292, should this comparison be < rather than <=. It seems like = would be an exact match to what is expected and the timestamps in the records are rounded to the nearest microsecond.

If one is going to call this 'delayed_recs' it might be best to explain why the code looks for < rather than simply != . It seems like it should never be greater than as that would imply that a sample occurred before the next sampling interval. 

line 343 my sense is that this estimation of the frequency is correct, but the proof is in the tests

CodyCBakerPhD:
This current state of this appears to fix the memory issue for me!

JuliaSprenger:
Ready for review!
The current version still has very different performance in terms of initialization time: on windows this seems 10x slower than on linux. This should be adressed in another PR.

@PeterNSteinmetz Thanks for your earlier comments. I improved the code accordingly.

Issue #979, opened 2021-04-23
many warnings in our test


samuelgarcia:
When I set tests on github actions saw many warnings.
We should fix then before any problems occurs.

samuelgarcia:
See this https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

We have several choice to replace np.int:
  1. int
  2. np.int_
  3. np.int32 or np.int64 depending the context and wanted precision.

Choice 1. et 2. don't change the behavior : the precission depend on the platform.

I would prefer choice 3. and fix the precision in the code for reader (IO implementation)
In many situtaion np.int64 is  needed and it is better to explicit.

In our test I will left np.int_ and np.float_ depending the context.

@apdavison @JuliaSprenger @mdenker ?

JuliaSprenger:
Hi @samuelgarcia.
Thanks for looking into the warning, the output is indeed very verbose currently.
Regarding your preferred choice option 3, we should be careful not not use np.int64 in places where it unnecessarily drastically increases memory usage. So a reader specific choice is also the best option in my mind.

Regarding options 1 and 2: Why would you go for the more verbose option of `np.int_` in the tests if `int` would do the same job? In terms of readability `int` should be better, no?

samuelgarcia:
Hi Julia,
I propose you have a look to the PR and make comments case by case because it depends.
you will see #991

samuelgarcia:
close by #991 and #1010

PR #991, opened 2021-05-26
Remove some np.int and np.float because of numpy 0.20.0

samuelgarcia:
#979 

See https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated


neo/core/analogsignal.py
```diff
@@ -448,7 +448,7 @@ def _pp(line):
     def time_index(self, t):
         """Return the array index (or indices) corresponding to the time (or times) `t`"""
         i = (t - self.t_start) * self.sampling_rate
-        i = np.rint(i.simplified.magnitude).astype(np.int)
+        i = np.rint(i.simplified.magnitude).astype(np.int64)
```
JuliaSprenger, line 451: @samuelgarcia in case just `int` is provided as type here, wouldn't this automatically pick the suitable integer length that fits the data?

JuliaSprenger (approved):

neo/core/analogsignal.py
samuelgarcia: Handling time with int32 is super dangerous.
I think forcing to int64 is a good choice.

JuliaSprenger: Ok, then let's go for int64 for AnalogSignals times for now. If this becomes a memory issue in the future we might reconsider this choice.

PR #993, opened 2021-05-31
Fix/allow missing bin file (axonarawio)

sbuergers:
This is actually the same as https://github.com/NeuralEnsemble/python-neo/pull/986, 

but with the additional fixes https://github.com/JuliaSprenger/python-neo/pull/4 and https://github.com/JuliaSprenger/python-neo/pull/3. 

Since Julia is on leave for the next few weeks, can we merge this one in place of https://github.com/NeuralEnsemble/python-neo/pull/986? 

@samuelgarcia I am not sure if this is the right way to overtake her PR, happy for any feedback. :-)


sbuergers:
Oh, actually trying to accomodate the situation of not having a `.bin` file, which contains the raw data traces leads to errors when testing the `axonaio` in `test_axonaio`. This is because the test wants to read the raw signal with `get_analogsignal_chunk()`, but no such signal exists. 

What would be the preferred solution for this? Change the `test_axonaio` to deal with missing `.bin` data?

JuliaSprenger:
Hi @sbuergers. I had a look and left some comments for the changes in https://github.com/JuliaSprenger/python-neo/pull/4 and https://github.com/JuliaSprenger/python-neo/pull/3 to keep the different issues separate.

sbuergers:
I am closing this PR in favor of https://github.com/NeuralEnsemble/python-neo/pull/986

PR #997, opened 2021-06-08
[phyio] Fixed spiketrain annotations

rjurkus:
This PR fixes some issues with PhyIO. 
Changes:
* Numeric spiketrain annotations are int or float (used to be str)
* Added `cluster_id` annotation (corresponds to id annotation, but is int, not str)
* PEP 8 fixes

### DIFF
--- a/neo/rawio/phyrawio.py
+++ b/neo/rawio/phyrawio.py
@@ -9,7 +9,7 @@ Author: Regimantas Jurkus
 """
 
 from .baserawio import (BaseRawIO, _signal_channel_dtype, _signal_stream_dtype,
-                _spike_channel_dtype, _event_channel_dtype)
+                        _spike_channel_dtype, _event_channel_dtype)
 
 import numpy as np
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


neo/rawio/phyrawio.py
```diff
@@ -132,6 +133,9 @@ def _parse_header(self):
         for index, clust_id in enumerate(clust_ids):
             spiketrain_an = seg_ann['spikes'][index]
 
+            # Add cluster_id annotation
+            spiketrain_an['cluster_id'] = clust_id
+
```
samuelgarcia, line 138: I am OK with this but we already have have this information in `self.header['spike_channels']['id']`  no ?
Maybe the SpikeTrainProxy object is not transmiting this info correctly in  basefromrawio.

samuelgarcia:
Hi Regi.
Now the `ìd` for signal/event/spike channel are handled with str in rawio.
Here you force a custum annotation to be int and having a special name `cluster_id`.

I think we should enhance the SpikeTrainProxy to retrieve the original neo `id`.

If you really need this cluster_id maybe you should rename it to phy_cluster_id to show that this annoation "cluster_id" is specific to phyio.

neo/rawio/phyrawio.py
rjurkus: Just to be clear, the id in `unit_name` and the `id` annotation is the same as `cluster_id`, but the pair is `str` and the 'cluster_id' is `int`. They indicate the id of the cluster and _are not necessarily sequential_ (e.g. `id` 56, 67, 87 are missing because those ids point to empty clusters).

PR #998, opened 2021-06-09
[spikeglxrawio] .meta file parsing fix

rjurkus:
This PR fixes .meta line parsing using the solution from the [Probe Interface project](
https://github.com/SpikeInterface/probeinterface/blob/19d6518fbc67daca71aba5e99d8aa0d445b75eb7/probeinterface/io.py#L649-L662). I also make parsing of .meta conditional, since it should only be done when .bin with a matching name is present.


samuelgarcia:
Hi Regi.
Thanks for this.

PR #1000, opened 2021-06-11
Implement SpikeTrainList class, and make Segment.spiketrains an instance of this class

apdavison:
Please note that the internal implementation can probably be improved/optimised, but that is likely to need some profiling.

The general idea is to create SpikeTrain objects only when necessary, and to provide the "multiplexed" property to allow treating spike trains directly as (spike_time_array, channel_id_array)


apdavison:
tests all passing now

JuliaSprenger (approved):

JuliaSprenger:
@apdavison Thanks for the new feature worthy of the 1000th PR / issue :)
I think SpikeTrainLists can be very usefull when dealing with multiplexed data, however with the current implementation any full neo structure (including segments & groups) will only use the original item-based SpikeTrain representation, as both parent objects contain links to different spiketrains. Is this correct? I am not sure if we can improve this somehow to still benefit from the multiplexing also in this scenario. Do you have any ideas for this?

apdavison:
@JuliaSprenger that's a good point. We could perhaps generalise ChannelView to index specific spike trains within the SpikeTrainList.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+                dim = self._items[0].units.dimensionality
+                for i, spiketrain in enumerate(self._items):
+                    if hasattr(spiketrain, "load"):  # proxy object
+                        spiketrain = spiketrain.load()
+                    if spiketrain.times.dimensionality.items() == dim.items():
+                        # no need to rescale
+                        spike_times.append(spiketrain.times)
+                    else:
+                        spike_times.append(spiketrain.times.rescale(dim))
+                    channel_ids.append(i * np.ones(spiketrain.shape))
```
samuelgarcia, line 290: we should force the dtype to int64 (the default is float64)

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+    def multiplexed(self):
+        """Return spike trains as a pair of arrays.
+
+        The first (plain NumPy) array contains the ids of the channels/neurons that produced
+        each spike, the second (Quantity) array contains the times of the spikes.
+        """
+        if self._spike_time_array is None:
+            # need to convert list of SpikeTrains into multiplexed spike times array
+            if self._items is None:
+                return np.array([]), np.array([])
```
samuelgarcia, line 277: I think we should not allow self._spike_time_array is None and self._items is None.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+        ...     SpikeTrain([], units="ms", t_start=0 * pq.ms, t_stop=100.0 * pq.ms),
+        ... ))
+        >>> stl.multiplexed
+        (array([0., 0., 0., 0., 1., 1., 2., 2.]),
+         array([ 0.5,  0.6, 23.6, 99.2,  0.7, 11.2,  1.1, 88.5]) * ms)
+
+        # Create from a pair of arrays
+
+        >>> stl = SpikeTrainList.from_spike_time_array(
+        ...     np.array([0.5, 0.6, 0.7, 1.1, 11.2, 23.6, 88.5, 99.2]),
```
samuelgarcia, line 46: label as float is a bit strange. Maube we should force to int, no ?
The boolean mask will be faster and unambiguous.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+                # todo: adjust times and t_start of other to be compatible with self
+            if self._spiketrain_metadata['t_stop'] != other._spiketrain_metadata['t_stop']:
+                raise ValueError("Incompatible t_stop")
+                # todo: adjust t_stop of self and other as necessary
+            combined_spike_time_array = np.hstack(
+                (self._spike_time_array, other._spike_time_array))
+            combined_channel_id_array = np.hstack(
+                (self._channel_id_array, other._channel_id_array))
+            combined_channel_ids = set(list(self._all_channel_ids) + other._all_channel_ids)
+            if (len(combined_channel_ids) !=
```
samuelgarcia, line 145: maybe we should add a `remap_ids` option to automatocally remap ids from the second spiketrainlist

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+            self._items.extend(other)
+            return self
+        elif len(other) == 0:
+            return self
+        else:
+            if self._items is None:
+                self._spiketrains_from_array()
+            return other + self._items
+
+    def append(self, obj):
```
samuelgarcia, line 207: If the inetrnal state is multiplexed then this implementation cost a lot.
Maybe we could add an option to transform other to spiketrainlist+remapp ids+concatenate.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+        """L.append(object) -> None -- append object to end"""
+        if not is_spiketrain_or_proxy(obj):
+            raise ValueError("Can only append SpikeTrain objects")
+        if self._items is None:
+            self._spiketrains_from_array()
+        obj.segment = self.segment
+        self._items.append(obj)
+
+    def extend(self, iterable):
```
samuelgarcia, line 216: Same comment than append()

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+        """L.extend(iterable) -> None -- extend list by appending elements from the iterable"""
+        if self._items is None:
+            self._spiketrains_from_array()
+        for obj in iterable:
+            obj.segment = self.segment
+        self._items.extend(iterable)
+
+    @classmethod
+    def from_spike_time_array(cls, spike_time_array, channel_id_array,
+                              all_channel_ids=None, units=None,
```
samuelgarcia, line 226: I think that all_channel_ids should be a necessary arguments.
Making `unique()` is dangerous when dealing with smal chunks a  neuron can have no spike in one segment.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+        for name, ann_value in annotations.items():
+            if len(ann_value) != len(obj):
+                raise ValueError(f"incorrect length for annotation '{name}'")
+        obj._annotations = annotations
+        return obj
+
+    def _spiketrains_from_array(self):
+        """Convert multiplexed spike time data into a list of SpikeTrain objects"""
+        if self._spike_time_array is None:
+            self._items = []
```
samuelgarcia, line 248: I think we should not allow self._spike_time_array is None and self._items is None.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+        else:
+            if self._all_channel_ids is None:
+                all_channel_ids = np.unique(self._channel_id_array)
```
samuelgarcia, line 251: even it is tedious I think we should avoid this magic unique()

neo/core/spiketrain.py
```diff
@@ -110,6 +110,56 @@ def _new_spiketrain(cls, signal, t_stop, units=None, dtype=None, copy=True,
     return obj
 
 
+def normalise_times_array(times, units, dtype=None, copy=True):
```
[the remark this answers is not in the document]
apdavison: done

neo/core/spiketrain.py
```diff
@@ -110,6 +110,56 @@ def _new_spiketrain(cls, signal, t_stop, units=None, dtype=None, copy=True,
...
+    D. times (NumPy array), units=None
+
+    In scenarios A-C we return a tuple (times as a Quantity array, dimensionality)
+    In scenario C, we rescale the original array to match `units`
+    In scenario D, we raise a ValueError
+    """
+    if dtype is None:
+        if not hasattr(times, 'dtype'):
+            dtype = np.float
+    if units is None:
```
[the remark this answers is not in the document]
apdavison: done

neo/core/spiketrain.py
```diff
@@ -110,6 +110,56 @@ def _new_spiketrain(cls, signal, t_stop, units=None, dtype=None, copy=True,
...
+        # No keyword units, so get from `times`
+        try:
+            dim = times.units.dimensionality
+        except AttributeError:
+            raise ValueError('you must specify units')
+    else:
+        if hasattr(units, 'dimensionality'):
+            dim = units.dimensionality
+        else:
+            dim = pq.quantity.validate_dimensionality(units)
```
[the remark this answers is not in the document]
apdavison: we've been using this for long time in Neo, so I think it's pretty stable.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+                raise ValueError("Incompatible t_stop")
+                # todo: adjust t_stop of self and other as necessary
+            combined_spike_time_array = np.hstack(
+                (self._spike_time_array, other._spike_time_array))
+            combined_channel_id_array = np.hstack(
+                (self._channel_id_array, other._channel_id_array))
+            combined_channel_ids = set(list(self._all_channel_ids) + other._all_channel_ids)
+            if (len(combined_channel_ids) !=
+                    len(self._all_channel_ids) + len(other._all_channel_ids)):
+                raise ValueError("Duplicate channel ids, please rename channels before adding")
```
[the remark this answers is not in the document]
apdavison: I think it's safest to ask the user to handle this manually by renaming channels explicitly.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+            if in_place:
+                self._items.extend(other._items)
+                return self
+            else:
+                return self.__class__(items=self._items[:] + other._items)
+        else:
+            # both self and other are storing multiplexed spike trains
+            if self._spiketrain_metadata['t_start'] != other._spiketrain_metadata['t_start']:
+                raise ValueError("Incompatible t_start")
+                # todo: adjust times and t_start of other to be compatible with self
```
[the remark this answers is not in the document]
apdavison: I think a later one.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+            self._items = []
+            for i, channel_id in enumerate(all_channel_ids):
+                mask = self._channel_id_array == channel_id
+                times = self._spike_time_array[mask]
+                spiketrain = SpikeTrain(times, **self._spiketrain_metadata)
+                spiketrain.annotations = {
+                    name: value[i]
+                    for name, value in self._annotations.items()
+                }
+                # todo: consider adding channel id as metadata
```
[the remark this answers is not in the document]
apdavison: reserved names in annotations is something to consider, but should probably be discussed in a separate issue.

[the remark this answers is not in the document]
apdavison: for now, just adding it as a plain annotation "channel_id"

neo/core/spiketrainlist.py
apdavison: done

neo/core/spiketrainlist.py
apdavison: we need to allow an empty list of spiketrains

neo/core/spiketrainlist.py
apdavison: I think a lot of internal optimisations are possible, but they should be done in separate PRs, with careful profiling

neo/core/spiketrainlist.py
apdavison: agreed. done

neo/core/spiketrainlist.py
apdavison: done.

neo/core/spiketrainlist.py
```diff
@@ -0,0 +1,294 @@
...
+         <SpikeTrain(array([ 0.7, 11.2]) * ms, [0.0 ms, 100.0 ms])>,
+         <SpikeTrain(array([ 1.1, 88.5]) * ms, [0.0 ms, 100.0 ms])>,
+         <SpikeTrain(array([], dtype=float64) * ms, [0.0 ms, 100.0 ms])>]
+
+    """
+
+    def __init__(self, items=None, segment=None):
+        """Initialize self"""
+        # todo: handle annotations
+        # todo: handle waveforms
```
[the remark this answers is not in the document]
apdavison: I actually don't remember what I meant by those todos.

apdavison:
if the tests all pass, this is ready to merge, I think

neo/core/spiketrainlist.py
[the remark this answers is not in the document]
JuliaSprenger: I don't see where the `channel_id` annotation is added, as it doesn't seem to be part of `self._spiketrain_metadata[i]`...

neo/core/spiketrainlist.py
JuliaSprenger: I also think this is necessarly, as also `block.segments[0].spiketrains` can be an empty list. @samuelgarcia or what did you have in mind?

neo/core/spiketrainlist.py
JuliaSprenger: so the difference is that `channel_id_array` has the same dimension as `spike_time_array` and `all_channel_ids` has to contain all unique entries present in `channel_id_array` plus some additional ones if a neuron didn't spike? Maybe it would be good to add an argument description in the docstring to avoid confusion....

neo/core/spiketrainlist.py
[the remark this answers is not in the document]
apdavison: [here](https://github.com/NeuralEnsemble/python-neo/pull/1000/files#diff-2aee3c2241ed9bda08d91f3a7aade21be02bdbc246949786a28f1a1f941e7036R258)

neo/core/spiketrainlist.py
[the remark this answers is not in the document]
JuliaSprenger: I opened an issue #1002

neo/core/spiketrainlist.py
JuliaSprenger: I think this has been solved by removal, @samuelgarcia correct?

neo/core/spiketrainlist.py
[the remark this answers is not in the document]
JuliaSprenger: ok, maybe we have to update this at another timepoint if it proves very inconvenient...

neo/core/spiketrainlist.py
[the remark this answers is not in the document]
JuliaSprenger: Yes, thanks!

neo/core/spiketrainlist.py
JuliaSprenger: that's a convenience behavior to be added later?

neo/core/spiketrainlist.py
JuliaSprenger: We need some equivalent to an empty spiketrain list. So would you like to add the default values or do you want to not allow an empty `spiketrainlist` object?

neo/core/spiketrainlist.py
```diff
@@ -222,10 +223,36 @@ def extend(self, iterable):
...
+                :class:`~quantities.Quantity`.
+        :t_stop: (quantity scalar, numpy scalar, or float) Time at which
+            spike recording ended. This will be converted to the
+            same units as :attr:`spike_time_array` or :attr:`units`.
+
+        *Recommended attributes/properties*:
+        :t_start: (quantity scalar, numpy scalar, or float) Time at which
+            spike recording began. This will be converted to the
+            same units as :attr:`spike_time_array` or :attr:`units`.
+            Default: 0.0 seconds.
```
JuliaSprenger, line 251: Would it make sense to make this default explicit and set is already in the method header?

neo/core/spiketrainlist.py
```diff
@@ -222,10 +223,36 @@ def extend(self, iterable):
...
+        and an array of channel ids.
+
+        *Required attributes/properties*:
+
+        :spike_time_array: (quantity array 1D, numpy array 1D, or list) The times of
+            all spikes.
+        :channel_id_array: (numpy array 1D, dtype int) The id of the channel (e.g. the
+            neuron) to which each spike belongs. This array should have the same length
+            as :attr:`spike_time_array`
+        :all_channel_ids`: (list, tuple, or numpy array 1D containing integers) All
```
JuliaSprenger, line 238:
```suggestion
        :all_channel_ids: (list, tuple, or numpy array 1D containing integers) All
```

neo/core/spiketrainlist.py
```diff
@@ -222,10 +223,36 @@ def extend(self, iterable):
...
+                              t_start=None, **annotations):
         """Create a SpikeTrainList object from an array of spike times
-        and an array of channel ids."""
+        and an array of channel ids.
+
+        *Required attributes/properties*:
+
+        :spike_time_array: (quantity array 1D, numpy array 1D, or list) The times of
+            all spikes.
+        :channel_id_array: (numpy array 1D, dtype int) The id of the channel (e.g. the
```
JuliaSprenger, line 235:
```suggestion
        :channel_id_array: (numpy array 1D of dtype int) The id of the channel (e.g. the
```

neo/core/spiketrainlist.py
JuliaSprenger: Clarification: In case of non-empty channel_id overlap the user should have the option to provide a remapping dictionary.

JuliaSprenger: This will be added later.

neo/core/spiketrainlist.py
JuliaSprenger: This performance improvement will be taken care of in another PR.

PR #944, opened 2021-02-10
Adding Support for Neuralynx Header using ``Date''

fccaa:
In the data set I am working with, I found some nev file having a header like this:

\######## Neuralynx Data File Header
\## File Name: V:\not_to_show.nev
\## Date Opened: (mm/dd/yyy): 07/04/2015 At Time: 17:01:39
\## Date Closed: (mm/dd/yyy): 07/04/2015 At Time: 17:01:39

However, the version in the official repository supports only "## Time Opened". After making the change, I can open the directory without the error reported in https://github.com/SpikeInterface/spikeextractors/issues/480 .

PR #1005, opened 2021-06-29
[neuralynx] add support for neuraview file header

JuliaSprenger:
This is an alternative implementation to #944 that introduced neuraview file header interpretation and a corresponding first simple unit test. @fccaa Can you confirm this version is working for your files?


fccaa:
> This is an alternative implementation to #944 that introduced neuraview file header interpretation and a corresponding first simple unit test. @fccaa Can you confirm this version is working for your files?

Hi. The version here behaves the same as #944 . I have no error message to open my files. I think this version is working.

JuliaSprenger:
@fccaa Thanks for checking. I would prefer to have the new header format properly embedded in the existing header routines. I hope you don't mind my alternative implementation. Thanks again for pointing this out and taking the initiative.
@samuelgarcia Ready for review from your side.

samuelgarcia:
OK for me.

PR #1009, opened 2021-07-01
Signal array annotations in NixIOfr

JuliaSprenger:
This PR adds support for annotations and array annotations for signals on NixIOfr level.

### DIFF
--- a/neo/rawio/nixrawio.py
+++ b/neo/rawio/nixrawio.py
@@ -52,7 +52,7 @@ class NIXRawIO(BaseRawIO):
     def _parse_header(self):
         self.file = nix.File.open(self.filename, nix.FileMode.ReadOnly)
         signal_channels = []
-        anasig_ids = {0: []} # ids of analogsignals by segment
+        anasig_ids = {0: []}  # ids of analogsignals by segment
         stream_ids = []
         for bl in self.file.blocks:
             for seg in bl.groups:
--- a/neo/rawio/nixrawio.py
+++ b/neo/rawio/nixrawio.py
@@ -4,15 +4,17 @@ RawIO Class for NIX files
 The RawIO assumes all segments and all blocks have the same structure.
 It supports all kinds of NEO objects.
 
-Author: Chek Yin Choi
+Author: Chek Yin Choi, Julia Sprenger
 """
 
-from .baserawio import (BaseRawIO, _signal_channel_dtype, _signal_stream_dtype,
-                _spike_channel_dtype, _event_channel_dtype)
+import os.path
 
-from ..io.nixio import NixIO
-from ..io.nixio import check_nix_version
 import numpy as np
+
+from .baserawio import (BaseRawIO, _signal_channel_dtype, _signal_stream_dtype,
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
@samuelgarcia What do you think of this?

neo/rawio/nixrawio.py
```diff
@@ -242,6 +228,36 @@ def _parse_header(self):
...
+                    props = [p for p in da.metadata.props
+                             if p.type != 'ARRAYANNOTATION']
+                    props_dict = self._filter_properties(props, "analogsignal")
+                    sig_ann[stream_id].update(props_dict)
+
+                    # collect array annotation properties
+                    props = [p for p in da.metadata.props
+                             if p.type == 'ARRAYANNOTATION']
+                    props_dict = self._filter_properties(props, "analogsignal")
+                    sig_ann[stream_id]['__array_annotations__'].update(
```
samuelgarcia, line 256: I think anderstand this.

I think I don't anderstand this loop because I don't anderstand the underlying struture of nix.

`sig_ann[stream_id]['__array_annotations__'][key]` is supposed to be an array with size of channel count but here it is update insid ethe channel loop !!!

neo/test/iotest/test_nixio_fr.py
```diff
@@ -123,14 +124,16 @@ def test_annotations(self):
...
         with NixIOfr(filename=self.testfilename) as frio:
             frbl = frio.read_block()
             assert 'my_custom_annotation' in frbl.annotations
             assert 'something' in frbl.segments[0].annotations
-            # assert 'ansigrandom' in frbl.segments[0].analogsignals[0].annotations
+            assert 'ansigrandom' in frbl.segments[0].analogsignals[0].annotations
             assert 'railway' in frbl.segments[0].spiketrains[0].annotations
             assert 'venue' in frbl.segments[0].events[0].annotations
             assert 'evven' in frbl.segments[0].events[1].annotations
+            assert 'custom_id' in frbl.segments[0].analogsignals[0].array_annotations
```
samuelgarcia, line 136: Could you also check for the size of of the value which must be channel count ?

neo/rawio/nixrawio.py
JuliaSprenger: In the current nix implementation the array_annotations are duplicated and saved for each channel individually. Therefore I only use the first channel of each analogsignal detected to add the array annotations to it.

JuliaSprenger: Once the array annotations were added to an analogsignal it won't be reannotated as it is listed in the `annotated_anasigs`

neo/test/iotest/test_nixio_fr.py
JuliaSprenger: I thought that's part of the array_annotation tests. But I added a couple of lines to also cover this here.

neo/rawio/nixrawio.py
JuliaSprenger: I added a clarification in the comments. Is it now easier to understand?

PR #1012, opened 2021-07-06
[neuralynx] reintroduce spiketime and recording trace tests

### LINKED ISSUE #907: [NeuralynxIO] Tests of keep_original_spike_times failing
The tests in test_neuralyxio.test_keep_original_spike_times are failing for the test data in Cheetah_v5.5.1 on the .nse files. 

It is not clear to me why these have been hidden previously, though that might have to do with the code which only ran the first two blocks of tests.

JuliaSprenger:
fixes #907

### DIFF
--- a/neo/test/iotest/test_neuralynxio.py
+++ b/neo/test/iotest/test_neuralynxio.py
@@ -2,7 +2,7 @@
 Tests of neo.io.neuralynxio.py
 """
 
-import time
+import os
 import warnings
 
 import unittest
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
@PeterNSteinmetz @samuelgarcia What do you think?

PeterNSteinmetz:
The question will be -- do these pass? My recollection is that there is some difference in the format of the test plain text data files in terms of column structure between the difference Neuralynx versions being tested. I believe these files were generated with several different programs.

JuliaSprenger:
The tests pass, so it seems the inconsistency was fixed. I only had to adjust the timestamp column id for nse files.

PeterNSteinmetz:
Then seems good by me.

samuelgarcia:
Hi Julia and Peter.
To be honest, I don't remember about this test removal, so it is OK for me.
Tell me if i can merge.

JuliaSprenger:
Ready to merge :)

Issue #936, opened 2021-01-20
`cut_segment_by_epoch` returns segment with wrong names when using data from NIX files


kohlerca:
Hello,

I am working with data read from a NIX file using NixIO, which also contain epochs for the individual trials. 

However, when I cut the main segment using one epoch, the `nix_name` annotation in the resulting segment is the same as the `Epoch` object used to cut the data.

The code below reproduces the behavior with generated data.

```
import neo
from neo import utils as neo_utils
import numpy as np
import quantities as pq

# Generate a segment with data. Annotations as expected from reading a NIX file

signal = neo.AnalogSignal(np.random.random((90000, 5)), units=pq.uV,
                          sampling_rate=30000*pq.Hz, name="Electrode data",
                          nix_name="neo.analogsignal.ca2b24032c0b4c0497856e24b69268c0",
                          implantation_site="M1")

segment = neo.Segment(name="Segment 0", description="Test segment",
                      nix_name="neo.segment.1dab77a06b734b638f13e9c19d76d6df")

segment.analogsignals = [signal]


# Generate epoch to cut between 1 and 2 s, with annotations

epoch = neo.Epoch([1]*pq.s, [1]*pq.s, name="Epoch 1", description="Test epoch",
                  trial_number=1, trial_protocol=10,
                  nix_name="neo.epoch.da832a0d90c8479abd2247aff4840133")


# Cut the segment using the epoch

trial_segment = neo_utils.cut_segment_by_epoch(segment, epoch)[0]


# Inspect the annotations

print(trial_segment.annotations)
```
The output is

```
{'nix_name': 'neo.epoch.da832a0d90c8479abd2247aff4840133', 'trial_number': 1, 'trial_protocol': 10}
```

One solution would be to store all the available identification attributes of the `Epoch` object (`name`, `description`, `nix_name`) as a `dict` in an annotation `source_epoch`. Or to explicitly ignore the `nix_name` annotations during the cut.

Thank you,
Cristiano

JuliaSprenger:
I @kohlerca,
thanks for raising the issue. That is indeed a tricky one as we need to weigh the user-friendliness of automatically copying the annotations from the source epoch to the segment against keeping the code general and not introducing extra rules just for the `nix_name` annotation.
With `cut_segment_by_epoch` being just a utility and not a core function I would tend to go for the latter and introduce an special case for the `nix_name` annotation here. @samuelgarcia @apdavison Do you have stronger opinions here?

[...]

PR #1016, opened 2021-07-09
[utils] do not copy nix_name annotation to different object types

JuliaSprenger:
solves #936

### DIFF
--- a/neo/test/utils/test_misc.py
+++ b/neo/test/utils/test_misc.py
@@ -3,6 +3,7 @@ Tests of the neo.utils.misc module
 """
 
 import unittest
+import warnings
 
 import numpy as np
 import quantities as pq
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi Julia.
Fixing this here is OK but doing this we inject very specific IO stuf into cor/utils.
I am super happy with this.
And moreover, if a user inject annotation in between object by itself the problem will still be here.

Could it be possible instead to make "annotation cleaner" inside NixIO before writting to remove annotation that have the wrong "type" ?

JuliaSprenger:
> Hi Julia.
> Fixing this here is OK but doing this we inject very specific IO stuf into cor/utils.
> I am super happy with this.
Did you maybe forget a `not` in this sentence?
> And moreover, if a user inject annotation in between object by itself the problem will still be here.
Yes if the user manually messes with the nix specific annotations, then the resulting nix file will be inconsistent, removing annotations of the wrong type wouldn't prevent this...
> 
> Could it be possible instead to make "annotation cleaner" inside NixIO before writting to remove annotation that have the wrong "type" ?
For now I will refactor the cleaning of the annotations into a separate utility function, then maybe when updating the nixio we can incorporate it there each time when writing to disk.

samuelgarcia:
> Did you maybe forget a `not` in this sentence?

Yes as you guess I gorgot the negation....

> For now I will refactor the cleaning of the annotations into a separate utility function, then maybe when updating the nixio we can incorporate it there each time when writing to disk.


Ok.
Shal we merge so ?

JuliaSprenger:
> For now I will refactor the cleaning of the annotations into a separate utility function, then maybe when updating the nixio we can incorporate it there each time when writing to disk.

I did the refactoring in this module. @samuelgarcia Does this look better to you?

samuelgarcia:
This lookfs good.

PR #1018, opened 2021-07-09
Fix Maxwell gain when 'lsb' is not in settings

alejoe91:
In some older files, the `lsb` field might not be present in settings. In that case, the `gain_uV` can be computed as:

`gain_uV = 3.3 / (1024 * gain) * 1e6`

where `gain` is the amplifier gain, 3.3 is the ADC voltage range, and 1024 (10^10), is the number of bits of the ADC.

### DIFF
--- a/neo/rawio/maxwellrawio.py
+++ b/neo/rawio/maxwellrawio.py
@@ -93,14 +93,14 @@ class MaxwellRawIO(BaseRawIO):
                 sr = 20000.
                 settings = h5["settings"]
                 if 'lsb' in settings:
-                    gain_uV = settings['lsb'][:][0] * 1e6
+                    gain_uV = settings['lsb'][0] * 1e6
                 else:
                     if "gain" not in settings:
                         print("'gain' amd 'lsb' not found in settings. "
                               "Setting gain to 512 (default)")
                         gain = 512
                     else:
-                        gain = settings['gain'][:][0]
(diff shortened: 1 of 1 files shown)


neo/rawio/maxwellrawio.py
```diff
@@ -91,7 +91,17 @@ def _parse_header(self):
         for stream_id in signal_streams['id']:
             if int(version) == 20160704:
                 sr = 20000.
-                gain_uV = h5['settings']['lsb'][0] * 1e6
+                settings = h5["settings"]
+                if 'lsb' in settings:
+                    gain_uV = settings['lsb'][:][0] * 1e6
```
JuliaSprenger, line 96: any reason for the additional `[:]`?
```suggestion
                    gain_uV = settings['lsb'][0] * 1e6
```

alejoe91: @JuliaSprenger no, but a difference between the 2 versions was already there (https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/maxwellrawio.py#L103). Should we stick to the `settings['lsb'][0]` notation for all?

JuliaSprenger: If it's working in all cases then I would argue to simplify the code in all places.

JuliaSprenger:
@samuelgarcia Any last comments from your side before we merge?

PR #1021, opened 2021-07-22
Small fix for winwcp files.

samuelgarcia:
Small fix for winwcp files.
The sampling rate can be slightly over segment due to float precision.
The previous implementation  raise error for some random files.

### DIFF
--- a/neo/rawio/winwcprawio.py
+++ b/neo/rawio/winwcprawio.py
@@ -72,9 +72,9 @@ class WinWcpRawIO(BaseRawIO):
 
                 all_sampling_interval.append(analysisHeader['SamplingInterval'])
 
-        assert np.unique(all_sampling_interval).size == 1
-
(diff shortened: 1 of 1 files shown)


samuelgarcia:
@ludo67100 : this fix the problem on winwcp

ludo67100:
@samuelgarcia : thanks a lot !!!!

neo/rawio/winwcprawio.py
```diff
@@ -72,9 +72,9 @@ def _parse_header(self):
 
                 all_sampling_interval.append(analysisHeader['SamplingInterval'])
 
-        assert np.unique(all_sampling_interval).size == 1
-
-        self._sampling_rate = 1. / all_sampling_interval[0]
+        # sampling interval can be slightly due to float precision
```
JuliaSprenger, line 75:
```suggestion
        # sampling interval can be slightly varying due to float precision
```

PR #1026, opened 2021-09-03
Update neo structure

### LINKED ISSUE #1024: remaining `fake_neo` reference
It's still referenced here:

https://github.com/NeuralEnsemble/python-neo/blob/bc6037048fc8d15919d51f2c67a4ab75df40dacc/doc/source/images/generate_diagram.py#L16

JuliaSprenger:
This PR is a suggestion on how to update the neo object linking to include RegionsOfInterest and ChannelView in a consistent manner.
Currently, `Group`s accept also `Segment`s as child objects, which I think is confounding the neo structure. Also `RegionOfInterest` was not considered as part of a neo structure yet. Since these are supposed to basically fulfill a masking functionality for `ImageSequence` their functionality is very similar to the one of `ChannelView`s for `Signal`s. So my suggestion is to have `Group` and `Segment` at the same level and `RegionOfInterest` as well as `ChannelView` as optional level before linking to data object. This results in the following neo dependencies:

This PR also updates the corresponding neo structure diagram (as well as generation script).

@samuelgarcia @apdavison What do you think of this?

Fixes #1024


JuliaSprenger:
The failing test is already fixed in #1025

JuliaSprenger:
I think it would be good to discuss the missing links and pieces in the Neo structure rather sooner than later. @apdavison What is your opinion on the suggestions here? Do you have an alternative organization in mind?

JuliaSprenger:
Here's a summary of the last discussion with @samuelgarcia @apdavison and @mdenker on this topic
- there is no consensus on whether `Group`s should permit `Segment`s as child objects. The strongest argument so far is to keep the object structure as little convoluted as possible for discovery/export/import purposes. Therefore I suggest keeping the current version in which `Group`s only can contain other `Group`s, `DataObject`s or masks of data objects (`RegionOfInterest`, `ChannelView`).
- Another important point to be added (in another PR) would be the (optional) automatic linking of `Spiketrain`s and `Analogsignal` traces upon loading. This link will not be explicitly saved when the block exported, but is on request generated when a block is loaded. We still need to decide on what criteria need to be fulfilled for a link to be generated. Is an identical `channel_index` annotation sufficient?

neo/core/regionofinterest.py
```diff
@@ -1,9 +1,32 @@
 from math import floor, ceil
+from neo.core.container import Container
+from neo.core.imagesequence import ImageSequence
 
 
-class RegionOfInterest:
+class RegionOfInterest(Container):
```
JuliaSprenger, line 6: This should be `BaseNeo`

neo/core/view.py
```diff
@@ -7,11 +7,12 @@
 
 import numpy as np
 from .baseneo import BaseNeo
+from .container import Container
 from .basesignal import BaseSignal
 from .dataobject import ArrayDict
 
 
-class ChannelView(BaseNeo):
```
JuliaSprenger, line 14: to be reverted or inherit from a `BaseView` class

apdavison:
This PR is doing a lot of things. I think we should break it up into several PRs, then we can discuss each of them separately.

The first one is for making `RegionOfInterest` consistent with `ChannelView`: #1313 

I'll follow up with PRs to:
- remove Segment from the allowed types in Group
- fix the figure generation script
- remove old documentation

alejoe91:
@apdavison can we clos this in favor of #1313 and #1314 ?

apdavison:
@alejoe91 we're still missing a PR to fix the figure generation script (the one which shows the Neo object relationship diagram). (That would need #1313 to be merged first.) I'd prefer to keep this PR open until we have either a PR or an issue for the figure generation, in case it gets forgotten.

apdavison:
#1384 fixes the figure generation script, and is the final PR needed to replace this one.

PR #1034, opened 2021-09-28
Spikeglx filesize check

Saksham20:
I came across some pair of spikeglx imec output files that have different value of filesize (of the `.bin`) file than that retrieved from the `.meta` file.
While this may be a faulty .meta file, but I think an exception should be thrown while parsing such a file.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -248,6 +248,9 @@ def scan_files(dirname):
                 info['bin_file'] = str(bin_filename)
                 info_list.append(info)
 
+            if bin_filename.stat().st_size != meta['fileSizeBytes']:
+                raise Exception('.meta file has faulty value for .bin filesize')
+
     return info_list
 
 
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -250,7 +250,7 @@ def scan_files(dirname):
                 info_list.append(info)
 
             if bin_filename.stat().st_size != meta['fileSizeBytes']:
-                raise warnings.warn('.meta file has faulty value for .bin file size on disc')
+                warnings.warn('.meta file has faulty value for .bin file size on disc')
 
     return info_list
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Saksham,
I would put an option to control this and by default I would not for ce the checking.
very often we reduce the bin without changing the meta file.
And it is the case for the file on tests.

JuliaSprenger:
@samuelgarcia Is the file reduction only for test files or do you do this also for files that are used in analysis? In the first case I would still think that we should always assume the data and metadata info is compatible, i.e. we should update the test files according to the file reduction.

JuliaSprenger:
@Saksham20 can you rebase on master? The failing tests should be fixed now.

samuelgarcia:
@JuliaSprenger : some people reduce the file itself when it is too long. Sometimes just to try sorting.
Forcing the check of file size will be annoying. And this will lead to lot lot github issue.

Saksham20:
> @JuliaSprenger : some people reduce the file itself when it is too long. Sometimes just to try sorting. Forcing the check of file size will be annoying. And this will lead to lot lot github issue.
> 
> I would prefer a warning rather than a eroor for this.

But if a different file size, then it will construct wrong dimensions of the raw array from the .bin file no?

samuelgarcia:
No because it use the size from the memmap not from the meta.

JuliaSprenger:
@samuelgarcia I changed the error to a warning. Do you think we can merge this?

samuelgarcia:
I still think that forcing this check is a bad thing.
Very often the bin file is manually cut.
We should had this test as an option False by default.
I think using size from the real data (memmap) is better that the one from meta file.

JuliaSprenger:
Hi Sam,
for exactly these reasons there is only a warning when .meta and .bin contain inconsistent information and not an error. If the user works on cut files than this warning would be expected, but for complete sessions it can still detecting flawed recording sessions.

samuelgarcia (approved):

PR #1036, opened 2021-09-29
[CI] pin datalad version

JuliaSprenger:
Datalad version 0.15.0 and later requires a git-annex version that is not in the ubuntu default packages yet.
See also https://github.com/datalad/datalad/commit/315911ca76687ef0b8a529f0dfa99bf25fe040ba

### DIFF
--- a/requirements_testing.txt
+++ b/requirements_testing.txt
@@ -1,6 +1,6 @@
 pytest
 pytest-cov
-datalad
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
In case the ubuntu default packages are not updated soon, we should consider using a conda environment to install git-annex, as conda provides a more recent git-annex version (8.20210903).

PR #1049, opened 2021-11-17
AlphaOmega file map version 4 raw reader

thperret:
This PR implements an AlphaOmega mpx file format reader.
This was mainly implemented following the specifications of the file format in the `AlphaRS User Manual V1.0.1.pdf` (sadly not publishable).
This reader can read files from AlphaRS and AlphaLab SnR hardware.

It's still missing the `AlphaOmegaIO` class but I wanted to discuss the fate of the current AlphaOmegaIO implementation which seems to support quite old and ill-documented format. Also from what I can read from the current code, it looks it was implemented with python2 and will not work using python3 (because the variable `m_TypeBlock` was a byte in python2 and compared to a string which is a byte object in python2 but a **string object in python3** (old unicode in python2) so the comparison will always be `False` and will never parse any data.
My take is: remove this old implementation and use this new one. The main drawback is that we lose the ability to read old map file format (but it would need fixing to *really* lose anything).
@samuelgarcia @JuliaSprenger can you check this implementation (I published test data ~~[here](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/62)~~ edit: [here](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/63)) and give your position on the old IO implementation?

Cheers,
Thomas


samuelgarcia:
Whaooooo! Impressive.
I will have look soon.
Thank you to revive this old reader.

thperret:
I implemented the part to read the spikes detected by the AlphaOmega logging software and improved a little bit the documentation of the module.

There's still the `AlphaOmegaIO` part that need to be discussed: 1) do we remove (or rename) the old `IO` reader or 2) do we add the new under another name?

I just realized that there are a lot of commits (I tried to make atomic commits) so if there's too much, I can try to rewrite the history as needed.

Cheers,
Thomas

JuliaSprenger:
@thperret Don't worry about cleaning up the history - in the worst case we can squash when merging.

I am not sure about keeping or overwriting the old IO. It seems at least the comon io tests are completing, so it's not completely broken, but I have never tested it with another dataset. In doubt I would tend to keep it in case someone is depending on it. @samuelgarcia What do you think?

thperret:
Hi @JuliaSprenger 

As I was saying in my first comment, it looks like the old IO reader was implemented using python2 an doesn't read anything (the tests pass because they don't check that actual data is read):
```python3
>>> from neo.io.alphaomegaio import AlphaOmegaIO
>>> r = AlphaOmegaIO(filename="File_AlphaOmega_1.map")
>>> b = r.read_block()
>>> print(b.segments[0].analogsignals)
[]
```
So from my point of view, it either need to be fixed or removed (or renamed)

JuliaSprenger:
Yes, I can confirm that the old implementation is not loading data any more. I short try on fixing it also didn't work (it seems the header format of the test data we have is not matching what is expected). @samuelgarcia You are listed as one of the original authors. How would you prefer to proceed?

samuelgarcia:
Hi Julia and Thomas.
Sorry I did not had time to read the PR yet.
Abut remove ing old code about this : yes!
The old code is not working since long time I think and never totally worked worrrectly.
It was started by Florent Jaillet (in your lab Julia) and he left the field long time ago.
So lets go for a big clean!!

thperret:
Gentle ping? @samuelgarcia @JuliaSprenger

JuliaSprenger:
@thperret Can you fix the residual profiling comments in your PR?

thperret:
@JuliaSprenger  oh, sorry for that. It's fixed

JuliaSprenger:
@thperret Thanks for coming back to this. There's also a few pep8 issues still open (see pep8speaks comment above)

neo/test/iotest/test_alphaomegaio.py
```diff
@@ -13,8 +13,7 @@ class TestAlphaOmegaIO(BaseTestIO, unittest.TestCase):
         'alphaomega'
     ]
     entities_to_test = [
-        'alphaomega/File_AlphaOmega_1.map',
```
JuliaSprenger, line 16: So the `map` files are not supported any more? I think this would be worth a sentence in the docstring, i.e 'For loading of .map files use an old Neo version (<0.11.0)' @samuelgarcia Do you agree?

neo/rawio/alphaomegarawio.py
```diff
@@ -0,0 +1,1252 @@
...
+                if filenames:
+                    self.logger.info(
+                        "Some *.mpx files not referenced. Will try to load them."
+                    )
+                    self._filenames[""].extend(filenames)
+
+    def _source_name(self):
+        return str(self.dirname)
+
+    def _read_file_blocks(self, filename, prune_channels=True):
```
samuelgarcia, line 155: To avoid any confusion with neo.Block I would use "datablock"' instead of block.

neo/rawio/alphaomegarawio.py
```diff
@@ -0,0 +1,1252 @@
...
+            recordings are longer than a day or run over the night
+            - The next segment must have a datetime greater or equal than the
+            previous segment (datetime : YEAR-MONTH-DAY-HOUR-MINUTE-SECOND)
+
+        :param factor_period: how many sample period to consider the next segment
+            is part of the previous one. This should be 1 <= `factor_period` < 2,
+            defaults to 1.5
+        :type factor_period: float
+        """
+        for block in self._blocks:
```
samuelgarcia, line 596: Same here To avoid any confusion with neo.Block I would use "datablock"' instead of block.

JuliaSprenger:
If @samuelgarcia is happy with the small changes requested we can merge this very soon. @thperret Could you address the small comments above?

samuelgarcia:
@JuliaSprenger : We had a call with Thomas on change need to be done.
He is quite busy but will do changes in the following weeks.

thperret:
Hi,

I just pushed the commits including the comments in your review @samuelgarcia:
- rename AlphaOmega blocks recording to datablock to prevent confusion with neo Block class
- add lsx_files __init__ parameter that expect a list of string of lsx file (those are text files listing mpx files of a recording session). Defaults to None which loads all mpx files in dirname, otherwise loads all mpx files references by all lsx files

samuelgarcia:
Cool. Merci Thomas. C'est parfait.
Tu as un héros!
Je merge dès que les tests sont au vert.

JuliaSprenger:
Thanks for the contribution @thperret. Test are passing. I merge first :)

samuelgarcia:
voleuse!

PR #1051, opened 2021-12-02
Fix exception raised when calling pyplot.plot(sig.times, sig) if the signal has array annotations

apdavison:
The traceback was:
```
  File "/Users/andrew/dev/simulation/PyNN/pyNN/utility/plotting.py", line 89, in plot_signals
    ax.plot(signal.times.rescale(ms), signal, label=label, **options)
  File "/Users/andrew/.conda/envs/simulation/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 1743, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/Users/andrew/.conda/envs/simulation/lib/python3.9/site-packages/matplotlib/axes/_base.py", line 273, in __call__
    yield from self._plot_args(this, kwargs)
  File "/Users/andrew/.conda/envs/simulation/lib/python3.9/site-packages/matplotlib/axes/_base.py", line 389, in _plot_args
    y = _check_1d(tup[-1])
  File "/Users/andrew/.conda/envs/simulation/lib/python3.9/site-packages/matplotlib/cbook/__init__.py", line 1318, in _check_1d
    ndim = x[:, None].ndim
  File "/Users/andrew/dev/analysis/neo/neo/core/analogsignal.py", line 270, in __getitem__
    obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 188, in array_annotate
    self.array_annotations.update(array_annotations)
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 400, in update
    self[key] = other[key]
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 389, in __setitem__
    value = self.check_function({key: value}, self.length)[key]
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 39, in _normalize_array_annotations
    value[key] = _normalize_array_annotations(value[key], length)
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 98, in _normalize_array_annotations
    _check_single_elem(value[0])
  File "/Users/andrew/dev/analysis/neo/neo/core/dataobject.py", line 81, in _check_single_elem
    raise ValueError("Array annotations should only be 1-dimensional")
ValueError: Array annotations should only be 1-dimensional
```

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -265,8 +265,11 @@ class AnalogSignal(BaseSignal):
                     raise TypeError("%s not supported" % type(j))
                 if isinstance(k, (int, np.integer)):
                     obj = obj.reshape(-1, 1)
-                if k is not None:  # matplotlib _check_1d() calls__getitem__ with (:, None)
-                    obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
+                elif k is None:
+                    # matplotlib _check_1d() calls__getitem__ with (:, None) and
+                    # reacts appropriately if an IndexError or ValueError is raised
+                    raise IndexError("Cannot add dimensions to an AnalogSignal")
+                obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
         elif isinstance(i, slice):
             obj = super().__getitem__(i)
             if i.start:
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -265,7 +265,8 @@ class AnalogSignal(BaseSignal):
                     raise TypeError("%s not supported" % type(j))
                 if isinstance(k, (int, np.integer)):
                     obj = obj.reshape(-1, 1)
-                obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
+                if k is not None:  # matplotlib _check_1d() calls__getitem__ with (:, None)
+                    obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
         elif isinstance(i, slice):
             obj = super().__getitem__(i)
             if i.start:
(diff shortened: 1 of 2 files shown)


neo/core/analogsignal.py
```diff
@@ -265,7 +265,8 @@ def __getitem__(self, i):
                     raise TypeError("%s not supported" % type(j))
                 if isinstance(k, (int, np.integer)):
                     obj = obj.reshape(-1, 1)
-                obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
+                if k is not None:  # matplotlib _check_1d() calls__getitem__ with (:, None)
+                    obj.array_annotate(**deepcopy(self.array_annotations_at_index(k)))
```
JuliaSprenger, line 269: Shouldn't the array_annotation for the case `k is None` still be added up to the last index available?

JuliaSprenger:
Hi @apdavison Was this fixed in another PR in the meantime? If not could you comment on the review?

neo/core/analogsignal.py
apdavison: On reflection, `None` shouldn't be allowed at all. `x[:, None]` adds an empty dimension to a NumPy array `x`. This doesn't make sense for `AnalogSignal`, which can only be 2D, so I've updated the PR to have Neo raise an IndexError (which also solves the original problem with matplotlib).

The alternative would be to cast the object to a 3D Quantities array, but I think that would be unexpected behaviour.

JuliaSprenger: I think a lot of the `AnalogSignal` code is written with 2 or more dimensions in mind, only assuming the first and the last dimension have a fixed meaning. So maybe adding a new dimension in case of `[ : , None]` would be the cleaner way? @samuelgarcia Any strong opinion on that?

JuliaSprenger: According to the AnalogSignal documentation these have to be 2D. So we should rather raise an error than adding a new dimension.

JuliaSprenger:
@apdavison Thanks for the fix! Looks good to me, I will merge this now.

PR #1052, opened 2021-12-06
Updates for data conversion

legouee:


neo/io/nwbio.py
```diff
@@ -434,10 +435,15 @@ def write_all_blocks(self, blocks, **kwargs):
         if "session_description" not in annotations:
             annotations["session_description"] = blocks[0].description or self.filename
             # todo: concatenate descriptions of multiple blocks if different
+        
         if "session_start_time" not in annotations:
-            raise Exception("Writing to NWB requires an annotation 'session_start_time'")
+            annotations["session_start_time"] = blocks[0].rec_datetime
+            if annotations["session_start_time"]==None:
+                annotations["session_start_time"] = datetime.now()
```
apdavison, line 442: I don't think we should use `datetime.now()`, as it is probably not the correct  session start time. We should raise an Exception instead

neo/io/nwbio.py
```diff
@@ -434,10 +435,15 @@ def write_all_blocks(self, blocks, **kwargs):
         if "session_description" not in annotations:
             annotations["session_description"] = blocks[0].description or self.filename
             # todo: concatenate descriptions of multiple blocks if different
+        
         if "session_start_time" not in annotations:
-            raise Exception("Writing to NWB requires an annotation 'session_start_time'")
+            annotations["session_start_time"] = blocks[0].rec_datetime
+            if annotations["session_start_time"]==None:
```
apdavison, line 441: use "is None"

neo/io/nwbio.py
```diff
@@ -514,8 +520,9 @@ def _write_segment(self, nwbfile, segment, electrodes):
         for i, signal in enumerate(
                 chain(segment.analogsignals, segment.irregularlysampledsignals)):
             assert signal.segment is segment
+            signal.name = "%s %s %i" % (signal.name, segment.name, i)
```
apdavison, line 523: if the signal already has a name, we should not over-write it

neo/io/nwbio.py
```diff
@@ -526,6 +533,7 @@ def _write_segment(self, nwbfile, segment, electrodes):
 
         for i, event in enumerate(segment.events):
             assert event.segment is segment
+            event.name = "%s  %s %i" % (event.name, segment.name, i)
```
apdavison, line 536: same comment as above for signals

neo/test/iotest/test_nwbio.py
```diff
@@ -36,7 +36,8 @@ class TestNWBIO(BaseTestIO, unittest.TestCase):
     entities_to_download = ["nwb"]
     entities_to_test = [
         # Files from Allen Institute:
-        "nwb/H19.29.141.11.21.01.nwb",  # 7 MB
+#        "nwb/H19.29.141.11.21.01.nwb",  # 7 MB
+        "/Users/legouee/Desktop/Brainhack/Neo_NWB/H19.29.141.11.21.01.nwb"
```
apdavison, line 40: user-specific paths shouldn't appear here

neo/test/iotest/test_nwbio.py
```diff
@@ -63,16 +64,19 @@ def test_roundtrip(self):
...
+                                 signal=np.random.randn(64, num_chan) * pq.mV,
                                  sampling_rate=8 * pq.kHz,
                                  t_start=40 * pq.ms)
-                c = AnalogSignal(np.random.randn(33, num_chan) * pq.uA,
+                c = AnalogSignal(name = 'Signal_c %s' % (seg.name),
+                                 signal=np.random.randn(33, num_chan) * pq.uA,
                                  sampling_rate=10 * pq.kHz,
                                  t_start=120 * pq.ms)
-
+                print("a = ", a)
```
apdavison, line 79: best to avoid `print()` in released code

apdavison (changes requested):
looks good, just a few changes

neo/io/nwbio.py
```diff
@@ -434,10 +435,15 @@ def write_all_blocks(self, blocks, **kwargs):
         if "session_description" not in annotations:
             annotations["session_description"] = blocks[0].description or self.filename
             # todo: concatenate descriptions of multiple blocks if different
+        
         if "session_start_time" not in annotations:
-            raise Exception("Writing to NWB requires an annotation 'session_start_time'")
+            annotations["session_start_time"] = blocks[0].rec_datetime
+            if annotations["session_start_time"] is None:
+                raise Exception("Give a session start time")
```
apdavison, line 442: please use the same exception text as previously, i.e. "Writing to NWB requires an annotation 'session_start_time'"

apdavison:
please could you also fix the PEP8 problems reported by pep8speaks? (see above)

PR #1053, opened 2021-12-07
Add a warning about changing the name of the signal (to avoid signals with the same name)

legouee:


legouee:
A new PR #1054 has been done to solve conflicts.

PR #1054, opened 2021-12-07
[NWBIO] Add a warning to avoid signals with the same name

legouee:


JuliaSprenger:
Hi @legouee,
sorry for the failing tests, we fixed those in master. If you rebase or pull from there and the test in your PR should hopefully also succeed and I can merge.

JuliaSprenger:
Hi @legouee I rebased your branch on the current master. Can you check all your changes are still as you intended them?

neo/io/nwbio.py
```diff
@@ -517,7 +517,9 @@ def _write_segment(self, nwbfile, segment, electrodes):
         for i, signal in enumerate(
                 chain(segment.analogsignals, segment.irregularlysampledsignals)):
             assert signal.segment is segment
-            signal.name = "%s %s %i" % (signal.name, segment.name, i)
+            if hasattr(signal, 'name'):
+                signal.name = "%s %s %i" % (signal.name, segment.name, i)
+                logging.warning("Warning signal name exists. New name: %s" % (signal.name))
             if not signal.name:
```
JuliaSprenger, line 523:
```suggestion
            else signal.name:
```

neo/io/nwbio.py
```diff
@@ -530,7 +532,9 @@ def _write_segment(self, nwbfile, segment, electrodes):
 
         for i, event in enumerate(segment.events):
             assert event.segment is segment
-            event.name = "%s  %s %i" % (event.name, segment.name, i)
+            if hasattr(event, 'name'):
+                event.name = "%s  %s %i" % (event.name, segment.name, i)
+                logging.warning("Warning event name exists. New name: %s" % (event.name))
             if not event.name:
```
JuliaSprenger, line 538:
```suggestion
            else event.name:
```

legouee:
Hi @JuliaSprenger Thank you very much for your review. I added your comments to the code.

neo/io/nwbio.py
```diff
@@ -517,8 +517,10 @@ def _write_segment(self, nwbfile, segment, electrodes):
         for i, signal in enumerate(
                 chain(segment.analogsignals, segment.irregularlysampledsignals)):
             assert signal.segment is segment
-            signal.name = "%s %s %i" % (signal.name, segment.name, i)
-            if not signal.name:
+            if hasattr(signal, 'name'):
+                signal.name = "%s %s %i" % (signal.name, segment.name, i)
+                logging.warning("Warning signal name exists. New name: %s" % (signal.name))
+            else:
                 signal.name = "%s : analogsignal%d %i" % (segment.name, i, i)
```
JuliaSprenger, line 524: Is it intentional that you compose the names differently for existing and non-existing name attributes? Wouldn't it make sense to always use a 'segment signal idx' ordering in the name?

JuliaSprenger: and the same of course also for events.

legouee: Yes, that was intentional as I ran into problems with the signal name and the event name when testing with NWB files.

neo/test/iotest/test_asciisignalio.py
```diff
@@ -13,6 +13,7 @@
 from neo.test.iotest.common_io_test import BaseTestIO
 from neo.core import AnalogSignal, Segment, Block
 
+import string
```
JuliaSprenger, line 16: I don't think this change is related to the NWBIO change you suggest. In that case can you remove it here?

neo/test/iotest/test_asciisignalio.py
```diff
@@ -133,8 +134,8 @@ def reader(filename, comment_rows):
 
         os.remove(filename)
 
-    # test usecols
-    # test skiprows
+    # test_skiprows
+    # test_usecols
```
JuliaSprenger, line 138: I don't think this change is related to the NWBIO change you suggest. In that case can you remove it here?

neo/test/iotest/test_asciisignalio.py
legouee: Yes. It is done. I removed it.

neo/test/iotest/test_asciisignalio.py
legouee: Yes. I had a mix of several of my branches. It is done. I removed it.

JuliaSprenger:
Looks good to me now. Thanks for the fix!

PR #1057, opened 2021-12-24
Extend tdtrawio to support single tdt block reading and empty streams

JuliaSprenger:
Currently the tdtrawio expects a specific folder structure containing multiple tdt blocks (folders). This does not permit to load an individual tdt (block) dataset that is not stored in such a folder structure. This PR permits to load such individual datasets by providing one of contained data files as `dirname` argument. In this case all accompanying Tdt files with the same file stem are loaded and `sev` files are detected flexibly.

This PR also
- adds test for the functionality described above
- permits to load datasets in which a stream does not contain samples in a neo segment (no test dataset yet)
- internally uses pathlib for path manupulation


JuliaSprenger:
The failing tests are not related to the changes here and should be fixed in master first.

JuliaSprenger:
@samuelgarcia What do you think of this?

JuliaSprenger:
@samuelgarcia I rebased on master and the tests should hopefully succeed now.

neo/rawio/tdtrawio.py
```diff
@@ -181,15 +209,24 @@ def _parse_header(self):
                         assert self._sigs_lengths[seg_index][stream_index] == size
 
                     # signal start time, relative to start of segment
-                    t_start = data_index['timestamp'][0]
+                    if len(data_index['timestamp']):
+                        t_start = data_index['timestamp'][0]
+                    else:
+                        # if no signal present use segment t_start as dummy value
+                        t_start = self._seg_t_starts[seg_index]
```
samuelgarcia, line 216: Is it initialize

neo/rawio/tdtrawio.py
```diff
@@ -181,15 +209,24 @@ def _parse_header(self):
...
                     # sampling_rate and dtype
-                    _sampling_rate = float(data_index['frequency'][0])
-                    _dtype = data_formats[data_index['dataformat'][0]]
+                    if len(data_index):
+                        _sampling_rate = float(data_index['frequency'][0])
+                        _dtype = data_formats[data_index['dataformat'][0]]
+                    else:
+                        # if no signal present use dummy values
+                        _sampling_rate = 1.
+                        _dtype = int
```
samuelgarcia, line 229: Is it safe.

neo/rawio/tdtrawio.py
JuliaSprenger: Yes, `seg_t_starts` is already initialized in line 128 & 130 and contains either a number or np.nan.

JuliaSprenger: If no event or signal or other timepoints are present in the segment, then this will probably lead to an error as a Segment shouldn't have `np.nan` as t_start. However, this was already previously the case (see lines 128 & 130).

neo/rawio/tdtrawio.py
JuliaSprenger: These dummy values are used when within a given stream and segment no data were recorded on a channel. In that case no data should be read from file in any case, so the value of the dummy dtype does not matter. It is only relevant for ensuring that all channels in a stream in a segment have the same characteristics (dtype, sampling_rate).

JuliaSprenger:
@samuelgarcia Thanks for merging :)

PR #1063, opened 2022-01-18
[nix] pin nixio to avoid buggy 1.5.2 version

JuliaSprenger:
The latest nixio version (1.5.2) raises an error when rewriting string annotations.
This should be resolved on the nixio level.


JuliaSprenger:
@samuelgarcia This is a fix for the failing tests in master. Can you merge this one first, so we can rebase other PRs?

PR #1068, opened 2022-02-04
Feature/new filter

kloss-o:
Until now it was not so easy to filter for multiple values with the filter function in neo. To solve this problem, I introduced "FilterConditions". These can be provided to the filter function as the value of a dictionary. 

An example of the syntax: 

seg.filter({“electrode_id”: FilterCondition) or seg.filter(electrode_id = FilterCondition) 

In order to be able to filter for several electrode_ids, there are some objects that inherit from FilterCondition and provide different filtering options. 

For example, seg.filter({"electrode_id": neo.FilterIsIn([0, 1, 5]) can be used to filter for all objects with electrode_id 0, 1 or 5. 

List of all currently added FilterConditions: 

-FilterEqual(x), returns id == x 

-FilterIsNot(x), returns id != x 

-FilterLessThan(x), returns id < x 

-FilterLessThanEqual(x), returns id <= x 

-FilterGreaterThan(x), returns id > x 

-FilterGreaterThanEqual(x), returns id >= x 

-FilterIsIn([x, y, z]), returns id == x or id == y or id == z 

-FilterInRange(a, b), returns a <= id <= b 

 
Additional FilterConditions can be added at any time and without much effort.

### DIFF
--- a/neo/core/__init__.py
+++ b/neo/core/__init__.py
@@ -10,7 +10,9 @@ or more class from this module.
 Classes:
 
 .. autoclass:: Block
+.. automethod:: Block.filter
 .. autoclass:: Segment
+.. automethod:: Segment.filter
 .. autoclass:: Group
 
 .. autoclass:: AnalogSignal
--- a/neo/core/__init__.py
+++ b/neo/core/__init__.py
@@ -37,7 +37,9 @@ from neo.core.segment import Segment
 from neo.core.analogsignal import AnalogSignal
 from neo.core.irregularlysampledsignal import IrregularlySampledSignal
 
-from neo.core.container import FilterEqual, FilterIsIn, FilterIsNot, FilterInRange, FilterGreaterThan, FilterLessThan, FilterGreaterThanEqual, FilterLessThanEqual
+from neo.core.container import FilterEqual, FilterIsIn, FilterIsNot,\
+    FilterInRange, FilterGreaterThan, FilterLessThan,\
+    FilterGreaterThanEqual, FilterLessThanEqual
 
 from neo.core.event import Event
 from neo.core.epoch import Epoch
--- a/neo/core/__init__.py
+++ b/neo/core/__init__.py
@@ -37,6 +37,8 @@ from neo.core.segment import Segment
 from neo.core.analogsignal import AnalogSignal
 from neo.core.irregularlysampledsignal import IrregularlySampledSignal
 
+from neo.core.container import FilterEqual, FilterIsIn, FilterIsNot, FilterInRange, FilterGreaterThan, FilterLessThan, FilterGreaterThanEqual, FilterLessThanEqual
+
 from neo.core.event import Event
 from neo.core.epoch import Epoch
 
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -77,7 +77,7 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
                         all([obj is not res for res in results])):
                     results.append(obj)
                 elif (key in obj.annotations and value.test(obj.annotations[key]) and
-                          all([obj is not res for res in results])):    #Benchmark -> hauefigkeit, time
+                          all([obj is not res for res in results])):
                     results.append(obj)
 
     # keep only objects of the correct classes
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -76,7 +76,7 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
                 if (hasattr(obj, key) and getattr(obj, key) == value and
                         all([obj is not res for res in results])):
                     results.append(obj)
-                elif (key in obj.annotations and obj.annotations[key] == value and
+                elif (key in obj.annotations and obj.annotations[key].test(key) and #elif (key in obj.annotations and obj.annotations[key].test(key) and AttributeError: 'int' object has no attribute 'test'
                           all([obj is not res for res in results])):
                     results.append(obj)
 
@@ -98,14 +98,6 @@ class FilterCondition():
     def test(self, x):
         return True
 
-class less_than(FilterCondition):
-
-    def __init__(self, z):
-        self.control = z
-
-    def test(self, x):
-        return x < self.control
-
 class equal(FilterCondition):
 
     def __init__(self, z):
@@ -114,7 +106,49 @@ class equal(FilterCondition):
     def test(self, x):
         return x == self.control
 
-    
+class less_than(FilterCondition):
+
+    def __init__(self, z):
+        self.control = z
+
+    def test(self, x):
+        return x < self.control
+
+class more_than(FilterCondition):
+
+    def __init__(self, z):
+        self.control=z
+
+    def test(self, x):
+        return x > self.control
+
+class is_in(FilterCondition):
+
+    def __init__(self, z):
+        self.control = z
+
+    def test(self, x):
+        if(type(x) == list):
+            return x.contains(self.control)
+        elif(type(x) == int):
+            return x == self.control
+        else:
+            raise SyntaxError('parameter not of type list or int')
+
+class in_range(FilterCondition):
+
+    def __init__(self, a: int, b: int):
+        if(type(a) != int or type(b) != int):
+            raise SyntaxError("parameters not of type int")
+        else:
+            self.control=[]
+            for num in range(a, b+1):
+                self.control.append(num)
+
+    def test(self, x):
+        is_in(self.control)
+
+
 
 class Container(BaseNeo):
     """
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -76,8 +76,8 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
                 if (hasattr(obj, key) and getattr(obj, key) == value and
                         all([obj is not res for res in results])):
                     results.append(obj)
-                elif (key in obj.annotations and obj.annotations[key].test(key) and #elif (key in obj.annotations and obj.annotations[key].test(key) and AttributeError: 'int' object has no attribute 'test'
-                          all([obj is not res for res in results])):
+                elif (key in obj.annotations and value.test(obj.annotations[key]) and
+                          all([obj is not res for res in results])):    #Benchmark -> hauefigkeit, time
                     results.append(obj)
 
     # keep only objects of the correct classes
@@ -106,6 +106,27 @@ class equal(FilterCondition):
     def test(self, x):
         return x == self.control
 
+class is_not(FilterCondition):
+    def __init__(self, z):
+        self.control = z
+
+    def test(self, x):
+        return x != self.control
+
+class less_than_equal(FilterCondition):
+    def __init__(self, z):
+        self.control = z
+
+    def test(self, x):
+        return x <= self.control
+
+class greater_than_equal(FilterCondition):
+    def __init__(self, z):
+        self.control = z
+
+    def test(self, x):
+        return x >= self.control
+
 class less_than(FilterCondition):
 
     def __init__(self, z):
@@ -114,7 +135,7 @@ class less_than(FilterCondition):
     def test(self, x):
         return x < self.control
 
-class more_than(FilterCondition):
+class greater_than(FilterCondition):
 
     def __init__(self, z):
         self.control=z
@@ -128,25 +149,34 @@ class is_in(FilterCondition):
         self.control = z
 
     def test(self, x):
-        if(type(x) == list):
-            return x.contains(self.control)
-        elif(type(x) == int):
+        if(type(self.control) == list):
+            return x in self.control
+        elif(type(self.control) == int):
             return x == self.control
         else:
             raise SyntaxError('parameter not of type list or int')
 
 class in_range(FilterCondition):
 
-    def __init__(self, a: int, b: int):
+    def __init__(self, a, b, left_closed=False, right_closed=False):
         if(type(a) != int or type(b) != int):
             raise SyntaxError("parameters not of type int")
         else:
-            self.control=[]
-            for num in range(a, b+1):
-                self.control.append(num)
+            self.a = a
+            self.b = b
+            self.left_closed = left_closed
+            self.right_closed = right_closed
 
     def test(self, x):
-        is_in(self.control)
+        if (not self.left_closed and not self.right_closed):
+            return (x >= self.a and x <= self.b)
+        elif (not self.left_closed and self.right_closed):
+            return (x >= self.a and x < self.b)
+        elif (self.left_closed and not self.right_closed):
+            return (x > self.a and x <= self.b)
+        else:
+            return (x > self.a and x < self.b)
+
 
 
 
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -27,7 +27,8 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
         in either their attributes or annotations.  Search terms can be
         provided as keyword arguments or a dictionary, either as a positional
         argument after data or to the argument targdict.
-        A key of a provided dictionary is the name of the requested annotation and the value is a FilterCondition object.
+        A key of a provided dictionary is the name of the requested annotation
+        and the value is a FilterCondition object.
         E.g.: FilterEqual(x), FilterLessThan(x), FilterInRange(x, y).
 
         targdict can also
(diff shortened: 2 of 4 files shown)


neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
         results = []
         for key, value in sorted(targdict.items()):
             for obj in data:
-                if (hasattr(obj, key) and getattr(obj, key) == value and
-                        all([obj is not res for res in results])):
+                if (hasattr(obj, key) and getattr(obj, key) == value and all(
+                        [obj is not res for res in results])):
```
JuliaSprenger, line 79: I don't think this change improves readability and the line also doesn't seem >100 characters. Was this a pep8speaks suggestion?

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
                     results.append(obj)
-                elif (key in obj.annotations and obj.annotations[key] == value and
-                          all([obj is not res for res in results])):
+                elif (isinstance(value, FilterCondition)):
+                    if (key in obj.annotations and value.test(obj.annotations[key]) and all(
+                            [obj is not res for res in results])):
+                        results.append(obj)
+                elif key in obj.annotations and obj.annotations[key] == value and all(
+                        [obj is not res for res in results]):
```
JuliaSprenger, line 86: same comment as above.

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
     else:
         return results
 
 
+class FilterCondition():
+    """
+        FilterCondition object is given as parameter to container.filter():
+
+        example.filter(annotation=<FilterCondition>) or
+        example=filter({'annotation': <FilterCondition>})
```
JuliaSprenger, line 105:
```suggestion
        segment.filter(my_annotation=<FilterCondition>) or
        segment=filter({'my_annotation': <FilterCondition>})
```

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
+    """
+
+    def __init__(self, z):
+        pass
+
+    def test(self, x):
+        return True
```
JuliaSprenger, line 112: You expect this to be overwritten by each subclass, in that case wouldn't it be better to do:
```suggestion
    def test(self, x):
        raise NotImplementedError()
```

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
+
+
+class FilterEqual(FilterCondition):
```
JuliaSprenger, line 115: Can you please also add docstrings to your FilterCondition classes here and below?

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
+        elif (type(self.control) == int):
+            return x == self.control
+        else:
+            raise SyntaxError('parameter not of type list or int')
+
+
+class FilterInRange(FilterCondition):
+
+    def __init__(self, a, b, left_closed=False, right_closed=False):
+        if (type(a) != int or type(b) != int):
```
JuliaSprenger, line 185: can't this also be border of type `float`?

neo/core/container.py
```diff
@@ -17,27 +17,29 @@ def unique_objs(objs):
...
+        targdict can also
+        be a list of dictionaries, in which case the filters are applied
+        sequentially.
+
+        A list of disctionaries is handled as follows: [ { or } and { or } ]
+        If targdict and kwargs are both supplied, the
+        targdict filters are applied first, followed by the kwarg filters.
+        A targdict of None or {} corresponds to no filters applied, therefore
+        returning all child objects. Default targdict is None.
+
```
JuliaSprenger, line 42: This seems overindented.

neo/core/container.py
```diff
@@ -73,25 +75,132 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
+        FilterCondition object is given as parameter to container.filter():
+
+        example.filter(annotation=<FilterCondition>) or
+        example=filter({'annotation': <FilterCondition>})
+    """
+
+    def __init__(self, z):
+        pass
+
+    def test(self, x):
```
JuliaSprenger, line 111: not sure if `test` is the right name here. Maybe rather something like `evaluate` or `get_mask`?

neo/test/coretest/test_container.py
```diff
@@ -4,8 +4,12 @@
 
 import unittest
 
+import quantities as q
```
JuliaSprenger, line 7: Conventionally this should be pq
```suggestion
import quantities as pq
```

neo/test/coretest/test_container.py
```diff
@@ -98,10 +102,190 @@ def test__children(self):
...
+        st2.annotate(test=6)
+        seg.spiketrains.append(st1)
+        seg.spiketrains.append(st2)
+
+        self.assertEqual(2, len(seg.filter(test=neo.FilterInRange(5, 6, False, False))))
+        self.assertEqual(1, len(seg.filter(test=neo.FilterInRange(5, 6, True, False))))
+        self.assertEqual(1, len(seg.filter(test=neo.FilterInRange(5, 6, False, True))))
+        self.assertEqual(0, len(seg.filter(test=neo.FilterInRange(5, 6, True, True))))
+
+    def test_filter_connectivity(self):
```
JuliaSprenger, line 270:
```suggestion
    def test_filter_filter_consistency(self):
```

neo/test/coretest/test_container.py
```diff
@@ -98,10 +102,190 @@ def test__children(self):
...
+        st2.annotate(name='st_num_1')
+        seg.spiketrains.append(st1)
+        seg.spiketrains.append(st2)
+
+        self.assertEqual(2, len(seg.filter({'test': neo.FilterEqual(5),
+                                            'filt': neo.FilterEqual(6)})))
+        self.assertEqual(0, len(seg.filter([{'test': neo.FilterEqual(5)},
+                                            {'filt': neo.FilterEqual(6)}])))
+        self.assertEqual(1, len(seg.filter(name='st_num_1')))
+
```
JuliaSprenger, line 288: You don't test cases yet where the annotations could be a list or similar.

JuliaSprenger (commented):
Thanks for looking into improving the filter functionality! This has been on our todo list for quite some time. Your approach is adding a number of new classes to neo. Intuitively this looks like an approach that might add a bit of overhead to filtering. Did you also run some performance tests with your new implementation? @samuelgarcia What do you think?

neo/core/container.py
```diff
@@ -372,18 +470,25 @@ def size(self):
...
+        A key of a provided dictionary is the name of the requested annotation
+        and the value is a FilterCondition object.
+        E.g.: equal(x), less_than(x), FilterInRange(x, y).
+
+        targdict can also
         be a list of dictionaries, in which case the filters are applied
-        sequentially.  If targdict and kwargs are both supplied, the
+        sequentially.
+
+        A list of disctionaries is handled as follows: [ { or } and { or } ]
```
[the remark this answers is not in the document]
kloss-o: I did run some tests and found out that the filtering time is pretty much the same

mdenker:
Thanks for the PR, Oliver, and for reviewing the code, Julia. 

Regarding performance, there is actually another follow-up PR by @kloss-o in prep  that would improve filtering speed quite a bit, but unrelated to this concrete implementation. @kloss-o , I think when you did some benchmarking the `FilterCondition` implementation (without further optimization) was comparable in speed to the traditional implementation?

The idea behind the different Filter classes was, for one, that it would allow users to also define their own "custom" filters as a filter function, like (in pseudocode):
```python
class FilterGoodElectrodesSurroundedByGoodOnes(FilterCondition)
    def test(electrode):
       return is_good_quality[electrode] and np.all(is_good_quality[i] for i in neighbor_electrodes(electrode])
```
Secondly, that it would not require any new or ambiguous syntax (e.g., something like `["property", "<=", 5]`) in the function call to `filter()`. In particular, the neo filter function would continue to work as before with the existing syntax, but adds additional functionality.

However, I think there are many other options. An alternative could be to revert to, e.g., simple filter functions, at the expense of being slightly more ambiguous.

neo/core/container.py
kloss-o: That was a pep8speaks suggestion. In my IDE the lines would be 115 characters long when combined

neo/core/container.py
kloss-o: I renamed it to evaluate

neo/test/coretest/test_container.py
kloss-o: I added tests for a list and for a dictionary

Moritz-Alexander-Kern:
The following notebook is intended to show the application of the 'new filter' function in practice.
The use cases were derived from the Neo - Elephant tutorials and aim to demonstrate how the suggested 'new filter' improves the workflow. 

Link to notebook `tutorial_new_filter_examples.ipynb`: 

GitHub gist: https://gist.github.com/Moritz-Alexander-Kern/afdfb707bde9052c6cd02ccddd2c4b9e
GIN: https://gin.g-node.org/NeuralEnsemble/neo_elephant_teaching/src/new_filter/tutorial_new_filter_examples.ipynb

JuliaSprenger:
Here some [notes](https://github.com/NeuralEnsemble/python-neo/files/8672726/2022-05-11_neo_core_meeting.md) on the discussion of this PR during the neo core meeting today.

JuliaSprenger:
Hi @kloss-o @Moritz-Alexander-Kern any news on this? Tell me if the notes from the last meeting are not detailed enough for you to continue on this topic.

kloss-o:
-I moved the filter classes to a separate module.

-I also managed to decrease the filtering time. For comparison, I created one Segment that contains x Spiketrains and use a filter that applies to every Spiketrain. (Tested on multiple combinations of Segments, Blocks, Spiketrains, Analogsignals, the result is nearly the same).

Old filtering method:
one_segment_only_spiketrains_success_average

New filtering method:
one_segment_only_spiketrains_success_average_new

neo/core/__init__.py
```diff
@@ -35,6 +37,11 @@
 from neo.core.analogsignal import AnalogSignal
 from neo.core.irregularlysampledsignal import IrregularlySampledSignal
 
+# Import FilterClasses
+from neo.core.filters import Equal, IsIn, IsNot,\
```
apdavison, line 41: I don't think the filters should be imported into the top-level namespace,
rather just `from neo.core import filters`

neo/core/container.py
```diff
@@ -18,27 +20,28 @@ def unique_objs(objs):
     using the "is" test.
     """
     seen = set()
-    return [obj for obj in objs
-            if id(obj) not in seen and not seen.add(id(obj))]
+    return [obj for obj in objs if id(obj) not in seen and not seen.add(id(obj))]
```
apdavison, line 23: please revert all white-space only changes, they make it harder to see the substantial changes.

neo/core/container.py
```diff
@@ -18,27 +20,28 @@ def unique_objs(objs):
...
 
 
 def filterdata(data, targdict=None, objects=None, **kwargs):
     """
-    Return a list of the objects in data matching *any* of the search terms
+    Return a list of child objects matching *any* of the search terms
```
apdavison, line 28: child of what? I think the original docstring is more understandable, perhaps put "objects in `data`" to make it clear that we're referring to the variable named `data`.

neo/core/container.py
```diff
@@ -74,23 +77,27 @@ def filterdata(data, targdict=None, objects=None, **kwargs):
...
+                        key in obj.annotations and value.evaluate(obj.annotations[key])):
                     results.append(obj)
-                elif (key in obj.annotations and obj.annotations[key] == value and
-                          all([obj is not res for res in results])):
+                if key in obj.annotations and obj.annotations[key] == value:
                     results.append(obj)
 
+    # remove duplicates from results
+    res = []
+    [res.append(result) for result in results if not any([result is elem for elem in res])]
```
apdavison, line 90: This looks as though it will be slow - perhaps use a `set` to eliminate duplicates?

neo/core/container.py
```diff
@@ -188,13 +195,12 @@ class Container(BaseNeo):
     # Containers that are listed when pretty-printing
     _repr_pretty_containers = ()
 
-    def __init__(self, name=None, description=None, file_origin=None,
-                 **annotations):
+    def __init__(self, name=None, description=None, file_origin=None, **annotations):
```
apdavison, line 198: same comment as above about whitespace-only changes

neo/core/filters.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    def __init__(self, z):
+        pass
+
+    def evaluate(self, x):
+        raise NotImplementedError()
+
+
+class Equal(FilterCondition):
```
apdavison, line 22: I would call this `Equals` or `EqualTo`

neo/core/filters.py
```diff
@@ -0,0 +1,109 @@
...
+class IsNot(FilterCondition):
+
+    def __init__(self, z):
+        self.control = z
+
+    def evaluate(self, x):
+        return x != self.control
+
+
+class LessThanEqual(FilterCondition):
```
apdavison, line 40: `LessThanOrEquals` or `LessThanOrEqualTo`

neo/core/filters.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __init__(self, z):
+        self.control = z
+
+    def evaluate(self, x):
+        return x <= self.control
+
+
+class GreaterThanEqual(FilterCondition):
```
apdavison, line 49: `GreaterThanOrEquals` or `GreaterThanOrEqualTo`

apdavison (changes requested):
I think this is in good shape now and almost ready to merge. I have suggested some changes to filter class names, and I would also prefer it if all whitespace-only changes were reverted, to make it easier to see where the code has actually changed. The most important change, I think, is to keep the functions in their own module and do not import them into the top-level namespace, i.e. they should be used as 
```
from neo.filters import LessThan
```
or
```
import neo.filters as nf

...nf.LessThan...
```

neo/core/container.py
apdavison: perhaps use the Python `id()` function?, e.g., 
```
unique_results = { id(res): res for res in results }.values()
```

Moritz-Alexander-Kern: Perfect, thanks!

neo/core/container.py
Moritz-Alexander-Kern: see: f8a56ea337b7cbbf4b43c8d60ae57d88a5233193

neo/core/__init__.py
Moritz-Alexander-Kern: see: bf1c51da51def0edc9eaf2cbe6d8e1eb4e2931e3

neo/core/container.py
Moritz-Alexander-Kern: see: f9e1d290ad23ad0e393cbf86fd1fd56b38caeeb3

neo/core/container.py
Moritz-Alexander-Kern: See: 4e1df3bf2baa5fc6435fe77356693c35358227b6

neo/core/container.py
Moritz-Alexander-Kern: See: f9e1d290ad23ad0e393cbf86fd1fd56b38caeeb3

neo/core/filters.py
Moritz-Alexander-Kern: See: 21f829af55ae3861942c58c2e6419152227476c1

neo/core/filters.py
Moritz-Alexander-Kern: See: 21f829af55ae3861942c58c2e6419152227476c1

neo/core/filters.py
Moritz-Alexander-Kern: See: 21f829af55ae3861942c58c2e6419152227476c1

neo/core/container.py
Moritz-Alexander-Kern: Agreed, the border type was adjusted to `numbers`.

Moritz-Alexander-Kern:
Hey @apdavison and @JuliaSprenger ,
Thank you for the reviews and the valuable feedback you provided. 
I considered all your comments and made changes to the code. 

If you have any further suggestions or if there are any remaining issues, please don't hesitate to let me know.

apdavison (approved):

PR #1069, opened 2022-02-05
Fix gains for SpikeGLX

alejoe91:
In the SpikeGLX parser, gains are wrongly set.

In the SpikeGLX documentation (https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_20.md#channel-entries-by-type):
`g = Vmax / Imax / gain`
but in the code it was set as (https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/spikeglxrawio.py#L287):
`g = gain * Vmax / Imax` 

where `Vmax / Imax` is `gain_factor `. Instead, it should be:
`g = Vmax / Imax / gain`

This is tested on a NP2 file, where the computed gainwas over 4000! With the corrected gain, it is now ~0.7, which gives more reasonable uV values.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -284,18 +284,18 @@ def scan_files(dirname):
                         v = meta['imroTbl'][c].split(' ')[index_imroTbl]
                         per_channel_gain[c] = 1. / float(v)
                     gain_factor = float(meta['imAiRangeMax']) / 512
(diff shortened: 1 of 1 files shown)


alejoe91:
@samuelgarcia @JuliaSprenger please validate this and let's merge soon :)

samuelgarcia:
@alejoe91 : I think the NP1 patch you propose is wring but the second one is OK. Let me check

PR #1079, opened 2022-03-01
Fixes from SpikeInterface CI

alejoe91:
A couple of small fixes that came up when addin tests in SI:

- BiocamRawIO: remove squeeze when returning traces
- AxonaRawIO: construct list of channels also when `channel_indexes` is a `slice`

### DIFF
--- a/neo/rawio/axonarawio.py
+++ b/neo/rawio/axonarawio.py
@@ -302,7 +302,7 @@ class AxonaRawIO(BaseRawIO):
             i_start = 0
         if i_stop is None:
             i_stop = bin_dict['num_total_samples']
-        if channel_indexes is None:
+        if channel_indexes is None or isinstance(channel_indexes, slice):
             channel_indexes = [i for i in range(bin_dict['num_channels'])]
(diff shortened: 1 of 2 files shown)


alejoe91:
@samuelgarcia @JuliaSprenger

neo/rawio/axonarawio.py
```diff
@@ -304,6 +304,9 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
             i_stop = bin_dict['num_total_samples']
         if channel_indexes is None:
             channel_indexes = [i for i in range(bin_dict['num_channels'])]
+        if isinstance(channel_indexes, slice):
```
samuelgarcia, line 307: I would put elif

alejoe91, line 307:
```suggestion
        elif isinstance(channel_indexes, slice):
```

samuelgarcia:
Thanks for the fix.
This reinforce friendship between peoples.

PR #1081, opened 2022-03-02
0.10.1 release updates

JuliaSprenger:
Currently the documentation is not building due to the documentation requirements being unaccessible. For me it seems the file is there (https://readthedocs.org/projects/neo/builds/16245520/). @samuelgarcia Any idea?

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.9"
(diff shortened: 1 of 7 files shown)


samuelgarcia:
Hi Julia,
I think we need the **.readthedocs.yaml** I will give a try.

samuelgarcia (approved):

samuelgarcia:
Hi Julia,
I think we need the **.readthedocs.yaml** I will give a try.

PR #1093, opened 2022-03-08
Add new github actions.

samuelgarcia:
@JuliaSprenger @apdavison  @mdenker 
This GH actions will replace travis one.
It run on several os/python/numpy.


samuelgarcia:
If tests pass I will remove tarvis

samuelgarcia:
This is now 45 jobs (3 * 3 * 5) for the core.
Maybe we could remove some numpy version.
@apdavison : do you remember what was the rules for numpy version testing ?
Here I already change the list compared to the travis CI

.github/workflows/core-test.yml
```diff
@@ -0,0 +1,45 @@
...
+  multi-os-python-numpy:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: true
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        python-version: ['3.7', '3.8', '3.9']
+        numpy-version: ['1.16.6', '1.19.5', '1.20.3', '1.21.5', '1.22.3']
+        exclude:
```
JuliaSprenger, line 18: What is the issue with this combination? Maybe put a comment here, so we know when to remove this constraint....

samuelgarcia: It is not an issue but the fact that we run 45 time sthe same test!!
Some numpy are not available for some python.

samuelgarcia:
Yes I was think something like this : run io and rawio only if core is passing (~3min).
I think we can serialize workflow, I will check.

JuliaSprenger:
Ok, then I will merge this in the current state and we can add chaining later.

apdavison:
For testing, we decided on [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) + 1 year, so we should be testing Python 3.7+ and NumPy 1.17+. I suggest we test all combinations for the core tests, but only the oldest and newest for io tests.

samuelgarcia:
thanks for merging.

Andrew, for cortest you suggest all numpy version from 1.17 to 1.22 ?
It is a lot no ?
If we do this maybe we could remove osx testing ?
otherwise it will be 66 jobs!! between 1min to 3min each!
numpy is not so instable regarding API and behavior even if from time to time we have a bad surprise.
I would test oldest/newest and problematic ones, no ?

For iotest I don't think we have to go for a matrix f os/python/numpy.
It already take 13min with chaching file and installation.

apdavison:
In this PR you're already testing 5 NumPy versions, so I'm suggesting only one more. I would agree about removing osx testing, and maybe we can also run a smaller matrix on Windows.

samuelgarcia:
Lets discuss in this PR #1096

PR #1094, opened 2022-03-08
auto push to pypi on push tags

samuelgarcia:
@JuliaSprenger @apdavison @mdenker : have a look.
This auto push to pypi when we tag.
We have the same on spikeinertface.
Here we don't run the full io test which too long. Only the neo.core.
I think this is enough.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,34 @@
+name:  Release to PyPI
+
+on:
+  push:
+    tags:
+       - '*'
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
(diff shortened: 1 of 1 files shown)


apdavison:
This is a definite no for me. We should not make a release until the _full_ test suite passes (including IO) and the docs build correctly.

Also, this PR should target test.pypi.org.

If everything works, then we switch to the real PyPI.

apdavison:
also, shouldn't this only be for tags that match a specific pattern? tags aren't only used for releases.

samuelgarcia:
Hi Andrew.
Thanks for the feedback.
1. OK I will integrate the full test suite so with all dependencies.
2. for teh doc, yes I forgot about it
3. The test.pypi I never play with this, I don't figure out how to do this. I will investigate.
4 For the tag pattern, yes we should do that. Any idea ?

samuelgarcia:
Also we should check that neo/version.py do not contain any '.dev0' as a security.

JuliaSprenger:
@samuelgarcia Any updates on this? Would be nice to use this already for the next release :)

samuelgarcia:
I propose to close and reopen something later

PR #1096, opened 2022-03-09
Fix gh actions

samuelgarcia:


samuelgarcia:
1. I remove osx.it is quite faster
2. I put all numpy version in range we decide.

To avoid too much electricity consumption at every push.
1. I would propose a full matrix on ubuntu. (like now)
2. Only a sparse matrix for windows. like py3.7+numpy1.16 / py3.8+numpy1.19 / py3.9+numpy1.21
   Installation on windows is quite slow.

JuliaSprenger:
Yes, reducing the number of py+numpy versions on windows seems a good compromise.

JuliaSprenger:
Hi @samuelgarcia Sorry I lost track of this PR. If it's still up-to date from your side I would merge it now.

samuelgarcia:
I also forgot it.
Yes please merge this.

PR #1103, opened 2022-03-21
Extend OE for Neuropixels

alejoe91:
- The Neuropix-PXI plugin saves TTL in a `states.npy`. Extended the list of `_possible_event_stream_names` to reflect this.

- In some OE versions, the `units` field in the `structure.oebin` is blank, but the gain is always to `uV`. So this is automatically inferred in case it's missing


samuelgarcia (approved):

samuelgarcia:
OK for me.
@JuliaSprenger @apdavison : do you want to have a look.

JuliaSprenger:
Yes, looks also good to me. @alejoe91 Thanks for the improvements!

PR #1112, opened 2022-04-15
SpikeTrainList improvements

apdavison:
bug fix for annotations plus more consistent interface

### DIFF
--- a/neo/core/spiketrainlist.py
+++ b/neo/core/spiketrainlist.py
@@ -6,6 +6,7 @@ which supports a multiplexed representation of spike trains
 neuron/channel the spike is from).
 """
 
+import warnings
 import numpy as np
 import quantities as pq
 from .spiketrain import SpikeTrain, normalize_times_array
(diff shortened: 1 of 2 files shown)


neo/core/spiketrainlist.py
```diff
@@ -278,10 +298,14 @@ def _spiketrains_from_array(self):
...
-                    for name, value in self._annotations.items()
-                }
+                for name, value in self._annotations.items():
+                    if (not isinstance(value, str)
+                        and hasattr(value, "__len__")
+                        and len(value) == len(self._all_channel_ids)
+                       ):
+                        spiketrain.annotate(**{name: value[i]})
+                    else:
+                        spiketrain.annotate(**{name: value})
```
JuliaSprenger, line 308: This is introducing a kind of `array_annotation` across spiketrains. Does this imply you want to remove the array_annotation mechanism for single spiketrains in the long run?

neo/core/spiketrainlist.py
```diff
@@ -319,3 +343,51 @@ def multiplexed(self):
...
                 self._channel_id_array = np.hstack(channel_ids)
         return self._channel_id_array, self._spike_time_array
+
+    @property
+    def t_start(self):
+        if "t_start" in self._spiketrain_metadata:
+            return self._spiketrain_metadata["t_start"]
+        else:
+            t_start_values = unique([item.t_start for item in self._items
+                                    if isinstance(item, SpikeTrain)])  # ignore proxy objects
```
JuliaSprenger, line 353: proxy objects also have t_start and t_stop properties. Would it be a problem to also include them here?

JuliaSprenger: Or do you plan a ProxySpikeTrainList object instead in the future?

PR #1113, opened 2022-04-27
Add support for `.eeg` and `lfp` formats with neuroscope

h-mayorquin:
Right now, the neuroscope reader we only supports the `.dat` format:

https://github.com/NeuralEnsemble/python-neo/blob/7cf65eb0af48fa0c2b245da0c02fee04f63ad4f1/neo/rawio/neuroscoperawio.py#L67

However, other two common formats in which the the data for this format comes are `.eeg` and `lfp`. As an example, in the gin test we have a file with the `.eeg` format:

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/neuroscope/dataset_1

The following PR adds support for the two aforementioned formats while preserving the old functionality.


JuliaSprenger:
Hi @h-mayorquin, thanks for extending the IO! Could you also add the `.eeg` test to the set of files that the neuroscope tests are run on? Are you sure the file structure is always the same for `.dat`, `.eeg` and `.lfp` files? If you have a small `.lfp` file that you want to run the tests on we can also add this to the set of test files on gin.

JuliaSprenger:
Hi @h-mayorquin, thanks for extending the IO! Could you also add the `.eeg` test to the set of files that the neuroscope tests are run on? Are you sure the file structure is always the same for `.dat`, `.eeg` and `.lfp` files? If you have a small `.lfp` file that you want to run the tests on we can also add this to the set of test files on gin.

h-mayorquin:
> Hi @h-mayorquin, thanks for extending the IO! Could you also add the `.eeg` test to the set of files that the neuroscope tests are run on? Are you sure the file structure is always the same for `.dat`, `.eeg` and `.lfp` files? If you have a small `.lfp` file that you want to run the tests on we can also add this to the set of test files on gin.

I added the test for the `.eeg` file. I also added a test where we call the file through their `.dat` format instead of by its stem form.

Concerning the file structure, maybe @CodyCBakerPhD can comment on this. He might also know if there is some small `.lfp` files somewhere there that we can add to the gin repository.

CodyCBakerPhD:
`.lfp` is literally just a different suffix for the same file and structure. If you really want duplicated tests just for that, you can simply make a copy of the `.eeg` and change the suffix to `.lfp`.

h-mayorquin:
> `.lfp` is literally just a different suffix for the same file and structure. If you really want duplicated tests just for that, you can simply make a copy of the `.eeg` and change the suffix to `.lfp`.

Thanks, @CodyCBakerPhD!
What do you think @JuliaSprenger?

JuliaSprenger:
Hi again, @CodyCBakerPhD Thanks for the hint. I duplicated the `.eeg` file in the test dataset since it's tiny, see https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/79. Do you think any part of the content needs to be adjusted for those two files to reside in the same dataset?

CodyCBakerPhD:
> Do you think any part of the content needs to be adjusted for those two files to reside in the same dataset?

Nope - the number of channels in the `anatomicalDescription` field of the `.xml` tells you the length of that dimension of all binary blobs (`.dat`, `.lfp`, `.eeg`), and the other dimension of frames is always inferred from the file size conditioned on the number of such channels.

And both `.lfp` and `.eeg` get their sampling rate from the `lfpSamplingRate` of the `.xml`, or whatever that field is named.

h-mayorquin:
@CodyCBakerPhD @JuliaSprenger 
I am hoping that we can add a `xml_file` with non-canonical file name in the gin-repo (by canonical I mean something not equally named as the data file as it is right now: `file_name.dat` -> `file_name.xml`). This can be done by simply creating a copy of the currently existing file with a different name. We often have the situation where the file name is not in the canonical format, and having it on gin will allow us to test against this case.

samuelgarcia:
Hi Herberto,
This is ok for me.
Maybe we could have an implentation that loop over possible extension and llok for the good one.
So the "main" entry point would remain the xml file.
What do you think ?
But if you need this quickly I can merge this now.

CodyCBakerPhD:
@h-mayorquin @samuelgarcia Sorry, just seeing this now - I had hoped for this PR to add an optional input argument identical to the previous treatment in spikeextractors (https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py#L77).

Reason being...

(a) in numerous datasets, we've seen multiple .xml files in the same folder each corresponding to different uses, and only one of which is the header associated with the .dat or .eeg/.lfp file.

(b) the 'correct' .xml header does not always share the same file stem as the corresponding .dat or .eeg/.lfp file.

(c) .xml header might even be in a different folder (not adjacent to .dat)

I think adding it as an optional argument whose default behaves the same as the current reader would allow the greater user control/flexibility with the least amount of magic and assumptions.

samuelgarcia:
Hi Cody and Herberto,
in neo I think that main entry used to be the xml file.
Maybe itcould be cool to keep this to not break previous code.

So we could maybe make the same as spikeextractors but with the resverse in mind.
xml file is the mandatory first arg.
binary_file (dat/eeg/lfp) could be the optional secondary args. If it is none this one is automatic like before but we loop over some extension to find it.

So we would have the same flexibility with before retro compatibility on neo side.

CodyCBakerPhD:
> Should we add some non-standardly named xml file to the tests?

You can always just copy/rename files in individual tests on the fly. Technically could have done that for the .lfp data on GIN as well.

JuliaSprenger:
Hi @h-mayorquin I think you can do all the restructuring of the input handling in this PR, unless merging this is very urgent for you.

For the additional cases of input files I opened a PR to add another fake dataset to the collection on gin, have a look here: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/80 Is this covering all cases you want to test for? If you want to test files from different directories, you could even load the xml from dataset `test1` and dat from dataset `test2`.

h-mayorquin:
> Hi @h-mayorquin I think you can do all the restructuring of the input handling in this PR, unless merging this is very urgent for you.
> 
> For the additional cases of input files I opened a PR to add another fake dataset to the collection on gin, have a look here: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/80 Is this covering all cases you want to test for? If you want to test files from different directories, you could even load the xml from dataset `test1` and dat from dataset `test2`.

Yes, that's fine. We can test any other case with symbolic links magic on the fly as Cody is suggesting. I will just wait for the tests to be re-calculated so I can build against them. Is it @samuelgarcia who can review the gin repo?

samuelgarcia:
HI,
gin PR is merged now.

h-mayorquin:
@samuelgarcia @JuliaSprenger 
OK, this is working locally for me. I added another test for the discussed scenario. An `xml` file with different name than the binary file and specified the latter through the new argument which I called `binary_file`. 

Let me know if something else is missing.

neo/test/rawiotest/test_neuroscoperawio.py
```diff
@@ -1,18 +1,61 @@
...
+        assert 'nb_block' not in txt, msg
+
+        reader.parse_header()
+
+        txt = reader.__repr__()
+        msg = 'After parser_header() nb_block should be known'
+        assert 'nb_block' in txt, msg
+        # print(txt)
+
+        #
```
JuliaSprenger, line 40:
```suggestion

```

neo/test/rawiotest/test_neuroscoperawio.py
```diff
@@ -1,18 +1,61 @@
...
+        compliance.read_spike_times(reader)
+        compliance.read_spike_waveforms(reader)
+        compliance.read_events(reader)
+        compliance.has_annotations(reader)
+
+        # basic benchmark
+        level = logging.getLogger().getEffectiveLevel()
+        logging.getLogger().setLevel(logging.INFO)
+        compliance.benchmark_speed_read_signals(reader)
+        logging.getLogger().setLevel(level)
```
JuliaSprenger, line 57: I am not sure if we should run the speed test for every pytest execution. @samuelgarcia What do you think?

JuliaSprenger (commented):
Thanks for integrating the new testdataset. I made some minor suggestion for improvements, but in general it looks pretty ready to merge for me.

neo/test/rawiotest/test_neuroscoperawio.py
```diff
@@ -1,18 +1,61 @@
...
+        'neuroscope/test1/test1',
+        'neuroscope/test1/test1.dat',
+        'neuroscope/dataset_1/YutaMouse42-151117.eeg'
     ]
 
+    def test_binary_argument_with_non_canonical_xml_file(self):
+
+        local_test_dir = get_local_testing_data_folder()
+        filename = local_test_dir / "neuroscope/test2/recording.xml"
+        binary_file = local_test_dir / "neuroscope/test2/signal1.dat"
```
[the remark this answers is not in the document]
h-mayorquin: Any standard way to do this?
The first thing that comes to my head is just to load this:

```python
   np.memmap(data_file_path, dtype=sig_dtype, mode='r', offset=0).reshape(-1, nb_channel)
```

And compare this to the reader `self._raw_signals` 
What do you think?

neo/rawio/neuroscoperawio.py
```diff
@@ -64,8 +105,9 @@ def _parse_header(self):
         else:
             raise (NotImplementedError)
 
-        self._raw_signals = np.memmap(filename + '.dat', dtype=sig_dtype,
-                                      mode='r', offset=0).reshape(-1, nb_channel)
+        self._raw_signals = np.memmap(data_file_path, dtype=sig_dtype,
+                                      mode='r',
+                                      offset=0).reshape(-1, nb_channel)
```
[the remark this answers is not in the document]
h-mayorquin: Done

neo/test/rawiotest/test_neuroscoperawio.py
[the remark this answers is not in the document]
JuliaSprenger: Maybe its least confusing if in the `parse_header`  `self.binary_file` actually contains the filename of the file that is loaded? I.e. after the filename resolution running `self.binary_file = data_file_path`?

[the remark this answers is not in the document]
h-mayorquin: Ok, I was thinking that we wanted to check the data somehow but if filename should be enough I added a comparison along the lines that you described. I also separated the logic to resolve the data and xml file paths in a function of its own. 

Check the new changes and let me know if that works for you.

neo/test/rawiotest/test_neuroscoperawio.py
JuliaSprenger: @samuelgarcia ping?

neo/rawio/neuroscoperawio.py
```diff
@@ -122,3 +137,38 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
...
+        elif suffix in supported_data_extensions:
+            xml_file_path = file_path.with_suffix(".xml")
+            data_file_path = file_path
+        else:
+            error_string = (
+                f"Format {suffix} not supported "
+                f"filename format should be {supported_data_extensions} or .xml"
+            )
+            raise KeyError(error_string)
+
```
JuliaSprenger, line 172:
```suggestion
assert xml_file_path.is_file(), f"xml file not found at  expected location {xml_file_path}"
```

neo/test/rawiotest/test_neuroscoperawio.py
```diff
@@ -1,18 +1,61 @@
...
+        reader = NeuroScopeRawIO(filename=filename, binary_file=binary_file)
+
+        txt = reader.__repr__()
+        msg = 'Before parser_header() nb_block should be NOT known'
+        assert 'nb_block' not in txt, msg
```
JuliaSprenger, line 31: Instead of checking the representation of the reader, wouldn't it be cleaner to check if the attribute `nb_block` is set, or in general no header information is present?

neo/test/rawiotest/test_neuroscoperawio.py
JuliaSprenger: @samuelgarcia ping?

neo/test/rawiotest/test_neuroscoperawio.py
JuliaSprenger:
```suggestion
        msg = 'Before parser_header() no header information should be present'
        assert reader.header is None, msg
```

JuliaSprenger (commented):
Hi @h-mayorquin Thanks for the update, looks very nice now. I only made a few tiny line change suggestions and pinged @samuelgarcia again for the benchmarking in the tests.

neo/test/rawiotest/test_neuroscoperawio.py
JuliaSprenger:
```suggestion
```

JuliaSprenger:
@h-mayorquin Thanks for the final fixes. I will merge this now :fireworks:

PR #1117, opened 2022-05-12
[CI] use conda environment for IO tests

### LINKED ISSUE #1115: partially broken download of test data
In the latest runs of GH action the first attempt to fetch a data file from GIN fails with the error
```
datalad.support.exceptions.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none cat-file blob git-annex:remote.log' failed with exitcode 128 under /home/runner/ephy_testing_data [err: 'fatal: path 'remote.log' does not exist in 'git-annex'']
```
See attempt 2 of https://github.com/NeuralEnsemble/python-neo/runs/6325659861?check_suite_focus=true
The same test run in attempt 1 were successfull 17 days earlier. 

The two attempts use the same versions of the datalad, git-annex are the same. `git` differs only in a minor version (`1:2.35.1-0ppa1~ubuntu20.04.1` vs `1:2.36.0-0ppa1~ubuntu20.04.1`). I didn't succeed in running the tests with the older `git` version again, but this seems to unlikely the reason for the failing tests, at it's on the git-annex-gin level. The respective test file on gin did not change in the last year.

@samuelgarcia Do you have any more ideas what to check? Should we removed the cached test files to check if this is the issue?

JuliaSprenger:
This should fix #1115, by installing more recent versions of datalad and git-annex.

Analogous to the current version, the conda environment is cached and only rebuild when one of the requirement files changes or a new month started. The execution time of the tests when everything is already cached is the same as before (~12 minutes).

### DIFF
--- a/.github/workflows/io-test.yml
+++ b/.github/workflows/io-test.yml
@@ -20,11 +20,11 @@ jobs:
         # "macos-latest", "windows-latest"
         os: ["ubuntu-latest", ]
         python-version: ['3.8', ]
+    defaults:
+      # by default run in bash mode (required for conda usage)
+      run:
+        shell: bash -l {0}
     steps:
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
 
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


JuliaSprenger:
@samuelgarcia Can you have a look at this?

.github/workflows/io-test.yml
```diff
@@ -33,49 +33,51 @@ jobs:
...
-          echo "::set-output name=HASH_EPHY_DATASET::$(git ls-remote https://gin.g-node.org/NeuralEnsemble/ephy_testing_data.git HEAD | cut -f1)"
-
-      - uses: actions/cache@v2
-        id: cache-datasets
-        with:
-          path: ~/ephy_testing_data
-          key: ${{ runner.os }}-datasets-${{ steps.vars.outputs.HASH_EPHY_DATASET }}
-          restore-keys: |
-            ${{ runner.os }}-datasets
+          pip install --upgrade -e .
```
samuelgarcia, line 77: Are you sure that this upgrade option do recurssively also upgrade numpy and quantities ?
The upgrade rules have change a bit.

environment_testing.yml
```diff
@@ -0,0 +1,8 @@
+name: neo-test-env
+channels:
+  - conda-forge
+dependencies:
+  - datalad
+  - pip
+  - pip:
+    - -r requirements_testing.txt
```
samuelgarcia, line 8: Why no remove this "requirements_testing.txt" and have everything at the same place ?

JuliaSprenger: To be nice to people who prefer to use pip for installing (almost) everything. You can still feed the `requirements_testing.txt` to pip if you prefer that, if I put everything in the yml this approach does not work any more..

.github/workflows/io-test.yml
JuliaSprenger: Even if it does not do recursive upgrades with the next month the default versions of these packages will be updated. Do you think this lag is too long? If you want to force recursive dependency upgrades I can add a `--upgrade-strategy "eager"`, but this means the tests might have a longer runtime until a new month starts.

samuelgarcia: ok

PR #1121, opened 2022-05-19
NWB Inspector Tests

legouee:
NWB Inspector is a new tool to help inspect NWB files for compliance with NWB Best Practices : 
https://github.com/NeurodataWithoutBorders/nwbinspector

The following pull request allows the validation of files created with Neo with respect to NWB Inspector.

### DIFF
--- a/neo/io/nwbio.py
+++ b/neo/io/nwbio.py
@@ -45,6 +45,8 @@ try:
     from pynwb.misc import AnnotationSeries
     from pynwb import image
     from pynwb.image import ImageSeries
+    from pynwb.file import Subject
+    from pynwb.epoch import TimeIntervals
     from pynwb.spec import NWBAttributeSpec, NWBDatasetSpec, NWBGroupSpec, NWBNamespace, \
         NWBNamespaceBuilder
     from pynwb.device import Device
@@ -54,6 +56,9 @@ try:
     have_pynwb = True
 except ImportError:
     have_pynwb = False
+
+import nwbinspector
+from nwbinspector import inspect_nwb, check_regular_timestamps
 
 # hdmf imports
 try:
@@ -275,6 +280,9 @@ class NWBIO(BaseIO):
         if "file_create_date" in self.global_block_metadata:
             self.global_block_metadata["file_datetime"] = self.global_block_metadata[
                 "rec_datetime"]
+        if "subject" in self.global_block_metadata:
+            self.global_block_metadata["subject"] = self.global_block_metadata[
+                "subject"]
 
         self._blocks = {}
         self._read_acquisition_group(lazy=lazy)
@@ -352,8 +360,6 @@ class NWBIO(BaseIO):
             except JSONDecodeError:
                 # For NWB files created with other applications, we put everything in a single
                 # segment in a single block
-                # todo: investigate whether there is a reliable way to create multiple segments,
-                #       e.g. using Trial information
                 block_name = "default"
                 segment_name = "default"
             else:
@@ -441,8 +447,16 @@ class NWBIO(BaseIO):
                 raise Exception("Writing to NWB requires an annotation 'session_start_time'")
         self.annotations = {"rec_datetime": "rec_datetime"}
         self.annotations["rec_datetime"] = blocks[0].rec_datetime
-        # todo: handle subject
+        self.annotations = {"subject": "subject"}
         nwbfile = NWBFile(**annotations)
+        if "subject" not in annotations:
+            nwbfile.subject = Subject(
+                                subject_id="subject_id",
+                                age="P0D", #Period x days old
+                                description="no description",
+                                species="Mus musculus", # by default
+                                sex="U", # unknown
+                            )
         assert self.nwb_file_mode in ('w',)  # possibly expand to 'a'ppend later
         if self.nwb_file_mode == "w" and os.path.exists(self.filename):
             os.remove(self.filename)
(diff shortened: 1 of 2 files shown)


neo/io/nwbio.py
```diff
@@ -441,8 +453,14 @@ def write_all_blocks(self, blocks, **kwargs):
...
         self.annotations["rec_datetime"] = blocks[0].rec_datetime
-        # todo: handle subject
+        self.annotations = {"subject": "subject"}
         nwbfile = NWBFile(**annotations)
+        if "subject" not in annotations:
+            nwbfile.subject = Subject(subject_id="subject_id",
+                                      age="P0D",  # Period x days old
+                                      description="no description",
+                                      species="Mus musculus",  # by default
+                                      sex="U")
```
JuliaSprenger, line 463: These are default placeholders, right? Maybe it would be good to explicitely use dummy labels here, e.g. `empty_neo_subject_id` or `neo_dummy_subject_id`?

neo/io/nwbio.py
```diff
@@ -458,15 +476,16 @@ def write_all_blocks(self, blocks, **kwargs):
...
 
         if sum(statistics(block)["Epoch"]["count"] for block in blocks) > 0:
-            nwbfile.add_epoch_column('_name', 'the name attribute of the Epoch')
-            # nwbfile.add_epoch_column('_description', 'the description attribute of the Epoch')
-            nwbfile.add_epoch_column(
-                'segment', 'the name of the Neo Segment to which the Epoch belongs')
-            nwbfile.add_epoch_column('block',
-                                     'the name of the Neo Block to which the Epoch belongs')
+            nwbfile.add_trial_column('segment', 'name of the Segment to which the Epoch belongs')
+            nwbfile.add_trial_column('block', 'name of the Block to which the Epoch belongs')
```
JuliaSprenger, line 480: if nwb is supporting the `epoch` concept, what is the reason to switch to `trials` now?

neo/io/nwbio.py
```diff
@@ -458,15 +476,16 @@ def write_all_blocks(self, blocks, **kwargs):
...
 
+        arr = [[], []]  # epoch array for ascending t_start and t_stop
         for i, block in enumerate(blocks):
-            self.write_block(nwbfile, block)
+            block_name = block.name
+            self.write_block(nwbfile, block, arr)
+        arr2 = np.sort(arr)
+        self._write_epoch(nwbfile, arr2, block)
```
JuliaSprenger, line 487: `self.write_block` and `self._write_epoch` receive the same arguments, but in different order. Maybe this would be good to harmonize?

neo/io/nwbio.py
```diff
@@ -611,23 +658,54 @@ def _write_spiketrain(self, nwbfile, spiketrain):
...
+                                timestamps=event.times.rescale('second').magnitude,
+                                unit=str(event.units),
+                                description=event.description or "",
+                                comments=json.dumps(hierarchy))
+
         nwbfile.add_acquisition(tS_evt)
         return tS_evt
 
-    def _write_epoch(self, nwbfile, epoch):
+    def _write_manage_epoch(self, nwbfile, segment, epoch, arr):
```
JuliaSprenger, line 682: This method could benefit from having a docstring. What exactly is `arr` here?

JuliaSprenger (commented):
It seems like the mapping between the neo nwb structure is still developing. Maybe it would be good to introduce a mapping version that facilitates the correct reading of the nwb file?

neo/io/nwbio.py
```diff
@@ -55,6 +57,13 @@
 except ImportError:
     have_pynwb = False
 
+try:
+    import nwbinspector
```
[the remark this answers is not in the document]
legouee: I added the extra installation requirement in the setup.py file.

neo/io/nwbio.py
```diff
@@ -441,8 +453,14 @@ def write_all_blocks(self, blocks, **kwargs):
                 raise Exception("Writing to NWB requires an annotation 'session_start_time'")
         self.annotations = {"rec_datetime": "rec_datetime"}
         self.annotations["rec_datetime"] = blocks[0].rec_datetime
-        # todo: handle subject
+        self.annotations = {"subject": "subject"}
```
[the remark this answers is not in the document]
legouee: This annotation is not necessary.

neo/io/nwbio.py
legouee: arr is an epoch array with t_start and t_stop sorted in ascending order.
I don't think it's necessary to create a docstring for this function. It's just an intermediate step in order to respect NWBInspector.

neo/io/nwbio.py
legouee: Yes. You are right. As it supports the epoch concept, it is not necessary to have the trials function.
I put it back as before.

neo/io/nwbio.py
```diff
@@ -620,26 +681,56 @@ def _write_event(self, nwbfile, event):
...
-                                            epoch.labels):
-            nwbfile.add_epoch(t_start, t_start + duration, [label], [],
-                              _name=epoch.name,
-                              segment=segment.name,
-                              block=segment.block.name)
+                                            epoch.labels,
+                                            ):
+            for j in [label]:
+                t_stop = t_start + duration
+                seg_name = "%s %s" % (epoch.segment.name, label)
```
JuliaSprenger, line 714: Here occurs the same error as with the events earlier. You have to keep a reference to the segment from before the proxy object is loaded.

neo/io/nwbio.py
JuliaSprenger: If you sort t_start and t_stop independently, doesn't this corrupt data in case one epoch duration is contained in another?

PR #1130, opened 2022-06-21
Enable NWBIO to write lazy data objects

JuliaSprenger:
@legouee @apdavison What do you think of these changes?


legouee:
It's great. Thank you Julia.

PR #1132, opened 2022-07-01
import datatlad only when running download_dataset()

samuelgarcia:
This remove the systematic import od datalad which can be annoying in notebook for some version because it trigger lot of errors.
datalad will be now imported on demand only.

### DIFF
--- a/neo/utils/__init__.py
+++ b/neo/utils/__init__.py
@@ -5,5 +5,4 @@ Utils package
 from .misc import (get_events, get_epochs, add_epoch,
     match_events, cut_block_by_epochs, cut_segment_by_epoch,
     is_block_rawio_compatible)
-from .datasets import (download_dataset,
-    get_local_testing_data_folder, HAVE_DATALAD)
+from .datasets import download_dataset, get_local_testing_data_folder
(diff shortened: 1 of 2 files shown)


JuliaSprenger:
Hi Samuel, I am a bit confused as in the title it sounds like you only want to import datalad in `download_dataset`, but your changes try to import datalad also in multiple other modules.
Maybe a better approach would be to try to download the dataset and catch the `datalad missing` exception when this fails?

samuelgarcia:
the `try .. except` is only at in subfolder tests.
The only real place is in the function.

neo/test/rawiotest/tools.py
```diff
@@ -5,15 +5,19 @@
...
 def can_use_network():
     """
     Return True if network access is allowed
     """
+    try:
+        import datalad
+        HAVE_DATALAD = True
+    except:
+        HAVE_DATALAD = False
+
```
JuliaSprenger, line 20:
```suggestion

```

JuliaSprenger:
Ok, I will merge this as soon as the tests are completed.

PR #1137, opened 2022-07-14
[neuralynx] expose parsed version of file headers

JuliaSprenger:
This permits extracting file metadata as io attributes. External tools don't need to parse file headers to retrieve additional, non-annotated metadata.


JuliaSprenger:
@samuelgarcia What do you think of this? This way the header only needs to be read once if subsequent tools also require header information.

samuelgarcia:
OK.
Can I merge ?

JuliaSprenger:
Yes!

PR #1140, opened 2022-07-22
Fix proble when priting the number of channel and channel streams

samuelgarcia:
Fix proble when priting the number of channel and channel streams

### DIFF
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -844,6 +844,6 @@ def pprint_vector(vector, lim=8):
         part2 = ' , '.join(e for e in vector[-lim // 2:])
         txt = f"[{part1} ... {part2}]"
     else:
-        part1 = ', '.join(e for e in vector[:lim // 2])
+        part1 = ', '.join(e for e in vector)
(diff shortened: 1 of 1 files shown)


neo/rawio/baserawio.py
```diff
@@ -844,6 +844,6 @@ def pprint_vector(vector, lim=8):
         part2 = ' , '.join(e for e in vector[-lim // 2:])
         txt = f"[{part1} ... {part2}]"
     else:
-        part1 = ', '.join(e for e in vector[:lim // 2])
+        part1 = ', '.join(e for e in vector)
```
JuliaSprenger, line 847: In this case you are only printing the first (by default 4) entries without indicating that there might be still more. Why don't you want to print all entries?

samuelgarcia: I want to print all entries but there were a but for `len(vector) <= lim` only half was printed because of a bad copy/paste.
now it is print entirly.
no ?

JuliaSprenger: sorry, I missread the change. My bad. Yes, this makes completely sense!

PR #1150, opened 2022-08-11
support for just .spikes in folder

ferchaure:
Hi I was trying to load an openephys _.spikes_ file but it required the .continuous file as well. To fix that, I separated a bit better the data loading, all the code was there. 
And I fixed a typo when the filename is read to check the number of segments.

This will work for my case, but probably a similar fix could be made to load only events files.

Cheers,
Fer

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -492,11 +492,11 @@ def explore_folder(dirname):
             if (seg_index + 1) > info['nb_segment']:
                 info['nb_segment'] += 1
         elif filename.endswith('.spikes'):
-            s = filename.replace('.spikes', '').split('_')
-            if len(s) == 2:
-                seg_index = 0
+            s = re.findall("(_\d+)$", filename.replace('.spikes', ''))
+            if s:
+                seg_index = int(s[0][1:]) - 1
             else:
-                seg_index = int(s[2]) - 1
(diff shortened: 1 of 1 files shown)


ferchaure:
Maybe the error is related to a unreported renaming in my data, I will check it that and fix it

samuelgarcia:
Hi fernando.
Thank you. this is Ok for me.
Can you fix the warning line 495 ?

neo/rawio/openephysrawio.py
```diff
@@ -485,11 +492,11 @@ def explore_folder(dirname):
             if (seg_index + 1) > info['nb_segment']:
                 info['nb_segment'] += 1
         elif filename.endswith('.spikes'):
-            s = filename.replace('.spikes', '').split('_')
-            if len(s) == 1:
-                seg_index = 0
+            s = re.findall("(_\d+)$", filename.replace('.spikes', ''))
```
samuelgarcia, line 495: pep8 give a warning. I guess maybe you have to do \\ no ?
can you check ?

JuliaSprenger: @samuelgarcia Feel free commit the suggestion and see if it resolves the pep8 issue.

alejoe91:
Closing in favor of #1377

PR #1152, opened 2022-08-13
improves blackrock spikes in nev

ferchaure:
Hi, 
I tried to open a .nev file (2.3 version) and I found t2o bugs and fix them:

-The timestamps gives one element less that the amount of spikes. I follow that issue to the _seg_t_stops definition and how is used to create a slice. Before the highest element in the nev file was defined as the t_stops and used to slice... keeping the last element outside the range. I just add a sample before converting to time and the appending to _seg_t_stops 

-It wasn't possible to load the waveforms. The reshape in _get_spike_raw_waveforms throw an error related to sizes mismatch. In a few words, it had to many bytes for the waveforms (in my case 96 when each waveform should have 92). I could increase the bytes for the reserved fields of the Spikes fields in the __nev_data_types ... but given that each channels has its own waveform_dtypes maybe it has to be resize for each case. For that reason I create that additional dt1 type, to remove the extra bytes knowing how many bytes should be there for the waveform.

Cheers,
Fernando


neo/rawio/blackrockrawio.py
```diff
@@ -645,8 +646,13 @@ def _get_spike_raw_waveforms(self, block_index, seg_index, unit_index, t_start,
 
         wf_dtype = self.__nev_params('waveform_dtypes')[channel_id]
         wf_size = self.__nev_params('waveform_size')[channel_id]
+        wf_byte_size = np.dtype(wf_dtype).itemsize * wf_size
 
-        waveforms = unit_spikes['waveform'].flatten().view(wf_dtype)
+        dt1 = [
+            ('extra', 'S{}'.format(unit_spikes['waveform'].dtype.itemsize - wf_byte_size)),
+            ('ch_waveform', 'S{}'.format(wf_byte_size))]
```
samuelgarcia, line 653: How did you get this specification of extra bytes ? is it in the official documents ?

ferchaure: Not really, the code looked like it follows the documentation, but I saw that the available bytes for the waveforms are bigger that the ones expected given the dtype.  Maybe the reserved bytes ( the ones before the waveform data) must be dynamic and not always uint8 (but I couldn't find documentation to support that theory). For that reason I did that dt1 type, just to extract the expected bytes

samuelgarcia: But how we can be sure that extra bytes are before ?
And what if extra size == 0 ?

ferchaure: good question, I don't know, I just tested it in a file and there is the relationship with the reserved bytes that are before. (I can share the file if you want)
I checked the case when extra size == 0 and it's fine, it just creates an empty binary field without errors

ferchaure:
The last commit edits the segment stop just to extract spikes times (adding a sample to let it work fine with the stop of a slice), it fixes the same without editing the _seg_t_stops attribute

mdenker:
Decision was made during a Neo discussion to add a flag to the loader that can be used to select the alternate waveform loading.

samuelgarcia (approved):

PR #1153, opened 2022-08-16
Read multi-column csvs in PhyIO

Kleinjohann:
PhyIO automatically loads csv files in the data folder, but assumes that they only have two columns. This causes an error e.g. when loading a SpikeInterface folder with quality metrics which are saved in a single multi-column 'metrics.csv' file.

I extended the csv loading routine to allow for multiple columns, and while doing that I noticed that negative floats currently cause an error since they are not covered by the regular expression that is being used to detect floats, so I fixed that, too. 
Additionally, I added tests covering the multi-column and negative float cases.


JuliaSprenger:
I had to add an dummy commit to retrigger CI , but tests are passing now and the changes look good.

PR #1158, opened 2022-08-26
Array-annotate Amplitudes and PCs in PhyIO

Kleinjohann:
PhyIO was written before array_annotations were fully functional, so it had TODOs in the comments to deal with data which should be array_annotated once that's possible.
I added array_annotations for PCs and amplitudes to all spikes, and since the PCs can be quite a lot of data I added a flag that can be toggled to load them.

### DIFF
--- a/neo/io/phyio.py
+++ b/neo/io/phyio.py
@@ -7,6 +7,6 @@ class PhyIO(PhyRawIO, BaseFromRaw):
     description = "Phy IO"
     mode = 'dir'
 
-    def __init__(self, dirname):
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi Alexander,
This is OK for me.
Maybe I would also add an option (false by default) to load or not the amplitudes of spikes like you did for PCs.
I guess that amplitudes can be an extra big vector.

Kleinjohann:
Hi Samuel, thanks for having a look! I added a flag for the amplitudes as well.

PR #1159, opened 2022-08-29
Openephys check consistency experiments

alejoe91:
This PR improves the checks the `OpenEphysBinaryRawIO` by:
1. pre-mapping folder structure in terms of recording nodes, experiments, and recordings
2. assigning `block_index` and `seg_index` after pre-mappping (it is possible that some experiments/recordings are deleted by the user because of errors etc.)
3. adds checks to ensure that:
    - experiments (blocks) are consistent acrosse multiple nodes (if present)
    - streams are consistent across recordings (segments)
    - the same streams are present in all experiments (blocks)
In case the latter check fails, there is a new option to use the argument `experiment_names` to select a subset of experiments with consistent streams to be loaded.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -485,7 +485,7 @@ def check_folder_consistency(folder_structure, possible_experiment_names=None):
         segment_stream_names = None
         if len(experiment['recordings']) > 1:
             for recording in experiment['recordings']:
-                stream_names = sorted(list(recording['streams'].keys()))
+                stream_names = sorted(list(recording['streams']['continuous'].keys()))
                 if segment_stream_names is None:
                     segment_stream_names = stream_names
                 assert segment_stream_names == stream_names, \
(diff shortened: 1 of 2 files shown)


alejoe91:
@samuelgarcia i don't know why core tests failed! I only change one rawio file... can you check and in case retrigger?

neo/rawio/openephysbinaryrawio.py
```diff
@@ -349,72 +355,155 @@ def explore_folder(dirname):
...
-                            event_stream[f'{name}_npy'] = str(npz_filename)
-
-                    all_streams[block_index][seg_index]['events'][stream_name] = event_stream
-
-    # TODO for later: check stream / channel consistency across segment
-
-    return all_streams, nb_block, nb_segment_per_block
+            if node_name not in folder_structure:
+                folder_structure[node_name] = {}
+                folder_structure[node_name]['experiments'] = []
```
samuelgarcia, line 378: Do we naturally sort this list after exploring the folder ?

alejoe91: mmm maybe we should actually!

alejoe91: @samuelgarcia done: both for experiments (block_index) and for recordings (seg_index)

samuelgarcia:
Thanks Alessio.
This is really cool.

For me this OK to be merge.

@JuliaSprenger : I don't known if you want to have a look.
We need it for the release.

samuelgarcia:
Test fails from something unrelated (neurlynx) I re run the test to check.
Unfortunatly the entire dataset will be download again (because test did not finish...)

Issue #1160, opened 2022-08-29
Remove the HAVE_PACKAGE logic and module import of extra dependency.


samuelgarcia:
We have this kind of logic for many IOs
```python
try:
    import sonpy
    HAVE_SONPY = True
except ImportError:
    HAVE_SONPY = False

class SomeClass
   def __init__(self):
        assert HAVE_SONPY, 'sonpy must be installed'
```

This make importing many modules for every `import neo` because all extra dependencies are at module level.
I propose to move to "import on demand" logic with this

```python

class SomeClass
   def __init__(self):
        import sonpy
```
I propose to do this (it is easy) for the next release 0.11.0

@JuliaSprenger @apdavison : are you OK with this ?

apdavison:
Yes, for IOs I think this makes sense, since it is rare to use more than one or two IOs in a given script

samuelgarcia:
OK I start this soon.

PR #1163, opened 2022-08-30
Local import and remove HAVE_PACKAGE logic

samuelgarcia:
This PR remove unnecessary import at module level and import extra dependency only when they are needed when instanciating an IO class.

See #1160

This should:
  * help in complicated environement (by not importing everything)
  * decrease import duration

The drawback:
  * it raise error only when instanciate (run time) and not at load time.

### DIFF
--- a/neo/io/igorproio.py
+++ b/neo/io/igorproio.py
@@ -17,13 +17,6 @@ import quantities as pq
 from neo.io.baseio import BaseIO
 from neo.core import Block, Segment, AnalogSignal
 
-try:
-    import igor.binarywave as bw
-    import igor.packed as pxp
-    from igor.record.wave import WaveRecord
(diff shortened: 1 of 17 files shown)


samuelgarcia:
@alejoe91

alejoe91 (approved):

JuliaSprenger:
Hi @samuelgarcia 
Thanks for the simplifications. Did you confirm this is reducing the import time?
I think we are loosing at tiny bit of user friendliness here, but it also simplifies the code, so let's go with it.

JuliaSprenger:
For reference: This change improves the import time on my system from 1.6s to 0.5s. @samuelgarcia It works :+1:

samuelgarcia:
cool.
To be totaly honest I did not make benchmark :)

Issue #1118, opened 2022-05-13
Spiketrainlist object comparison fails in rare cases


JuliaSprenger:
In some cases the neo core tests fail due to an object comparison of a list and a spiketrainlist, see 
- https://github.com/NeuralEnsemble/python-neo/runs/6422039865?check_suite_focus=true
- https://github.com/NeuralEnsemble/python-neo/runs/5718538078?check_suite_focus=true 
- https://github.com/NeuralEnsemble/python-neo/runs/5659519369?check_suite_focus=true.
This error is only happening in some of the test runs, it seems only to occur in tests on windows, but not linked to one python version. @apdavison: Do you have any idea why this fails only in some cases and can we do adjust the spiketrainlist in some way to permit an object comparison like this?

JuliaSprenger:
And here it happened again: https://github.com/NeuralEnsemble/python-neo/runs/6524684429?check_suite_focus=true

JuliaSprenger:
@apdavison And here it happened again: https://github.com/NeuralEnsemble/python-neo/runs/7540753511

Is this tests supposed to fail with a low probability? There is a comment here `# for debugging occasional test failure`

apdavison:
@JuliaSprenger I spent some time looking into this a few weeks ago, which is when I added that debugging line (which shouldn't have been merged/pushed, sorry). If I remember correctly, it occurs when you happen to get a spike train list of length 0, and the error is somewhere in the filter functionality, but I didn't get as far as fixing it.

JuliaSprenger:
Tests failed again twice for the current master, see https://github.com/NeuralEnsemble/python-neo/actions/runs/3331652474/jobs/5609080600

PR #1169, opened 2022-09-26
Update Open Ephys events

alejoe91:
### Updates for v0.6:

In the new OE version, `timestamps.npy` are already in seconds and do not need scaling by the sampling frequency (https://open-ephys.github.io/gui-docs/User-Manual/Recording-data/Binary-format.html#events)

### Other updates:

In case the event stream has `states.npy`, the event channel type is an `epoch` since the duration can be computed using positive and negative states for each TTL channel.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -66,6 +66,7 @@ class OpenEphysBinaryRawIO(BaseRawIO):
         self.experiment_names = experiment_names
         self.load_sync_channel = load_sync_channel
         self.folder_structure = None
+        self._use_direct_evt_timestamps = None
 
     def _source_name(self):
         return self.dirname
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @alejoe91 Thanks for the update. Unfortunately it seems like your changes are messing up the array_annotations of old style datasets. Could you have a look at this?
Also we can add a dataset in the new format to gin for testing. Do you have a small one available?

alejoe91:
@JuliaSprenger I think I fixed it :) however, another core test seems to be failing..can you take a look?

neo/rawio/openephysbinaryrawio.py
```diff
@@ -204,6 +204,32 @@ def _parse_header(self):
...
 
+                    # # If available, use 'states' to compute event duration
+                    if 'states' in d:
+                        states = d["states"]
+                        timestamps = d["timestamps"]
+                        labels = d["labels"]
+                        rising = np.where(states > 0)[0]
+                        falling = np.where(states < 0)[0]
+                        # make sure first event is rising and last is falling
+                        if states[0] < 0:
```
JuliaSprenger, line 215: What about the case where states == 0 or multiple values in the beginning or end are smaller or bigger than 0 respectively?

JuliaSprenger:
Tests are passing now, the failure was one of those rare spiketrainlist errors, see #1118

neo/rawio/openephysbinaryrawio.py
alejoe91: I thin that `states` are by definition either `+channel_number` or `-channel_number`.

@jsiegle could you confirm that?

alejoe91:
> Hi @alejoe91 Thanks for the update. Unfortunately it seems like your changes are messing up the array_annotations of old style datasets. Could you have a look at this? Also we can add a dataset in the new format to gin for testing. Do you have a small one available?

The `v0.6.x_neuropixels_multiexp_multistream` is already from the new version. You now have `sample_numbers.npy` (samples) and `timestamps.npy` (seconds).

neo/rawio/openephysbinaryrawio.py
jsiegle: That's right, `states` can never be zero. However, it is possible that first event can be falling, e.g. if a TTL line was high when the recording started. Same with the last event being rising.

alejoe91: That's taken care of :) in that case the first/last event is dropped

JuliaSprenger:
> > Hi @alejoe91 Thanks for the update. Unfortunately it seems like your changes are messing up the array_annotations of old style datasets. Could you have a look at this? Also we can add a dataset in the new format to gin for testing. Do you have a small one available?
> 
> The `v0.6.x_neuropixels_multiexp_multistream` is already from the new version. You now have `sample_numbers.npy` (samples) and `timestamps.npy` (seconds).

I don't think this session is included in tests yet. Could you add it to also ensure the new format is read correctly?

alejoe91:
It is tested already! Both for [rawio](https://github.com/alejoe91/python-neo/blob/update_oe_events/neo/test/rawiotest/test_openephysbinaryrawio.py#L16) and [iotests](https://github.com/alejoe91/python-neo/blob/97509e6c11a9a56cda9ac11786f99c4019c265b6/neo/test/rawiotest/test_openephysbinaryrawio.py#L16)

JuliaSprenger:
Sorry, I missed that reference. Then this is complete for me.
Merging now :)

PR #1170, opened 2022-09-29
Improved pretty

apdavison:


JuliaSprenger:
Hi @apdavison These pretty representations are only used by the ipython lib or is there any way on how these are used by python itself? What is the advantage of having this verbose representation instead of the `__repr__` implementation?

apdavison:
Yes, these are used in IPython and Jupyter notebooks. Python also has the `pprint` module. This PR is basically just improving the "pretty" representations for Epoch, Event and SpikeTrain; the representations for AnalogSignal, Segment, Block are already ok.

PR #1171, opened 2022-09-29
Allow epoch duration to be a float

apdavison:
in which case the duration units will be taken from the `units` argument or from the units of the `times` argument.


JuliaSprenger:
Nice catch, the `units` were not properly documented yet. Looks good to me.

PR #1173, opened 2022-09-30
Create issue templates

apdavison:


JuliaSprenger:
This looks like a good first set of issue templates to me. I guess we might want to update / add more details later on, but this is a good start already. :+1:

PR #1174, opened 2022-10-01
BlackrockRawIO FileSpec 3.0

### LINKED ISSUE #795: Updates to Blackrock File Format (3.0)
Blackrock is expecting a change to our file format in an upcoming release (I think at least a couple of months away), so I wanted to reach out to you to make sure it is on your radar for incorporation into neo. As always, we will allow people access to our previous file formats, and the adoption will be slow, so this would be supported in addition to the current ones that you load. 

The primary change is a move to 64 bit timestamps. The other change is the inclusion of an additional event type in our Neural Events File: an indicator for the recording pausing, resuming, and stopping. 

We can provide the file spec as soon as it is all signed off on our side, which should be in just a couple of weeks, but I just wanted to throw this in here so that it could be planned on. 

Let me know if there is anything else you'll need from me besides example files.

cboulay:
Closes #795 
[Blackrock_FileSpec_3_0.zip](https://github.com/NeuralEnsemble/python-neo/files/9689684/Blackrock_FileSpec_3_0.zip)

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -179,44 +179,64 @@ class BlackrockRawIO(BaseRawIO):
         self.__nsx_header_reader = {
             '2.1': self.__read_nsx_header_variant_a,
             '2.2': self.__read_nsx_header_variant_b,
-            '2.3': self.__read_nsx_header_variant_b}
+            '2.3': self.__read_nsx_header_variant_b,
+            '3.0': self.__read_nsx_header_variant_b,
+        }
         self.__nsx_dataheader_reader = {
             '2.1': self.__read_nsx_dataheader_variant_a,
             '2.2': self.__read_nsx_dataheader_variant_b,
-            '2.3': self.__read_nsx_dataheader_variant_b}
+            '2.3': self.__read_nsx_dataheader_variant_b,
+            '3.0': self.__read_nsx_dataheader_variant_b
+        }
         self.__nsx_data_reader = {
             '2.1': self.__read_nsx_data_variant_a,
             '2.2': self.__read_nsx_data_variant_b,
-            '2.3': self.__read_nsx_data_variant_b}
+            '2.3': self.__read_nsx_data_variant_b,
+            '3.0': self.__read_nsx_data_variant_b
+        }
         self.__nsx_params = {
             '2.1': self.__get_nsx_param_variant_a,
             '2.2': self.__get_nsx_param_variant_b,
-            '2.3': self.__get_nsx_param_variant_b}
+            '2.3': self.__get_nsx_param_variant_b,
+            '3.0': self.__get_nsx_param_variant_b
+        }
         # NEV
         self.__nev_header_reader = {
             '2.1': self.__read_nev_header_variant_a,
             '2.2': self.__read_nev_header_variant_b,
-            '2.3': self.__read_nev_header_variant_c}
+            '2.3': self.__read_nev_header_variant_c,
+            '3.0': self.__read_nev_header_variant_c,
+        }
         self.__nev_data_reader = {
             '2.1': self.__read_nev_data_variant_a,
             '2.2': self.__read_nev_data_variant_a,
-            '2.3': self.__read_nev_data_variant_b}
+            '2.3': self.__read_nev_data_variant_b,
+            '3.0': self.__read_nev_data_variant_b
+        }
         self.__waveform_size = {
             '2.1': self.__get_waveform_size_variant_a,
             '2.2': self.__get_waveform_size_variant_a,
-            '2.3': self.__get_waveform_size_variant_b}
+            '2.3': self.__get_waveform_size_variant_b,
+            '3.0': self.__get_waveform_size_variant_b
+        }
         self.__channel_labels = {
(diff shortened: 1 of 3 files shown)


JuliaSprenger:
Hi @cboulay Thanks for adding support for the version 3.0. Do you mind if I add the files you provided to our set of test files on [gin](https://gin.g-node.org/Neuralensemble/ephy_testing_data), so we can continually test these? In that case whom can I list as official contributor of the files? Also I noticed the corresponding ccf file is relatively large for not being actually used by the reader. Is there some redundant information in there that we could remove for the usage as test file?

samuelgarcia:
Hi Chadwick, this is really really cool.

JuliaSprenger:
I added the files to the gin repository (https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/86). As soon as that PR is merged it would be good to include the new dataset in the neo tests. Could you then add the corresponding reference in the [RawIO tests here](https://github.com/NeuralEnsemble/python-neo/blob/714c9c29c3cfb2244839a2a341b6b4a0e95aea60/neo/test/rawiotest/test_blackrockrawio.py#L26) and in the [IO tests here](https://github.com/NeuralEnsemble/python-neo/blob/714c9c29c3cfb2244839a2a341b6b4a0e95aea60/neo/test/iotest/test_blackrockio.py#L45)?

JuliaSprenger:
The PR on gin is merged. You should now be able to add the files to the tests.

cboulay:
Thanks @JuliaSprenger .

I don't love the idea of comparing to the Matlab loader. I actually don't have a Matlab license at the moment.

Would it be sufficient to test the shape, the type, the range of values, and other metrics like that?
If the values need to be tested directly then I can do that for 1-2 channels.

JuliaSprenger:
Yes, I agree. Comparing the data loaded to an alternative way to access the same data is would be a very advanced test already. As a first step just verifying that the IO can load these files and that some data values are loaded is completely fine. If you are motivated to also add a test checking for the correctness of these values then storing some of these as plain text is also an option (as it's done e.g. in the [NeuralynxIO tests](https://github.com/NeuralEnsemble/python-neo/blob/714c9c29c3cfb2244839a2a341b6b4a0e95aea60/neo/test/iotest/test_neuralynxio.py#L286)).

cboulay:
I got around to attempting to write the unit tests but I failed. It seems I can't run unit tests for a module unless I have the complete environment created. I cannot install `git-annex` using the DataLad installer instructions because I get `datalad_installer.MethodNotSupportedError: Darwin OS not supported`.

I put the files in the correct location manually but then they were erased when I attempted to run a test. That seems intentional.
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/utils/datasets.py#L64-L67

Is there another way?

cboulay:
I spun up my Linux box and I was able to run the tests. I identified an issue. There seems to be a large number of segments in the .nev file but not in the .ns6 file. Currently this is manifesting as an error when testing spike count compliance, but that's a misdirection.

The real reason is that some events are written with newer timestamps than the spikes that are being written. I will open a new issue about that.

cboulay:
Tests pending outcome of #1181

cboulay:
[375e474](https://github.com/NeuralEnsemble/python-neo/pull/1174/commits/375e474bfa3328ced07afc892e74d46e9c0a6447) fix for #1181 is necessary for the tests to pass if we want to use this same example file. If the proposed solution of ignoring all unhandled events is too broad then we can add the problematic event types to the list of handled events then explicitly ignore those problematic types only.

Even if Blackrock decides to stop storing the events that are creating the issue, there might still be some files in the wild that have this issue so we need some kind of solution.

cboulay:
Hi @JuliaSprenger , I decided not to add anymore tests than just the common ones unless you have any specific requests. Please let me know if anything else is required before merging.

neo/rawio/blackrockrawio.py
```diff
@@ -1113,14 +1142,18 @@ def __get_event_segment_ids(self, raw_event_data, masks, nev_data_masks):
...
             reset_ev_mask = self.__get_reset_event_mask(raw_event_data, masks, nev_data_masks)
             reset_ev_ids = np.where(reset_ev_mask)[0]
 
             # consistency check for monotone increasing time stamps
-            # explicitly converting to int to allow for negative diff values
-            jump_ids = \
-                np.where(np.diff(np.asarray(raw_event_data['timestamp'], dtype=int)) < 0)[0] + 1
+            # - Use logical comparator (instead of np.diff) to avoid unsigned dtype issues.
+            # - Only consider handled/known event types.
+            mask_handled = np.any(np.vstack([value[nev_data_masks[key]] for key, value in masks.items()]), axis=0)
```
JuliaSprenger, line 1152: Are you sure the `vstack` is needed here? I think you don't need it, unless you noticed a performance improvement when using it.

JuliaSprenger:
> Hi @JuliaSprenger , I decided not to add anymore tests than just the common ones unless you have any specific requests. Please let me know if anything else is required before merging.
The common tests are fine for now as these automatically cover the new lines you added

JuliaSprenger:
Tests are passing now. @cboulay Thanks for the work you put into this!

PR #1180, opened 2022-10-05
Draft to move from setup.py to pyproject.toml

samuelgarcia:
Hi @apdavison @JuliaSprenger 
I think the new trend is to move from dynamic setup (setup.py) to static declaration (pyproject.toml).
There somes peps about this and I think it help many tools for building and packaging.

Here a first draft for this.
Normally everything is centralize in pyproject.toml.

It still use setuptools for building but the medata come from pyproject.toml.

If you have time to make feeback on this. It would be cool.


.github/workflows/io-test.yml
```diff
@@ -74,6 +74,7 @@ jobs:
 
       - name: Install neo
         run: |
+          pip install .[test]
```
apdavison, line 77: shouldn't this go after the next line? Or just add `[test]` at the end of the next line?

neo/__init__.py
```diff
@@ -2,12 +2,12 @@
 Neo is a package for representing electrophysiology data in Python,
 together with support for reading a wide range of neurophysiology file formats
 '''
+import importlib.metadata
+__version__ = importlib.metadata.version("neo")
```
apdavison, line 6: I think this line should go after all the import statements.

pyproject.toml
```diff
@@ -0,0 +1,84 @@
+[project]
+name = "neo"
+version = "0.12.0.dev0"
+
+authors = [
```
apdavison, line 5: I think either we list _all_ authors, or we use the same approach as in the current setup.py, and give the name as "Neo authors and contributors"

pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+  { name="Samuel Garcia", email="samuel.garcia@cnrs.fr" },
+  { name="Andrew Davison", email="andrew.davison@cnrs.fr" },
+  { name="Julia Sprenger", email="julia.sprenger@univ-amu.fr" },
+  { name="Michael Denker", email="m.denker@fz-juelich.de" },
+]
+
+description = "Neo is a package for representing electrophysiology data in Python, together with support for reading a wide range of neurophysiology file formats"
+
+readme = "README.rst"
```
apdavison, line 14: I presume this replaces the "long_description" field?

apdavison (commented):
This looks good. I think doc/source/conf.py will need to be updated to get the version string from the metadata file.

.github/workflows/io-test.yml
samuelgarcia: I did not wanted to upgrade the [test] dependencies.
It can be put after.

pyproject.toml
samuelgarcia: Sure. I forgot about that.

pyproject.toml
samuelgarcia: description is the short one from what I understand.
For the long one. I don't known how to include yet.
I forgot also about that.
Maybe the readme replace it no ?

samuelgarcia:
>  doc/source/conf.py will need to be updated to get the version string from the metadata file.

Oups. I also forgot that.
Is it still necessary in sphinx to put this variable or is it automatic now ?
Do when parse the pyproject.toml for this ?

samuelgarcia:
@apdavison : other question.
Do we keep this minimal setup.py ?

Or do we switch  to ?
  * `pip install .` only instead of `python setup.py install`
  * `pip install -e .` only instead of `python setup.py develop`
  * `build` package of python with `python -m build --wheel`  instead of `python setup.py sdist`

I will no modify NOT the doc because you are working inanother PR  on it but maybe you could propagate this also no ?

pyproject.toml
```diff
@@ -0,0 +1,80 @@
...
+nixio = ['nixio>=1.5.0']
+stimfitio = ['stfio']
+tiffio = ['pillow']
+edf = ['pyedflib']
+ced = ['sonpy']
+nwb = ['pynwb']
+maxwell = ['h5py']
+biocam = ['h5py']
+
+# TODO [all]
```
samuelgarcia, line 77: @JuliaSprenger please note that I have not find a way to do the all the moment.

JuliaSprenger: I think for implementing the 'all' keyword in this framework we need to duplicate the requirements here. It's not nice, but acceptable I think.

PR #1183, opened 2022-10-11
Better `id` for Open Ephys Events

alejoe91:
The Open Ephys event channel ids were set to the `stream_index`, which made it not very informative.
This PR sets the event `id` to be the same as the event `name`: e.g. `Message Center`, `PXIe`, `RhythmFPGA`, etc.


samuelgarcia:
OK for me.
@JuliaSprenger  ?

PR #1187, opened 2022-10-21
Fix installation of "all" dependencies

### LINKED ISSUE #1184: not all extra_requirements pip installable
**What is the problem?**
With the new extra_requirements["all"] field users will try to install neo from pypi using `pip install neo[all]`. However at least one dependency listed can not be installed via this, as `stfio` is not a PyPI package.

**Suggestions for fixing the problem**
We should find a way to either specify the sftio dependency in a way that it can be automatically installed or exceptionally exclude it from the list of "all" extra requirements.

JuliaSprenger:
Fixes #1184

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,9 @@ extras_require = {
 }
 extras_require["all"] = sum(extras_require.values(), [])
 
+# explicitly removing stfio from list of all as it is not pip installable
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
@samuelgarcia Can we merge this before the release?

samuelgarcia (approved):

PR #1190, opened 2022-10-25
[IO] Extend utilities to find suitable IO for a dataset

JuliaSprenger:
This PR adds an additional utility function for finding a suitable IO for a given dataset. In addition to the `get_io` functions, there's now also an `list_candidate_ios` function, that provides a list of all IOs supporting the formats detected in a given path. This list is generated based on the file extension of the file or the extensions of the matching files in a given folder, potentially also scanning subfolders if required.

I added a common test to verify the correct IO for each test dataset is found, extended the list of linked extensions for IOs where required and removed some outdated test utility functions.


JuliaSprenger:
@samuelgarcia What do you think of this approach? With some file formats being supported by multiple IOs (e.g. `.txt` for the AsciiSignal  and AsciiSpikeTrainIO) we can not always determine 100% the suitable IO, but with this approach users can still retrieve a very reduced set of IOs to test.

neo/test/iotest/common_io_test.py
```diff
@@ -518,4 +514,11 @@ def test__handle_pathlib_filename(self):
             elif self.ioclass.mode == 'dir':
                 self.ioclass(dirname=pathlib_filename,
                              *self.default_arguments,
-                             **self.default_keyword_arguments)
+                             **self.default_keyword_arguments))
```
JuliaSprenger, line 517:
```suggestion
                             **self.default_keyword_arguments)
```

PR #1204, opened 2022-12-13
Tdt format compatibility

JulesLebert:
Hi, I am not sure if that should be a pull request or just a question on my end for the compatibility spikeinterface/neo. If these changes are too specific to my data format, I will just keep using my forked repository of Neo. I am just following @samuelgarcia suggestion to continue this discussion on the neo side (https://github.com/SpikeInterface/spikeinterface/pull/1138)

There are two changes in this pull request:

The first one is that the regular expression to find the .sev files was not working for me, even for the test data (https://gin.g-node.org/NeuralEnsemble/ephy_testing_data)

sev_regex =

With changes in the regular expression, I was able to load the test data:

sev reqex

The other changes are specific to my data (but are also compatible with the test data).
My tdt data are in the following format:

Screenshot 2022-12-13 at 14 31 29

This is one recording session (or one neo segment), where BB_2 and BB_3 are one 32 channel array and BB_4 and BB_5 another 32 channel array. Both 32 channel arrays are split into two streams (or TDT ‘stores’) of 16 channels.

Some changes I had to do to load the data in neo are a few changes in the lines following the regular expression. Because the regular expression was returning the .sev files from all stores.

The other change is an addition in the conditions of the mask for the tsq data. I'm not sure why, but even though the TankEvType of the different stream in info_channel_group corresponds to the default value of EVTYPE_STREAM (33025), the ‘evtype’ in the tsq files for my data containing the analog signals are 33041 instead of 33025. Therefore, adding another condition in the mask allows me to load the data (otherwise data_index returns an empty array)

array( 33041 , dtype-int32)

After all these changes, test data can be loaded as usual:

Screenshot 2022-12-13 at 15 06 22

As well as my own tdt data:

Screenshot 2022-12-13 at 15 07 55

And when I compare the Neo data to the data loaded by the tdt python package (https://www.tdt.com/docs/sdk/offline-data-analysis/offline-data-python/00_Intro/) they are the same:

sampling rate

With all this in mind, and following my question on the spikeinterface repository, is there a way to load the different streams into one, which would allow me to spike sort everything in one go through spikeinterface instead of having to run 4 different spike sorter with one for each stream?


neo/rawio/tdtrawio.py
```diff
@@ -248,14 +249,21 @@ def _parse_header(self):
                         sev_filename = (path / sev_stem).with_suffix('.sev')
                     else:
                         # for single block datasets the exact name of sev files in not known
-                        sev_regex = f".*_ch{chan_id}.sev"
+                        sev_regex = f"*_Ch{chan_id}.sev"
```
JuliaSprenger, line 252: I think it would be good to also preserve the old matching behaviour here:
```suggestion
                        sev_regex = f"*_[cC]h{chan_id}.sev"
```

neo/rawio/tdtrawio.py
```diff
@@ -248,14 +249,21 @@ def _parse_header(self):
...
                         sev_filename = list(self.dirname.parent.glob(str(sev_regex)))
+                        # in case multiple sev files are found, try to find the one for current stream
+                        if len(sev_filename) > 1:
```
JuliaSprenger, line 255: I think this can be simplified a bit via
```suggestion
                        if sev_filename:
```

neo/rawio/tdtrawio.py
```diff
@@ -197,7 +197,8 @@ def _parse_header(self):
                 for seg_index, segment_name in enumerate(segment_names):
                     # get data index
                     tsq = self._tsq[seg_index]
-                    mask = (tsq['evtype'] == EVTYPE_STREAM) & \
+                    mask = ((tsq['evtype'] == EVTYPE_STREAM) | \
```
JuliaSprenger, line 200: Do you have an idea why the Stream code in your system is different? In any case the new code should be defined on the class level as the current `EVTYPE_STREAM`.

JuliaSprenger (commented):
Hi @JulesLebert Thanks for your detailed description of the issue and your fix for it! I added a few suggestions. Can you have a look and see if you can still load your files including my suggestions?

neo/rawio/tdtrawio.py
JulesLebert: This seems to work fine with my files

neo/rawio/tdtrawio.py
JulesLebert: This will not work with the test data, as when len(sev_filename) is exactly one, the format of the file is not necessarily f'*_{store}_Ch{chan_id}.sev'. I think the initial suggestion is more appropriate

neo/rawio/tdtrawio.py
JulesLebert: I am not sure why it is different to be honest. What's puzzling me is that the 'TankEvType' in the TBK file is the default EVTYPE_STREAM (or 33025) but the data themselves in the tsq files have an 'evtype' of 33041 for my neural data streams. Therefore, just changing the value of the current EVTYPE_STREAM will not work. I could add a class EVTYPE_STREAM_2 or something similar and check if evtype is equal to either of the EVTYPE_STREAM variables?

JulesLebert:
Hi, sorry for the late response. Here are some answers to the suggestions.

neo/rawio/tdtrawio.py
```diff
@@ -197,7 +197,8 @@ def _parse_header(self):
...
+                           (tsq['evtype'] == int(33041))) & \
```
samuelgarcia, line 201: Instead of of hardcoded I would create a global vriable for this number ssomething like 
`EVTYPE_STREAM_VARIANT=33041` to avoid hidden code.

JuliaSprenger:
Hi @JulesLebert, any news on this?

JulesLebert:
Hi @JuliaSprenger I have just made the suggested changes and it seems to be compatible with both the test data and my data

JuliaSprenger (approved):

neo/rawio/tdtrawio.py
```diff
@@ -562,6 +570,7 @@ def read_tbk(tbk_filename):
 EVTYPE_STROFF = int('00000102', 16)  # 258
 EVTYPE_SCALAR = int('00000201', 16)  # 513
 EVTYPE_STREAM = int('00008101', 16)  # 33025
+EVTYPE_STREAM_VARIANT = int(33041)
```
JuliaSprenger, line 573:
```suggestion
EVTYPE_STREAM_VARIANT = int('00008111', 16)  # 33041
```

neo/rawio/tdtrawio.py
```diff
@@ -594,4 +603,4 @@ def is_tdtblock(blockpath):
     if file_ext >= tdt_ext:  # if containing all the necessary files
         return True
     else:
-        return False
+        return False
```
JuliaSprenger, line 606:
```suggestion
        return False

```

JuliaSprenger:
Hi @JulesLebert. Thanks for including the comments. I made some minor reformatting. Looks ready to merge for me now.

PR #1213, opened 2023-02-02
Warning instead of assertion if legacy Open Ephys timestamps are not monotonically increasing

### LINKED ISSUE #1210: Error when loading OpenEphys .continuous files, but only in some recording sessions
Hey!

I’m In the process of analysing data recorded with the Open Ephys system (v0.4.6) using 16 tetrode shuttle drives (64ch). I've used the Spikeinterface toolbox and I’m having problems loading .continuous files in some recordings (while other sessions have no issues loading at all). The error exists within the neo site-package, so I was hoping someone here would have some advice for how I can solve this. 

When I execute “recordings.append(si.read_openephys(recording_path[i], stream_name='Signals CH'))” I get the following error:

```
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\IPython\core\interactiveshell.py", line 3433, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-314ef08b03de>", line 53, in <module>
    recordings.append(si.read_openephys(recording_path[i], stream_name='Signals CH'))
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\spikeinterface\extractors\neoextractors\openephys.py", line 220, in read_openephys
    recording = OpenEphysLegacyRecordingExtractor(folder_path, **kwargs)
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\spikeinterface\extractors\neoextractors\openephys.py", line 56, in __init__
    NeoBaseRecordingExtractor.__init__(self, stream_id=stream_id,
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 43, in __init__
    _NeoBaseExtractor.__init__(self, block_index, **neo_kwargs)
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 25, in __init__
    self.neo_reader = get_reader(self.NeoRawIOClass, **neo_kwargs)
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 15, in get_reader
    neo_reader.parse_header()
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\neo\rawio\baserawio.py", line 179, in parse_header
    self._parse_header()
  File "C:\Users\joach\anaconda3\envs\si_env\lib\site-packages\neo\rawio\openephysrawio.py", line 117, in _parse_header
    assert np.all(diff == RECORD_SIZE), \
AssertionError: Not continuous timestamps for 100_CH1.continuous. Maybe because recording was paused/stopped.
```

I don’t remember pausing or stopping this recording, so I don’t think this is the reason for the error. I also don’t think the data is corrupted, as I am able to read it in my Matlab pipieline using OpenEphys’ own loading function “[data, timestamps, info] = load_open_ephys_data(inputdata)”.

I get this error with a handful of recording sessions in my dataset, and I really need to get it to work but I have no clue how to fix this error. Any insight as to how it could be fixed would be very much appreciated.

I've uploaded one of my .continuous files that elicits this error, along with the respective settings file here:
https://drive.google.com/drive/folders/1LR4zth7zluBofwv2QDiPo2ArugkyF5x9?usp=sharing

**Environment:**
 - OS: Windows 11
 - Python version 3.9.15
 - Neo version 0.11.1
 - NumPy version 1.21.6

Best regards
Jo

alejoe91:
fixes #1210


alejoe91:
Hi @JuliaSprenger 

Is it ok if we relax the assertion to a warning in this case? See https://github.com/NeuralEnsemble/python-neo/issues/1210

I can change the failing test accordingly.

JuliaSprenger:
Hi @alejoe91 I think a warning might be too relaxed in this case as even for significant pauses this might be missed by the user. What the deviation of timestamps you observed in the test files provided by @ThePuppetShow? Maybe we can make a reasonable estimate of what deviations are acceptable and should only raise a warning?

alejoe91:
If I remember correctly, there were a bunch of zeros at the beginning of the timestamps file, but we might need to double check.

On the other hand, the assertion would completely prevent from loading the data in case something is wrong with timestamps. I think that the API should allow to at least load the data, even just to check what is wrong and maybe cut out "good" portions of the recording. No?

ThePuppetShow:
Hey, and thanks a lot for looking into this

I got this issue in some but not all my recordings. To my knowledge none of the recordings should not have started with any pauses, so I'm not sure why sometimes the timestamps were not directly from 0. Maybe a lag of sorts? 

In one of these sessions the first timestamp was at 1.536 s. I made some plots here https://github.com/NeuralEnsemble/python-neo/issues/1210 

Would it be possible to at least allow for the data to be accessed, but give a warning? Then I can just adjust it myself, if the timestamps would get mismatched somehow. Or just automatically scale it if it starts with a slight delay?

Again, thanks a lot for taking the time to look into this : ) 

Best 
J

JuliaSprenger:
Hi @alejoe91 I would argue that small deviations could pass with just a warning, but for bigger deviations an error should be raised. For debugging purposes I would propose to add a keyword to the IO that permits to ignore that error and still return the data (with wrong timestamps). This way the user would need to actively decide to load incorrect timestamps and it is not happening silently.

alejoe91:
That's fair! :) Something like `ignore_timestamps_errors`?

alejoe91:
@JuliaSprenger should be done

neo/rawio/openephysrawio.py
```diff
@@ -115,9 +116,11 @@ def _parse_header(self):
 
                 # check for continuity (no gaps)
                 diff = np.diff(data_chan['timestamp'])
-                assert np.all(diff == RECORD_SIZE), \
-                    'Not continuous timestamps for {}. ' \
-                    'Maybe because recording was paused/stopped.'.format(continuous_filename)
+                if not np.all(diff == RECORD_SIZE) and not self._ignore_timestamps_errors:
+                    raise Exception(
+                        'Not continuous timestamps for {}. ' \
```
JuliaSprenger, line 121:
```suggestion
                    raise ValueError(
                        'Not continuous timestamps for {}. ' \
```

neo/test/rawiotest/test_openephysrawio.py
```diff
@@ -22,6 +22,12 @@ def test_raise_error_if_discontinuous_files(self):
             'openephys/OpenEphys_SampleData_2_(multiple_starts)'))
         with self.assertRaises(Exception):
```
JuliaSprenger, line 23:
```suggestion
        with self.assertRaises(ValueError):
```

JuliaSprenger (commented):
Hi @alejoe91 thanks for implementing the new flag.
I noticed that until now this was raising a plain Exception and not a more specific Error. Do you agree this error should be more specific?

neo/test/rawiotest/test_openephysrawio.py
alejoe91: done

neo/rawio/openephysrawio.py
alejoe91: done

neo/rawio/openephysrawio.py
```diff
@@ -115,9 +116,11 @@ def _parse_header(self):
 
                 # check for continuity (no gaps)
                 diff = np.diff(data_chan['timestamp'])
```
samuelgarcia, line 118: We have to keep in mind that this diff have a very high cost when opening.
on spiekinterface side this means when parallel processing making n_jobs time this check!!!

JuliaSprenger:
Failing tests are due to #1227 and not related to this PR.

neo/rawio/openephysrawio.py
JuliaSprenger: @samuelgarcia Feel free to make a proposal on this in a separate PR.

PR #1219, opened 2023-02-16
added code and tests for MED codec

dancrepeau:


apdavison:
Many thanks for this contribution, @dancrepeau 

As you can see, the NeoIOTest check is failing, with `ModuleNotFoundError: No module named 'dhn_med_py'`

You will need to update `pyproject.toml` (probably you will have to update your master branch, this file was only added recently), and add all dependencies for MedIO to the `[project.optional-dependencies]` section, at least to the "test" and "all" lists.

PR #1220, opened 2023-02-17
Changes to support MED codec

dancrepeau:
Hello,

Here is the second attempt for the MED codec.  The pyproject.toml is now updated with the dhn_med_py dependency.  I specified >=0.9.5 for the version, but if the pip install always just uses the latest available, then perhaps that specification is unnecessary.  Thank you!

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -68,9 +68,10 @@ class OpenEphysRawIO(BaseRawIO):
     extensions = ['continuous', 'openephys', 'spikes', 'events', 'xml']
     rawmode = 'one-dir'
 
-    def __init__(self, dirname=''):
+    def __init__(self, dirname='', ignore_timestamps_errors=False):
         BaseRawIO.__init__(self)
         self.dirname = dirname
+        self._ignore_timestamps_errors = ignore_timestamps_errors
 
     def _source_name(self):
         return self.dirname
@@ -115,9 +116,11 @@ class OpenEphysRawIO(BaseRawIO):
 
                 # check for continuity (no gaps)
                 diff = np.diff(data_chan['timestamp'])
-                assert np.all(diff == RECORD_SIZE), \
-                    'Not continuous timestamps for {}. ' \
-                    'Maybe because recording was paused/stopped.'.format(continuous_filename)
+                if not np.all(diff == RECORD_SIZE) and not self._ignore_timestamps_errors:
+                    raise ValueError(
+                        'Not continuous timestamps for {}. ' \
+                        'Maybe because recording was paused/stopped.'.format(continuous_filename)
+                    )
 
                 if seg_index == 0:
                     # add in channel list
--- a/neo/test/rawiotest/test_openephysrawio.py
+++ b/neo/test/rawiotest/test_openephysrawio.py
@@ -20,8 +20,14 @@ class TestOpenEphysRawIO(BaseTestRawIO, unittest.TestCase, ):
         # It must raise an error
         reader = OpenEphysRawIO(dirname=self.get_local_path(
             'openephys/OpenEphys_SampleData_2_(multiple_starts)'))
-        with self.assertRaises(Exception):
+        with self.assertRaises(ValueError):
             reader.parse_header()
+        # if ignore_timestamps_errors=True, no exception is raised
+        reader = OpenEphysRawIO(dirname=self.get_local_path(
+            'openephys/OpenEphys_SampleData_2_(multiple_starts)'),
+                                ignore_timestamps_errors=True)
+        reader.parse_header()
+
 
     def test_raise_error_if_strange_timestamps(self):
         # In this dataset CH32 have strange timestamps


neo/rawio/medrawio.py
```diff
@@ -0,0 +1,238 @@
...
+    Uses the dhn-med-py MED python package (version >= 0.9.5), created by
+    Dark Horse Neuro, Inc.
+    
+    Currently reads the entire MED session into one block/segment/stream.  If
+    channels have differing sampling frequencies, all frequencies are upsampled
+    to the highest frequency prior to returning the chunk.
+    
+    Timestamps generated are referenced to the beginning of the session,
+    with the beginning of the session being timestamp zero.  Gaps in the data
+    are ignored for the purposes of timestamping.
```
JuliaSprenger, line 28: Ignoring gaps of arbitrary sizes can lead to very wrong timestamps without the user having a chance to notice this. If you want to permit loading data sets with small variations in the timestamps of a continuous signal I would propose to introduce a flag for this, e.g. `ignore-timestamp-gaps`. In my opinion if a gap is larger than the sampling rate this discrepancy should not be silently ignored, but either raise an error or be split into multiple segments. @samuelgarcia What do you think about this?

neo/rawio/medrawio.py
```diff
@@ -0,0 +1,238 @@
...
+            signal_channels.append((ch_name, chan_id, sr, dtype, units, gain, offset, stream_id))
+    
+        signal_channels = np.array(signal_channels, dtype=_signal_channel_dtype)
+
+        # no unit/epoch information contained in MED
+        spike_channels = []
+        spike_channels = np.array(spike_channels, dtype=_spike_channel_dtype)
+        
+        # Events
+        # events are not being read by this Neo wrapper
```
JuliaSprenger, line 112: How complicated would it be to add events here? Depending on the data these might be essential for preprocessing or analysis...

neo/rawio/medrawio.py
```diff
@@ -0,0 +1,238 @@
...
+
+        stream_id = self.header['signal_streams'][stream_index]['id']
+        
+        num_channels = 0
+        if channel_indexes is None:
+            self.sess.set_channel_active('all')
+            num_channels = self.num_channels_in_session
+        else:
+            if any(channel_indexes < 0):
+                raise IndexError('No channels selected!')
```
JuliaSprenger, line 180:
```suggestion
                raise IndexError(f'Can not index negative channels: {channel_indexes}')
```

neo/rawio/medrawio.py
```diff
@@ -0,0 +1,238 @@
...
+            # Set all channels to be inactive, then selectively set some of them to be active
+            self.sess.set_channel_inactive('all')
+            for i, channel_idx in enumerate(channel_indexes):
+                num_channels += 1
+                self.sess.set_channel_active(str(self.header['signal_channels'][channel_idx][0]))
```
JuliaSprenger, line 185: I think the channel name should be already a string, so you probably don't need to convert again here. Also to improve readability maybe write it like this:
```suggestion
                self.sess.set_channel_active(self.header['signal_channels']['channel_name'][channel_idx])
```

neo/test/iotest/test_medio.py
```diff
@@ -0,0 +1,128 @@
...
+from neo.test.iotest.tools import get_test_file_full_path
+from neo.io.proxyobjects import (AnalogSignalProxy,
+                SpikeTrainProxy, EventProxy, EpochProxy)
+from neo import (AnalogSignal, SpikeTrain)
+
+import quantities as pq
+import numpy as np
+
+
+# This run standart tests, this is mandatory for all IO
```
JuliaSprenger, line 19:
```suggestion
# This run standard tests, this is mandatory for all IOs
```

neo/test/iotest/test_medio.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertEqual(len(seg.analogsignals), 1)
+        # Test that the correct number of samples are read, 5760000 samps for 3 channels
+        self.assertEqual(seg.analogsignals[0].shape[0], 5760000)
+        self.assertEqual(seg.analogsignals[0].shape[1], 3)
+        
+        # Test the first sample value of all 3 channels, which are
+        # known to be [-1, -4, -4]
+        self.assertEqual(seg.analogsignals[0][0][0], -1)
+        self.assertEqual(seg.analogsignals[0][0][1], -4)
+        self.assertEqual(seg.analogsignals[0][0][2], -4)
```
JuliaSprenger, line 58: This can be simplified a bit using numpy
```suggestion
        np.testing.assert_array_equal(seg.analogsignals[0,0,:3], [-1, -4, -4])
```

neo/test/iotest/test_medio.py
```diff
@@ -0,0 +1,128 @@
...
+            assert isinstance(st, SpikeTrain)
+            self.assertNotEqual(st.size, 0)
+
+        # annotations
+        #assert 'seg_extra_info' in seg.annotations
+        assert seg.name == 'Seg #0 Block #0'
+        for anasig in seg.analogsignals:
+            assert anasig.name is not None
+        for st in seg.spiketrains:
+            assert st.name is not None
```
JuliaSprenger, line 73: This does not make sense since the IO does not generate spiketrains
```suggestion
```

JuliaSprenger (commented):
Hi @dancrepeau 
Thanks for implementing the new IO. I added a few comments to the implementation.
Is the underlying dhn-med-py package inherently loading the data in lazy mode when initialising a session or are there some setting that could improve the lazy loading performance of the IO?

dancrepeau:
Hi Julia, thank you for the wonderful feedback.  I will go through and review the comments.  As for your question about lazy mode and the underlying library, it is acting in lazy mode.  The MedSession() constructor in parse_header is simply reading metadata and opening file pointers, while the get_matrix_by_sample() function in get_analogsignal_chunk() is doing the actual reading/decoding of the data from the files.

dancrepeau:
I like all of the changes suggested, and will implement them (this might take a few weeks).  Thanks again for the good comments!

PR #1221, opened 2023-02-20
Fix caching of testing data in CI

JuliaSprenger:
Currently the cached ephys data do not contain the latest hash in the caching key as the hash is not correctly forwarded to the github output variable.


JuliaSprenger:
Hi @samuelgarcia Could you have a look at this minimal change and confirm the caching key is non-empty now?

PR #1225, opened 2023-02-27
Add a flag for intanrawio indicating whether intan hardware has already applied a notch filter to raw data

zm711:
This is my first stab at incorporating a flag so that users can see whether or not their data has already had a notch filter applied. I couldn't think of a way to get it incorporated during the flow of the rest of the script, so I added the check right before the return to generate the flag.

@samuelgarcia


neo/rawio/intanrawio.py
```diff
@@ -360,6 +365,7 @@ def read_rhs(filename):
     ('desired_upper_bandwidth', 'float32'),
 
     ('notch_filter_mode', 'int16'),
+
```
JuliaSprenger, line 368:
```suggestion
```

JuliaSprenger (approved):

JuliaSprenger:
Thanks for the changes, they look good to me. The failing tests unrelated (#1227) to the changes here.

Issue #1224, opened 2023-02-27
`joblib` dependency?


joschaschmiedt:
I'm currently developing a Neo reader for our in-house HDF5-based data format following the guidance provided in the [developer's guide](https://neo.readthedocs.io/en/stable/io_developers_guide.html#io-dev-guide). I noticed that `joblib` is used in `BaseRawIO.__init__`:
https://github.com/NeuralEnsemble/python-neo/blob/d550b80f08fd28570284413397ed3698b86f704e/neo/rawio/baserawio.py#L158-L162
but not listed as a dependency in [pyproject.toml](https://github.com/NeuralEnsemble/python-neo/blob/master/pyproject.toml). Is that on purpose?

JuliaSprenger:
Good catch, looks like we forgot to list it explicitly. Can you open a PR to add it?

joschaschmiedt:
Sure. I added it as a hard dependency to `pyproject.toml` and the docs as well: #1226

As I'm currently on Windows I cannot run all the tests as DataLad doesn't work at the moment.

PR #1226, opened 2023-02-28
Add joblib dependency

joschaschmiedt:
joblib is used in `BaseRawIO.__init__` and is a hard dependency if caching is used (#1224). I therefore added it to the project's dependencies.

### DIFF
--- a/doc/source/developers_guide.rst
+++ b/doc/source/developers_guide.rst
@@ -41,6 +41,7 @@ Requirements
     * Python_ 3.8 or later
     * numpy_ >= 1.19.5
     * quantities_ >= 0.14.1
+    * joblib >= 1.2.0
     * pytest (for running tests)
     * Sphinx_ (for building documentation)
 
--- a/doc/source/developers_guide.rst
+++ b/doc/source/developers_guide.rst
@@ -41,7 +41,7 @@ Requirements
     * Python_ 3.8 or later
     * numpy_ >= 1.19.5
     * quantities_ >= 0.14.1
-    * joblib >= 1.2.0
+    * joblib >= 1.0.0
     * pytest (for running tests)
     * Sphinx_ (for building documentation)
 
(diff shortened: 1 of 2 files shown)


doc/source/developers_guide.rst
```diff
@@ -41,6 +41,7 @@ Requirements
     * Python_ 3.8 or later
     * numpy_ >= 1.19.5
     * quantities_ >= 0.14.1
+    * joblib >= 1.2.0
```
JuliaSprenger, line 44:
```suggestion
    * joblib >= 1.0.0
```

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "packaging",
     "numpy>=1.19.5",
     "quantities>=0.14.1",
+    "joblib>=1.2.0"
```
JuliaSprenger, line 28:
```suggestion
    "joblib>=1.0.0"
```

JuliaSprenger (commented):
I would not restrict to the most recent `joblib` version as this might cause version conflicts on some systems. `joblib` version `1.0.0` was released end of 2020, which might be a better minimum requirement. We try to roughly follow the [Numpy policy](https://numpy.org/neps/nep-0029-deprecation_policy.html#nep29) for version dependencies.
@joschaschmiedt Or did you notice problems using `joblib<1.2.0`?

doc/source/developers_guide.rst
joschaschmiedt: I have not tested `joblib<1.2.0`. Looking at [joblib's changelog](https://github.com/joblib/joblib/blob/master/CHANGES.rst) I'm not sure if the bugs fixed since 1.0.0 affect the caching in neo, but I'd guess no and hope that a test will show if it does.

Speaking of tests, as the missing dependency did not pop up before, is there actually any test for the caching?

PR #1228, opened 2023-03-01
Minor typos in Maxwell

CodyCBakerPhD:
Fixing some typos and improving error message


CodyCBakerPhD:
Also @JuliaSprenger, I'm not having any luck getting the Maxwell decompression plugin to work on Windows or Mac - I notice it is in your CI though, was there some magic you had to do to get that to work?

In both platforms, I'm calling the automagical function here and checking that the environment variable was set, but it still fails when it tries to read the first analog signal chunk

JuliaSprenger (approved):

Issue #1227, opened 2023-03-01
NWB test failing


JuliaSprenger:
With recent GH action runs NWB tests are failing:
 ```
neo/test/iotest/test_nwbio.py::TestNWBIO::test_roundtrip_with_annotations - ValueError: CurrentClampStimulusSeries.__init__: incorrect shape for 'data' (got '(100, 1)', expected '(None,)')
```
For a full test run see [here](https://github.com/NeuralEnsemble/python-neo/actions/runs/4288338958/jobs/7475817893)

@apdavison Was there an update on the NWB side potentially causing this?

Update: This issue occurs with pynwb version `2.3.1`

PR #1231, opened 2023-03-08
Option to not load nev in blacrock to speedup the opening

samuelgarcia:
When we want to load only traces, inspecting the nev file make the opening very slow.
This new option make the opening really faster.
This is usefull on spikeinterface side for parallel computing.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -108,6 +108,9 @@ class BlackrockRawIO(BaseRawIO):
             If 'all', then all nsX will be loaded.
             Contrary to previous version of the IO  (<0.7), nsx_to_load
             must be set at the init before parse_header().
+        load_nev (bool):
+            Load (or not) events/spikes by ignoring or not the nev file.
+            Default: True
 
     Examples:
         >>> reader = BlackrockRawIO(filename='FileSpec2.3001', nsx_to_load=5)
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @samuelgarcia. I see why you would like to improve the loading time, however the IO uses `.nev` is used for identifying pauses in the recording that are not documented in the metadata or `nsX` files. So if you are going to introduce this option is should come with a huge warning that this might result in incorrect AnalogSignals timestamps for some recording sessions.

samuelgarcia:
Hi Julia,
For me pauses are segments.
From what I remember, the number of segment is driven by the nsX [here](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/blackrockrawio.py#L414)
And the nev spike/event are asigned to a specific segment depending there timestamps and this operation cost a lot at opening.
So I do not think that the represention of signal alone is changed by reading or not event/spikes.
Is there some hiden logic I do not remember in bloackrock ?

JuliaSprenger (approved):
Hi @samuelgarcia Yes, you are right. I think I mixed up the mechanism of the NeuralynxIO and the BlackrockIO.

JuliaSprenger:
Failing tests are related to #1227, not to changes here.

samuelgarcia:
thank you.

PR #1233, opened 2023-03-10
Make the master dev version greater than the latest released.

samuelgarcia:


JuliaSprenger:
Hi @samuelgarcia You are right, but shouldn't master be actually already `0.13.0.dev0` since `0.13.0` will probably be the next bigger release. I think we mixed this up a bit when switching to the `toml` requirements.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "neo"
-version = "0.12.0.dev0"
+version = "0.12.1.dev0"
```
JuliaSprenger, line 3:
```suggestion
version = "0.13.0.dev0"
```

samuelgarcia: ok

PR #1234, opened 2023-03-16
added RAWDATAFILE to neuralynx type_of_recording

GeezleCode:
Neo didn't recognize the "RawDataFile" acquisition type in the header string. This is set when ".ncs" files are created by an offline replay of Neuralynx raw data (".nrd") files. 
I also commented on the bug in issue#1202


JuliaSprenger:
Hi @GeezleCode! Thanks for improving the IO, the changes look good to me.

PR #1236, opened 2023-03-21
[NixIO] use OS independent datetime formatting

JuliaSprenger:
The datetime formatter `%s` is not available on Windows systems, `int(datetime.timestamp())` yields the same information OS independently.


JuliaSprenger:
@apdavison Tests are still failing for #1227

samuelgarcia (approved):

PR #1248, opened 2023-04-03
Allow folders without events/continuous streams to be loaded

### LINKED ISSUE #1199: Error reading open ephys binary data if no event data is present
**Describe the bug**
An error will occur when reading binary open ephys data if no event data is present

**To Reproduce**
Remove the entry in the `events` field of `structure.oebin` and then try to load a binary open ephys recording

**Expected behaviour**
The recording should still be loaded successfully

**Environment:**
 - OS: Windows
 - 3.8
 - Neo version: 0.12.0.dev0
 - NumPy version: 1.21.0

**Additional context**
The `all_stream` returned from `explore_folder` may not have an `event` field, but in the following line it always assumes it is present without any check.  https://github.com/NeuralEnsemble/python-neo/blob/7fc30cb7e3aea52e9c405475c5cbb7ea96c60f3c/neo/rawio/openephysbinaryrawio.py#L82

alejoe91:
Fix #1199

- Allow loading of OE bunary fodlers without events or consinuous streams
- More informative error message if `structure.oebin` files are not found ("not a valid Open Ephys binary folder")

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -77,9 +77,16 @@ class OpenEphysBinaryRawIO(BaseRawIO):
         check_folder_consistency(folder_structure, possible_experiments)
         self.folder_structure = folder_structure
 
-        # all streams are consistent across blocks and segments
-        sig_stream_names = sorted(list(all_streams[0][0]['continuous'].keys()))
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @alejoe91! Thanks for implementing this. Maybe it would make sense to also add a test for the error raised on non-valid OE folder?

Issue #792, opened 2020-02-08
Read digital channels from rhd


luiztauffer:
Hi, I’m using `IntanIO` to read form `rhd` files, which works well but does not read digital channels. Am I missing something or is this still to-be-implemented?

Just as a reference, Intantech provides python scripts that read those files and extract basically everything in nice dictionaries: http://intantech.com/downloads.html?tabSelect=Software

Ashkees:
Hi.  I am also interested in this feature if it becomes available.  Thanks!

samuelgarcia:
Hi all.
I am the autor of this IO but I never use it.
Reading this
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/intanrawio.py#L522
suggest that it is implemented at header parsing level but thoses channel are not exposed to 
_get_analogsignal_chunk and not expose in the header['signal_channels']
If someone have time to do it it would be great.
Do you plan to have expose thoeses channels as analogsignal (with 0 and 1) or via events (that would need a full read of the entire file) ?

Ashkees:
For me, either an analogsignal or events would be fine.  If I had it as as analogsignal, most likely I would then convert it into an events, but this should be straightforward.

CodyCBakerPhD:
Some progress towards this in #1249

bendichter:
@CodyCBakerPhD , @luiztauffer , #1249 is now merged. What are the next steps here?

CodyCBakerPhD:
@luiztauffer what project did you need this for? (If you can remember) - this issue can likely be closed

A new release of neo would be great @samuelgarcia 

I suppose confirmation that it works in SI to close https://github.com/SpikeInterface/spikeinterface/issues/1876 is the small 'next step'

Otherwise, @weiglszonja has a current project that needs this so they will be making the data interface on NeuroConv: https://github.com/catalystneuro/neuroconv/issues/519

PR #1249, opened 2023-04-03
Add Access to Intan Digital-In and Digital-Out from the RawIO API

zm711:
This might just be me, but I was hoping to also use Neo to obtain `digital in` information from Intan recordings. The ADC stream worked great, so I tweaked the code to allow access to digital streams as well.  Based on my local testing it worked fine and the dig-in channel is actually encoded by the height of the bar (so users could do their own post-processing to separate out into dig-1, dig-2 etc. --see attach figure with x-axis being given in samples).  I don't have any files with `dig-out` though .

I also put up a first draft of a doc-string. Happy to receive any comments. 

I know @samuelgarcia wrote the implementation so if there is a reason why you'd prefer to not include this just let me know.

image


zm711:
One other possibility that I was thinking about would be to create a function like below:

```python
def _get_digitalsignal_chunk(self, i_start=None, i_stop=None):
        
        digital_memmap = self._raw_data["DIGITAL-IN"]
        dig_size = digital_memmap.size
        dig_shape = digital_memmap.shape
        
        if i_start is None:
            i_start=0
        if i_stop is None:
            i_stop =dig_size
        
        block_size = dig_shape[1]
        block_start = i_start // block_size
        block_stop = i_stop // block_size + 1

        sl0 = i_start % block_size

        digital_data = np.squeeze(digital_memmap[block_start:block_stop].flatten()[sl0:sl1])
        
        native_order = self._global_info['dig_native_order']
        
        digital_chunk = np.zeros((len(native_order), (i_stop-i_start)), dytpe='uint16')
        
        for value in native_order:
            digital_chunk[value, :] = np.not_equal(np.bitwise_and(digital_data,(1 << value)),0,)
            
        return digital_chunk
```

This would mirror the `analogsignal_chunk`. Then instead I would  add the `native_order` of the channels which are already obtained while parsing the header (I would add a couple lines to add them into `_global_info`.) Then we don't touch `ordered channels` which might confuse users who try to use the normal analog functions on the digital data. This function would let us  handle returning the digital without messing with mixing digital and analog.

Let me know if you'd be willing to implement something like this Sam @samuelgarcia for Neo or whether you think it's out of scope. If you like this idea better (or hate it) I'm happy to tweak the PR.

CodyCBakerPhD:
resolves #792

@JuliaSprenger @samuelgarcia anything holding up this PR? Do you need example data for it?

samuelgarcia:
Hi Zach.
I am ultra sorry for the long delay.
Thanks for the patch.
This OK for me in actual state but I have to admit that I would be more happy to have this
```def _get_digitalsignal_chunk(self, i_start=None, i_stop=None):```
but this need modification also in baserawio.py and add a new entry in the self.header.

So maybethe best is to merge now and postpone a bit the writting of "digitalsignal" concept.

zm711:
@samuelgarcia 

That works for me. I have my workaround function already outside of neo, so I'm in good shape. This merging would at least allow me to stay in neo rather than hack the memory map structure.

zm711:
I'll fix the pep8speaks.

zm711:
Sam so long as the tests pass, I'm good with this as is. Thanks.

samuelgarcia:
Ok. Thank you Zach.

PR #1251, opened 2023-04-04
Add codemeta.json

apdavison:
see https://codemeta.github.io

This is to help automate the registration of Neo in repositories of scientific software, e.g. EBRAINS.

In the future, I plan to automate updates to this file using GH Actions. Until that's done, it will need to be updated by hand for each release.


apdavison:
I'm not aware of any existing tools.

JuliaSprenger:
Ok, then we duplicate for now, but have to pay attention to keep everything synced manually. @samuelgarcia Do you agree?

PR #1252, opened 2023-04-04
Make joblib an optional dependency

apdavison:
Since we only use a single feature of joblib, and in only one place, where it is optional, I thought it would be better not to install it by default.


JuliaSprenger (approved):
Looks even better to me.

PR #1246, opened 2023-04-03
spikeglx naming debug #1240

samuelgarcia:
#1240

PR #1254, opened 2023-04-04
Fix for spikeglx to repair #1246

samuelgarcia:
#1246 have broken spikeinterface and neuroconv tests.
Here a fix that should be OK.


JuliaSprenger:
Does it make sense to add a test similar to the failing ones in SI & neuroconv?

JuliaSprenger:
Ok for now it seems to fix the downstream issues. @samuelgarcia Thanks for the fix

Issue #1202, opened 2022-11-28
An OSError about header when importing neuralynx data


redagavin:
I was trying to import neuralynx data into python using neo. But, I got an OSError.
This is a screenshot of the error.
![Screen Shot 2022-11-27 at 5 07 43 PM](https://user-images.githubusercontent.com/39891180/204162273-7ef68c45-302c-4e80-bfd6-bac6da9e915b.png)

Also, I noticed that there is a small bug in the line that raise the OSError. I think `.format(an, av)` is supposed to fill the first {} with `an` and the second {} with `av`. However, only the second string after the `+` calls `.format(an, av)` in the source code, which is why the error message is weird.
<img width="800" alt="Screen Shot 2022-11-27 at 5 19 11 PM" src="https://user-images.githubusercontent.com/39891180/204162647-a07972fd-f759-4677-8209-702cb060a4aa.png">

redagavin:
@samuelgarcia @JuliaSprenger @apdavison Could you please take a look at this? Thanks!

samuelgarcia:
Hi,
thanks for the report.
Could you send us a file to debug this ?

@JuliaSprenger @PeterNSteinmetz: would you have time to handle this, you are our best neuralynx expert!

redagavin:
> Hi, thanks for the report. Could you send us a file to debug this ?
> 
> @JuliaSprenger @PeterNSteinmetz: would you have time to handle this, you are our best neuralynx expert!

@samuelgarcia @JuliaSprenger @PeterNSteinmetz Yes, I can send you a file to debug this. Please provide me with a link so that I can share the file privately.

PeterNSteinmetz:
Gavin, I can take a look at that. Can you send via email to PeterNSteinmetz@steinmetz.org ? Alternatively, do you use Google Drive? If so you could share with me there - username is PeterNSteinmetz

redagavin:
> Gavin, I can take a look at that. Can you send via email to [PeterNSteinmetz@steinmetz.org](mailto:PeterNSteinmetz@steinmetz.org) ? Alternatively, do you use Google Drive? If so you could share with me there - username is PeterNSteinmetz

I sent the files to you through email and google drive. Please take a look. Thanks!

samuelgarcia:
Thank you Peter!

PeterNSteinmetz:
[...]

PR #1257, opened 2023-04-13
Cheetah v5.6.0 fixes only

PeterNSteinmetz:
The fairly small fixes to handle issue #1202. Warning, this does not include the test data or code because we could not get permission to include the data. 

In truth the whole handling of the date format parsing should be generalized a bit.

### DIFF
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -850,20 +850,21 @@ def get_nse_or_ntt_dtype(info, ext):
     """
     dtype = [('timestamp', 'uint64'), ('channel_id', 'uint32'), ('unit_id', 'uint32')]
 
-    # count feature
-    nb_feature = 0
-    for k in info.keys():
-        if k.startswith('Feature '):
-            nb_feature += 1
+    # for purpose of dtypes, features in the file are always fixed 8 presently,
+    # whether mentioned in the header or not. Features may not be listed in the header
+    # if no feature names are assigned in Neuralynx software.
+    nb_feature = 8
     dtype += [('features', 'int32', (nb_feature,))]
 
-    # count sample
+    # Number of samples are fixed in the file at 32 for .nse 32 * 4 for .ntt.
+    # WaveformLength may or may not be listed in the file depending on settings
+    # in the Neuralynx software, so don't try retrieving it.
     if ext == 'nse':
-        nb_sample = info['WaveformLength']
+        nb_sample = 32
         dtype += [('samples', 'int16', (nb_sample,))]
     elif ext == 'ntt':
-        nb_sample = info['WaveformLength']
-        nb_chan = 4  # check this if not tetrode
+        nb_sample = 32
+        nb_chan = 4
         dtype += [('samples', 'int16', (nb_sample, nb_chan))]
 
     return dtype
--- a/neo/rawio/neuralynxrawio/nlxheader.py
+++ b/neo/rawio/neuralynxrawio/nlxheader.py
@@ -97,6 +97,12 @@ class NlxHeader(OrderedDict):
                             r'  At Time: (?P<time>\S+)',
             filename_regex=r'## File Name: (?P<filename>\S+)',
             datetimeformat='%m/%d/%Y %H:%M:%S.%f'),
+        # Cheetah version 5.6.0, some range of versions in between
+        'v5.6.0': dict(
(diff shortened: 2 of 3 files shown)


neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -850,20 +850,21 @@ def get_nse_or_ntt_dtype(info, ext):
...
 
-    # count feature
-    nb_feature = 0
-    for k in info.keys():
-        if k.startswith('Feature '):
-            nb_feature += 1
+    # for purpose of dtypes, features in the file are always fixed 8 presently,
+    # whether mentioned in the header or not. Features may not be listed in the header
+    # if no feature names are assigned in Neuralynx software.
+    nb_feature = 8
```
JuliaSprenger, line 856: Hi @PeterNSteinmetz Hard coding the number of features, tetrode channels and number of samples per waveform here seems risky, are you sure this is really true for all Neuralynx files ever recorded? Maybe it would be better to use the default values only if that information can not be retrieved from the info dictionary.

neo/rawio/neuralynxrawio/nlxheader.py
```diff
@@ -223,6 +229,8 @@ def __init__(self, filename):
                 hpd = NlxHeader.header_pattern_dicts['bv5']
             elif av <= Version('5.4.0'):
                 hpd = NlxHeader.header_pattern_dicts['v5.4.0']
+            elif av == Version('5.6.0'):
```
JuliaSprenger, line 232: @PeterNSteinmetz Do you have any idea if this style was also used in other versions between 5.4.0 and 5.6.4?

JuliaSprenger:
Thanks for fixing this for the provided dataset.

> In truth the whole handling of the date format parsing should be generalized a bit.

Feel free to start a PR on this :)

neo/rawio/neuralynxrawio/nlxheader.py
PeterNSteinmetz: Unfortunately no. The manufacturer never documented these changes and has not responded to requests for better clarification. Thus we are stuck with observing changes as they come in.

neo/rawio/neuralynxrawio/neuralynxrawio.py
PeterNSteinmetz: These number are part of the hard definition of the file format as provided by the manufacturer. These are the maximum numbers possible. The normal result of assuming 8 when a smaller number are used would be to have some features all 0.

neo/rawio/neuralynxrawio/nlxheader.py
JuliaSprenger: Ok, then we have to keep guessing based on the data files we come across :/

neo/rawio/neuralynxrawio/neuralynxrawio.py
JuliaSprenger: And the same is true for the waveform shape? It's not possible to record fewer samples per waveform?

PeterNSteinmetz: Yes, .Nse records are always of fixed length. Again I think some of the latter part of the record could just be 0.

JuliaSprenger: In that case of 0-padding of the waveform it would still be more user friendly to return only the non-padded waveform data, so to use the indicated waveformshape and not the default of 32 samples, no?

PeterNSteinmetz: .Nse files are always 32 sample points. I am not aware of anything in the files documenting more or less than 32 being used.

JuliaSprenger: Since that information is in the header for at least some files, it might be possible to adjust that setting in some files. It might still be safer to use that information if available and only use the default when no info on this is present in the header. @samuelgarcia Do you have an opinion on that? See also lines 863 and 866 below.

PeterNSteinmetz: I don't know that there is any information in the headers of any files on there being fewer than 32 sample points. 

The features are a separate item. I also don't know where the code which counts them and you note above came from. There are a lot of files from older versions which have features embedded but no mention of them in the headers. I think that was a more recent addition to the headers. And such changes are again undocumented by Neuralynx so one would have to guess at which versions include it or not based on extant examples.

apdavison:
Can this be merged? @JuliaSprenger @samuelgarcia ?

PR #1258, opened 2023-04-18
Add option to `MEArecRawIO` for loading only recordings or only sorting data

h-mayorquin:
Right now `MEArecRawIO` loads all the data (recording and sorting) by default. This PR leaves that default as it is but also adds the option of only loading recording or sorting data with a boolean value at `__init__`. 

With this, the intialization of the object -specially the recoder- should be lighter as it does not need to go through the spiking data in mearec.

I added tests for both of these modes.

### DIFF
--- a/neo/rawio/mearecrawio.py
+++ b/neo/rawio/mearecrawio.py
@@ -69,8 +69,13 @@ class MEArecRawIO(BaseRawIO):
                                           check_suffix=False,
                                           load=load,
                                           load_waveforms=False)
-        
-        self.info_dict = self._recgen.info
+
+        self.info_dict = deepcopy(self._recgen.info)
+        if self.load_recordings:
+            self._recordings = self._recgen.recordings
+        if self.load_spiketrains:
+            self._spiketrains = self._recgen.spiketrains
+
         self._sampling_rate = self.info_dict['recordings']['fs']
         self.duration_seconds = self.info_dict["recordings"]["duration"]
         self._num_frames = int(self._sampling_rate * self.duration_seconds)
@@ -80,8 +85,6 @@ class MEArecRawIO(BaseRawIO):
         signals = [('Signals', '0')] 
         signal_streams = np.array(signals, dtype=_signal_stream_dtype)
 
-        if self.load_recordings:
-            self._recordings = self._recgen.recordings
         
         sig_channels = []
         for c in range(self._num_channels):
@@ -100,7 +103,6 @@ class MEArecRawIO(BaseRawIO):
         # creating units channels
         spike_channels = []
         if self.load_spiketrains:
-            self._spiketrains = self._recgen.spiketrains
             for c in range(len(self._spiketrains)):
                 unit_name = 'unit{}'.format(c)
                 unit_id = '#{}'.format(c)
@@ -129,7 +131,7 @@ class MEArecRawIO(BaseRawIO):
         self._generate_minimal_annotations()
         for block_index in range(1):
             bl_ann = self.raw_annotations['blocks'][block_index]
-            bl_ann['mearec_info'] = deepcopy(self._recgen.info)
+            bl_ann['mearec_info'] = self.info_dict
 
     def _segment_t_start(self, block_index, seg_index):
         all_starts = [[0.]]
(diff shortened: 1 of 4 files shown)


alejoe91 (approved):

alejoe91 (approved):

alejoe91:
Thanks @h-mayorquin! This will make reading MEArec files much faster and getting the number of channels in a more robust way

JuliaSprenger:
Hi @h-mayorquin Could you also add the new arguments to the [MeaRecIO](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/io/mearecio.py), to be available also for the non-rawio?

h-mayorquin:
@JuliaSprenger should be done.

DradeAW:
If this is done, could it be merged? :)

Thanks

h-mayorquin:
@DradeAW We are waiting for Neo side to merge the fix there.

samuelgarcia:
This is cool.
Here in neo we do not have the naming recording which is very very confusing.
I would change it to `load_analogsignal'`

neo/rawio/mearecrawio.py
```diff
@@ -36,52 +48,74 @@ class MEArecRawIO(BaseRawIO):
...
+            self._spiketrains = self._recgen.spiketrains
+
+        self._sampling_rate = self.info_dict['recordings']['fs']
+        self.duration_seconds = self.info_dict["recordings"]["duration"]
+        self._num_frames = int(self._sampling_rate * self.duration_seconds)
+        self._num_channels = self.channel_positions.shape[0]
+        self._dtype = self.info_dict["recordings"]["dtype"]
+        
+        signals = [('Signals', '0')] 
+        signal_streams = np.array(signals, dtype=_signal_stream_dtype)
```
samuelgarcia, line 87: With load_recording=False this shoudl be empty.

neo/rawio/mearecrawio.py
```diff
@@ -36,52 +48,74 @@ class MEArecRawIO(BaseRawIO):
...
             sr = self._sampling_rate  # Hz
-            dtype = self._recordings.dtype
+            dtype = self._dtype
             units = 'uV'
             gain = 1.
             offset = 0.
             stream_id = '0'
             sig_channels.append((ch_name, chan_id, sr, dtype, units, gain, offset, stream_id))
+    
         sig_channels = np.array(sig_channels, dtype=_signal_channel_dtype)
```
samuelgarcia, line 102: With load_recording=False this should be empty.

neo/rawio/mearecrawio.py
h-mayorquin: I tried this, but then if someone calls `get_analog_signal_chunk` in that configuration it will throw a criptic error.

```
    def _get_stream_index_from_arg(self, stream_index_arg):
        if stream_index_arg is None:
>           assert self.header['signal_streams'].size == 1
E           AssertionError
```

This is generated because the general rawio. calls:

https://github.com/h-mayorquin/python-neo/blob/ac508e117097f44ea82b4058c4a4d4f6d1e7c2eb/neo/rawio/baserawio.py#L531-L553

Which in turns call `get_stream_index_from_arg` which then asks for a header.

If we only get rid of the channels we still get an error:

```
        if signal_streams.size > 0:
>           assert signal_channels.size > 0, 'Signal stream but no signal_channels!!!'
E           AssertionError: Signal stream but no signal_channels!!!
```

I prefer to have informative error messages but I don't know the cost of keeping those references open. What do you think?

h-mayorquin:
I changed `load_recordings` to `load_analogsignal`

neo/rawio/mearecrawio.py
h-mayorquin: This is empty now.

h-mayorquin:
I implemented Sam's feedback now.

JuliaSprenger (approved):

JuliaSprenger:
@samuelgarcia Are you happy with the changes? Then feel to merge.

samuelgarcia (approved):

samuelgarcia:
merci @h-mayorquin

PR #1260, opened 2023-04-24
Gain on Sync channel missing from NP2.0 probes (Spikeglx)

JoeZiminski:
The gain on NP2.0 is fixed at 80 and applied in `per_channel_gain[:-1] = 1 / 80.` in spikeglxrawio.py.

However, I just tested against a NP2 (1-shank) file using the syn channel and the scaling was not correct for the sync channel (as compared to SpikeGLX (output saved in .csv).

Changing the above line to include the sync channel: `per_channel_gain[:] = 1 / 80.` aligned with SpikeGLX scaling and passes the test on the file I have.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -404,7 +404,7 @@ def extract_stream_info(meta_file, meta):
             # https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_20.md#channel-entries-by-type
             # https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_20.md#imec
             # https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_30.md#imec
-            per_channel_gain[:-1] = 1 / 80.
+            per_channel_gain[:] = 1 / 80.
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Joe.
Thank you for this.
Do you have any odea why the synch channel also need the gain ?

JoeZiminski:
Hi Samuel, I'm not sure why but testing against a NP2.0 file (1 shank) that uses the sync channel every other channel matches SpikeGLX exported file except for the sync. Applying the `1/80` gain to the sync channel brings them back into alignment.

I tried this out on a NP1 recording that uses sync channel, and it loaded properly in SpikeInterface. It may be the switch from (if I understand correctly) customisable gain per channel in NP1 to fixed gain across channels in NP2 that the new `1/80` gain was applied uniformly across all channels including sync. 

I think the safest thing to do is check this assumption, I can send Bill Karsh a message and check. Will let you know!

JoeZiminski:
Hi @samuelgarcia, @JuliaSprenger very sorry for delay here. It turns out that scaling is applied to the sync channel when exporting data from some spikeglx versions, but it is only on export that should be changed in future versions. In this case it is not necessary to scale the sync channel, I will close this PR.

Issue #1261, opened 2023-04-25
PlexonRawIO _parse_header() uses a lot of memory


benjamin-heasly:
**Describe the bug**
When reading a Plexon `.plx` file with `PlexonRawIO`, Python consumes a lot of memory (greater than the entire `.plx` file size).  This can cause the system to hang or the process to crash.

**To Reproduce**
This can be reproduced just by constructing a `PlexonRawIO`:

```
plx_file = 'my_plexon_file.plx'
neo_reader = PlexonRawIO(filename=plx_file)
```

In my case, the `my_plexon_file.plx` file is 1.1GB total and contains about 16 million blocks.  Constructing the `PlexonRawIO` instance takes several minutes.  After about 4 minutes, the OS slows down and becomes unstable/unusable.  In `top` I can see the Python process consuming more and more memory, reaching about 80-90% of available memory (~10GB) before becoming unstable/unusable.

**Expected behaviour**
I would expect the `PlexonRawIO` to initialize without consuming so much memory -- perhaps less than or close to the overall data file size.

**Environment:**
 - OS: Linux Mint 19.3 Cinnamon
 - Python version: Python 3.11.3
 - Neo version: 0.12.0
 - NumPy version: 1.23.5
 - Hardware: Thinkpad T450s
 - Physical RAM: 11.6 GiB

**Additional context**
I think the memory is being used up during the initial scan of block headers, [here](https://github.com/NeuralEnsemble/python-neo/blob/v0.12.0/neo/rawio/plexonrawio.py#L94).  Since the block headers themselves are only 16 bytes, I would not expect this step to take so much memory -- on the order of 300 MB in my file (16 bytes * 16 million blocks) as opposed to 10+ GB.

Could it be the `view(DataBlockHeader)`s are taking excessive space to store?  Or is there an unexpected interaction between  `view` and `memmap`?  Maybe the requirement that `view`s are backed by the original buffer is preventing `memmap` from releasing mapped pages, and these pages are ranging over the entire file -- or something along those lines?

In any event refactoring to avoid storing all the `view(DataBlockHeader)`s seems to work and also keeps memory usage in check.  I'll link a PR with such a refactor, in case you're interested!

Thank you for this useful project!

PR #1262, opened 2023-04-25
Refactor to use less memory parsing Plexon block headers.

benjamin-heasly:
This would address https://github.com/NeuralEnsemble/python-neo/issues/1261.

In summary:

When reading a Plexon `.plx` file with `PlexonRawIO`, Python consumes a lot of memory (greater than the entire `.plx` file size). This can cause the system to hang or the process to crash.

To Reproduce
This can be reproduced just by constructing a `PlexonRawIO`:

```
plx_file = 'my_plexon_file.plx'
neo_reader = PlexonRawIO(filename=plx_file)
```

I think the memory is being used up during the initial scan of block headers, [here](https://github.com/NeuralEnsemble/python-neo/blob/v0.12.0/neo/rawio/plexonrawio.py#L94).

Since the block headers themselves are only 16 bytes, I would not expect this step to take so much memory.  In my local example, the `.plx` file has about 16 million blocks.  I'd expect this to consume on the order of 300 MB (16 bytes * 16 million blocks).  When I run, I see memory usage climb to ~10 GB before the system becomes unstable.

Could it be the `view(DataBlockHeader)`s are taking excessive space to store? Or is there an unexpected interaction between `view` and `memmap`? Maybe the requirement that `view`s are backed by the original buffer is preventing `memmap` from releasing mapped pages, and these pages are ranging over the entire file -- or something along those lines?

In any event refactoring to avoid storing all the `view(DataBlockHeader)`s seems to work and also keeps memory usage in check.

### DIFF
--- a/neo/rawio/plexonrawio.py
+++ b/neo/rawio/plexonrawio.py
@@ -75,11 +75,7 @@ class PlexonRawIO(BaseRawIO):
 
         offset4 = offset3 + np.dtype(SlowChannelHeader).itemsize * nb_sig_chan
 
-        # loop over data blocks and put them by type and channel
-        block_headers = {1: {c: [] for c in dspChannelHeaders['Channel']},
-                         4: {c: [] for c in eventHeaders['Channel']},
-                         5: {c: [] for c in slowChannelHeaders['Channel']},
-                         }
+        # locate data blocks and group them by type and channel
         block_pos = {1: {c: [] for c in dspChannelHeaders['Channel']},
                      4: {c: [] for c in eventHeaders['Channel']},
                      5: {c: [] for c in slowChannelHeaders['Channel']},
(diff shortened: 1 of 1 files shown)


JuliaSprenger:
Hi @benjamin-heasly Thanks for looking into this.
 I think you are right about some reference to the header causing the loading a bigger chunk of data. I have seen similar effects for different formats, but did not fully understand the mechanics behind it as it might also depend on your file system architecture and caching settings.
The changes look good to me and tests are passing. What is the current memory usage for the file that used 10GB earlier?

benjamin-heasly:
Hi @JuliaSprenger -- thanks for taking a look at this PR!  I agree, the details might be system-dependent.

I ran that same file again, on the same machine, and watched `top`.  I saw memory usage peak at 20%, which on this machine is ~2.5 GB.

JuliaSprenger (approved):

JuliaSprenger:
That's still more than ideally expected, but already a good improvement. Thanks for the contribution!

samuelgarcia:
Hi @benjamin-heasly.
I am a bit late to read this sorry.
I am OK with this if it helps regarding the memory.
Just one question does the extra explicit loop make the code as fast as before ? Even for long files ?

benjamin-heasly:
Hi @samuelgarcia it's hard for me to tell because beforehand I was unable to finish parsing the file header.  I'd run out of memory after ~4 minutes.  With this refactor parsing my file header takes about 10 minutes (1.1GB file, ~16 million blocks).  My understanding is there would be no change in behavior or timing after the header parsing is complete because it produces the same data structures.

4-10 minutes does seem slow to me.  But if the goal is to index every block, I'm not sure we can have it both ways.

Issue #1037, opened 2021-10-11
446 tests fail with neo-0.10.0


TheChymera:
This is the full build and test log: https://ppb.chymera.eu/f0c5ba.log
I assume some failures at least are due to network access (is there any way to easily skip tests which require network access), and some due to missing dependencies.

One thing I spotted is ipython — this is an interactive shell. Is any functionality of python-neo really dependent on an *interactive* Python shell?

TheChymera:
Hi @samuelgarcia thank you for looking into this! The way many packages do this is that they either:
1. have an environment variable such as `PACKAGENAME_TESTS_NONETWORK` (datalad for instance uses `DATALAD_TESTS_NONETWORK`)
2. mark the tests with e.g. `@pytest.mark.network` so that they can be deselected when running pytest via `pytest -m "not network"`.

samuelgarcia:
OK
I put that on my todo list.
I won't do it immediatly because I am a bit busy next 3 weeks but will do it soon.

JuliaSprenger:
@samuelgarcia Any news on this?

TheChymera:
Still failing very many tests (506) on `0.11.1`: https://ppb.chymera.eu/65897d.log

TheChymera:
Just to update, though it's probably “as expected” given the milestone, network tests still fail in a network sandbox and I can't see how to auto-disable them on 
`0.12.0` https://ppb.chymera.eu/6fae8d.log

PR #1263, opened 2023-04-26
Introduce `NEO_TESTS_NONETWORK` switch to skip network tests

JuliaSprenger:
As requested by #1037 this is a proposal for a mechanism to skip tests that require network access. @samuelgarcia Is this what you had originally in mind?


samuelgarcia:
Hi Julia.
Sorry for the enormous delay.
I think I was happy with function apporach.
We could also have put this env variable in the function no ?
Not having datalad is a good reason to not launch test no ?

PeterNSteinmetz:
I think another idea as well would be to move the check on the network to a setupClass routine rather than setup. That way it would run only once per testing session rather than for each test.

neo/test/rawiotest/tools.py
```diff
@@ -3,24 +3,5 @@
...
-        HAVE_DATALAD = True
-    except:
-        HAVE_DATALAD = False
-    if not HAVE_DATALAD:
-        return False
-    if os.environ.get('NOSETESTS_NO_NETWORK', False):
-        return False
-    if os.environ.get('TRAVIS') == 'true':
-        return False
-    return True
```
zm711, line 26: @samuelgarcia, the functional approach can actually fail because on one my computers I have datalad but I don't have git-annex correct. So I think importing datalad would still be possible, but downloading the files would not be possible. So I think an environment variable is probably safer, no?

zm711: I got this work actually so never mind.

alejoe91:
Closing in favor of #1420

Issue #684, opened 2019-07-06
Circular references to proxy objects cause files to be locked even after leaving scope


jpgill86:
With code like this,

```python
import neo
io = neo.io.get_io(filename)
blk = io.read_block(lazy=True)
del io
del blk
```

the file `filename` will remain locked until the Python process ends.

I believe the cause is circular references to proxy objects, such as this:

https://github.com/NeuralEnsemble/python-neo/blob/70f78117bc2173a25e21149f3cdc69262df6e446/neo/io/basefromrawio.py#L264-L265

These circular references appear all over the place in Neo and are not unique to proxy objects. I suspect that because of them, most Neo objects in memory are not destroyed even after commands like `del io; del blk`. Not only is this a source of a memory leak, but in the case of lazy loading using proxy objects, the file becomes locked (e.g., it cannot be deleted in Windows at least).

Manually performing garbage collection is a workaround:

```python
import gc
gc.collect()
```

However, the best solution I think would be to use [weak references](https://docs.python.org/3/library/weakref.html).

samuelgarcia:
Hi jeffrey,
thank you to catch this.Andrew have a long date plan to make a pseudo list object to deal with relationship in neo. It is catually a mess.
At implementtation time, we could consider make weakref at least for one way bottum up.

The fact that file are still open when del to all object is strange.
Of course when rawio many object have an internal memmap to data.
So yes the file stay open when object is still alive.
Note that for proxyobject there is a link to the IO : **self._rawio**
I guess this could be a weakref. The ptach is easy to do.

jpgill86:
> Andrew have a long date plan to make a pseudo list object to deal with relationship in neo. It is catually a mess.
At implementtation time, we could consider make weakref at least for one way bottum up.

Yes, this sounds like a good way to do it eventually.

> Of course when rawio many object have an internal memmap to data.
So yes the file stay open when object is still alive.

Exactly. A strong reference to the RawIO persists in each proxy object, and the memmaps keep the file open. Each proxy object persists, even after `del io; del blk`, because of the circular references, until garbage collection runs.

> Note that for proxyobject there is a link to the IO : self._rawio
I guess this could be a weakref. The ptach is easy to do.

Good point, it may be easier to solve the file lock issue (not the memory leak problem) with a few targeted weakrefs.

However, the first few things I tried didn't work.

I tried replacing in each proxy object

```python
self._rawio = rawio
```

with

```python
[...]

PR #1265, opened 2023-05-03
memmap refactoring

samuelgarcia:
This is a proposal to change the actual memap behavior in rawio layer.

spikeinterface use neo.rawio a lot with extenssive IO demand with parralel read/computing.
Many users experience big memmory saturation even on big machine when the file are enormous.
This quite a bit OS dependant, mainly windows but even some linux have this problem).
Somehow, the memmap do not release page after the use of one buffer and so the memory increase forever.

The actual strategy we now have in almost all IO when it is binary base :
 1. In  parse_header, open one or several numpy.memmap object and set then as attribute or nested dict
 2. In get_analogsignal_chunk : lazly fetch buffer via numpy slicing using this memmap arrays
After consumming the traces chunk the memory should be relased and so the memmory usage should low but in fact in some case it increase until saturating.

@h-mayorquin and @samuelgarcia made some benchmark to find a better strategy.
Here some code examples:
https://gist.github.com/h-mayorquin/8b8d1899c724690ce7c18fb49f56af82

The best strategy we find to try to fix is:
  1. In the parse header, open the file with mode='rb'
  2. In get_analogsignal_chunk : create a new memmap object using the opened file

For the 2. we mimic the numpy.memmap but a but simplify to be faster.
The speed of reading is almost the same and the memory problem seems to be fixed.

@maxjuv @TomBugnon @vncntprvst @juliencarponcy: we need more tests!
I started to change spikeglx and openephysrawbinary.
Could you try to combine this branch and also this [PR on spikeinterface](https://github.com/SpikeInterface/spikeinterface/pull/1602)
And then:
  * read a recording with read_spikeglx() or read_openephys()
  * make a simple preprocessing
  * rec.save(folder=folder, n_jobs=64, chunk_duration='1s', progress_bar=True)
And monitor the memory usage.
Do you still have an infinite memory increase or not ?

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -52,6 +52,7 @@ Some functions are copied from Graham Findlay
 
 from .baserawio import (BaseRawIO, _signal_channel_dtype, _signal_stream_dtype,
                 _spike_channel_dtype, _event_channel_dtype)
+from .utils import create_memmap_buffer, get_memmap_shape
 
 from pathlib import Path
 import os
@@ -91,7 +92,8 @@ class SpikeGLXRawIO(BaseRawIO):
 
         nb_segment = np.unique([info['seg_index'] for info in self.signals_info_list]).size
 
-        self._memmaps = {}
+        # self._memmaps = {}
+        self._memmap_args = {}
         self.signals_info_dict = {}
         for info in self.signals_info_list:
             # key is (seg_index, stream_name)
@@ -100,11 +102,14 @@ class SpikeGLXRawIO(BaseRawIO):
             self.signals_info_dict[key] = info
 
             # create memmap
-            data = np.memmap(info['bin_file'], dtype='int16', mode='r', offset=0, order='C')
-            # this should be (info['sample_length'], info['num_chan'])
-            # be some file are shorten
-            data = data.reshape(-1, info['num_chan'])
-            self._memmaps[key] = data
+            #~ data = np.memmap(info['bin_file'], dtype='int16', mode='r', offset=0, order='C')
+            #~ # this should be (info['sample_length'], info['num_chan'])
+            #~ # be some file are shorten
+            #~ data = data.reshape(-1, info['num_chan'])
+            #~ self._memmaps[key] = data
+            shape = get_memmap_shape(info['bin_file'], 'int16', num_channels= info['num_chan'])
+            fid = open(info['bin_file'], "rb")
+            self._memmap_args[key] = (fid, shape, np.dtype('int16'), 0)
 
         # create channel header
         signal_streams = []
@@ -182,7 +187,13 @@ class SpikeGLXRawIO(BaseRawIO):
                             loc = np.concatenate((loc, [[0., 0.]]), axis=0)
                         for ndim in range(loc.shape[1]):
                             sig_ann['__array_annotations__'][f'channel_location_{ndim}'] = loc[:, ndim]
-
+    
+    def __del__(self):
+        # need an explicit close
+        for k, args in self._memmap_args.items():
+            fid, *_ = args
+            fid.close()
+    
     def _segment_t_start(self, block_index, seg_index):
         return 0.
 
@@ -191,7 +202,9 @@ class SpikeGLXRawIO(BaseRawIO):
 
     def _get_signal_size(self, block_index, seg_index, stream_index):
         stream_id = self.header['signal_streams'][stream_index]['id']
-        memmap = self._memmaps[seg_index, stream_id]
+        #~ memmap = self._memmaps[seg_index, stream_id]
+        key = (seg_index, stream_id)
+        memmap = create_memmap_buffer(*self._memmap_args[key])
         return int(memmap.shape[0])
 
     def _get_signal_t_start(self, block_index, seg_index, stream_index):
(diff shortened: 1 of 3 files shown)


zm711:
Hey Sam, just as an fyi and to support the need for this refactor I was doing some emg processing (outside of the spike sorting world) and saw the same thing where I import neo in a function create a reader and then return the analog emg data for post-processing and it seems that as long as my IPython kernel is active the memmap and header information aren't releasing (ie, saturating the RAM eventually). Despite the reader being "cleaned up" after the function if I call the same function on the binary file it runs significantly faster (seems like the header and memmap have been cached). Closing the session finally releases the RAM. On a windows for this work currently (which does annoyingly hold onto files way longer than Mac and Linux do).

samuelgarcia:
@zm711 thanks for the feedback. So do you this this patch will help your case ?

zm711:
@samuelgarcia 
I'm using the IntanRawIO, so I can't specifically test this, and this issue only occurs when I'm trying to batch process data, so it's a rarer issue for me, (so I never reported it I just dealt with my computer crashing every once in a while). If this test actually works for the more extreme spikeinterface case, then I would predict it would also be helpful in the edge cases of data analysis I run into. But I would just need to actually see if this is a general windows issue, ipython issue, or np.memmap issue.

jpgill86:
@zm711, I wonder if your memory release issue is related to #684. Manual garbage collection might help:
```python
import gc
gc.collect()
```

zm711:
@jpgill86 

I tried the `gc.collect()` and the memory usage didn't change. I tried to two rounds of experiments with the first `gc.collect()` returning 23 unreachable objects and the second round returning 86. 

I do have some `try` `except` blocks for some of this analysis, so the main memory lock occurs if parts of the meta-function fails and moves onto trying a different sub-function due to needing something like #1249 for gaining digital access in addition to analog access (which again is IntanRawIO specific). 

The memory getting locked up is a bit inconsistent so it's hard for me to know for sure what is causing it and it is rare enough that hunting it down isn't hugely important to us. So I'm happy to try new implementations to see if anything helps, but I don't want to bog down this PR with memory leak stuff so if we want to continue this we could chat on #684?

maxjuv:
Hello, 
I tried theses Neo & Spikeinterface branches on my Windows. While the memory used is a bit high (7G) for 14 cores, 1s chunksize and the processing consisted to highpass/phase shift/ detect_bad channels/CMR on a Neuropixel dataset. However, the memory stays stable at 7G, so it is way better than before.

JuliaSprenger:
Hi @samuelgarcia The tests in the current master are passing again. You might need to rebase / merge for proper CI testing.

TomBugnon:
Hi @samuelgarcia 
Sorry for the late reply.

I did not check the memory usage (which I didn't specifically have issues with), but relative to the parent commit (2d63e18) this fixes a `OSError: cannot allocate memory` I'm getting when too many memmaps/recordings are opened relying on the same file!

For instance this would raise the following error:
```
import spikeinterface.extractors as se
path = "/path/to/sglx"
extractors = []
for i in range(50):
    extr = se.SpikeGLXRecordingExtractor(, stream_id="imec0.ap")
    extractors.append(extr)
```

> Traceback (most recent call last):
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/pipeline_tests/test_cannot_allocate_memory_error.py", line 49, in <module>
>     extr = se.SpikeGLXRecordingExtractor("/Volumes/ceph-tononi/npx_archive/CNPIX2-Segundo/1-21-2020/SpikeGLX/1-21-2020_g0/1-21-2020_g0_imec0", stream_id="imec0.ap")
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/spikeinterface/src/spikeinterface/extractors/neoextractors/spikeglx.py", line 51, in __init__
>     NeoBaseRecordingExtractor.__init__(
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/spikeinterface/src/spikeinterface/extractors/neoextractors/neobaseextractor.py", line 94, in __init__
>     _NeoBaseExtractor.__init__(self, block_index, **neo_kwargs)
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/spikeinterface/src/spikeinterface/extractors/neoextractors/neobaseextractor.py", line 48, in __init__
>     self.neo_reader = get_neo_io_reader(self.NeoRawIOClass, **neo_kwargs)
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/spikeinterface/src/spikeinterface/extractors/neoextractors/neobaseextractor.py", line 39, in get_neo_io_reader
>     neo_reader.parse_header()
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/python-neo/neo/rawio/baserawio.py", line 178, in parse_header
>     self._parse_header()
>   File "/home/tbugnon/projects/ecephys_dev_si_pipeline/python-neo/neo/rawio/spikeglxrawio.py", line 103, in _parse_header
>     data = np.memmap(info['bin_file'], dtype='int16', mode='r', offset=0, order='C')
>   File "/home/tbugnon/miniconda3/envs/ecephys_dev_si_pipeline/lib/python3.10/site-packages/numpy/core/memmap.py", line 267, in __new__
>     mm = mmap.mmap(fid.fileno(), bytes, access=acc, offset=start)
> OSError: [Errno 12] Cannot allocate memory
> 


And is fixed by this PR.
This happens for our very large recordings. I was NOT able to replicate it on the short recordings from `spikeinterface.core.datasets` (maybe a size dependence of the number of memmaps?)

I did not have to check out PR 1602 on spikeinterface for this behavior.

Let me know if I can help somehow! Looking forward to seeing it merged
Best
Tom

samuelgarcia:
Salut Tom.
Merci pour le feedback.
I will try to finalize this soon.

TomBugnon:
Hi @samuelgarcia 
do you have any ETA regarding this PR? And/or is this branch safe to use as such or does it need more testing?

neo/rawio/spikeglxrawio.py
```diff
@@ -191,7 +202,9 @@ def _segment_t_stop(self, block_index, seg_index):
 
     def _get_signal_size(self, block_index, seg_index, stream_index):
         stream_id = self.header['signal_streams'][stream_index]['id']
-        memmap = self._memmaps[seg_index, stream_id]
+        #~ memmap = self._memmaps[seg_index, stream_id]
+        key = (seg_index, stream_id)
+        memmap = create_memmap_buffer(*self._memmap_args[key])
```
h-mayorquin, line 207: Can't you use `get_memmap_shape` to avoid creating a buffer here?

PR #1268, opened 2023-05-04
deprecate igorio as is is not maintained

JuliaSprenger:


apdavison:
It looks as though igor has been forked - https://github.com/NeuralEnsemble/python-neo/pull/1268 - so I think we can just replace "igor" with "igor2" as a dependency.

JuliaSprenger:
Good to know. Nice to see the project is still kept alive.
I will update the PR accordingly.

PR #1271, opened 2023-05-10
Use igor2 in place of igor

apdavison:
Because the latter is no longer maintained, and produces errors with recent versions of NumPy


JuliaSprenger:
@apdavison  `igor2` needs to be added to the test environment for the tests to pass on github actions.

JuliaSprenger:
Thanks for fixing the igor tests. Failing tests are not related to the igorio
This closes #1268.

PR #1274, opened 2023-05-15
[AsciiIO] Fix handling of handle channel index array annotations

JuliaSprenger:
This PR clarifies the relation between `timecolumn` and `usecol` and covers the case in which `timecolumn` is included in `usecol`. In this case the time column channel index does not require to be annotated. This leads to incorrect dimensions of the array annotations in the current master branch.


JuliaSprenger:
Hi @apdavison! I extended the existing tests to explicitely cover the content of the channel_index array annotation.

samuelgarcia (approved):

PR #1276, opened 2023-05-16
Fix DeprecationWarning during Testing due to invalid escape character

zm711:
When running pytest I keep getting this:

``` python
\neo\rawio\alphaomegarawio.py:1: DeprecationWarning: invalid escape sequence \*
```

So this is a pull request just to fix the docstring. It is the only DeprecationWarning I'm getting currently so should be an easy fix.


JuliaSprenger:
Hi @zm711 Thanks for fixing this. I don't see any reason why we need to keep the current notation, so I will merge this now as the failing tests are not related to the changes here.

PR #1278, opened 2023-05-17
Fix bug in ASCIISIGNAL IO causing failing tests

zm711:
When I tested this locally it worked for the asciisignalio failing tests. I want to see what happens with the neo tests.

### DIFF
--- a/neo/io/asciisignalio.py
+++ b/neo/io/asciisignalio.py
@@ -269,7 +269,11 @@ class AsciiSignalIO(BaseIO):
                 ana_sig = AnalogSignal(signal * self.units, sampling_rate=sampling_rate,
                                        t_start=t_start,
                                        name='multichannel')
-                ana_sig.array_annotate(channel_index=self.usecols or np.arange(signal.shape[1]))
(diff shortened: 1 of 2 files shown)


zm711:
So basically I found a problem where the usecols could include the timecolumn which meant that annotations would be different lengths. In the example
there were 3 columns used, but one was a timecolumn. So the length of the data was 2 columns, but the annotation was trying to occur over 3 columns, which was failing. 

To summarize I added in an `if else` to check if usecols is not `None` and that timecolumn is not `None` if that is the case you need to correct usecols based on timecolumn, but if either is None then you just annotate with usecols.

I also made one change to the test. The csv docs from python recommend setting the csvwriter with newline=''". This is important for windows since running the test on my windows computer would actually fail if the csvwriter was missing this line. It also didn't affect the ubuntu based test, so just to be safe I think it is better to include. @JuliaSprenger (not sure who to tag to review, but I know you respond quickly).

zm711:
Missed that you already had fix. Sorry.

PR #1290, opened 2023-06-01
Improve neuroscope documentation and refactor path resolution

h-mayorquin:
I am working on using this reader for a conversion and I found the logic for deciding how to load the binary file and the xml file quite confusing. To improve this I have done two things:

1) Add a descrition of the path resolution logic in the docstring.
2) Refactor the path resolution logic (without changing) to make it more readable.

Also, I don't know why but the filename was equal to an empty string by default. I don't know if anyone use this default but that would have resulted in an error on the previous implementation. Is this something that is intentional for reasons of the non-raw-io part of neo?


JuliaSprenger:
Hi @h-mayorquin Thanks for the cleaner rewrite. The default filename as empty string is a legacy from earlier implementations of the IO. I don't think we need to cover that case any more.

PR #1297, opened 2023-06-19
update get_io to return ios for files that do not exist yet

### LINKED ISSUE #1287: neo.get_io fails on file yet to be written
**Describe the bug**
Calling neo.get_io no longer works with the name of a file yet to be written
which means it can no longer be uses before a 
io.write(... call

**To Reproduce**
import neo
neo.get_io("tobewritten.pkl")

**Expected behaviour**
return an io

**Environment:**
 - OS: [e.g. macOS, Linux, Windows]
 i think all but found with linux
 - Python version
 found with 3.11
 - Neo version
0.12.0
Worked in neo 0.10
 - 
 - NumPy version
not relevant

**Additional context**
our workaround is
def get_neo_io(file_or_folder):
    try:
        return neo.get_io(file_or_folder)
    except ValueError as ex:
        # As neo.get_io only works with existinf files
        _, suffix = os.path.splitext(file_or_folder)
        suffix = suffix[1:].lower()
        if suffix in neo.io_by_extension:
            writer_list = neo.io_by_extension[suffix]
            return writer_list[0](file_or_folder)

zm711:
Fixes #1287 

Rather than check if the file exists it now just checks that a suffix is provided. If a suffix is provided whether the file already exists or is a future file it returns the potential ios that would support that file format. The rest of the logic for checking for potential ios remains the same.

### DIFF
--- a/neo/io/__init__.py
+++ b/neo/io/__init__.py
@@ -429,7 +429,7 @@ def list_candidate_ios(file_or_folder, ignore_patterns=['*.ini', 'README.txt', '
     """
     file_or_folder = pathlib.Path(file_or_folder)
 
-    if file_or_folder.suffix:
+    if file_or_folder.is_file():
         suffix = file_or_folder.suffix[1:].lower()
         if suffix not in io_by_extension:
             raise ValueError(f'{suffix} is not a supported format of any IO.')
@@ -452,6 +452,12 @@ def list_candidate_ios(file_or_folder, ignore_patterns=['*.ini', 'README.txt', '
     # to select all files sharing the `session1-` prefix
     elif file_or_folder.parent.exists():
         filenames = file_or_folder.parent.glob(file_or_folder.name + '*')
+
+    elif file_or_folder.suffix:
+        suffix = file_or_folder.suffix[1:].lower()
+        if suffix not in io_by_extension:
+            raise ValueError(f'{suffix} is not a supported format of any IO.')
+        return io_by_extension[suffix]
     
     else:
         raise ValueError(f'{file_or_folder} does not contain data files of a supported format')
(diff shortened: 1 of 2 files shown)


zm711:
So checking if `is_file` is important because some files don't have 'traditional' file suffixes (suffix of .1000 for example caused my initial logic to fail). So instead I just moved the not written file logic to the end of if-elif- checks.

JuliaSprenger:
Hi @zm711 Thanks for looking into this. Your approach does not solve #1287 as this function tries to initialize an io, which is by default failing for read-only ios and non-existant files.
I added some tests and changed the code to do the initial checking based on the suffix as I am not sure what exactly what you observed with 'non-traditional' file suffixes. Could you maybe also add a test case to make this more explicit?

zm711:
@JuliaSprenger Thanks so much. I wanted to add tests, but haven't really used unit test so didn't want to mess with any of the test classes. In the future I'll just add some pytest-like tests instead (like you did). As usual with file systems it turns out there needed to be a couple more checks (we both had the suffix at the beginning instinct).  Thanks for the feedback and great work on Neo :)

I guess one other question that isn't currently being addressed is whether a user might want a double-suffix format. For instance if I wanted to tarball my nix file my non-existent file would be `not_data_yet.nix.tar`, which this code might fail for. It seems pretty niche to want to generate something like that, so might just want to let the code fail rather than add more complexity, but just wanted to check. (suffix in this case would return `'.tar'` whereas suffixes would return `['nix', 'tar']`).

JuliaSprenger:
Hi @zm711 So far we didn't look into additional compression / archival formats that might add multiple suffixes. To be honest, I think this is a bit out of scope for this utility function, but if you want to invest time into developing the ultimate io finder feel free to do so.
The current changes look good to me, I only added some more cleanup for the testfiles.

zm711:
Haha. I agree it's out of scope. I also mainly work in the rawio layer, so I doubt I even have the correct mental schema to figure out all possible finders.

Issue #1017, opened 2021-07-09
Duration of time slices of an AnalogSignal differs when using or not lazy loading


kohlerca:
Hi All,

I am using `cut_segment_by_epoch` to extract trials from a Blackrock Microsystems dataset (`i140703-001.ns6` and `i140703-001.nev` publicly available [here](https://doi.gin.g-node.org/10.12751/g-node.f83565/)).

When cutting a fixed duration after the times defined in an `Event` object, the first dimension of the resulting `AnalogSignal` objects varies across the slices by 1 sample. This is odd, as the signal is sampled at 30 KHz and the interval is an integer number in ms (i.e. `300*pq.ms`), which should yield 9000 samples. Data was loaded using `lazy=True`.

Apparently, there is a bug when getting time slices using `AnalogSignalProxy` objects. In the code, `AnalogSignal` and `AnalogSignalProxy` use different algorithms to find the values of the start/end indexes in the data array:

1. The `AnalogSignal.time_slice` function finds the number of samples corresponding to the time delta in the slice, and finds the end index by adding it to the start index;
 
2. The `AnalogSignalProxy.load` function finds the indexes that corresponds to `t_start` and `t_stop` in the slice. Since the `Segment` in this dataset does not have `t_start == 0`, the truncation using `int()` may result in one less sample due to floating-point error when correcting the offset.

I used the code below to cut a single slice and compare the behavior when using or not `lazy` loading, and the slices do indeed differ by the last sample. The one with `lazy=False` is the expected output, with 300 ms duration.

```
import neo
import quantities as pq
from neo.utils import add_epoch, cut_segment_by_epoch

file = "i140703-001"
io = neo.BlackrockIO(file, nsx_to_load=6)

for lazy in True, False:

    print(f"Cutting with lazy={lazy}\n\n")

    block = io.read_block(lazy=lazy, load_waveforms=False)

    events = block.segments[0].events[0]

    if lazy:
        events = events.load()

    # As no annotations are present, use the index of one of the 'CUE-ON'
    # events and create an Event object with the epoch start time
    event_time = events[144]
    cut_event = neo.Event([event_time.magnitude] * event_time.units)

    epoch = add_epoch(block.segments[0], cut_event,
                      pre=0*pq.ms, post=300*pq.ms, attach_result=False)

    trial = cut_segment_by_epoch(block.segments[0], epoch, reset_time=True)[0]

    print(f"Shape: {trial.analogsignals[0].shape}\n")
[...]

PR #1298, opened 2023-06-20
ensure analog signal proxy and analog signal behave the same for time slices 

zm711:
In #1017 there is a discrepancy between 
`lazy=True` vs `lazy=False` because one uses `AnalogSignal` and one uses `AnalogSignalProxy`. This is likely due to the fact that `AnalogSignalProxy` is missing an `np.ceil` when calculating the `i_stop` which is included when calculating `i_start`. `AnalogSignal` accomplishes this with `np.rint` which when including the `astype(int64)` does the same as `int(np.ceil())`

https://github.com/NeuralEnsemble/python-neo/blob/53b202b2a6ae306392063d1dd2438323b4815de3/neo/io/proxyobjects.py#L191-L204

Let me know if you want a test added for this.

### DIFF
--- a/neo/test/iotest/test_proxyobjects.py
+++ b/neo/test/iotest/test_proxyobjects.py
@@ -54,8 +54,8 @@ class TestAnalogSignalProxy(BaseProxyTest):
         # ceil next sample when slicing
         anasig = proxy_anasig.load(time_slice=(1.99999 * pq.s, 5.000001 * pq.s))
         assert anasig.t_start == 2. * pq.s
-        assert anasig.duration == 3.0001 * pq.s
-        assert anasig.shape == (30001, 8)
+        assert anasig.duration == 3. * pq.s
+        assert anasig.shape == (30000, 8)
 
         # buggy time slice
         with self.assertRaises(AssertionError):
--- a/neo/test/iotest/test_proxyobjects.py
+++ b/neo/test/iotest/test_proxyobjects.py
@@ -54,8 +54,8 @@ class TestAnalogSignalProxy(BaseProxyTest):
         # ceil next sample when slicing
         anasig = proxy_anasig.load(time_slice=(1.99999 * pq.s, 5.000001 * pq.s))
         assert anasig.t_start == 2. * pq.s
-        assert anasig.duration == 3. * pq.s
-        assert anasig.shape == (30000, 8)
+        assert anasig.duration == 3.0001 * pq.s
+        assert anasig.shape == (30001, 8)
 
         # buggy time slice
         with self.assertRaises(AssertionError):
--- a/neo/io/proxyobjects.py
+++ b/neo/io/proxyobjects.py
@@ -187,10 +187,10 @@ class AnalogSignalProxy(BaseProxy):
                     assert self.t_start <= t_start <= self.t_stop, 't_start is outside'
                 else:
                     t_start = max(t_start, self.t_start)
-                # the i_start is necessary ceil
+                # the i_start is rounded to the nearest sample
                 i_start = np.rint((t_start - self.t_start).magnitude * sr.magnitude).astype(np.int64)
-                # this needed to get the real t_start of the first sample
-                # because do not necessary match what is demanded
+                # this is needed to get the real t_start of the first sample
+                # because it does not necessary match what is demanded
                 sig_t_start = self.t_start + i_start / sr
 
             if t_stop is None:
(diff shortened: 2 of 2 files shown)


zm711:
Actually the tests failed because currently the `AnalogSignalProxy` rounds down, but if the point of time slice is to include any information from within the time slice then even if we go slightly over then we need to round up to the next `i_index`. So in the tests our `t_stop = 5.00001` which is much less than a sample which for the test is `0.0001`. In the old version of the `AnalogSignalProxy` because 0.000001<0.0001 it rounds down and only includes signal information up to 5 which means we are losing one "index" of information. 

I would say based on #1017 this is not the logic users expect. If I want a time slice of 5.00001 seconds then I want the whole extra sample which would be 0.0001. 

But if the original behavior is desired then this can pull request can be closed and it needs to be made clear that the `AnalogSignalProxy` will only round down to the nearest sample rather than round up like `AnalogSignal`.

neo/test/iotest/test_proxyobjects.py
```diff
@@ -54,8 +54,8 @@ def test_AnalogSignalProxy(self):
         # ceil next sample when slicing
         anasig = proxy_anasig.load(time_slice=(1.99999 * pq.s, 5.000001 * pq.s))
         assert anasig.t_start == 2. * pq.s
-        assert anasig.duration == 3. * pq.s
-        assert anasig.shape == (30000, 8)
+        assert anasig.duration == 3.0001 * pq.s
+        assert anasig.shape == (30001, 8)
```
samuelgarcia, line 58: I think theses test were design with np.round in mind.

samuelgarcia:
Hi Zach,
thanks a lot for finding the little but very important detail!
I was not aware that astype(int64) was equivalent to int(np.ceil()). I was thinking it was rounding.
Do you have reference for that ?

JuliaSprenger:
Hi! I am still a bit confused as `np.rint(a).astype('int64') == int(np.ceil(a))` does only hold for numbers with a decimal component `>=0.5`.  I agree that we should have the same behaviour for proxy and non-proxy objects. But why can't we use the same mechanism for both implementations?

zm711:
Howdy Sam @samuelgarcia  & @JuliaSprenger 

Response to Sam:
Actually I was referring to the np.rint which would improve the behavior and less the casting with astype (although np.rint can also round down as opposed to np.ceil)—sorry for being unclear. The alternative would be change both int+np.ceil into np.rint. Which is what you’ve just said @JuliaSprenger 

Response to Julia:
Happy to switch them all to np.rint if that is the preferred choice. I think the np.ceil gives the user the sample they request even if it is a partial sample, but since neo has never guaranteed that it is fine to use the original behavior.

neo/test/iotest/test_proxyobjects.py
zm711: Reverted the tests and switched to rint

JuliaSprenger:
I would be happy with this solution. @samuelgarcia You too?

samuelgarcia:
I reading in the comment `# the i_start is necessary ceil` I think I wrote this but do not remember exactly the why.

I think maybe one possible behavior I add in mind was : "ceil the t_start and round the duration" (and then compute i_stop). The actual proposal is round both t_start and t_stop which is not exactly the same.

For sure we need to harmonize between all part of neo : neo object and proxy object.

The actual AnalogSignal.time_slice have round t_start and then round the duration. Which is also another implementation!

I think rounding the duration is important. The question is more rounding t_start of ceiling t_start.

samuelgarcia:
OK. Lets align `AnalogSignalProxy` with the historical `AnalogSignal.time_slice()`
Which is round(t_start) and round(duration).
I let you double check this behavior:
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/core/analogsignal.py#L443
https://github.com/NeuralEnsemble/python-neo/blob/master/neo/core/analogsignal.py#L437

Thanks a lot for caching and patching this and sorry for so long latencies.
Detail are super importants!

zm711:
@samuelgarcia ready for you to check. I kept the syntax the same as AnalogSignal to make it easier to compare in the future.

samuelgarcia:
Merci  beaucoup.
This OK for me.
@JuliaSprenger do you want to have a look ?

JuliaSprenger (approved):

JuliaSprenger:
Looks also good to me! Thanks for making the slicing more consistent!

zm711:
@JuliaSprenger just in case you like cleaning up the issue tracker this as far as I can tell fixes #1017.  Thanks for all help!

PR #1299, opened 2023-06-22
add workflow dispatch to manually trigger io tests on non-ubuntu os

JuliaSprenger:
This PR adds the option to manually trigger the IO test suite and select the operation system to run the test on.
The automatic triggering of the IO tests always uses 'ubuntu-latest' as operating system. In the manual trigger it is possible to specify also `macos-latest` and `windows-latest` as operation systems.

Most likely the test suite won't pass for the other operation systems with the current test setup, but this will need a 2nd PR for debugging.

### DIFF
--- a/.github/workflows/io-test.yml
+++ b/.github/workflows/io-test.yml
@@ -1,14 +1,15 @@
 name: NeoIoTest
 
-workflow_call:
-  inputs:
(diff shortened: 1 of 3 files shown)


samuelgarcia:
cool.

PR #1303, opened 2023-07-06
[WIP] Fix ephys channel ids in spikegadgetsrawio

### LINKED ISSUE #1215: Incorrect channel IDs in `SpikeGadgetsRecordingExtractor`
**Describe the bug**
originally here https://github.com/SpikeInterface/spikeinterface/issues/1260
Recordings from SpikeGadgets hardware have two possible channel maps: HWChan (a channel map defined by the amplifier) and Trodes channel map (for displaying on Trodes, the data acquisition software for SpikeGadgets). The mapping between these are in the header of the file that contains the recording (in .rec format). The channel IDs one gets with get_channel_ids from a SpikeGadgetsRecordingExtractor is in the Trodes order (though the channel names are HWChan names), which is sensible. But the order of the traces is completely scrambled (neither Trodes nor HWChan). 

**To Reproduce**
- Get example data here https://ucsf.box.com/s/ah2u2avamtf4hj54mryxqo1kvl09t7tp
- Load it
```
recording_rec = si.read_spikegadgets('.../20220919_130120/20220919_130120.rec',
                                     stream_name='trodes')
```
- Download latest version of Trodes https://bitbucket.org/mkarlsso/trodes/downloads/
- this contains a useful util called `trodesexport` which converts data to binary
- run `trodesexport` to convert the file to binary: 
```
trodesexport -rec path_to_file.rec -kilosort
```
- load binary
```
recording_bin_ks = si.BinaryRecordingExtractor(file_paths='.../20220919_130120/20220919_130120.kilosort/20220919_130120.group0.dat', sampling_frequency=30e3, num_chan=128, dtype='int16')
```
- compare. e.g. plot the first channel of `recording_rec` and `recording_bin_ks`
- extra: to get the channel map info, look in the header of rec file

**Expected behaviour**
they should match (but they don't)

**Environment:**
latest version of Trodes and spikeinterface

rly:
Fix #1215. Using code from @khl02007 - thanks!

This needs some more testing, but I am proposing this PR to continue the conversation.

### DIFF
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -54,6 +54,21 @@ class SpikeGadgetsRawIO(BaseRawIO):
     def _source_name(self):
         return self.filename
 
+    def _produce_ephys_channel_ids(self, n_total_channels, n_channels_per_chip):
+      """Compute the channel ID labels
+      The ephys channels in the .rec file are stored in the following order: 
+      hwChan ID of channel 0 of first chip, hwChan ID of channel 0 of second chip, ..., hwChan ID of channel 0 of Nth chip, 
+      hwChan ID of channel 1 of first chip, hwChan ID of channel 1 of second chip, ..., hwChan ID of channel 1 of Nth chip,
+      ...
+      So if there are 32 channels per chip and 128 channels (4 chips), then the channel IDs are:
+      0, 32, 64, 96, 1, 33, 65, 97, ..., 128
+      See also: https://github.com/NeuralEnsemble/python-neo/issues/1215
+      """
+      x = []
+      for k in range(n_channels_per_chip):
+          x.append([k+i*n_channels_per_chip for i in range(int(n_total_channels/n_channels_per_chip))])
+      return [item for sublist in x for item in sublist]
+
     def _parse_header(self):
         # parse file until "</Configuration>"
         header_size = None
@@ -77,7 +92,8 @@ class SpikeGadgetsRawIO(BaseRawIO):
         sconf = root.find('SpikeConfiguration')
 
         self._sampling_rate = float(hconf.attrib['samplingRate'])
-        num_ephy_channels = int(hconf .attrib['numChannels'])
+        num_ephy_channels = int(hconf.attrib['numChannels'])
+        num_chan_per_chip = int(sconf.attrib['chanPerChip'])
 
         # explore sub stream and count packet size
         # first bytes is 0x55
@@ -147,11 +163,14 @@ class SpikeGadgetsRawIO(BaseRawIO):
             signal_streams.append((stream_name, stream_id))
             self._mask_channels_bytes[stream_id] = []
 
+            channel_ids = self._produce_channel_ids(num_ephy_channels, num_chan_per_chip)
+
             chan_ind = 0
             for trode in sconf:
                 for schan in trode:
-                    name = 'trode' + trode.attrib['id'] + 'chan' + schan.attrib['hwChan']
-                    chan_id = schan.attrib['hwChan']
+                    chan_id = str(channel_ids[chan_ind])
+                    name = 'chan' + chan_id
+                    
                     # TODO LATER : handle gain correctly according the file version
                     units = ''
                     gain = 1.


khl02007:
I think setting the channel names to `HWChan N` (where N is channel index) would be clearer.

khl02007:
@samuelgarcia @JuliaSprenger could you please review this? would like this functionality on spikeinterface asap

alejoe91:
Hi @khl02007 

Can you remove `[WIP]` from the PR name and see why tests are failing for spikegadgets?

neo/rawio/spikegadgetsrawio.py
```diff
@@ -183,10 +353,10 @@ def _parse_header(self):
                 self.selected_streams = [self.selected_streams]
             assert isinstance(self.selected_streams, list)
 
-            keep = np.isin(signal_streams['id'], self.selected_streams)
+            keep = np.in1d(signal_streams["id"], self.selected_streams)
```
zm711, line 356:
```suggestion
            keep = np.isin(signal_streams["id"], self.selected_streams)
```

`in1d` will be deprecated and removed in future NumPy.

neo/rawio/spikegadgetsrawio.py
```diff
@@ -183,10 +353,10 @@ def _parse_header(self):
...
             signal_streams = signal_streams[keep]
 
-            keep = np.isin(signal_channels['stream_id'], self.selected_streams)
+            keep = np.in1d(signal_channels["stream_id"], self.selected_streams)
```
zm711, line 359:
```suggestion
            keep = np.isin(signal_channels["stream_id"], self.selected_streams)
```

zm711 (commented):
@rly, just want to ping this to see if you want to figure out why tests are failing for when you have time :)

zm711:
@rly we also ran black on the whole code base and did some cleanup with docstrings and asserts which have led to merge conflicts. If you have the bandwidth to fix those really quick that would be awesome!

zm711:
> raw_unit16 = raw_unit8_mask.reshape(-1).view("int16").reshape(shape)
E       ValueError: cannot reshape array of size 512 into shape (1024,0)

This is the error causing tests to fail.

zm711:
@khl02007 , thanks for checking. The way the testing works is it auto-detects what the possible streams are based on the reader and then iterates through them. So that means that the way the reader is implemented in this PR is not quite following the appropriate strategy for a Neo reader (ie this would need to be fixed at the Neo reader level). If this is the case it seems like you would need to add some more logic in the reader to prevent the reader itself from making mistakes between analog and digital signaling. Does that make sense?

khl02007:
@zm711 I think I understand, but what exactly needs to be done to address this concern then? How does the reader automatically detect the possible streams? What is the appropriate strategy for a neo reader?

zm711:
I just talked to @samuelgarcia and he said he worked don this io before. He might be best to help a bit. He said he was having some struggles trying to get the additional streams due to the way the data is set up so it might be a bit more work to get this to work within the Neo structure. I would either reach out to him directly or ping him every once in a while.

zm711:
@rly and @khl02007.

We discussed this a bit and from what Sam remembers this format would be very hard to fit within the model of `analogsignal` chunk that Neo uses because of the way that digital channels are stored. The hack that we used for another IO was to extract those values, but not to expose the stream at the Neo machinery level. Then for users that need that info they could just grab the private stream information from a private attribute.

So more concretely previously for the intanrawio to get access for the digital information you would have to do:

```python

reader = IntanRawIO()
reader.parse_header()

digital_data = reader._raw_data['DIGITAL-IN']
```

So you could do something similar where you store these channels in the raw data container, but you don't add them to channel info for now.

The future goal is to have a separate digitalsignal stream that would allow us more flexibility in loading these types of signals, but that is on the backburner for now.

neo/rawio/spikegadgetsrawio.py
```diff
@@ -75,10 +81,31 @@ def __init__(self, filename="", selected_streams=None):
...
+        x = []
+        for k in range(n_channels_per_chip):
+            x.append(
+                [
+                    k + i * n_channels_per_chip
+                    for i in range(int(n_total_channels / n_channels_per_chip))
+                ]
+            )
+        return [item for sublist in x for item in sublist]
+
```
zm711, line 108: This is now in main. Let us know if you want to keep this open and try to get the digital channels working with Neo or whether you want to close this PR now?

khl02007: It's up to @rly, but I think it may be best to just keep this open and work towards getting the analog and digital IOs loadable with Neo.

Issue #1306, opened 2023-07-14
Why not `segment.add()`?


kjohnsen:
**Is your feature request related to a problem? Please describe.**
I'm trying to take a bunch of data objects and stick them in a segment. Looks like I need to recreate the functionality of `Group.add` which automatically finds the right container to add to if I don't know the class a priori.

**Describe the solution you'd like**
I'd like to be able to call `segment.add(object_of_unknown_class)` like I can with a `Group`. Is there a reason this isn't currently supported? Why isn't `add` defined at the `Container` level instead of `Group`?

**Describe alternatives you've considered**
Copying and pasting `Group`'s code. This is my workaround
```python
def add_to_neo_segment(
    segment: neo.core.Segment, *objects: neo.core.dataobject.DataObject
):
    """Taken from neo.core.group.Group."""
    container_lookup = {
        cls_name: getattr(segment, container_name)
        for cls_name, container_name in zip(
            segment._child_objects, segment._child_containers
        )
    }

    for obj in objects:
        cls = obj.__class__
        if hasattr(cls, "proxy_for"):
            cls = cls.proxy_for
        container = container_lookup[cls.__name__]
        container.append(obj)
```

JuliaSprenger:
Hi @kjohnsen This sound like a very good idea to me. Feel free to open a PR to add the new utility functionality.

apdavison:
Implemented in #1307

PR #1307, opened 2023-07-14
add `add()` function to `Container` for use on `Segment` and `Block`

kjohnsen:
As suggested in #1306

### DIFF
--- a/neo/test/coretest/test_block.py
+++ b/neo/test/coretest/test_block.py
@@ -466,6 +466,32 @@ class TestBlock(unittest.TestCase):
         blk.segments += [Segment(), Segment()]
         assert len(blk.segments) == 3
 
+    def test_add(self):
+        blk = self.blocks[0]
+        new_blk = simple_block()
+        n_groups_start = len(new_blk.groups)
+        for group in blk.groups:
+            assert group not in new_blk.groups
+            new_blk.add(group)
+            assert group in new_blk.groups
+        assert len(new_blk.groups) == n_groups_start + len(blk.groups)
+
+        n_segs_start = len(new_blk.segments)
+        for seg in blk.segments:
+            assert seg not in new_blk.segments
+            new_blk.add(seg)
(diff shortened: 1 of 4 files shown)


kjohnsen:
This should simplify the API a tad. Users can call `seg.add(obj1, obj2, ...)` instead of `seg.whatevercontainer.append(obj)` one at a time

apdavison (approved):

apdavison:
Very nice. Please could you also add some tests for `Block.add()` ?

kjohnsen:
As you wish

apdavison:
Many thanks!

neo/test/coretest/test_block.py
```diff
@@ -466,6 +466,33 @@ def test_segment_list(self):
...
+
+        n_segs_start = len(new_blk.segments)
+        for seg in blk.segments:
+            assert seg not in new_blk.segments
+            new_blk.add(seg)
+            assert seg in new_blk.segments
+        assert len(new_blk.segments) == n_segs_start + len(blk.segments)
+
+        # test adding multiple at once
+        blk = self.blocks[1]
```
JuliaSprenger, line 487: For completeness it would be good to also check here if the new object to add is already part of the list.
```suggestion
        blk = self.blocks[1]
        for group in blk.groups:
            assert group not in new_blk.groups
```

neo/test/coretest/test_block.py
```diff
@@ -466,6 +466,33 @@ def test_segment_list(self):
...
+        n_groups_start = len(new_blk.groups)
+        new_blk.add(*blk.groups)
+        assert len(new_blk.groups) == n_groups_start + len(blk.groups)
+
+        n_segs_start = len(new_blk.segments)
```
JuliaSprenger, line 492:
```suggestion
        for seg in blk.segments:
            assert seg not in new_blk.segments
        n_segs_start = len(new_blk.segments)
```

Issue #1308, opened 2023-07-14
merge fails for object in both Segment and Group


kjohnsen:
**Describe the bug**
I'm trying to merge two blocks together, each having one segment, into one block with two segments. But I have problems calling merge since I have spike trains in groups. Why am I told SpikeTrain can only have Segment, not Group parents? No errors popped up before merging. From what I understand, an object can be in more than one container, but only have one *parent*.

**To Reproduce**
```python
#%%
import neo
#%%
blocks = []
for i in range(2):
    block = neo.Block()
    seg = neo.Segment(name=f'segment {i}')
    block.segments.append(seg)
    g = neo.Group()
    st = neo.SpikeTrain([1,2,3], t_stop=4, units='ms')
    g.spiketrains.append(st)
    block.groups.append(g)
    seg.spiketrains.append(st)
    blocks.append(block)
blocks
# %%
blocks[0].merge(blocks[1])
blocks[0]
```

<details>
<summary>Output:</summary>

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/home/kyle/scratch/python-neo/test.py in line 2
      [16](file:///home/kyle/scratch/python-neo/test.py?line=15) # %%
----> [17](file:///home/kyle/scratch/python-neo/test.py?line=16) blocks[0].merge(blocks[1])
      [18](file:///home/kyle/scratch/python-neo/test.py?line=17) blocks[0]

File [~/scratch/python-neo/neo/core/container.py:491](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/kyle/scratch/python-neo/~/scratch/python-neo/neo/core/container.py:491), in Container.merge(self, other)
    [489](file:///home/kyle/scratch/python-neo/neo/core/container.py?line=488)     continue
    [490](file:///home/kyle/scratch/python-neo/neo/core/container.py?line=489) if obj.name in lookup:
--> [491](file:///home/kyle/scratch/python-neo/neo/core/container.py?line=490)     lookup[obj.name].merge(obj)
    [492](file:///home/kyle/scratch/python-neo/neo/core/container.py?line=491) else:
    [493](file:///home/kyle/scratch/python-neo/neo/core/container.py?line=492)     lookup[obj.name] = obj

[...]

PR #1309, opened 2023-07-14
enable merging of groups

kjohnsen:
As described in #1308.

I don't understand the merge process well enough to know if this behavior is desired for merging. But if it is, this PR should probably add to the test suite for before merging.

### DIFF
--- a/neo/core/container.py
+++ b/neo/core/container.py
@@ -514,7 +514,8 @@ class Container(BaseNeo):
                     lookup[obj.name] = obj
                     ids.append(id(obj))
                     getattr(self, container).append(obj)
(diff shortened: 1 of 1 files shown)


zm711:
@apdavison do you have any opinions on this? Do we want to move the milestones around? It is still a draft. @kjohnsen Did you want to receive actual feedback on this? Or add testing if Andrew approves of this? We are cleaning up PRs etc.

kjohnsen:
@zm711  I think this is best left to your judgment since this was a while ago and I don't know the Neo codebase very well in general. I do not have time to add testing myself right now

PR #1310, opened 2023-07-24
Fixed bug in maxwellrawio.py that shuffled channels 

hornauerp:
Fix for issue [https://github.com/SpikeInterface/spikeinterface/issues/1691](url) that shuffled channels when selecting and concatenating channels from MaxWell recordings with different mappings.

### DIFF
--- a/neo/rawio/maxwellrawio.py
+++ b/neo/rawio/maxwellrawio.py
@@ -183,9 +183,10 @@ class MaxwellRawIO(BaseRawIO):
             if np.array(channel_indexes).size > 1 and np.any(np.diff(channel_indexes) < 0):
                 # get around h5py constraint that it does not allow datasets
                 # to be indexed out of order
-                sorted_channel_indexes = np.sort(channel_indexes)
-                resorted_indexes = np.array(
-                    [list(sorted_channel_indexes).index(ch) for ch in channel_indexes])
+                order_f = np.argsort(channel_indexes) 
+                sorted_channel_indexes = channel_indexes[order_f] 
+                # use argsort again on order_f to obtain resorted_indexes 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Thank you for the patch.

@alejoe91 : can you check this ?

alejoe91:
Thanks @hornauerp!

Could you add a plot with the behavior before and after the fix? :)

neo/rawio/maxwellrawio.py
```diff
@@ -185,7 +185,7 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop,
                 # to be indexed out of order
                 sorted_channel_indexes = np.sort(channel_indexes)
                 resorted_indexes = np.array(
-                    [list(channel_indexes).index(ch) for ch in sorted_channel_indexes])
+                    [list(sorted_channel_indexes).index(ch) for ch in channel_indexes])
```
alejoe91, line 188: @hornauerp I think there is a better implementation.

Can you try if this gives the same result:
```
order_f = np.argsort(channel_indexes)
sorted_channel_idxeses = channel_indexes[order_f]
# use argsort again on order_f to obtain resorted_indexes
resorted_indexes = np.argsort(order_f)
```

hornauerp: Yes, same outcome, but slightly faster. (17 vs 50 us)

I just wanted to stick to the original code as much as possible to facilitate the review process, but yours is definitely the more optimal implementation.

hornauerp:
Before fix as shown in the original issue.

After fix (different recording though):
grafik

alejoe91 (approved):

PR #1314, opened 2023-07-27
do not allow Groups to contain Segments

apdavison:
as proposed in #1026, which this PR partially replaces


apdavison:
I know we've discussed it several times, but it's not clear to me what the disadvantages are of allowing grouping of Segments. Does it break anything?

The use case I have in mind is where a stimulus is repeated several times, and each recording is in a separate Segment, but then there is some artefact during one of the segments, so we want to group the "good" segments. (A similar thing could of course be achieved using annotations.)

PR #1329, opened 2023-09-14
Update calls to NumPy dtypes being removed in NumPy 2.0

zm711:
https://numpy.org/devdocs/release/2.0.0-notes.html

NumPy 2.0  will remove np.NaN, np.float_, and np.Inf see link for more removals. I searched the code base and updated the code that will break when NumPy 2.0 comes out later this year/early next year.


zm711:
In addition

`np.in1d` will be deprecated. I didn't change that, but can add another commit if you want me to just take care of that now.

zm711:
np.in1d taken care of in #1330 .

PR #1330, opened 2023-09-15
`np.in1d` to `np.isin` as the former will be deprecated 

h-mayorquin:
Quoting @zm711 in spikeinterface:

> Just wanted to put this on everyone's radar. Not sure when this will be a hard error. I assume in a few versions (similar to the np.bool/np.float which deprecated in 1.20 and hard errored in 1.24)

> https://numpy.org/devdocs/release/2.0.0-notes.html

This PR takes care of this:


zm711:
Thanks! I took care of the hard faults in #1329 (Neo uses some aliases that will be removed in the 2.0)

PR #1333, opened 2023-10-13
Update Examples on RTD

zm711:
@apdavison,

Where are the rtd logs for Neo? I'm trying to update some of the examples so I want to see why they are mostly failing on RTD (only one works currently).
I'm also not super familiar with using Group/ChannelView since I mostly use Neo from the rawio layer, but I'll try to add that into this example soon.


zm711:
I'll fix pep 8 stuff at the end.

apdavison:
@zm711 The rtd logs are here: https://readthedocs.org/projects/neo/builds/

zm711:
Moved this to #1429.

PR #1229, opened 2023-03-02
Add type hints to ExampleRawIO

joschaschmiedt:
As I was writing a new I/O module and figuring out what the to-be-overridden functions of `BaseRawIO` have to do , I added type hints to the `examplerawio.py` file. These may be useful to other developers as well.

Issue #1123, opened 2022-05-31
Adding type hints 


codingchipmunk:
Hello everyone,
I would like to go through the code in the current master-branch and add type hints to most stuff.
However I wanted to ask if this enhancement would be desired or categorically rejected before I put in the work.

I am currently developing a system which will rely heavly on neo and I believe adding type hints will be beneficial for developing systems based on neo. Also data scientists will most likely also profit from this, as  Visual Studio Code and other IDEs begin to support Jupyter notebooks and hence can also make use of the hints.

apdavison:
That would be very welcome.

samuelgarcia:
Hi all,
I will be the grumpy guy.
Sorry for that.
I do not like to much the type hint in python.

Maybe I am too old and do not adapt fast enought,  my feeling is that this python type hint is usefull mainly for vscode checking.
I don't see a clear advantage between the verbosity added in the code and the global benefit.

Further more when playing with args that can be numpy.array or list or None. It become more comlex.
Herewe also will have quantity object (scalar or vector).
neo is not only str/int/float
To cover all possible input, the machinery behind type hints could be become complex in neo.
Maybe I am wrong.

I would vote for a call for discussion.


Sam, the grumpy

apdavison:
I understand where you're coming from, Sam. For smaller projects, or for applications rather than libraries, I tend to agree that the cost sometimes exceeds the benefits.

For a project like Neo, however, which is a dependency for many other projects, I think 

1. the process of adding type hints tends to improve the code, as it catches bugs and exposes hidden assumptions and corner-cutting;
2. having an explicit interface makes it easier for projects that depend on Neo to use it correctly;
3. you get IDE support, as mentioned.

(by the way I am planning to add type hints to PyNN sometime in the next 12 months, and for that I'll need to at least have type stubs for Neo.)

As a compromise approach, how about putting the type information into .pyi stub files? This avoids adding verbosity to the main code base, but makes the type/interface information available to people who need it. (The disadvantage is that it's more effort to keep the .py and .pyi files in sync).

JuliaSprenger:
Let's wait for a quantities release including typing and then include it in the core library.

joschaschmiedt:
I added type hints to `ExampleRawIO` to make writing a new I/O module easier: #1229

JuliaSprenger:
> Let's wait for a quantities release including typing and then include it in the core library.
[...]

PR #1334, opened 2023-10-13
Add Typing to `baseio`, `baserawio`, and `exampleio`

zm711:
Replaces #1229 because that now has merge conflicts.

I know this has been discussed #1123. Sam is against, others are typically for. This is just a draft to think about (& correct or reject), but I think adding type hints to some places exposed to the user would be beneficial. By putting the hints into the base it propagates to all ios to minimize other typing needing to be added, although developers can override the typing in their specific io.

I know that neo has complicated inputs, but I think type hints work best to give a suggested appropriate input (And as a nod to Sam I didn't add type hints for the returns yet).

### DIFF
--- a/neo/io/baseio.py
+++ b/neo/io/baseio.py
@@ -98,7 +98,7 @@ class BaseIO:
 
     mode = 'file'  # or 'fake' or 'dir' or 'database'
 
-    def __init__(self, filename: str | Path =None, **kargs):
+    def __init__(self, filename: str | Path = None, **kargs):
         self.filename = str(filename)
         # create a logger for the IO class
         fullname = self.__class__.__module__ + '.' + self.__class__.__name__
@@ -113,7 +113,7 @@ class BaseIO:
             corelogger.addHandler(logging_handler)
 
     ######## General read/write methods #######################
-    def read(self, lazy:bool=False, **kargs):
+    def read(self, lazy: bool = False, **kargs):
         """
         Return all data from the file as a list of Blocks
         """
--- a/neo/io/baseio.py
+++ b/neo/io/baseio.py
@@ -10,6 +10,8 @@ BaseIO        - abstract class which should be overridden, managing how a
 
 If you want a model for developing a new IO start from exampleIO.
 """
+from __future__ import annotations
+from pathlib import Path
 
 try:
     from collections.abc import Sequence
@@ -96,7 +98,7 @@ class BaseIO:
 
     mode = 'file'  # or 'fake' or 'dir' or 'database'
 
-    def __init__(self, filename=None, **kargs):
+    def __init__(self, filename: str | Path =None, **kargs):
         self.filename = str(filename)
         # create a logger for the IO class
         fullname = self.__class__.__module__ + '.' + self.__class__.__name__
@@ -111,7 +113,7 @@ class BaseIO:
             corelogger.addHandler(logging_handler)
 
     ######## General read/write methods #######################
-    def read(self, lazy=False, **kargs):
+    def read(self, lazy:bool=False, **kargs):
         """
         Return all data from the file as a list of Blocks
         """
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -651,7 +651,7 @@ class BaseRawIO:
         return self._spike_count(block_index, seg_index, spike_channel_index)
 
     def get_spike_timestamps(self,
-                             block_index: int  = 0,
+                             block_index: int = 0,
                              seg_index: int = 0,
                              spike_channel_index: int = 0,
                              t_start: float | None = None,
@@ -725,8 +725,10 @@ class BaseRawIO:
             block_index, seg_index, event_channel_index, t_start, t_stop)
         return timestamp, durations, labels
 
-    def rescale_event_timestamp(self, event_timestamps: np.ndarray, dtype: np.dtype= 'float64',
-                    event_channel_index:int = 0):
+    def rescale_event_timestamp(self,
+                                event_timestamps: np.ndarray,
+                                dtype: np.dtype = 'float64',
+                                event_channel_index: int = 0):
         """
         Rescale event timestamps to seconds.
         """
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -134,7 +134,7 @@ class BaseRawIO:
 
     rawmode = None  # one key from possible_raw_modes
 
-    def __init__(self, use_cache: bool=False, cache_path='same_as_resource', **kargs):
+    def __init__(self, use_cache: bool = False, cache_path = 'same_as_resource', **kargs):
         """
         :TODO: Why multi-file would have a single filename is confusing here - shouldn't
         the name of this argument be filenames_list or filenames_base or similar?
@@ -474,7 +474,7 @@ class BaseRawIO:
         channel_indexes = np.array([chan_ids.index(chan_id) for chan_id in channel_ids])
         return channel_indexes
 
-    def _get_channel_indexes(self, stream_index:int, channel_indexes: list[int]|None, channel_names: list[str]|None, channel_ids: list[str]|None):
+    def _get_channel_indexes(self, stream_index: int, channel_indexes: list[int] | None, channel_names: list[str] | None, channel_ids: list[str] | None):
         """
         Select channel_indexes for a stream based on channel_indexes/channel_names/channel_ids
         depending which is not None.
@@ -485,7 +485,7 @@ class BaseRawIO:
             channel_indexes = self.channel_id_to_index(stream_index, channel_ids)
         return channel_indexes
 
-    def _get_stream_index_from_arg(self, stream_index_arg):
+    def _get_stream_index_from_arg(self, stream_index_arg: int | None):
         if stream_index_arg is None:
             assert self.header['signal_streams'].size == 1
             stream_index = 0
@@ -494,7 +494,7 @@ class BaseRawIO:
             stream_index = stream_index_arg
         return stream_index
 
-    def get_signal_size(self, block_index: int, seg_index: int, stream_index: int|None=None):
+    def get_signal_size(self, block_index: int, seg_index: int, stream_index: int | None = None):
         """
         Retrieve the length of a single section of the channels in a stream.
         :param block_index:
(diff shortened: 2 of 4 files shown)


zm711:
I'll fix formatting after I know whether this is supported or not.

JoeZiminski:
I would add my support to such a PR. Typing is fantastic for catching bugs, in general improving robustness. Since I started using type hints last year they have caught a lot of potential bugs and saved me a lot of trouble!

samuelgarcia:
Having the role of the old dinosaur is not so easy in a group but I am trying my best to maintain my public character.

zm711:
> Having the role of the old dinosaur is not so easy in a group but I am trying my best to maintain my public character.

@samuelgarcia never change. Your example error messages here and in probeinterface are always the highlight of my proofreading.

alejoe91:
@zm711 this looks good to me! 

@samuelgarcia can we merge?

zm711:
Thanks @alejoe91. I have a feeling when there is typing in the PR title @samuelgarcia has a filter for his alerts to jeter dans la poubelle.

samuelgarcia:
@zm711 it is even worst. typing PR are triggering in me a brain state nearby freezing.
I always need to turn off the computer for at least one hour to recover totally.

neo/io/baseio.py
```diff
@@ -96,7 +98,7 @@ class BaseIO:
 
     mode = 'file'  # or 'fake' or 'dir' or 'database'
 
-    def __init__(self, filename=None, **kargs):
+    def __init__(self, filename: str | Path = None, **kargs):
```
apdavison, line 101: The use of `|` in type hints was added in Python 3.10. Since we still support Python 3.8 and 3.9, this should be `Union[str, Path]`

neo/io/baseio.py
```diff
@@ -111,7 +113,7 @@ def __init__(self, filename=None, **kargs):
             corelogger.addHandler(logging_handler)
 
     ######## General read/write methods #######################
-    def read(self, lazy=False, **kargs):
+    def read(self, lazy: bool = False, **kargs):
```
apdavison, line 116: It would be nice to have an annotation for the return type, but perhaps that's for a future PR

neo/rawio/baserawio.py
```diff
@@ -369,7 +370,7 @@ def block_count(self):
         """return number of blocks"""
         return self.header['nb_block']
 
-    def segment_count(self, block_index):
+    def segment_count(self, block_index: int):
```
apdavison, line 373: add annotation for the return type?

neo/rawio/baserawio.py
```diff
@@ -473,7 +474,11 @@ def channel_id_to_index(self, stream_index, channel_ids):
         channel_indexes = np.array([chan_ids.index(chan_id) for chan_id in channel_ids])
         return channel_indexes
 
-    def _get_channel_indexes(self, stream_index, channel_indexes, channel_names, channel_ids):
+    def _get_channel_indexes(self,
+                             stream_index: int,
+                             channel_indexes: list[int] | None,
```
apdavison, line 479: same comment as above concerning `|` vs `Union`.
Strictly speaking this should also use `List`, since the use of `list` in this way was only introduced in Python 3.8, but since we'll be dropping 3.8 support around April 2024, maybe we can ignore that.

neo/rawio/baserawio.py
```diff
@@ -587,8 +602,13 @@ def get_analogsignal_chunk(self, block_index=0, seg_index=0, i_start=None, i_sto
 
         return raw_chunk
 
-    def rescale_signal_raw_to_float(self, raw_signal, dtype='float32', stream_index=None,
-                                    channel_indexes=None, channel_names=None, channel_ids=None):
+    def rescale_signal_raw_to_float(self,
+                                    raw_signal: np.ndarray,
+                                    dtype: np.dtype = 'float32',
```
apdavison, line 607: shouldn't this be `Union[np.dtype, str]`, since `'float32'` is a string?

neo/rawio/baserawio.py
```diff
@@ -693,21 +725,23 @@ def get_event_timestamps(self, block_index=0, seg_index=0, event_channel_index=0
...
-                    event_channel_index=0):
+    def rescale_epoch_duration(self, raw_duration: np.ndarray, dtype: np.dtype = 'float64',
+                    event_channel_index: int = 0):
         """
         Rescale epoch raw duration to seconds.
         """
         return self._rescale_epoch_duration(raw_duration, dtype, event_channel_index)
 
-    def setup_cache(self, cache_path, **init_kargs):
+    def setup_cache(self, cache_path: 'home' | 'same_as_resource', **init_kargs):
```
apdavison, line 744: shouldn't this be `cache_path: Literal['home', 'same_as_resource']` ?

neo/rawio/baserawio.py
```diff
@@ -735,7 +769,7 @@ def setup_cache(self, cache_path, **init_kargs):
             dirname = os.path.dirname(resource_name)
         else:
             assert os.path.exists(cache_path), \
-                'cache_path do not exists use "home" or "same_as_resource" to make this auto'
+                'cache_path does not exists use "home" or "same_as_resource" to make this auto'
```
apdavison, line 772: Not really related to the PR, but I don't understand what this error message is telling me. More punctutation and context needed.

apdavison (changes requested):
I am in favour of this PR, but some changes are needed for the type hints to be compatible with Python 3.8 and 3.9.

neo/rawio/baserawio.py
zm711: I thought np.dtype included "float32" since numpy accepts float32? But maybe I'm wrong. We don't accept any string (for example I can do "float100000") so I think the only floats accepted are numpy floats which are contained by the numpy.dtype.

neo/rawio/baserawio.py
zm711: Neither do I. But I think @samuelgarcia would have to comment on that.

neo/io/baseio.py
```diff
@@ -10,6 +10,8 @@
 
 If you want a model for developing a new IO start from exampleIO.
 """
+from __future__ import annotations
```
zm711, line 13: This allows 3.8 and 3.9 to accept the |

neo/io/baseio.py
zm711: Sam is so against typing I was trying to make small steps to start to get everyone on board. I think doing returns would be better in a second PR just to slowly move toward full typing.

neo/rawio/baserawio.py
```diff
@@ -67,6 +67,7 @@
 constructions of a RawIO for a given set of data.
 
 """
+from __future__ import annotations
```
zm711, line 70: same here this gives us 3.9 and 3.10 typing even in 3.8

zm711:
> I am in favour of this PR, but some changes are needed for the type hints to be compatible with Python 3.8 and 3.9.

I think I addressed all of this in comments. But just to overall explain for future reference

I have added: 

```python
from __future__ import annotations
```

to the three files with typing, which allows us to use type hints from 3.9 and 3.10 in 3.8, 3.9, and 3.10 to keep things clean. This means we can:

* use `|` instead of needing `Literal`, but if `Literal` is preferred I can switch to that. 
* use built-in types even in 3.8 `list` instead of `List`, `dict` instead of `Dict`

neo/rawio/baserawio.py
apdavison: I don't have a very deep understanding of type hints in Python, but my reasoning was based on this:
```
In [1]: import numpy as np

In [2]: x = np.arange(5)

In [3]: x.dtype
Out[3]: dtype('int64')

In [4]: isinstance(x.dtype, np.dtype)
Out[4]: True

In [5]: isinstance("int64", np.dtype)
Out[5]: False
```

apdavison:
> so if `|` was going to break the code we should've seen that during the testing.

yes, that confused me, but since you've explained about `from __future__ import annotations` it all makes sense!

There's one remaining point I'm not sure about (numpy.dtype), but otherwise this looks good.

apdavison (approved):

neo/rawio/baserawio.py
zm711: That's a great point I meant more in the sense that you can add `int64` as the dtype in any np.function.

```
import numpy as np

>>> x = np.arange(5, dtype='int64')
>>> x.dtype
dtype('int64')

>>> x = np.arange(5, dtype='int2')
  File "<stdin>", line 1, in <module>
TypeError: data type 'int2' not understood
```

It won't accept any string, but only the strings allowable by the numpy.dtype constructor. So you're right in that it is a string as its type, but it's only the strings accepted by numpy. Maybe there is a better way to represent that, but I thought the most succinct (and generally understandable) way to do this would be to tell people to just use `np.type`. I can do some reading and see if there is a more standard way to represent this though!

zm711:
Officially the better typing would be:

```python
import numpy.typing as npt

my_function(x, dtype: npt.DTypeLike = 'int32')
```

This indicates that something can be coerced into a numpy dtype. I currently have the dtype listed as the dtype constructor which is the constructor that takes an input and attempts to convert it to a numpy.dtype. So @apdavison you are correct that the current type is not strictly correct. Would you like me to change or do you think people will understand despite the mention of the constructor rather than the typing. I think the issue we will run into is that some old versions of numpy won't have the numpy typing module so it would be harder to support.

neo/rawio/baserawio.py
apdavison: Looking into numpy's own typing, it looks like maybe we should use `import numpy.typing as npt` and 
```dtype: npt.DTypeLike = 'float32'```

zm711: I just put the same comment below :) haha.

apdavison: sorry, just saw you posted the same thing below.

apdavison:
> I think the issue we will run into is that some old versions of numpy won't have the numpy typing module so it would be harder to support

That's a good point. For now we're not doing strict type checking with mypy or whatever, so I agree let's leave it as is for now.

zm711:
Also just for reference I checked and typing was added in 1.21 so we would need to wait until our minimum numpy is that to be safe.

Issue #1321, opened 2023-09-01
Correct handling of Sync channel for SpikeGLX and Open Ephys


alejoe91:
**Describe the bug**
Both SpikeGLX and Open Ephys - Neuropix-PXI can **optionally** save a 385th SYNC channel in the binary file.

While both IOs have a `load_sync_channel` option, all cases are not currently handled. In particular:
- SpikeGLX does not check wheter a SYNC channel is present, and just remove the last channel if `load_sync_channel=False` (see [here](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/spikeglxrawio.py#L127-L128)).
- Open Ephys handles it a bit betterr, BUT it doesn't throw an error if `load_sync_channel=True`, but not present in the data (same for SpikeGLX)

**Expected behaviour**
SpikeGLX and OpenEphys correctly handle the SYNC option of NP probes in a similar way, handling all possible cases.

**Additional context**
Currently on GIN we only have:
- SpikeGLX with SYNC
- OpenEphys without SYNC

We need to gather:
- SpikeGLX without SYNC (any volunteer? @DradeAW ? )
- Open Ephys with SYNC (I can get one)

samuelgarcia:
I do not think @DradeAW is the correct candidate for this because he is writting the thesis manuscript!

PR #1335, opened 2023-10-16
Add 2013 and 2014 neuropixels versions.

JoeZiminski:
Hi All,

Some newer versions of NP probes (I think 'commercial 2B probes') have the metadata probe type as '2013' or '2014' rather than '21' or '24' as described [here](https://billkarsh.github.io/SpikeGLX/Sgl_help/Metadata_30.html).

It seems the only changes are to the 'Reference IDs':

```
Type-2013,2014 reference ID values are {0=ext, 1=gnd, [2..5]=tip[0..3]}. On-shank reference electrodes are removed from commercial 2B probes.
```

I assume these are not relevant for neo as these reference channels are automatically set to off?
i.e. from [here](https://billkarsh.github.io/SpikeGLX/Sgl_help/UserManual.html)

```
You can mark a site used=0 if you know it is broken or disconnected. For imec probes, we automatically set used=0 for reference sites and those you have turned off (bad channels) in the IM Setup tab.
```

So, simply also accepting '2013' and '2014' probes seems to work, though it feels a little too good to be true 😅 

I have a '2013' file to test against, happy to cut down and add to the testing repo. Currently I do not have a '2014' so this could be removed from this conditional in the case we want to test against it before rolling out.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -399,7 +399,7 @@ def extract_stream_info(meta_file, meta):
                 per_channel_gain[c] = 1. / float(v)
             gain_factor = float(meta['imAiRangeMax']) / 512
             channel_gains = gain_factor * per_channel_gain * 1e6
-        elif meta['imDatPrb_type'] in ('21', '24') and stream_kind == 'ap':
+        elif meta['imDatPrb_type'] in ('21', '24', '2013', '2014') and stream_kind == 'ap':
             # This work with NP 2.0 case with different metadata versions
             # https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_20.md#channel-entries-by-type
             # https://github.com/billkarsh/SpikeGLX/blob/gh-pages/Support/Metadata_20.md#imec


neo/rawio/spikeglxrawio.py
```diff
@@ -399,7 +399,7 @@ def extract_stream_info(meta_file, meta):
                 per_channel_gain[c] = 1. / float(v)
             gain_factor = float(meta['imAiRangeMax']) / 512
             channel_gains = gain_factor * per_channel_gain * 1e6
-        elif meta['imDatPrb_type'] in ('21', '24') and stream_kind == 'ap':
+        elif meta['imDatPrb_type'] in ('21', '24', '2013', '2014') and stream_kind == 'ap':
```
alejoe91, line 402:
```suggestion
        elif meta['imDatPrb_type'] in ('21', '24', '2003', '2004', '2013', '2014') and stream_kind == 'ap':
```
@JoeZiminski 2003 and 2004 will be the commercial versions of the 1-shank design

samuelgarcia:
The PR #1321  also made the same fix so we can close this.
Thanks Joe for poiting this.

PR #1338, opened 2023-10-20
Add support for Blackrock NSx files with per-sample PTP nanosecond timestamps

### LINKED ISSUE #1332: BlackRock data format 3.0
**Describe the bug**
Read *.ns6 data from BlackRock causes a "Too many open files" error in __read_nsx_data_variant_b  function nearly immediately.

**To Reproduce**
Something  like this:

```
import neo

data_dir = "data/test_data/HubTests/"
path_gemini = data_dir + "233117-001.ns6"  # new data

reader = neo.io.BlackrockIO(filename=path_gemini)

```
(Data cannot be shared publicly.) 

**Expected behaviour**

Data is read into memory

**Environment:**
 - OS: Windows 10
 - Python version: 3.7.8 
 - Neo version: momentarily not available, most recent downloadable
 - NumPy version: 1.21.6

**Additional context**

New Blackrock device uses nsx-format version 3.0.
The neo code seems to have execute branches for version 3.0, 
but states in comments that it handles nsx-formats up to version 2.3.  

Prior devices (using nsx version 2.3) produced data that were read without problem,
so it seems that either blackrock is implementing the 3.0 format incorrectly, or neo
is handling it incorrectly.

There could be a marker, or signal, in the v3.0-data that is interpreted as 
a "recording pause" triggering a segment split by neo code? 

The arrival of nsx v3.0 is recent for us, with new equipment, but the version 
may have been available for a long while. 

Is this a known problem?

cboulay:
Fixes #1332

Blackrock's newest signal acquisition devices use nanosecond timestamps from clocks synchronized with PTP. This allows synchronization between devices that might have slightly different clock rates and not capture exactly 30_000 samples per reference second.

Individual per-sample timestamps are stored in a data header, per sample.

Previously, the presence of a header in the data block would indicate the start of a new segment, and methods would look for the header to start a new memmap. That previous solution, when applied to millions of headers in the PTP-enabled files, led to OS-level problems by creating too many memmaps.

This PR allows the new PTP-enabled NSx files to be loaded.

Note: the time range of data files no longer start at 0! They start at whatever the monotonic clock was when the NSx file was recorded. We can't simply re-zero the timestamps per file because there might be multiple sets of files that were recorded simultaneously (i.e., from different pieces of hardware) but didn't start at the exact same time. We could re-zero with the common minimum across all devices, but the BlackrockRawIO loader doesn't support multi-device ns6 files. For now, it's best to leave it to the user to load the K files manually then find their own common t-zero among those K files.

### DIFF
--- a/neo/test/rawiotest/test_blackrockrawio.py
+++ b/neo/test/rawiotest/test_blackrockrawio.py
@@ -186,7 +186,7 @@ class TestBlackrockRawIO(
                         assert_equal(python_digievents, matlab_digievents)
 
     def test_blackrockrawio_ptp_timestamps(self):
-        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp/20231027-125608")
+        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp")
         reader = BlackrockRawIO(filename=dirname)
         reader.parse_header()
 
--- a/neo/test/rawiotest/test_blackrockrawio.py
+++ b/neo/test/rawiotest/test_blackrockrawio.py
@@ -28,7 +28,7 @@ class TestBlackrockRawIO(
         "blackrock/FileSpec2.3001",
         "blackrock/blackrock_2_1/l101210-001",
         "blackrock/blackrock_3_0/file_spec_3_0",
-        "blackrock/blackrock_3_0_ptp",
+        "blackrock/blackrock_3_0_ptp/20231027-125608-001",
     ]
 
     @unittest.skipUnless(HAVE_SCIPY, "requires scipy")
@@ -186,7 +186,7 @@ class TestBlackrockRawIO(
                         assert_equal(python_digievents, matlab_digievents)
 
     def test_blackrockrawio_ptp_timestamps(self):
-        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp")
+        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp/20231027-125608-001")
         reader = BlackrockRawIO(filename=dirname)
         reader.parse_header()
 
--- a/neo/test/rawiotest/test_blackrockrawio.py
+++ b/neo/test/rawiotest/test_blackrockrawio.py
@@ -28,7 +28,7 @@ class TestBlackrockRawIO(
         "blackrock/FileSpec2.3001",
         "blackrock/blackrock_2_1/l101210-001",
         "blackrock/blackrock_3_0/file_spec_3_0",
-        "blackrock/blackrock_3_0_ptp/20231027-125608",
+        "blackrock/blackrock_3_0_ptp",
     ]
 
     @unittest.skipUnless(HAVE_SCIPY, "requires scipy")
@@ -186,7 +186,7 @@ class TestBlackrockRawIO(
                         assert_equal(python_digievents, matlab_digievents)
 
     def test_blackrockrawio_ptp_timestamps(self):
-        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp/20231027-125608")
+        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp")
         reader = BlackrockRawIO(filename=dirname)
         reader.parse_header()
 
--- a/neo/test/rawiotest/test_blackrockrawio.py
+++ b/neo/test/rawiotest/test_blackrockrawio.py
@@ -28,6 +28,7 @@ class TestBlackrockRawIO(
         "blackrock/FileSpec2.3001",
         "blackrock/blackrock_2_1/l101210-001",
         "blackrock/blackrock_3_0/file_spec_3_0",
+        "blackrock/blackrock_3_0_ptp/file_spec_3_0_ptp",
     ]
 
     @unittest.skipUnless(HAVE_SCIPY, "requires scipy")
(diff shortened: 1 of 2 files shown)


cboulay:
OK this is ready for review. If you want a sample data file then that will have to wait until Friday when I'm in front of the hardware needed to produce the 3.0-ptp variant file spec.

samuelgarcia:
Hi Chadwick.
An enormous thanks for figthing with this new version of blackrock.
I will have a look soon. Lets wait friday to get the the test file that corresponf to this.
Can you try to make a very very short one ?

cboulay:
I found and fixed a bug in my previous implementation thanks to unit tests I just added.

Here's the test file. If you can provide me instructions on how to created a pull request on GIN then I can go ahead and do that, but I'm also totally satisfied having someone else do it if you're already setup for that.

[blackrock_3_0_ptp.zip](https://github.com/NeuralEnsemble/python-neo/files/13196454/blackrock_3_0_ptp.zip)

samuelgarcia:
I push tests files here https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/121/files

cboulay:
OK I've rebased and fixed the conflicts along the way. Let's hope the tests pass. BTW, I don't mind if you squash the commits when merging... there was some back and forth on quotes that become obsolete when you adopted PEP8.

cboulay:
Actually I went ahead and squashed into 1 commit.

The unit test isn't running because it can't find the file. I guess I'm not providing its path correctly. Can anyone help me out with that part?

neo/test/rawiotest/test_blackrockrawio.py
```diff
@@ -184,6 +185,32 @@ def test_compare_blackrockio_with_matlabloader_v21(self):
                         matlab_digievents = mts_ml[mid_ml == int(label)]
                         assert_equal(python_digievents, matlab_digievents)
 
+    def test_blackrockrawio_ptp_timestamps(self):
+        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp/20231027-125608-001")
```
zm711, line 189:
```suggestion
        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp")
```

It looks to me like Sam's PR of your data removed this part of the file so the folder structure is just without the ending part.

zm711 (commented):
@cboulay see my comment and see if it looks reasonable.

See test structure here:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/blackrock/blackrock_3_0_ptp

cboulay:
I'm pretty lost trying to figure out what's going on the with file loading. I'm happy to yield to someone else.

neo/test/rawiotest/test_blackrockrawio.py
```diff
@@ -184,6 +185,32 @@ def test_compare_blackrockio_with_matlabloader_v21(self):
                         matlab_digievents = mts_ml[mid_ml == int(label)]
                         assert_equal(python_digievents, matlab_digievents)
 
+    def test_blackrockrawio_ptp_timestamps(self):
+        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp/20231027-125608")
```
zm711, line 189:
```suggestion
        dirname = self.get_local_path("blackrock/blackrock_3_0_ptp")
```

zm711: I'll push this change myself and see if it fixes it.

zm711: Actually do you have let maintainers push turned on? I can't edit this file myself.

zm711:
@cboulay,

We are planning on including this in the next release which will go out in the next few days so if you have the time to either implement my commit fix or give us access we will include otherwise we will plan on this for the next release.

cboulay:
I don't see the option to allow edits from maintainers. I found the [GH docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) that explains it but the option isn't on this page.

zm711:
Normally there is a little check box on the right side of the bag just below all the info. When I first looked at your PR you had it turned on but it got switched off at some point. Either way we will see if tests pass now!

zm711:
Seems like it failed. Could you doublecheck the gin link I gave you and make sure it has all the files necessary for blackrock?

cboulay:
Ah, it's because I made the PR from the lab org. [sigh](https://github.com/orgs/community/discussions/5634)

neo/test/rawiotest/test_blackrockrawio.py
```diff
@@ -28,6 +28,7 @@ class TestBlackrockRawIO(
         "blackrock/FileSpec2.3001",
         "blackrock/blackrock_2_1/l101210-001",
         "blackrock/blackrock_3_0/file_spec_3_0",
+        "blackrock/blackrock_3_0_ptp/20231027-125608",
```
zm711, line 31: I think this one also needed to be fixed!

zm711:
So I thin there are two paths that need to be fixed. One at the top and one at the bottom. Since I can't push stuff here do you want a PR on your branch with the two file path fixes I think we need and we can test?

zm711:
That's cool. One last check:

could you try this as the filename? Previously you didn't have the -001 right? so maybe it needs the full file name for the .nev file? 

20231027-125608-001

zm711:
Perfect that worked! thanks for bearing with me! The dirname messed with me! We should really call that filename in the tests for this IO (not your fault, something we've been discussing #1455 recently).

cboulay:
Seems to be working. Thanks for your help!

zm711:
@samuelgarcia tests are passing here, but you also said you would look over the code. Did you want to read this over first before we merge?

samuelgarcia:
Thanks a lot for this new add Chadwick.
And thanks a lot to wait for me.
this looks good to me.

PR #1339, opened 2023-10-20
Switch Neo specific Code to F strings

zm711:
I switched a lot of the `.format` and some of the `%` to f-strings. Neo is 3.8 + right now and f-strings were introduced in 3.6.  They are faster (~ 1.5-2x compared to format) and at least to some, myself included, more readable. 

I didn't touch any of the ios/rawios in this PR.
Take a read and let me know what you think.


samuelgarcia:
Thanks a lot Zach for refactoring this quite old style of writing.

zm711:
De rien.

PR #1340, opened 2023-10-20
Switch `np.math.ceil` to `math.ceil` as the former is deprecated as of NumPy 1.25

zm711:
This is just a minor deprecation fix PR that I saw while running the test suite.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -646,7 +646,7 @@ class BlackrockRawIO(BaseRawIO):
         if t_start is None:
             ind_start = None
         else:
-            ts = np.math.ceil(t_start * self.__nev_basic_header['timestamp_resolution'])
+            ts = int(np.ceil(t_start * self.__nev_basic_header['timestamp_resolution']))
             ind_start = np.searchsorted(timestamp, ts)
 
         if t_stop is None:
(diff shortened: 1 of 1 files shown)


zm711:
@samuelgarcia, I know you worked on this rawio so just FYI-ing you here for this deprecation fix PR.

neo/rawio/blackrockrawio.py
```diff
@@ -646,7 +646,7 @@ def _get_timestamp_slice(self, timestamp, seg_index, t_start, t_stop):
         if t_start is None:
             ind_start = None
         else:
-            ts = np.math.ceil(t_start * self.__nev_basic_header['timestamp_resolution'])
+            ts = int(np.ceil(t_start * self.__nev_basic_header['timestamp_resolution']))
```
apdavison, line 649: The deprecation warning suggests replacing `np.math` with the built-in Python `math` module, so I think this could just be `ts = math.ceil(...)` (with an import of `math` above), since `math.ceil` returns an integer.

apdavison (changes requested):

neo/rawio/blackrockrawio.py
zm711: Done!

apdavison (approved):

PR #1341, opened 2023-10-20
Fix `test_asciisignalio` to work on Windows

zm711:
Since Windows use a different newline in order for the csv to work the `newline` argument must be specified. Adding this doesn't affect Linux testing, but it makes my life better when running local tests on Windows.


apdavison:
For future reference, the reason for this is given in the Python docs here: https://docs.python.org/3/library/csv.html#id3

zm711:
Thanks @apdavison, next time I'll include the link documentation in the PR message itself.

PR #1326, opened 2023-09-06
Add support for Biocam brw v4.x files, closes #1324

mahlzahn:
* ~missing: proper test file~

PR #1348, opened 2023-11-15
Fix 2048 µv offset and signal inversion for BioCAM recordings

### LINKED ISSUE #1347: Filter-dependant voltage shift in BioCAM recordings
**Describe the bug**
Raw traces from BioCAM recording have an unexpected + 2048 applied to the voltage measurements, i.e. traces are all shifted 2048 µV higher than they actually are. This issue disappears when the recording is filtered through SpikeInterface.

**To Reproduce**
Using either version of BRW recordings available on [GIN](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/biocam/) : 
 
```
from pathlib import Path
import spikeinterface as si
import spikeinterface.extractors as se
import spikeinterface.preprocessing as spre

path = Path(r'/home/lumin/Documents/ephy/biocam_hw3.0_fw1.6.brw') # or biocam_hw3.0_fw1.7.0.12_raw.brw

recording = se.read_biocam(path, mea_pitch=81, electrode_width=21)
# recording = spre.bandpass_filter(recording=recording,\
#                                   freq_min=300, freq_max=3000)

traces = recording.get_traces(channel_ids=['35','26','15','23','8'],\
                               end_frame=4096) # Picking a few random channels
plt.plot(traces)
```
neo_raw

```
from pathlib import Path
import spikeinterface as si
import spikeinterface.extractors as se
import spikeinterface.preprocessing as spre

path = Path(r'/home/lumin/Documents/ephy/biocam_hw3.0_fw1.6.brw') # or biocam_hw3.0_fw1.7.0.12_raw.brw

recording = se.read_biocam(path, mea_pitch=81, electrode_width=21)
recording = spre.bandpass_filter(recording=recording,\
                                  freq_min=300, freq_max=3000)

traces = recording.get_traces(channel_ids=['35','26','15','23','8'],\
                               end_frame=4096) # Picking a few random channels
plt.plot(traces)
```
neo_bp

**Expected behaviour**
Raw baseline should be around zero.

**Environment:**
 - OS: Ubuntu
 - Python version : 3.10.10
 - Neo version : 0.13.0.dev0
 - NumPy version : 1.24.2

**Additional context**
I first reported the problem in a [SpikeInterface issue](https://github.com/SpikeInterface/spikeinterface/issues/728) a while ago, and detailed it a bit more in this [recent PR](https://github.com/NeuralEnsemble/python-neo/pull/1326).

Briefly, looking through @mhhennig 's [read function](https://github.com/mhhennig/HS2/blob/43f631652768017323d96e1a89eabfb981a46f90/herdingspikes/probe_functions/readUtils.py#L100-L106) in HerdingSpikes, signal is either read as `data - 2048` or `2048 - data` depending on file version and signal inversion.
There also seem to be some kind of mask to filter out certain values ? Not sure about that : `d[np.abs(d) > 1500] = 0`.

On the other hand, neo reads signal as either `data` or `4096 - data`, depending on signal inversion.
Furthermore, signal inversion itself isn't actually taken into account : 
```
    if format_100:
        if signal_inv == 1:
            read_function = readHDF5t_100
        elif signal_inv == 1:
            read_function = readHDF5t_100_i
        else:
            raise Exception("Unknown signal inversion")
    else:
        if signal_inv == 1:
            read_function = readHDF5t_101
        elif signal_inv == 1:
            read_function = readHDF5t_101_i
        else:
            raise Exception("Unknown signal inversion")
```

By quickly running through random data point in the HDF5 file, it looks to me that the `- 2048` is necessary : 

On `biocam_hw3.0_fw1.6.brw` :
```
>>> f['3BData']['Raw'][0]
2043
>>> f['3BData']['Raw'][758]
2045
>>> f['3BData']['Raw'][92475]
2037
>>> 
```

On `biocam_hw3.0_fw1.7.0.12_raw.brw` :
```
>>> f['Well_A1']['Raw'][0]
2048
>>> f['Well_A1']['Raw'][725]
2046
>>> f['Well_A1']['Raw'][98712]
2050
>>> 
```

BptGrm:
Closes #1347.
Will update with BRW v4.x support once #1326 is approved.

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -416,12 +416,9 @@ class AnalogSignal(BaseSignal):
         '''
         Handle pretty-printing the :class:`AnalogSignal`.
         '''
-        pp.text("{cls} with {channels} channels of length {length}; "
-                "units {units}; datatype {dtype} ".format(cls=self.__class__.__name__,
-                                                          channels=self.shape[1],
-                                                          length=self.shape[0],
-                                                          units=self.units.dimensionality.string,
-                                                          dtype=self.dtype))
+        pp.text(f"{self.__class__.__name__} with {self.shape[1]} channels of length "
+                f"{self.shape[0]}; units {self.units.dimesionality.string}; datatype "
+                f"{self.dtype}")
         if self._has_repr_pretty_attrs_():
             pp.breakable()
             self._repr_pretty_attrs_(pp, cycle)
@@ -431,8 +428,8 @@ class AnalogSignal(BaseSignal):
             with pp.group(indent=1):
                 pp.text(line)
 
-        _pp("sampling rate: {}".format(self.sampling_rate))
-        _pp("time: {} to {}".format(self.t_start, self.t_stop))
+        _pp(f"sampling rate: {self.sampling_rate}")
+        _pp(f"time: {self.t_start} to {self.t_stop}")
 
     def time_index(self, t):
         """Return the array index (or indices) corresponding to the time (or times) `t`"""
@@ -716,7 +713,7 @@ class AnalogSignal(BaseSignal):
             raise MergeError(f'Signals are not continuous. Can not concatenate signals with gaps. '
                              f'Please provide a padding value.')
         if padding is not False:
-            logger.warning('Signals will be padded using {}.'.format(padding))
+            logger.warning(f'Signals will be padded using {padding}.')
             if padding is True:
                 padding = np.nan * units
             if isinstance(padding, pq.Quantity):
--- a/neo/core/baseneo.py
+++ b/neo/core/baseneo.py
@@ -33,8 +33,8 @@ def _check_annotations(value):
     """
     if isinstance(value, np.ndarray):
         if not issubclass(value.dtype.type, ALLOWED_ANNOTATION_TYPES):
-            raise ValueError("Invalid annotation. NumPy arrays with dtype %s"
-                             "are not allowed" % value.dtype.type)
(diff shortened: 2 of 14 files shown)


alejoe91:
@BptGrm before you submit the PR. I discussed with @samuelgarcia about this.

We think that the biocam system has a 12-bit ADC, so that's where the 4096 come from. In general, NEO tends to keep the original data type (at least the RawIO), and use the `gain` and `offset` property to tell how to convert to uV. In this case, the offset should be half of the dynamic range: 2^11 = 2048.

Does it make sense? 
We should deal with the "conversion" later in the pipeline. SpikeInterface has a nice `unsigned_to_signed` function (see [here](https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.preprocessing.unsigned_to_signed)), that we could extend with an optional `bit_depth` (if None, it will take the bit depth from the data type, e.g., int16 -> 16). Does it make sense?

alejoe91:
In other words, the 2048 offset is correct. That is exactly how the data is saved to disk ;)

BptGrm:
Makes sense !

I guess the conversion should happen later on, probably in SpikeInterface's BioCAM extractor ? I'm not that familiar with those.

It would probably be safer to have it on by default with the option to turn it off.

Should this still be mentionned in the comments or the docs of the neo reader, in case it is used outside of SI ?

BptGrm:
I can also open a smaller PR just for handling the signal inversion mistake, that should still be fixed although I don't think it's a problem with current 3Brain recordings.

PR #1350, opened 2023-11-17
Download correct ddl for 64 system on windows with Plexon

h-mayorquin:
Currently plexon download is fixed to use the 32 bits architecture. This means that plexon is not working if the windows machine is 64 bits. This PR fixes this by adding code that checks the architecture and selects the appropriate file to download based on that query.

There is a warning that I think should be removed in the code. Let me know what you think.

[EDIT] it seems that wine on linux uses the 32 bit architecture (I wonder why?) so I added code that restricts the 64 dll only for windows which does make sense anyway.

### DIFF
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -301,7 +301,7 @@ class Plexon2RawIO(BaseRawIO):
         stream_id = self.header['signal_streams'][stream_index]['id']
         stream_characteristic = list(self.signal_stream_characteristics.values())[stream_index]
         assert stream_id == stream_characteristic.id
-        return stream_characteristic.n_samples
+        return int(stream_characteristic.n_samples)  # Avoids returning a numpy.int64 scalar
 
     def _get_signal_t_start(self, block_index, seg_index, stream_index):
         # This returns the t_start of signals as a float value in seconds
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -86,7 +86,7 @@ class Plexon2RawIO(BaseRawIO):
             pl2_dll_folder.mkdir(exist_ok=True)
             pl2_dll_file_path = pl2_dll_folder / file_name
 
-            if not pl2_dll_file_path.exists():
+            if pl2_dll_file_path.exists():
                 # I think this warning should be removed
                 # Warnings should provide a solution but this is just a reminder to the user of normal behavior
                 # Nothing to do
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -77,9 +77,9 @@ class Plexon2RawIO(BaseRawIO):
         # download default PL2 dll once if not yet available
         if pl2_dll_file_path is None:
             architecture = platform.architecture()[0]
-            if architecture == '64bit':
+            if architecture == '64bit' and platform.system() == 'Windows':
                 file_name = "PL2FileReader64.dll"
-            else:
+            else:  # Apparently wine uses the 32 bit version in linux
                 file_name = "PL2FileReader.dll"
             
             pl2_dll_folder = pathlib.Path.home() / '.plexon_dlls_for_neo'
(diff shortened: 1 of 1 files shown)


neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -74,14 +76,24 @@ def __init__(self, filename, pl2_dll_file_path=None):
...
+                file_name = "PL2FileReader.dll"
+            
+            pl2_dll_folder = pathlib.Path.home() / '.plexon_dlls_for_neo'
             pl2_dll_folder.mkdir(exist_ok=True)
-            pl2_dll_file_path = pl2_dll_folder / 'PL2FileReader.dll'
+            pl2_dll_file_path = pl2_dll_folder / file_name
 
-            if pl2_dll_file_path.exists():
-                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
+            if not pl2_dll_file_path.exists():
```
zm711, line 89: I'm confused why did you switch to the opposite logic here?

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -74,14 +76,24 @@ def __init__(self, filename, pl2_dll_file_path=None):
...
+                # I think this warning should be removed
+                # Warnings should provide a solution but this is just a reminder to the user of normal behavior
+                # Nothing to do
+                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
```
zm711, line 93: Super fair point. I don't think this warning is doing anything!

neo/rawio/plexon2rawio/plexon2rawio.py
zm711: Or did you mean to download with urlopen if it doesn't exist and then just delete the else since the warning is likely not needed.

Or maybe I'm just misunderstanding the flow?

zm711 (commented):
One question @h-mayorquin and I support the removal of the warning personally!

neo/rawio/plexon2rawio/plexon2rawio.py
h-mayorquin: No, that's a mistake. Thanks.

I removed the warning beforehand and then re-verted without re-re-verting the logic.

h-mayorquin, line 89:
```suggestion
            if pl2_dll_file_path.exists():
```

zm711: Ah cool cool. That makes sense.

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -288,7 +301,7 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
         stream_id = self.header['signal_streams'][stream_index]['id']
         stream_characteristic = list(self.signal_stream_characteristics.values())[stream_index]
         assert stream_id == stream_characteristic.id
-        return stream_characteristic.n_samples
+        return int(stream_characteristic.n_samples)  # Avoids returning a numpy.int64 scalar
```
h-mayorquin, line 304: This change is related to this:
https://github.com/SpikeInterface/spikeinterface/pull/2223

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -74,14 +76,25 @@ def __init__(self, filename, pl2_dll_file_path=None):
...
+            
+            pl2_dll_folder = pathlib.Path.home() / '.plexon_dlls_for_neo'
             pl2_dll_folder.mkdir(exist_ok=True)
-            pl2_dll_file_path = pl2_dll_folder / 'PL2FileReader.dll'
+            pl2_dll_file_path = pl2_dll_folder / file_name
 
             if pl2_dll_file_path.exists():
-                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
+                # I think this warning should be removed
+                # Warnings should provide a solution but this is
```
zm711, line 91:
```suggestion
                # Warnings should provide a solution but this is
```

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -74,14 +76,25 @@ def __init__(self, filename, pl2_dll_file_path=None):
...
             pl2_dll_folder.mkdir(exist_ok=True)
-            pl2_dll_file_path = pl2_dll_folder / 'PL2FileReader.dll'
+            pl2_dll_file_path = pl2_dll_folder / file_name
 
             if pl2_dll_file_path.exists():
-                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
+                # I think this warning should be removed
+                # Warnings should provide a solution but this is 
+                # just a reminder to the user of normal behavior
+                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
```
zm711, line 93:
```suggestion
                warnings.warn(f'Using cached plexon dll at {pl2_dll_file_path}')
```

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -74,14 +76,25 @@ def __init__(self, filename, pl2_dll_file_path=None):
 
         # download default PL2 dll once if not yet available
         if pl2_dll_file_path is None:
-            pl2_dll_folder = pathlib.Path .home() / '.plexon_dlls_for_neo'
+            architecture = platform.architecture()[0]
+            if architecture == '64bit' and platform.system() == 'Windows':
+                file_name = "PL2FileReader64.dll"
+            else:  # Apparently wine uses the 32 bit version in linux
+                file_name = "PL2FileReader.dll"
+
```
zm711, line 84:
```suggestion

```

zm711 (commented):
This should fix 3/4 of the pep8 issues. The line too long could be fixed, but then it would look super ugly. The L84 issue is the most annoying pep8 check ever. But there were spaces on the blank line, so I think I deleted them all.

PR #1361, opened 2023-12-08
Fix some tests that fail when running locally

apdavison:
In most cases, this is because a dependency isn't installed, in this case the tests are now skipped rather than failing.


zm711:
Thanks @apdavison, this has been such a pain when trying to run tests locally!

alejoe91 (approved):

samuelgarcia (approved):

PR #1362, opened 2023-12-12
Parse spikeglx nidq without gate or trigger number

guptadivyansh:
The 'easy' regex `in parse_spikeglx_fname in spikeglxraio.py` expects the fname to end in .ap or .lf, which doesn't work if the input is actually an NIDaq file from SpikeGLX.

I just added another regex that works for the .nidq files, and some example filenames to comment what the regex targets.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -337,7 +337,7 @@ def parse_spikeglx_fname(fname):
                 run_name, device, stream_kind = r[0]
                 gate_num, trigger_num = None, None
             else:
-                #easy case for nidaq, example: sglx_xxx.nidq
+                # easy case for nidaq, example: sglx_xxx.nidq
                 r = re.findall(r'(\S*)\.(\S*)', fname)
                 run_name, device = r[0]
                 gate_num, trigger_num, stream_kind = None, None, None
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -331,10 +331,16 @@ def parse_spikeglx_fname(fname):
             stream_kind = None
         else:
             # the naming do not correspond lets try something more easy
+            # example: sglx_xxx.imec0.ap
             r = re.findall(r'(\S*)\.(\S*).(ap|lf)', fname)
             if len(r) == 1:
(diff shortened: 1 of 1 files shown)


guptadivyansh:
@alejoe91 Please do have a look, thanks!

alejoe91 (approved):

neo/rawio/spikeglxrawio.py
```diff
@@ -331,10 +331,16 @@ def parse_spikeglx_fname(fname):
...
+            # example: sglx_xxx.imec0.ap
             r = re.findall(r'(\S*)\.(\S*).(ap|lf)', fname)
             if len(r) == 1:
                 run_name, device, stream_kind = r[0]
                 gate_num, trigger_num = None, None
+            else:
+                # easy case for nidaq, example: sglx_xxx.nidq
+                r = re.findall(r'(\S*)\.(\S*)', fname)
+                run_name, device = r[0]
+                gate_num, trigger_num, stream_kind = None, None, None
```
alejoe91, line 343: @guptadivyansh in this case shouldn't `stream_kind` be set to `nidq`? I guess this is the only case when we expect to get to this condition, right?
```suggestion
                stream_kind = "nidq"
                gate_num, trigger_num = None, None
```

guptadivyansh: `stream_kind` was set to None for the normal NI case above (line 331), so I didn't change it. `device` will be set to `nidq` by the regex parser.

alejoe91: @samuelgarcia what do you think?

alejoe91: Actually, this is perfectly in line with the "default" behavior as you suggested. Resolving

alejoe91:
> We should add a test to check this again the file "Noise4Sam_g0" wich has NIDQ

@samuelgarcia actually this only deals with nidq filenames without the gate/trigger pattern. In all our test files, the NIDQ streams are correctly loaded from the first condition ([here](https://github.com/NeuralEnsemble/python-neo/blob/3388557c4841c81638620e30da5e84f6663aed97/neo/rawio/spikeglxrawio.py#L327-L331))

PR #1363, opened 2023-12-12
Gain calculation for Neuropixel 2.0 might be wrong

guptadivyansh:
For commercial 2.0 probes, the manual says the ADC is 12 bits, which would make the highest positive value to be 2048, not 8192, as is hardcoded in the code. In the metadata of my recordings, `imMaxInt=2048`. 

Jennifer Colonnel's code [here](https://github.com/jenniferColonell/SpikeGLX_Datafile_Tools/blob/4b6c28658b715808e4941e43d5fa0c57198b09da/Python/DemoReadSGLXData/readSGLX.py#L78C29-L78C29) uses the imMaxInt to calculate the gain factor.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -409,7 +409,8 @@ def extract_stream_info(meta_file, meta):
             # We allow also LF streams for NP2.0 because CatGT can produce them
             # See: https://github.com/SpikeInterface/spikeinterface/issues/1949
             per_channel_gain[:-1] = 1 / 80.
-            gain_factor = float(meta['imAiRangeMax']) / 8192
+            max_int = int(meta['imMaxInt']) if 'imMaxInt' in meta else 8192
+            gain_factor = float(meta['imAiRangeMax']) / max_int
             channel_gains = gain_factor * per_channel_gain * 1e6
         else:
             raise NotImplementedError('This meta file version of spikeglx'
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -408,7 +408,10 @@ def extract_stream_info(meta_file, meta):
             # https://github.com/billkarsh/SpikeGLX/blob/15ec8898e17829f9f08c226bf04f46281f106e5f/Markdown/Metadata_30.md#imec
             # We allow also LF streams for NP2.0 because CatGT can produce them
             # See: https://github.com/SpikeInterface/spikeinterface/issues/1949
-            per_channel_gain[:-1] = 1 / 80.
+            if 'imChan0apGain' in meta:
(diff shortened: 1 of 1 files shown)


guptadivyansh:
@alejoe91 Please have a look, thanks! :)

neo/rawio/spikeglxrawio.py
```diff
@@ -408,8 +408,12 @@ def extract_stream_info(meta_file, meta):
             # https://github.com/billkarsh/SpikeGLX/blob/15ec8898e17829f9f08c226bf04f46281f106e5f/Markdown/Metadata_30.md#imec
             # We allow also LF streams for NP2.0 because CatGT can produce them
             # See: https://github.com/SpikeInterface/spikeinterface/issues/1949
-            per_channel_gain[:-1] = 1 / 80.
-            gain_factor = float(meta['imAiRangeMax']) / 8192
+            if 'imChan0apGain' in meta:
+                per_channel_gain[:-1] = 1 / float(meta['imChan0apGain'])
```
alejoe91, line 412: For NP2, the LF stream can be produced only by CatGT (see https://github.com/SpikeInterface/spikeinterface/issues/1949), so I believe it's ok to use the AP gain in that case too

alejoe91: since these changes are releated to the SpikeGLX version, should they be propaated to NP1 probes too? @guptadivyansh what do you think?

guptadivyansh:
The per channel gain part is already read from the imro table for 1.0, so that should be correct as is.

I didn't touch the immaxInt (512) for 1.0, because I assumed all 1.0 probe were 10bit and didn't want to introduce changes that might make people's old data inconsistent with new runs of Neo. Although it seems that immaxInt has been in SpikeGLX for a while. So we could also change it for 1.0, for consistency with 2.0 and only use the hardcoded 512 if the immaxInt is not in Meta. 

If 1.0 were all 10bit, and I would rather not change it. But if you think we should, I can update the PR later this week.

neo/rawio/spikeglxrawio.py
guptadivyansh: The per channel gain part is already read from the imro table for 1.0, so that should be correct as is.

I didn't touch the immaxInt (512) for 1.0, because I assumed all 1.0 probe were 10bit and didn't want to introduce changes that might make people's old data inconsistent with new runs of Neo. Although it seems that immaxInt has been in SpikeGLX for a while. So we could also change it for 1.0, for consistency with 2.0 and only use the hardcoded 512 if the immaxInt is not in Meta.

If 1.0 were all 10bit, and I would rather not change it. But if you think we should, I can update the PR later this week.

alejoe91: Makes sense. I think that all NP1.0 are 10bit so no need to update ;)

alejoe91 (approved):

PR #1368, opened 2024-01-09
Increase line limit for pep8 speaks

zm711:
Since there is no consensus on adding linting yet could we at least switch pep8speaks to something more reasonable. For me I prefer anything > 110, but I know @alejoe91 likes 120, so that is what I switched this to. Would you at least be okay with that @samuelgarcia ?


alejoe91:
@zm711 

@apdavison should have disabled PEP8 speaks. We decided to do a round of black formatting and to have maybe an action that reformats master every week or so. How does that sound?

zm711:
Thanks goodness. I'll close :)

PR #1372, opened 2024-01-23
Plexon2: Fix conversion between C++ tm structure and python datetime

### LINKED ISSUE #1371: Plexon .pl2 files fail to load if it is recorded in January; files from other months have month created off-by-one
**Describe the bug**
Plexon .pl2 files fail to load if they are recorded in January. File recorded in other months will have the month created be stored off-by-one.

.pl2 files store the date created according to the tm structure found in time.h: https://cplusplus.com/reference/ctime/tm/

The month in the tm struct is encoded as 0-11, but the Python DateTime object expects the month to be from 1-12. A ValueError exception is thrown when trying to load a .pl2 file recorded in January.

**To Reproduce**
To reproduce, try to load a .pl2 file recorded in January.
[JanuaryRecording.zip](https://github.com/NeuralEnsemble/python-neo/files/13992167/JanuaryRecording.zip)

**Environment:**
 - OS: Windows 11
 - Python version 3.9.5
 - Neo version 0.13.0.dev0
 - NumPy version 1.22.1

zm711:
Fixes #1371.

As explained in the issue the structures are indexed different so we need to +1 to correctly index. Add comment to explain this with links.

### DIFF
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -223,7 +223,11 @@ class Plexon2RawIO(BaseRawIO):
                 # invalid datetime information if year is <1
                 if tmo.tm_year != 0:
                     microseconds = block_info['m_CreatorDateTimeMilliseconds'] * 1000
-                    dt = datetime(year=tmo.tm_year, month=tmo.tm_mon, day=tmo.tm_mday, hour=tmo.tm_hour,
+                    # tm_mon range is 0..11 https://cplusplus.com/reference/ctime/tm/
+                    # python is 1..12 https://docs.python.org/3/library/datetime.html#datetime.datetime
+                    # so month needs to be tm_mon+1; also tm_sec could cause problems in the case of leap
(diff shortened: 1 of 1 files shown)


zm711:
@h-mayorquin, I know you read over and help with the `dll` for plexon2. Do you want to double check this fix?

alejoe91 (approved):

h-mayorquin (commented):
I think that being off by a second with the leap second is fine. Glad that @alejoe91 had the time to check it.

zm711:
@h-mayorquin, my point for leap second is that based on my reading Python datetime will not accept the leap second value

tm can return 60 or 61 to count as leap seconds, but datatime second only accepts 0..59. So in the case that someone records on a leap second then this IO could fail to annotate the date.

h-mayorquin:
What I mean is to just substract 1 and make it 60 in that case. The other solution would be to propagate all the way up to year in th worse case scenario which is indded difficult as I think you were pointing out.

But also, having the current code just failn and then have someone coming here to tell us so we can device a better solutio nwith an example in our hands seems like a good option : )

PR #1384, opened 2024-01-29
Updated "generate_diagram.py" script

apdavison:
based on changes by Julia Sprenger in #1026


zm711:
I'm still figuring out where all the figures live in the repo, but this figure also is outdated no (since it has groups containing multiple segments?)
image

apdavison:
> this figure also is outdated no (since it has groups containing multiple segments?)

no. There is no problem for a group to contain objects that belong to different segments, in fact that's one of the main use cases for Group.

(Groups can no longer contain Segment objects, but that is not shown in the figure)

zm711:
Cool. So it would be incorrect if the the orange box was over one or both black boxes. I see. Thanks @apdavison !

samuelgarcia:
Merci Andrew and Julia for this Heoric patch on this very old and dirty diagram generation code.

samuelgarcia (approved):

PR #1370, opened 2024-01-18
[IO] added support for new pattern of datetime when working with Neuralynx

Antho2422:
… data in version 2.2.2 of Pegasus. This is fixing the error when trying to parse the date time with the wrong format.

PR #1385, opened 2024-01-29
Handle Pegasus 2.1.1 Nev files and tests of same.

PeterNSteinmetz:
This is an older item that was due and is in preparation for dealing with issue #1370. 
Tests use the Pegasus 2.1.1 data already in the ephy_testing_data repository.


alejoe91 (approved):

samuelgarcia:
Merci Peter!

PR #1389, opened 2024-01-31
Better error message upon failure in loading `OpenEphysBinary` folder

alejoe91:
Something I've been wanting to do for a long time :)


neo/rawio/openephysbinaryrawio.py
```diff
@@ -481,105 +481,108 @@ def explore_folder(dirname, experiment_names=None):
...
 
-                    recording['streams']['events'][stream_name] = event_stream
+                        recording['streams']['events'][stream_name] = event_stream
 
-            folder_structure[node_name]['experiments'][experiment_id]['recordings'][recording_id] \
-                = recording
+                folder_structure[node_name]['experiments'][experiment_id]['recordings'][recording_id] \
+                    = recording
+    except Exception as e:
+        raise Exception(f"Error while exploring Open Ephys folder {dirname}. Invalid folder structure?")
```
zm711, line 585:
```suggestion
        raise Exception(f"Error {e} while exploring Open Ephys folder {dirname}. Invalid folder structure.")
```

Do we want to actually share the e for the user so they can see what exactly triggered the exception or do you think that is too much info? Also you mean `.` or `?` at the end.

zm711 (commented):
It's a big diff, but really you just wrapped with a try-except right?

neo/rawio/openephysbinaryrawio.py
zm711: True. The issue tracker often involves these folder issues. :)

zm711 (dismissed):
Looks good to me.

samuelgarcia:
I think I am clearly not agree with this.
This big try except is not good design.
We need to put try except on very targeted places that focus on a clear failure that could happen.

alejoe91:
Better than before at least. I agree that long term we need a better handling, but I'm not sure when I'll have time for it

zm711:
I also agree more targeted would be better, but at least what I see from the tracker the usual flow is this.

I have an error with openephysbinary (gives error)
That's probably related to folder structure could you list the files.

With this error you tell them that it is likely the folder structure to begin with. Maybe a comment to say in the future improve to more specific/more helpful errors (ie this file was not found--I don't know openephy very well so I'm not sure what the structure would really need to be).

samuelgarcia:
At least lets do the try in the other part:
```python
try:
    explore_folder(dirname, experiment_names=None)
except:
   ...
```

No ?

PR #1396, opened 2024-02-03
Convert codebase over to f-strings

### LINKED ISSUE #857: Consistent string formatting

zm711:
Fixes #857.

I think I got most of the codebase over to f-strings. There are a couple places that I didn't change (docstrings using %/format, one instance of a hex format that I didn't want to touch, and a join that I didn't touch).

Also found a format bug that I fixed when I changed to f-strings.


.gitignore
```diff
@@ -56,6 +56,7 @@ cover
 .directory
 .gdb_history
 .DS_Store?
```
zm711, line 58: Not sure if this one with a question mark is needed, but based on the fact I accidentally committed a few .DS_STORE without the question mark this needs to be added.

zm711:
As I thought 3.8 with the f-strings and dictionaries was not super-robust. So I rewrote those as well. I forget if it was 3.9 or 3.10 where they added in more robust f-string handling of dictionary indexing.

I also had an if-else in an f-string, but that didn't parse correctly so I had to revert that.

Accidentally uploaded some ds_store files, so deleted those and added it to the .gitignore for the future. 

And then fixed a typo.

samuelgarcia:
Merci Zach for this enormous and tedious task.
For the DSSTORE add and remove maybe we could merge and squash this time to avoid having then in history  no ?

neo/core/epoch.py
```diff
@@ -227,10 +227,9 @@ def __repr__(self):
         """
 
         objs = [
-            "%s@%s for %s" % (label, str(time), str(dur))
-            for label, time, dur in zip(self.labels, self.times, self.durations)
+            f"{label}@{str(time)} for {str(dur)}" for label, time, dur in zip(self.labels, self.times, self.durations)
         ]
-        return "<Epoch: %s>" % ", ".join(objs)
+        return f"<Epoch: %s>" % ", ".join(objs)
```
zm711, line 232:
```suggestion
        return "<Epoch: %s>" % ", ".join(objs)
```

leaving this one as % because I don't want to mess up this join dynamic

zm711:
I think so. That would probably be nice. I made one last change and then this is ready to merge on my end. I tested import locally on 3.11 and it also seemed to be fine. I was mostly worried about 3.8 and 3.9 struggling with some of the more complex f strings, but they are passing tests. 

I'm sure my hours of tedium will save some future person 200 ms, so it was all worth it in the end.

PR #1399, opened 2024-02-11
Check that keys of nested dictionaries in annotations are `str`

### LINKED ISSUE #1195: `check_annotations` doesn't check dict key type
Not that I'm in favor of these inherently flawed checks (e.g., I could make a perfectly valid dict that doesn't inherit from `dict` and Neo would reject it) but `check_annotations` doesn't check the dict key type:

```
(b4py9) robin@LAPTOP-JF6T3PSU:~/git/cerebellum$ python
Python 3.9.9 (main, Jan 12 2022, 15:37:14)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from neo import io, Block
>>> writer = io.NixIO("ello.nix", mode="ow")
>>> n = Block(my_meta={(lambda a: None): 5})
>>> n.annotations
{'my_meta': {<function <lambda> at 0x7f04967c8c10>: 5}}
>>> writer.write(n)
  File "<stdin>", line 1, in <module>
  File "/home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/neo/io/baseio.py", line 142, in write
    self.write_block(bl, **kargs)
  File "/home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/neo/io/nixio.py", line 641, in write_block
    self._write_property(metadata, k, v)
  File "/home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/neo/io/nixio.py", line 1284, in _write_property
    section.create_property(name, values)
  File "/home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/nixio/section.py", line 158, in create_property
    dtype = DataType.get_dtype(single_val)
  File "/home/robin/.pyenv/versions/b4py9/lib/python3.9/site-packages/nixio/datatype.py", line 47, in get_dtype
    raise ValueError("Unknown type for value {}".format(value))
ValueError: Unknown type for value <function <lambda> at 0x7f04967c8c10>
```

Using `neo==0.11.1`

zm711:
Fixes #1195.

Just adding a check for the key. Example from issue:
```
my_meta={(lambda a: None): 5}
```
Because the keys aren't checked this doesn't fail until writing. We should check to make sure the key is one of the [allowable annotations](https://github.com/NeuralEnsemble/python-neo/blob/c96ec936049a125d6bedfa26b67514f927bb1ab3/neo/core/baseneo.py#L14-L29).

We still have problems with writing some of the allowable annotations (see [this issue](1198)), but adding this check at least moves in the right direction.

### DIFF
--- a/neo/core/baseneo.py
+++ b/neo/core/baseneo.py
@@ -45,7 +45,8 @@ def _check_annotations(value):
         if not issubclass(value.dtype.type, ALLOWED_ANNOTATION_TYPES):
             raise ValueError(f"Invalid annotation. NumPy arrays with dtype {value.dtype.type}" f"are not allowed")
     elif isinstance(value, dict):
-        for element in value.values():
+        for key, element in value.items():
+            _check_annotations(key)
             _check_annotations(element)
     elif isinstance(value, (list, tuple)):
         for element in value:
--- a/neo/core/baseneo.py
+++ b/neo/core/baseneo.py
@@ -46,7 +46,8 @@ def _check_annotations(value):
             raise ValueError(f"Invalid annotation. NumPy arrays with dtype {value.dtype.type}" f"are not allowed")
     elif isinstance(value, dict):
         for key, element in value.items():
-            _check_annotations(key)
+            if not isinstance(key, str):
+                raise TypeError(f"Annotations keys must be strings not of type {type(key)}")
             _check_annotations(element)
     elif isinstance(value, (list, tuple)):
         for element in value:


apdavison:
I think we'd always assumed that annotation keys would be strings, since they represent names. I don't think anyone would be surprised if they got an error when using a lambda function as the name for an annotation.

If we're going to check this, then only `str` should be allowed for annotation keys and sub-keys, with a `TypeError` raised otherwise.

zm711:
That makes things easier, so I'm happy to switch that :)

PR #1400, opened 2024-02-12
Skip plexon 2 tests when wine is not present

h-mayorquin:
This is generating error when the tests are attempeded to run without wine.  I added an informative message. Plus, I also added the vscode files to the gitignore.

### DIFF
--- a/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
+++ b/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
@@ -9,14 +9,20 @@
 # copyright notice is kept intact.
 
 from sys import platform
-import os
+import subprocess
 import pathlib
 import warnings
 
 if any(platform.startswith(name) for name in ("linux", "darwin", "freebsd")):
+    try:
+        result_wine = subprocess.run(["dpkg", "-l", "wine"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError:
+        raise ImportError("Wine is not installed. Please install wine to use the PL2FileReader.dll")
+
     from zugbruecke import CtypesSession
 
     ctypes = CtypesSession(log_level=100)
+
 
 elif platform.startswith("win"):
     import ctypes
--- a/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
+++ b/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
@@ -19,13 +19,14 @@ plataform_is_windows = platform.system() == "Windows"
 if plataform_is_windows:
     import ctypes
 else:
-
-    try:
-        is_wine_available = subprocess.run(
-            ["wine", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False
-        )
-    except subprocess.CalledProcessError:
-        raise ImportError("Wine is not installed. Please install wine to use the PL2FileReader.dll")
+    pltaform_is_linux = platform.system() == "Linux"
+    if pltaform_is_linux:
+        try:
+            is_wine_available = subprocess.run(
+                ["wine", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False
+            )
+        except subprocess.CalledProcessError:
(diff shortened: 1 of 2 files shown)


zm711:
Downloading plexon dll to /Users/zacharymckenzie/.plexon_dlls_for_neo/PL2FileReader.dll
  Cell In[2], line 1
    test = Plexon2RawIO('/Users/zacharymckenzie/Downloads/4chDemoPL2.pl2')
  File ~/Documents/GitHub/python-neo/neo/rawio/plexon2rawio/plexon2rawio.py:104 in _init
    from neo.rawio.plexon2rawio.pypl2.pypl2lib import PyPL2FileReader
  File ~/Documents/GitHub/python-neo/neo/rawio/plexon2rawio/pypl2/pypl2lib.py:18
    result_wine = subprocess.run(["dpkg", "-l", "wine"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
  File ~/miniconda3/envs/spykes/lib/python3.11/subprocess.py:548 in run
    with Popen(*popenargs, **kwargs) as process:
  File ~/miniconda3/envs/spykes/lib/python3.11/subprocess.py:1026 in _init
    self._execute_child(args, executable, preexec_fn, close_fds,
  File ~/miniconda3/envs/spykes/lib/python3.11/subprocess.py:1950 in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'dpkg'
__

h-mayorquin:
Yes, I realize that also the spikeinterface code does not work for Mac. I think this should work but I don't have a mac so I can't test.

zm711:
@h-mayorquin, you need to switch your import to `import platform`. The sys.platform isn't what you want. But when I did the switch locally with this PR, it still gives a file not found.

```

  File ~/miniconda3/envs/spykes/lib/python3.11/subprocess.py:1950 in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)

FileNotFoundError: [Errno 2] No such file or directory: 'wine'
```

h-mayorquin:
Damn, well, I will restrict the wine check to linux. If someone with mac can expand it that would be great but I can't test that without a mac really.

zm711:
If you test this on linux and it works for you then I'm fine to merge this. I can confirm it worked on Mac:

```
   raise ImportError("Wine is not installed. Please install wine to use the PL2FileReader.dll")

ImportError: Wine is not installed. Please install wine to use the PL2FileReader.dll
```

h-mayorquin:
This works for me.

zm711:
Sorry one last question @h-mayorquin. You deleted the final else "system not supported". I agree that is probably okay not to have, but what was your logic for that just so we can record that.

neo/rawio/plexon2rawio/pypl2/pypl2lib.py
```diff
@@ -8,22 +8,26 @@
...
-    from zugbruecke import CtypesSession
+platform_is_windows = platform.system() == "Windows"
 
-    ctypes = CtypesSession(log_level=100)
-
-elif platform.startswith("win"):
+if platform_is_windows:
     import ctypes
 else:
-    raise SystemError("unsupported platform")
```
zm711, line 24: Could this ever happen? Are there other non-posix systems that we should put as an else?

h-mayorquin: I thought that it was unecessary but it does not hurt. Do you want me to add just in case?

zm711: Might as well just in case. I'm sure someone is using something crazy and just itching to try and fail to read .pl2 files.

h-mayorquin: Yeah, now that I think about it I don't think it fits the current branch structure. You want the second branch for something else. I rather have someone come with that problem and then we fix it.

zm711: That's fair. I mean they will still get an error. They just won't get your nice error message.

zm711 (approved):

PR #1402, opened 2024-02-13
Allow IntanRawIO & IntanIO to handle the headerless binary files

zm711:
Replaces #1318. 

After plenty of great discussion with @h-mayorquin I have changed my mind and think it might be better to just incorporate the new file types in the IntanIO family rather than make a new io. 

TODO:
- [x] Improve comments
- [x] Make sure tests pass

Also would fix (https://github.com/SpikeInterface/spikeinterface/issues/194)


zm711:
@h-mayorquin,

Feel free to give this a read. I think the diff will be easier to tell what was from the original implementation and what is new. Then if you want to improve something that was older we can do that too.

zm711:
Notes based on discussion with Heberto.
chan_id as native channel name vs chan_name as custom channel name
timestamp assert messaging improvement

neo/rawio/intanrawio.py
```diff
@@ -53,40 +56,90 @@ class IntanRawIO(BaseRawIO):
...
 
     def _parse_header(self):
 
+        filename = Path(self.filename)
+        
+        if not filename.exists() or not filename.is_file():
+            raise FileNotFoundError(f"{filename} does not exist")
+
         if self.filename.endswith(".rhs"):
+            self.file_type = 'header-attached'
```
h-mayorquin, line 78: What do you think of calling this format instead of file type?

neo/rawio/intanrawio.py
```diff
@@ -53,40 +56,90 @@ class IntanRawIO(BaseRawIO):
...
             self._global_info, self._ordered_channels, data_dtype, header_size, self._block_size = read_rhs(
                 self.filename
             )
+        # 3 possibilities for rhd files two use a plain header called 'info.rhd' and one format has the header
```
h-mayorquin, line 82:
```suggestion
        # 3 possibilities for rhd files, one combines the header and the data in the same file with suffix `rhd` while the other two separates the data from the header which is always called `info.rhd`
```

What do you think?

neo/rawio/intanrawio.py
```diff
@@ -53,40 +56,90 @@ class IntanRawIO(BaseRawIO):
...
+
             self._global_info, self._ordered_channels, data_dtype, header_size, self._block_size = read_rhd(
-                self.filename
+                self.filename, self.file_type
             )
 
         # memmap raw data with the complicated structured dtype
-        self._raw_data = np.memmap(self.filename, dtype=data_dtype, mode="r", offset=header_size)
+        # if header-attached there is one giant memory-map
+        if self.file_type == "header-attached":
```
h-mayorquin, line 104: What do you think of making `_raw_data` a dict so we always index by the available stream_index instead of neding the stream_index_key

neo/rawio/intanrawio.py
```diff
@@ -53,40 +56,90 @@ class IntanRawIO(BaseRawIO):
...
+            for stream_index, (stream_index_key, sub_datatype) in enumerate(data_dtype.items()):
+                # for 'one-file-per-signal' we have one memory map / neo stream
+                if self.file_type == "one-file-per-signal":
+                    self._raw_data[stream_index] = np.memmap(
+                        raw_file_paths_dict[stream_index_key], dtype=sub_datatype, mode="r"
+                    )
+                # for one-file-per-channel we have one memory map / channel stored as a list / neo stream
+                else:
+                    self._raw_data[stream_index] = []
+                    for channel_index, datatype in enumerate(sub_datatype):
```
h-mayorquin, line 117: This `datatype` is dangerously closed to `data_type`, maybe:
* data_type
* stream_data_type
* channel_data_type

h-mayorquin (commented):
Hey Zach, three small comments that are suggestions. Feel free to ignore them. 

More importantly, I am getting errors when I try to run some of my files. I will investigate this further and come back to you. Hopefully soon.

zm711:
Definitely let me know what the issue is. Do you know what the file structure is? Is there a particular signal missing? Maybe I can make a simulated dataset that is similar so I can test on my end too?

neo/rawio/intanrawio.py
zm711: Done.

neo/rawio/intanrawio.py
zm711: Done. Great idea to be clearer!

neo/rawio/intanrawio.py
zm711: I'm curious how you see this? I can see how a dict would be good for the file per channel, but how would I make the 'header-attached' be a dict? 

The issue is the user api must take an int. So we would have to incorporate a translation layer. I would have to think about this. If you want to build something off of this you could open a PR on this PR with your idea. Otherwise I'll need to think about this some more.

neo/rawio/intanrawio.py
```diff
@@ -496,39 +603,54 @@ def read_rhd(filename):
...
         chan_info["sampling_rate"] = sr
         chan_info["units"] = "uV"
         chan_info["gain"] = 0.195
         chan_info["offset"] = -32768 * 0.195
         ordered_channels.append(chan_info)
-        data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        if file_format == "header-attached":
+            data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        else:
+            data_dtype[0] += [(name, "uint16", BLOCK_SIZE)]
```
h-mayorquin, line 627: Reading the format again I think the probelm with my data is here. Two points:

1) The first is minor: for the other two formats, [the specification](https://intantech.com/files/Intan_RHD2000_data_file_formats.pdf) mentions that this is int16. Different from the header-attached case where the dtype is uint16. Can you double check that?
2) More importantly. I see no mention in the specification that the data is organized on blocks for the other two formats that we are adding here. For the data that I have, it does not seem to be the case. This is important, because Sam implementation of memmaps assumes that the total file size will be a multiple of the datatypes and if the datatypes includes the block size then this is unlikely to match. If I remove this assumption locally I can succesfully read my files.

That said, I am still curios on why it does work with your files. Is it just a coincidence? I feel that there is something that I am missing. On the other hand, at least for the `one-file-per-channel` case, the structured dtype with block_size is totally unecessary so I don't think there is a point in having it.

I think that following the structure dtype of the header-attached case will pose problems down the line.

zm711: For 1) You are right, but its not minor. I re-looked at the data and because it is signed it doesn't have an offset so not only does it need to be changed for the dtype it needs to have a different offset/gain setting. THANKS for pointing this out that is huge. 

For 2) You might be right. Let me read more about this tomorrow. I can try to make some more changes and then we can re-test. I'll push some changes for the weekend and then maybe next week we can strategize again.

neo/rawio/intanrawio.py
```diff
@@ -114,8 +179,23 @@ def _parse_header(self):
...
+                    for raw_data in self._raw_data.values()
+                ]
+            )
+        else:
+             self._max_sigs_length = max(
+                [
+                    len(raw_data) * raw_data[0].size
+                    for raw_data in self._raw_data.values()
+                ]
+            )
```
zm711, line 198: I was thinking about this? Maybe just the raw_data[0].size? since each is an individual file?

neo/rawio/intanrawio.py
```diff
@@ -148,7 +228,12 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
         signal_channels = self.header["signal_channels"][mask]
         channel_names = signal_channels["name"]
         chan_name0 = channel_names[0]
-        size = self._raw_data[chan_name0].size
+        if self.file_format == "header-attached":
+            size = self._raw_data[chan_name0].size
+        elif self.file_format == 'one-file-per-signal':
+            size = self._raw_data[stream_index][:,0].size
```
zm711, line 234: Same here, I was trying to think about whether this was the correct way. I think this one is okay, but want to read this carefully.

zm711 (commented):
@h-mayorquin also a couple comments on things to read carefully, because I myself and trying to make sure we are grabbing the correct dimension of the data.

neo/rawio/intanrawio.py
h-mayorquin: Do we need this? Can you just calculate the `get_signal_size` over all the stream indices either at the end of parsing or within `_segment_t_stop`?

neo/rawio/intanrawio.py
h-mayorquin: This function is mean to calculate the number of samples, right? For this file_format case you already loaded the memmap with a specific shape:

https://github.com/zm711/python-neo/blob/839060e5b910d878021348de34f50e7bf1ce343e/neo/rawio/intanrawio.py#L112-L120

So you know what the number of samples is. This will avoid you accessing the data of the memmap as you are currently doing which I think is bad because it loads  data into memory.

neo/rawio/intanrawio.py
```diff
@@ -496,39 +619,66 @@ def read_rhd(filename):
...
+        else:
+            chan_info["offset"] = 0.0
         ordered_channels.append(chan_info)
-        data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        if file_format == "header-attached":
+            data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        elif file_format == 'one-file-per-signal':
+            data_dtype[0] = "int16"
+        else:
+            data_dtype[0] += [(name, "int16")]
```
h-mayorquin, line 650: I don't think you need the structured data type for the "one-file-per-channel" format either. 

Getting rid of this will allow you to ommit the overspecification of the channel in the `get_analog_signal_chunk` function:

https://github.com/zm711/python-neo/blob/839060e5b910d878021348de34f50e7bf1ce343e/neo/rawio/intanrawio.py#L290-L294

There, you first access the channel data with the index (that is, you know the channel already) but then, because is a structured datatype you need to index again with the channel name.

h-mayorquin (commented):
I tested the current version on my data and it works. This is great.

neo/rawio/intanrawio.py
zm711: This is based on Sam's implementation. It occurs here so I kept it here for everything.

neo/rawio/intanrawio.py
h-mayorquin: Yeah, in the case of the `header_attached` it does not need to be a dict. Pretty much as we have right now.

The other suggestion would be to use the stream_names as key as conditions like this would be more readable:

```
https://github.com/zm711/python-neo/blob/839060e5b910d878021348de34f50e7bf1ce343e/neo/rawio/intanrawio.py#L114-L118
```

Right now, someone reading the specification would have to remember to which streams these two numbers corresponds two but they would not need two if we were to iterate over the stream_names.

That said, I think this is of little marginal gain and I can prototype in a future PR once we are done with this.

neo/rawio/intanrawio.py
h-mayorquin: Ok, I think you took care of this. Feel free to mark it as resolved.

neo/rawio/intanrawio.py
```diff
@@ -53,40 +57,101 @@ class IntanRawIO(BaseRawIO):
...
+        else:
+            self._raw_data = {}
+            for stream_index, (stream_index_key, stream_datatype) in enumerate(data_dtype.items()):
+                # for 'one-file-per-signal' we have one memory map / neo stream
+                if self.file_format == "one-file-per-signal":
+                    n_chans = channel_number_dict[stream_index_key]
+                    if stream_index_key == 4 or stream_index_key == 5:
+                        n_samples = int(os.path.getsize(raw_file_paths_dict[stream_index_key]) / 2) # uint16 2 bytes
+                    else:
+                        n_samples = int(os.path.getsize(raw_file_paths_dict[stream_index_key]) / (n_chans * 2))# unit16 2 bytes
```
h-mayorquin, line 117: Just for your info.

Pathlib objects have a method called stat that allows you to access the size directly without needing to call the `os` module:

https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat

So you could do something like this:
```
    file_path = raw_file_paths_dict[stream_index_key]
    size_in_bytes = file_path.stat().st_size
    size_in_bytes_of_dtype = np.dtype(stream_datatype).itemsize
    n_samples = stream_factor * size_in_bytes // size_in_bytes_of_dtype
```

neo/rawio/intanrawio.py
h-mayorquin: Yeah, I think he was just pre-computing this value to use it when `_segment_t_stop` is called. But we can discuss that in another PR.

neo/rawio/intanrawio.py
zm711: I knew pathlib had one. I just knew the os right away and was too lazy. Thanks though :) I can change that. I like pathlib better in general.

neo/rawio/intanrawio.py
zm711: True. You're right. I just really like naming memmaps. For no reason though. It's a strong point. I'll fix it.

neo/rawio/intanrawio.py
zm711: Great point. I wasn't even thinking about that.

zm711:
@h-mayorquin any chance you would run a cProfile on your data? I'm curious about performance with the many memmaps being created and then maybe reading one chunk of data....

h-mayorquin:
Yep, let me do some profiling.

neo/rawio/intanrawio.py
zm711: That sounds good. After this is merged then you can make some improvements!

neo/rawio/intanrawio.py
zm711: Made this change.

neo/rawio/intanrawio.py
```diff
@@ -167,7 +247,15 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
...
         channel_ids = signal_channels["id"]
         channel_id_0 = channel_ids[0]
+        if self.file_format == "header-attached":
+            size = self._raw_data[channel_id_0].size
+        elif self.file_format == 'one-file-per-signal':
+            size = self._raw_data[stream_index][:,0].size
+        else:
+            size = self._raw_data[stream_index][0].size
+
         size = self._raw_data[channel_id_0].size
```
h-mayorquin, line 257: I think this line got added in the merge but should not be here, right? line 257

zm711: Thanks. Fixed it. Happened when I was fixing the merge conflicts. So I think for the #1416 I'll wait until after this is merged.

neo/rawio/intanrawio.py
h-mayorquin: I mean just getting the first element of shape should do the deed, right?

neo/rawio/intanrawio.py
```diff
@@ -167,7 +247,15 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
         signal_channels = self.header["signal_channels"][mask]
         channel_ids = signal_channels["id"]
         channel_id_0 = channel_ids[0]
+        if self.file_format == "header-attached":
+            size = self._raw_data[channel_id_0].size
+        elif self.file_format == 'one-file-per-signal':
+            size = self._raw_data[stream_index][:,0].size
+        else:
+            size = self._raw_data[stream_index][0].size
```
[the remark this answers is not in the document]
zm711: raw_data[stream_index] is a list for `one-file-per-channel`.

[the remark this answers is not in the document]
h-mayorquin: Yeah, but the memory access is happening on the `one-file-per-signal`

neo/rawio/intanrawio.py
h-mayorquin: You got to master git, man, don't let git merge dictacte your workflow.

[JK, I can take care of the other one after this if you prefer : ) ]

zm711: You can host a git workshop for me. :)

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
zm711: The one file per signal was in L253. I changed that in my most recent commit. L255 is a list so it doesn't have a shape.

neo/rawio/intanrawio.py
zm711: Fixed. Now that I understand what you mean :)

neo/rawio/intanrawio.py
```diff
@@ -183,28 +285,57 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
+                    data_chan = self._raw_data[stream_index][:, channel_indexes[channel_index]]
+            else:
+                if channel_indexes_are_none:
+                    data_chan = self._raw_data[stream_index][channel_index]
+                else:
+                    data_chan = self._raw_data[stream_index][channel_indexes[channel_index]]
+
+            # if shape is 1 it is either temp a 1d or a 1d array from one-file-per-stream
+            # or one-file-per-channel. In which case we do not want to do the block math
+            if len(shape) == 1 or self.file_format != 'header-attached':
```
zm711, line 338: @h-mayorquin I just realized we (but really my fault since I wrote this :( ) had another tiny bug that may mess things up. We shouldn't get the signal chunk based on the block_start and block_stop because for the non-header-attached formats we don't need to do this . We can just grab the starting index and the stopping index.

zm711: And now I realize this comment is cutting off my point haha. If you look from line 336-345ish you'll see what I mean.

neo/rawio/intanrawio.py
```diff
@@ -72,34 +76,104 @@ class IntanRawIO(BaseRawIO):
...
+                if self.file_format == "one-file-per-signal":
+                    n_chans = channel_number_dict[stream_index_key]
+                    if stream_index_key == 4 or stream_index_key == 5:
+                        n_samples = int(os.path.getsize(raw_file_paths_dict[stream_index_key]) / 2) # uint16 2 bytes
+                    else:
+                        n_samples = int(os.path.getsize(raw_file_paths_dict[stream_index_key]) / (n_chans * 2))# unit16 2 bytes
+                    if stream_index_key != 6:
+                        self._raw_data[stream_index] = np.memmap(
+                            raw_file_paths_dict[stream_index_key], dtype=stream_datatype, shape = (n_chans, n_samples),  mode="r"
+                        ).T
```
zm711, line 140: @samuelgarcia,

neo is samples x channels  so I added the transpose. But only your final read through just double check me? The Intan documentation says that map should be made with n_chans x n_samples

neo/rawio/intanrawio.py
h-mayorquin: Yep. That makes sense.

neo/rawio/intanrawio.py
```diff
@@ -586,17 +766,81 @@ def read_rhd(filename):
...
 
-    if bool(global_info["notch_filter_mode"]) and version >= V("3.0"):
-        global_info["notch_filter_applied"] = True
+    if global_info["notch_filter_mode"] == 2 and version >= V("3.0"):
+        global_info["notch_filter"] = '60Hz'
+    elif global_info["notch_filter_mode"] == 1 and version >= V("3.0"):
+        global_info["notch_filter"] = '50Hz'
     else:
-        global_info["notch_filter_applied"] = False
+        global_info["notch_filter"] = False
```
zm711, line 791: Originally I added this code in because Intan was just returning the mode which isn't helpful. While I was at it I figured it would be even more helpful to say which notch filter was applied. But we could change this if it is unclear. I just wanted a clean way to represent that since neo was returning 0, 1, or 2 and so someone would need to go to the Intan documentation to know what these meant. Not relevant for this PR, but I want to make this fix now.

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
h-mayorquin: OK, so I see you changed. Can you mark this as resolved? I am unable to for some reason.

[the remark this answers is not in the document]
zm711: yes. will do. Sorry the fact that it was pointing at 253, but the commit suggestion was for 271 confused me originally so I fixed it myself.

neo/rawio/intanrawio.py
```diff
@@ -183,28 +285,57 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
+            # this is the general case with 2D (ie in the header-attached format)
             block_size = shape[1]
             block_start = i_start // block_size
             block_stop = i_stop // block_size + 1
 
             sl0 = i_start % block_size
             sl1 = sl0 + (i_stop - i_start)
 
-        sigs_chunk = np.zeros((i_stop - i_start, len(channel_ids)), dtype="uint16")
+        if self.file_format != 'header-attached' and stream_id == 'RHD2000 amplifier channel':
```
h-mayorquin, line 317: I think we really need something like what is on:

https://github.com/NeuralEnsemble/python-neo/pull/1416

The time for examples is int32 which is not included here.

neo/rawio/intanrawio.py
zm711: I was trying to decide if I wanted to do this now or later. Maybe resolved for now and we can think about improving later?

neo/rawio/intanrawio.py
zm711: I did this patch before you submitted your pr. Yours in general is better. No one should request the time. So really the only option was `uint16` but since amplifier is `int16` for the other formats I did mine. I agree yours is the better final solution.

zm711: Especially since it future proofs in case intan changes other dtypes in the future.

neo/rawio/intanrawio.py
h-mayorquin: Yeah, we can do it later if you are not fully convinced : )

zm711: I'm fully convinced I'm just lazy about this one....

neo/rawio/intanrawio.py
h-mayorquin: How comes the time is not requested? : O

neo/rawio/intanrawio.py
h-mayorquin: Oh, no, does this mean that we are supposed to wait for @samuelgarcia ? : P

Well, call me back in a couple of months and I will do the profile then : )

neo/rawio/intanrawio.py
h-mayorquin: I can do it after.

neo/rawio/intanrawio.py
zm711: Most people just want the data and convert to times themselves by dividing by sample_rate. Also Intan timestamps aren't great if I remember (but I never look at them, except once when they weren't useful for me).

 It's just tricky because I have to do the two different patches so your PR fixes the main case and then I back fix and I clearly get cocky and make mistakes rather than testing locally. I've got the Fridays.... Sorry, but you're right. We should be ready in the case that someone requests time so that we don't give the wrong numbers.

neo/rawio/intanrawio.py
zm711: Haha. After our extensive testings I think he can do a lighter read. We've worked out most of the kinks and tests are passing, so we are in a good place.

neo/rawio/intanrawio.py
zm711: Actually I don't believe time is accessible. For something to be accessible for Intan it has to be stored in the ordered_channels. Which if we look in the `read_rhd` and the `read_rhs` functions we add timestamp to the dtype but not to ordered_channels. This was the same problem that was happening with digital data previously before I patched it. Once could reach into the memmap and grab the data, but one couldn't use the machinery.

neo/rawio/intanrawio.py
```diff
@@ -516,19 +661,35 @@ def read_rhd(filename):
...
+        else:
+            data_dtype[6] = [("timestamp", "int32",)]
+            channel_number_dict[6] = 1
     else:
-        data_dtype = [("timestamp", "uint32", BLOCK_SIZE)]
+        if file_format == "header-attached":
+            data_dtype = [("timestamp", "uint32", BLOCK_SIZE)]
+        else:
+            data_dtype[6] = [("timestamp", "uint32",)]
+            channel_number_dict[6] = 1
```
zm711, line 674: If we look here @h-mayorquin you can see that we add timestamp to our dtype for memmapping, but we don't add it to `ordered_channels`. So timestamps can't be accessible.

neo/rawio/intanrawio.py
h-mayorquin: Yes, it is not accessible as of now as you say. I still think that it should be done. But this is also problematic for other datatypes anyway.

neo/rawio/intanrawio.py
h-mayorquin: Reading this function again I would suggest completly separating the workflow for `_get_analogsignal_chunk` for the three formats:

1) The header-attached version needs to block logic because the data is interleaved but is not necessary for the other formats.
2) The one-file-per-signal version does not need the loop over the channels as all the channels are in the memmap already and I think we can optimize access by using this layout.
3) The one-file-per-channel version requires iterating over channels for accessing the different memmaps but does not require block logic.

I personally would create three private methods:
1) `_get_analogsignal_chunk_header_attached`, `_get_analogsignal_chunk_one_file_per_signal`, `_get_analogsignal_chunk_one_file_per_channel` and just call them after the channel_index logic depending on the format.

You can also just three big if if you don't like adding methods. 

But I think this would be better for debugging, optimizing and profiling. I think that the fact that the other two modes do not require the block logic escape both of us because the current implementation is a little bit too complicated with so many branches. 

What do you think?

zm711: I agree with that. Let's merge #1416 and then I'll draft that change while incorporating the typecasting changes at the same time.

h-mayorquin: Do you want me to do this? I have been making a bunch of request. I can open a PR with this as a base and you can review it and merge to this branch if you aprove.

h-mayorquin: Yes, just push your latest changes if you have anything to the remote of this branch and I will take care of it. I will also propagate the merge of the dtype stuff so you don't have to take care of it.

zm711: I have no remote you're good to go. And #1416 is merged.

zm711: Thanks Heberto!

h-mayorquin:
> > Why did you decided to go back to structured dtypes for the one-file-per-channel format in this commit? : O
> 
> I didn't. It's because we iterate through each channel above so I need this in a giant list so that I can iterate through the list. It is only put in as the actual dtype not a structured.

Ah, OK, since you added the dictionary of number of channels (great idea btw!) we don't need this. See:
https://github.com/zm711/python-neo/pull/3

I also did the pathlib discussed above.

neo/rawio/intanrawio.py
h-mayorquin: OK, now in a PR to this branch:
https://github.com/zm711/python-neo/pull/3

zm711:
I see what you mean. That was great. Thanks really cleans things up.

zm711:
@h-mayorquin, Sam is busy for the next few weeks so I doubt he will get to this in that time frame, but are you good with this just so I can ping him to let him know this can be reviewed whenever.

h-mayorquin:
Yeah, I was aware that he has his hands busy with the updates in Spikeinterface and some events. 

This is ready on my side. Let's ping him on two weeks (I will put in my calendar) as I would not like to stress him : )

zm711:
Sounds good. Thanks for all the great feedback and additions during the process. It's been great :)

ChauhanT:
Sorry to be bouncing from one pull to another. This is in response to @zm711 [here](https://github.com/NeuralEnsemble/python-neo/pull/1318#issuecomment-1979333240). 

Sorry, I forgot this was a neo pull and not a spike-interface pull !

**What are the second to last elements?**

The second to last elements in the header are usually the lower-limit of the dtype of the channel. This is very useful for low-level operations. In this pull request it is set to 0. for all channels. Here's a snippet from a 64-channel probe I use, first using the proposed code :  

```
 ('A-063', 'A-063', 25000.    , 'uint16', 'uV', 1.950e-01,   0.  , '0')
 ('A-AUX1', 'A-AUX1',  6250.    , 'uint16', 'V', 3.740e-05,   0.  , '1')
 ('A-AUX2', 'A-AUX2',  6250.    , 'uint16', 'V', 3.740e-05,   0.  , '1')
 ('A-AUX3', 'A-AUX3',  6250.    , 'uint16', 'V', 3.740e-05,   0.  , '1')
 ('A-VDD1', 'A-VDD1',   195.3125, 'uint16', 'V', 7.480e-05,   0.  , '2')
 ('ANALOG-IN-1', 'ANALOG-IN-1', 25000.    , 'uint16', 'V', 3.125e-04, -10.24, '3')
 ('DIGITAL-IN-01', 'DIGITAL-IN', 25000.    , 'uint16', 'TTL', 1.000e+00,   0.  , '4')]
```

Next, using the current code: 
```
('A-061', '61', 25000.    , 'uint16', 'uV', 1.950e-01, -6389.76, '0')
 ('A-062', '62', 25000.    , 'uint16', 'uV', 1.950e-01, -6389.76, '0')
 ('A-063', '63', 25000.    , 'uint16', 'uV', 1.950e-01, -6389.76, '0')
 ('A-AUX1', '64',  6250.    , 'uint16', 'V', 3.740e-05,     0.  , '1')
 ('A-AUX2', '65',  6250.    , 'uint16', 'V', 3.740e-05,     0.  , '1')
 ('A-AUX3', '66',  6250.    , 'uint16', 'V', 3.740e-05,     0.  , '1')
 ('A-VDD1', '67',   195.3125, 'uint16', 'V', 7.480e-05,     0.  , '2')
 ('ANALOG-IN-1', '68', 25000.    , 'uint16', 'V', 3.125e-04,   -10.24, '3')
 ('DIGITAL-IN', '69', 25000.    , 'uint16', 'TTL', 1.000e+00,     0.  , '4')]
```

**Channel IDs**

I think addressing legacy code is extremely important. Maybe providing very verbose error messages which notify the user of the issue, and/or repeatedly listing the changes in release changelogs for at least 2 releases.

h-mayorquin:
The channel ids issue should be discussed outside of here. It was changed in https://github.com/NeuralEnsemble/python-neo/pull/1415 so it will need to be reverted outside of this. 

My initial intuition is that downstream code should not depend on channel_ids. I would like to know use cases for that so we can be more mindful about this in the future if they exists though.

ChauhanT:
Sorry, I also have to add - this is awesome work.  I/O off of proprietary hardware is probably one of the most paint-staking things to code. Thank you very much for taking this on.

zm711:
@ChauhanT, the minimum value for an unsigned integer is 0? I'm not sure what the problem is for the dtype question.

As far as the channels this may be a situation where a bug has become a feature. It was not the right choice originally to do it the way we did, but I do understand the concern. I personally use channel_indexes and for most people they could use channel_indexes in place of what they did for chan_ids in the past, but with ints instead of strings. But if this is a huge concern please open an issue @ChauhanT and we three can discuss there and revert if we absolutely need to in another PR.

ChauhanT:
@zm711  Ah, there might be an issue there. IIrc the dtype for analogue channels is int16 and not unsigned. At least for the one-file-per-channel format.

h-mayorquin:
I still don't understand your proposal above. Is that the header for the signals, the dtype for the signals is here:

https://github.com/NeuralEnsemble/python-neo/blob/6cd8158d84a004217a9318c5aecc3e21df539ee0/neo/rawio/baserawio.py#L94-L103

The second to last element is the offset.

neo/rawio/intanrawio.py
```diff
@@ -519,19 +679,35 @@ def read_rhd(filename):
...
+            chan_info["offset"] = -32768 * 0.195
+        else:
+            chan_info["offset"] = 0.0
         ordered_channels.append(chan_info)
-        data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        
+        if file_format == "header-attached":
+            data_dtype += [(name, "uint16", BLOCK_SIZE)]
+        else:
+            data_dtype[0] = "int16"
```
h-mayorquin, line 709: @ChauhanT is this what you mean?

zm711:
It's not a minimum value it's an offset. From Intan's documentation there is no offset....
image

neo/rawio/intanrawio.py
ChauhanT: Sorry, my bad. Two things I had not checked : 
1. The interface proposed by neo. I thought every reader was doing its own thing.
2. I hadn't realised it's an offset as @zm711 has pointed out below.

ChauhanT:
@zm711 Gotcha. Thanks for explaining this. And what about the dtype of the amp data ? I just downloaded the manual. Afaict they use int16 for the one-file-per-channel format and uint for the one-file-per-recording formats. Do you reckon that might be an issue ? Without an offset, you'd probably have to use the sign of the signal.

neo/rawio/intanrawio.py
h-mayorquin: Ah, OK. Yes, if you look at the bottom of the headers that you printed above you can see that they have dtypes. Those dtypes indicate what every column means. 

Also, taking this asside. in a spikeinterface object you can get the header like this:

```
recording_extractor.neo_reader.header
```

All the best,

h-mayorquin: Ah, this code that I am referencing here addressess your last question btw:

https://github.com/NeuralEnsemble/python-neo/pull/1402#issuecomment-1979427022

Header-attached is what you call "one-file-per-recording" (which is also a good name btw!). So here, we have a different dtype depending on the format.

ChauhanT: Many thanks ! Really appreciate the help from both.

ChauhanT: Thanks again @h-mayorquin . I think the datatype is definitely read in correctly, because spikeinterface has the correct dtype. I'll try to work out why the header still reports it as uint.

neo/rawio/intanrawio.py
```diff
@@ -72,42 +76,106 @@ class IntanRawIO(BaseRawIO):
...
 
         # signals
         signal_channels = []
         for c, chan_info in enumerate(self._ordered_channels):
             name = chan_info["custom_channel_name"]
             channel_id = chan_info["native_channel_name"]
-            if chan_info["signal_type"] == 20:
-                # exception for temperature
+            if chan_info["signal_type"] == 20 or (self.file_format != "header-attached" and chan_info["signal_type"] == 0):
+                # exception for temperature and for amplifier in non-header attached files
```
zm711, line 178: @ChauhanT 

Good catch for the header info. We hadn't updated this yet for our updates (@h-mayorquin). So you were right that the info was being displayed wrong, but it wasn't hurting anything. You can test this new commit to see if it fixes the header display, but when I tested it locally it worked. :)

zm711: @h-mayorquin if we decide to allow timestamps in the future, then we will need to fix this solution more robustly then what I just did. (So if you have a brilliant idea for a better fix, just let me know).

h-mayorquin: Add the dtype to `chan_info` in the `read_rhd` function and then use it here:

https://github.com/zm711/python-neo/blob/4283378cadc734b844cbc757d65e39c58d1a93d1/neo/rawio/intanrawio.py#L786-L791

zm711: smart. I did it for rhs too, because this is a common function for both formats.

h-mayorquin:
pinging @samuelgarcia on this is on my to-do.

samuelgarcia:
Salut Zach.
I am really sorry for taking so long to read tis pull request.
It is super well written and easy (and so quick) to read.
For me this OK to merge.
Merci infiniment! Big thanks to Ramon that follow this very closly.

samuelgarcia (approved):

zm711:
Salut Sam! Thanks for reading it over!

samuelgarcia:
I think we can merge. @alejoe91  : do you want to have look ?

alejoe91:
Ok for me!

Issue #1352, opened 2023-11-23
Add new header pattern to loading of Neuralynx files


Antho2422:
Hello all, 

I am using spikeinterface and Neo 0.12.0. I get an error concerning the time informations in the header due to, I guess, new patterns not supported yet by Neo. One is from new recordings and one from very old recording. 

1. First case:
I get an error on the following line :
```python
 info = NlxHeader(filename)
``` 
Error trace:
```python
NlxHeader.__init__(self, filename)
    246 else:
    247     dt1 = sr.groupdict()
--> 248     self['recording_opened'] = datetime.datetime.strptime(
    249         dt1['date'] + ' ' + dt1['time'], hpd['datetimeformat'])
    251 # close time, if available
    252 if 'datetime2_regex' in hpd:

File c:\Users\agnes.kandracs\AppData\Local\anaconda3\envs\si_env\lib\_strptime.py:568, in _strptime_datetime(cls, data_string, format)
    565 def _strptime_datetime(cls, data_string, format="%a %b %d %H:%M:%S %Y"):
    566     """Return a class cls instance based on the input string and the
    567     format string."""
--> 568     tt, fraction, gmtoff_fraction = _strptime(data_string, format)
    569     tzname, gmtoff = tt[-2:]
    570     args = tt[:6] + (fraction,)

File c:\Users\agnes.kandracs\AppData\Local\anaconda3\envs\si_env\lib\_strptime.py:352, in _strptime(data_string, format)
    349     raise ValueError("time data %r does not match format %r" %
    350                      (data_string, format))
    351 if len(data_string) != found.end():
--> 352     raise ValueError("unconverted data remains: %s" %
    353                       data_string[found.end():])
    355 iso_year = year = None
    356 month = day = 1
 ValueError: unconverted data remains: .000
``` 

And here is the associated header:
```
######## Neuralynx Data File Header 
-FileType NCS 
-FileVersion 3.4 
-FileUUID 189b12c7-ac30-4a8e-9e03-ecd6bfead8b7 
-SessionUUID 2c4921e7-e248-409b-ae37-b98584ea243b 
-ProbeName  
-OriginalFileName "/network/lustre/iss02/epimicro/patients/raw/pat_03249_1711/eeg/03249_2023-01-04_13-50/03249_2023-01-04_13-50_AmT2_1.ncs" 
-TimeCreated 2023/01/04 13:50:16.000 
-TimeClosed 2023/01/04 15:50:15.999 
 
[...]

PR #1408, opened 2024-02-16
Fix up changes from antho2422 Pegasus support pull request

PeterNSteinmetz:
This incorporates the changes from PR #1370 but fixes them to work with current testing data. 

The issue here was that the Pegasus v2.2.2 uses one datetime format in the opening time entry and another in the closing entry. This PR also includes support for that and should ensure that this is used properly when Pegasus v2.2.2 files are tested.


zm711:
Fixes #1352.

samuelgarcia (approved):

PR #1415, opened 2024-02-23
IntanRawIO: Change channel id to native-channel-name in intan

h-mayorquin:
From the discussion in #1402.

I think this will be distracting over there and be solved as an small PR here.

From a discussion with Sam, the purpose of an id is to be unique whereas name is meant to be user determined or more descriptive. 

Right now, we are using an index as channel id. The present PR uses the `native_channel_name` which accomplishes the design purpose of channel id while at the same time is more meaningful.

The relevant section in the specification:

image

@zm711 I just realized this as I was working on the PR.
The memmap field should be named as the channel ids as they are unique. Right now we are using names which is wrong as field names in structured dtypes have to be unique.


zm711:
@h-mayorquin , I already made that change in the other PR. Upon your recommendation. :)

zm711:
I see what you changed though!

h-mayorquin:
@zm711 Yeah, I am aware, I just think that it will simplify the PR if this small issues is addressed separatley. Plus, I think we should change the way we are naming the memmap fields.

zm711:
I agree with the memmap field name. You are right. The names can change the ids can't so we should do the ids to be safe.

neo/rawio/intanrawio.py
```diff
@@ -104,8 +104,8 @@ def _parse_header(self):
         # signals
         signal_channels = []
         for c, chan_info in enumerate(self._ordered_channels):
-            name = chan_info["native_channel_name"]
-            chan_id = str(c)  # the chan_id have no meaning in intan
+            name = chan_info["custom_channel_name"]
+            chan_id = chan_info["native_channel_name"]
```
zm711, line 108: Should we call this channel_id so that we have the same name here as later in the code so we don't call this chan_id here and channel_id later in the code?

h-mayorquin: I would prefer this.

zm711 (approved):

PR #1416, opened 2024-02-23
IntanRawIO: type casting bug in `get_analogsignal_chunk`

h-mayorquin:
Currently, when calling `get_analog_signal_chunk` with Intan a numpy array is created an array with dtype `uint16` is created. However, some of the streams do not have that dtype and then an implicit casting that I think is buggy occurs.

This PR instead determines the dtype of the output numpy array by accessing the dtype from the corresponding memmap region.

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -187,7 +187,8 @@ class IntanRawIO(BaseRawIO):
             channel_indexes = slice(None)
         channel_ids = signal_channels["id"][channel_indexes]
 
-        shape = self._raw_data[channel_ids[0]].shape
+        channel_id_0 = channel_ids[0]
+        shape = self._raw_data[channel_id_0].shape
 
         # some channel (temperature) have 1D field so shape 1D
         # because 1 sample per block
@@ -200,7 +201,9 @@ class IntanRawIO(BaseRawIO):
             sl0 = i_start % block_size
             sl1 = sl0 + (i_stop - i_start)
 
-        sigs_chunk = np.zeros((i_stop - i_start, len(channel_ids)), dtype="uint16")
+        # Create an array to store the requested chunk
+        dtype = self._raw_data[channel_id_0].dtype
+        sigs_chunk = np.zeros((i_stop - i_start, len(channel_ids)), dtype=dtype)
         for channel_index, channel_id in enumerate(channel_ids):
             # Memmap fields are the channel_ids for unique channels
             data_chan = self._raw_data[channel_id]


zm711:
I think it surprises everyone including Sam himself that the function is actually called `get_analogsignal_chunk` and not `get_analog_signal_chunk`. I'll reread this carefully on Monday and we can think about merging it and then propagating the fixes to #1402.

neo/rawio/intanrawio.py
```diff
@@ -200,7 +201,9 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
             sl0 = i_start % block_size
             sl1 = sl0 + (i_stop - i_start)
 
-        sigs_chunk = np.zeros((i_stop - i_start, len(channel_ids)), dtype="uint16")
```
zm711, line 203: So the original idea for this was that timestamps aren't not accessible so the only originally real data were all `uint16` (except temperature which I think was just ignored). So it is technically faster to just do it this way. I guess the question becomes do we want to allow access to timestamps? If not then this PR is really just in prep for #1402 once we have to worry about different dtypes for the amplifier data.

zm711 (commented):
Philosophically for this PR we are trying to future proof to make sure we have the correct dtype. My original solution in #1402 was to just extend to the current space with the assumption that someone in the future would have to modify it more if new dtypes were added. I think this is probably the better solution though overall. What do you think about the timestamps @h-mayorquin?

neo/rawio/intanrawio.py
h-mayorquin: As you mention, this is fixing the dtype for the temperature as well. So, it is fixing a present bug. 

If you mean computationally faster, I don't think there is any meaningful difference between getting the dtype from the data and assume uint16.

Concerning the timestamps. I don't see why they should not be made available as a separate stream. But I think that should be done in a different PR.

h-mayorquin:
> I think it surprises everyone including Sam himself that the function is actually called `get_analogsignal_chunk` and not `get_analog_signal_chunk`. I'll reread this carefully on Monday and we can think about merging it and then propagating the fixes to #1402.

Given that these are a private methods, maybe we should change those ones? 

I guess the private one could be really slowly deprecated if you guys wanted it.

zm711:
@h-mayorquin,

Actually all rawios need a private _get_analogsignal_chunk. The BaseRawIO has the public get_analogsignal_chunk and it calls the private _get_analogsignal_chunk so that each rawio uses the same base. So I don't think a change would be made.

neo/rawio/intanrawio.py
zm711: I agree. It's a theoretical speedup rather than a practical/meaningful speedup.

h-mayorquin:
> @h-mayorquin,
> 
> Actually all rawios need a private _get_analogsignal_chunk. The BaseRawIO has the public get_analogsignal_chunk and it calls the private _get_analogsignal_chunk so that each rawio uses the same base. So I don't think a change would be made.

Yeah, I mean that you can change:

https://github.com/zm711/python-neo/blob/6f49801ec8b839ed7c5e8a44668bb3a0671f4b33/neo/rawio/baserawio.py#L1280-L1298

Without affecting users (maybe would be the time to use a proper abstract method as well?).  It is the public use the one that should not be changed.

h-mayorquin:
> But in that case we would need to change the private method in each rawio. It's not hard work but it's just a tedious change.

What IDE do yo use? It would take me less than five minutes on mine (vscode).

h-mayorquin:
That said, such a small inconsistency is not really problematic in my view.

zm711:
I'm going to merge this, but just to be clear for reference, this is a current bug because temperature would be in the wrong format. It is also an enhancement, because it allows us to add access to timestamps with the correct format (in a future PR) as well as in preparation for #1402 which will have the amplifier data in a different dtypes.

PR #1418, opened 2024-02-27
Fix Neuralynx gaps detection

samuelgarcia:
@PeterNSteinmetz :

I am fixing some gaps detection problem I have for some files. The actual is detecting too much gaps.

This PR:
  * Put back numpy approach for gap detections speedup x50
  * relax the fact that prediction timestamp must exactly the same as following block. I guess that the internal clock is not precise and the actual gap detector creat too much gaps

@yger : can you check this ?
@Antho2422 : can you check this on data by trying gap_detection_strict=False/True  and tell me how many segments you have when reading your data ?

### DIFF
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -87,7 +87,7 @@ class NeuralynxRawIO(BaseRawIO):
         Otherwise set 0 of time to first time in dataset
     gap_detection_strict: bool, default: False
         If False then the gap detection with timestamp of data block is relaxed.
-        It True (old behavior), gap depection need extact timestamp in data block otherwise gaps (and so segment)
+        If True (old behavior), gap detection requires exact timestamps within the data block otherwise gaps (and so segments)
         are created.
     Notes
     -----
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -85,7 +85,10 @@ class NeuralynxRawIO(BaseRawIO):
     keep_original_times: bool, default: False
         If True, keep original start time as in files,
         Otherwise set 0 of time to first time in dataset
-
+    gap_detection_strict: bool, default: False
+        If False then the gap detection with timestamp of data block is relaxed.
+        It True (old behavior), gap depection need extact timestamp in data block otherwise gaps (and so segment)
+        are created.
     Notes
     -----
     * This IO supports NCS, NEV, NSE and NTT file formats (not NVT or NRD yet)
@@ -125,7 +128,7 @@ class NeuralynxRawIO(BaseRawIO):
         ("samples", "int16", (NcsSection._RECORD_SIZE)),
     ]
 
-    def __init__(self, dirname="", filename="", exclude_filename=None, keep_original_times=False, **kargs):
+    def __init__(self, dirname="", filename="", exclude_filename=None, keep_original_times=False, gap_detection_strict=False, **kargs):
 
         if dirname != "":
             self.dirname = dirname
@@ -137,6 +140,7 @@ class NeuralynxRawIO(BaseRawIO):
             raise ValueError("One of dirname or filename must be provided.")
 
         self.keep_original_times = keep_original_times
+        self.gap_detection_strict = gap_detection_strict
         self.exclude_filename = exclude_filename
         BaseRawIO.__init__(self, **kargs)
 
@@ -788,9 +792,8 @@ class NeuralynxRawIO(BaseRawIO):
             data = self._get_file_map(ncs_filename)
             nlxHeader = NlxHeader(ncs_filename)
 
-            verify_sec_struct = NcsSectionsFactory._verifySectionsStructure
-            if not chanSectMap or (not verify_sec_struct(data, chan_ncs_sections)):
-                chan_ncs_sections = NcsSectionsFactory.build_for_ncs_file(data, nlxHeader)
+            if not chanSectMap or (not NcsSectionsFactory._verifySectionsStructure(data, chan_ncs_sections)):
+                chan_ncs_sections = NcsSectionsFactory.build_for_ncs_file(data, nlxHeader, gap_detection_strict=self.gap_detection_strict)
 
             # register file section structure for all contained channels
             for chan_uid in zip(nlxHeader["channel_names"], np.asarray(nlxHeader["channel_ids"], dtype=str)):
--- a/neo/rawio/neuralynxrawio/ncssections.py
+++ b/neo/rawio/neuralynxrawio/ncssections.py
@@ -136,7 +136,7 @@ class NcsSectionsFactory:
         return round(startTime + NcsSectionsFactory.get_micros_per_samp_for_freq(sampFr) * posn)
 
     @staticmethod
-    def _parseGivenActualFrequency(ncsMemMap, ncsSects, chanNum, reqFreq, blkOnePredTime):
+    def _parseGivenActualFrequency(ncsMemMap, ncsSects, chanNum, reqFreq, blkOnePredTime, gap_detection_strict=False):
         """
         Parse sections in memory mapped file when microsPerSampUsed and sampFreqUsed are known,
         filling in an NcsSections object.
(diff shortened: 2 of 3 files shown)


zm711:
Sam I just fixed your comment to actually tag Anthony so he can test :)

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -85,7 +85,10 @@ class NeuralynxRawIO(BaseRawIO):
     keep_original_times: bool, default: False
         If True, keep original start time as in files,
         Otherwise set 0 of time to first time in dataset
-
+    gap_detection_strict: bool, default: False
+        If False then the gap detection with timestamp of data block is relaxed.
+        It True (old behavior), gap depection need extact timestamp in data block otherwise gaps (and so segment)
```
zm711, line 90:
```suggestion
        It True (old behavior), gap detection requires exact timestamps within the data block otherwise gaps (and so segments)
```

samuelgarcia:
@zm711 : merci

PeterNSteinmetz:
@samuelgarcia I have had a look at this and believe there are several issues to think about. I will comment on them in several comments here. 

First is the question of these files with the gaps. If these are the same as provided by @Antho2422 or from the same system, I have looked closely at them. The gaps in these files are somewhat peculiar. The normal step between the timestamps of sequential CscRecords in the file is 125,000 microseconds, but occasionally it is instead something like 38 or 22 microseconds longer. Previously with .ncs files this occurred due to misconfiguration of the hardware. Thus I suspect this may be unique to this system rather than an intended feature of hardware operation. 

This then raises the question of how best to deal with it? My view would be that the best solution is to fix the hardware or its configuration. However, if we want to provide a software method of dealing with this, I think this raises a question of at what level do we want to be dealing with this? It is a question of the philosophy of the library I suppose. I tend to view covering over a hardware or configuration problem as something that might be best dealt with at a higher than RawIO level. Essentially what one wants to do is string together the sections that are present in the file. So would some sort of either separate adaptor or stringing together at the Neo level be more appropriate? 

I would also suggest that the option be called something like "ignoreGaps" and that the default be False, so that users are made aware of the issue before deciding to ignore it.

PeterNSteinmetz:
@samuelgarcia In terms of compensating for the gaps. Normally how the file is parsed for gaps is determined by the header and file type. This is critical given the variations which have been present over time. The method of parsing is chosen in the NcsSections.build_for_ncs_file method. This chooses presently amongst the possible _buildForMaxGap, _buildGivenActualFrequency methods. 

It is unclear to me why _buildForMaxGap with appropriate parameters would not work here for these files. Thus I would suggest that rather than changing the lowest level code, that the higher level build_for_ncs_file method be changed to accept the option and make appropriate choices of the lower level method to call.

PeterNSteinmetz:
@samuelgarcia In terms of the change to NlxHeader in commit 6ba4a90be4421d1d8db3abc50cd29e5210d4f631, this is quite similar to the change just pulled for NlxHeader. I would suggest testing with the new pulled change and without this change. I am starting to restructure the header parsing so that may have already taken care of this.

Antho2422:
@samuelgarcia @PeterNSteinmetz @zm711 

Thank you. 
I will give a try using my data and `gap_detection_strict=False/True`. The gaps are in fact an issue that we need to deal with. Unfortunately I would not be able to fix it on the hardware level so if we find a solution at the soft level I would be more than happy.

PeterNSteinmetz:
@Antho2422 I think it can be fixed by providing some method of stitching together the various sections. I believe you could do that presently with your own code. Just concatenate all the samples for each file into one array. 

The trouble I suspect is with the configuration of the hardware, not the hardware itself. This has happened before. I would suggest talking to your hardware support person or the vendor.

Antho2422:
@PeterNSteinmetz Okay I will tell them about that thank you.

@samuelgarcia it seems to work fine.

I get `nb_segment = 1` when putting `gap_detection_strict=False` and I get `nb_segment = 10652` when `gap_detection_strict=True` . 

Is that all you wanted me to test ?

samuelgarcia:
Hi @PeterNSteinmetz

I was not aware of this `_buildForMaxGap()` I will have a look maybe this is the place I should have made the patch.
In the data I have clearly the timestamp is not precise and so we have a distribution around theexpected next timestamp. So we can put some euristic on the limit.
I do not know the system at all. My guess is that the timestamp is done by a secondary clock (some hib of data) and so this is not precise at all. So we need to relax to avoid creating too much gaps.

Normaly this branch is build on top of all your PR because I merged then before.


@Antho2422 : exactly, it was the feedback I wanted. Cool.

PeterNSteinmetz:
@samuelgarcia I don't think you will need to change _buildForMaxGap, rather just to call it from the _build_for_ncs_file with appropriate parameters based on the option. 

Can you send me one of the files you are looking at? What is happening in the ones from @Antho2422 is that about every 8 gaps are 22 - 38 microseconds too long. In the older Neuralynx Cheetah hardware this had to do with incorrect configuration of the analog to digital convertor and its buffering. Given that this is still a Neuralynx product, I would expect a similar cause.

samuelgarcia:
The data come from "atlas" so this should be `_parseGivenActualFrequency()`
I read the code again `_parseForMaxGap()` and `_parseGivenActualFrequency()` are pretty similar in logic except that the delta used to be strict in `_parseGivenActualFrequency()` and have tolerance in `_parseGivenActualFrequency()`.

`_parseForMaxGap()` was done with numpy vector in mind.
 `_parseGivenActualFrequency()`  was done in loop (in this PR I switch to numpy). Maybe the 2 functions could merged in one function with diffrents option no ?

I am not in lab where this data was produced so I cannot change the setting but I need to fix the reader logic in neo to be able to analyzer then.

I think here `ignoreGaps` would be different, the option I introduced is more very small gaps are ignored.
When the gaps duration is less than 2 data timestamp then it is not a gap it is a problem of none accurate clock tagging data paquet which is also a problem!

I will try to improve this PR.

What do you think rewrite `_parseForMaxGap()` and  `_parseGivenActualFrequency()` in more somilar way ?

PeterNSteinmetz:
@samuelgarcia I think perhaps this is best broken up? I just created a new pull request #1241 for the speed up. That tests fine for me. It looks like the NlxHeader change was somehow already included in master?

PeterNSteinmetz:
@samuelgarcia I can't say that I care for further pushing the distinction between exact matching and allowing gaps down in the call tree. That strikes me as going the wrong direction. 

It looks like to allow a gap of a fixed amount that _parseForMaxGap() should work. It is just a question of what gap or tolerance should be allowed and provided in the argument to that function. The decision about how large a gap to allow is presently made in _buildForMaxGap and uses a default of 0.2 sample deltas. I would suggest changing this method to have a parameter for the tolerance, rather than just the constant. And then passing the desired value from the code in _build_for_ncs_file when it is an ATLAS acqType and tolerating the gap is desired. This should not be needed for BML acqType. Perhaps the desired gap tolerance could be a parameter to this method call, and if none, it would not allow gap tolerance, which is the more normal case. 

Perhaps the desired gap tolerance can be the parameter to the NeuralynxRawIO constructor with a default value of None and the comments on that parameter noting a suggested value if the user runs into troubles with a large number of segments? 

In addition the NcsSections factory could have the desired gap tolerance as a setting in its constructor and then be built by the NeuralynxRawIO, though this is a somewhat larger refactorization.

PeterNSteinmetz:
@samuelgarcia I just added a PR #1422 which parameterizes the gap tolerance as I suggested above. If this looks like it will work for your files, then perhaps this can be incorporated into a new parameter for the NeuralynxRawIO constructor. 

This keeps the decision making about what sort of parsing and gap tolerance to use in one method.

PeterNSteinmetz:
@samuelgarcia The 17 sample (I assume not microsecond gap) you mention in the other PR is definitely quite a bit larger than others I have observed. I agree it must be due to some change in how the timestamp collection is occurring. I can easily see this happening if, as you say, they are using some other system based clock and then time-stamping the records when they come out of a queue. 

I have just added PR 1424 to add a module comment for NcsSections.py which discusses in detail all the various things which may lead to apparent time gaps. I would suggest pulling that as well prior to further work. 

As noted in that new comment, functionally there are two types of gaps. Those due to some hardware, software, or file format issue and the larger ones created by the operator. I tend to think the job of the RawIO level is to represent the latter larger ones as Neo Segments. The job of NcsSections is to provide the information needed to do that job. 

Thus I agree with your idea of having one appropriately parameterized _parse method. Then the decision on what those parameters should be can be made in the build_for_ncs_file method which takes the memmap and NlxHeader to produce the NcsSections. It could also take other parameters about desired tolerances to decide on the parameters; or perhaps better given that a factory will likely be used with the same desired tolerances for multiple files, would be to turn this into a proper object which stores those items and is created by the RawIO for its use. 

I think the AcqType should characterize all the information which depends on the NlxHeader for a file. Perhaps a new Pegasus type is needed. The logic in build_for_ncs_file would then combine the AcqType with the requested tolerances and choose appropriate parameters for the _parse method. 

Please also note that the RawIO likely should use the actual sampling frequency to avoid errors over longer recordings. 

In truth the existing multiple _parse methods are simply evolved over time to handle the various cases without breaking anything. So explicitly commenting on the cause of these cases and dealing with them is a very good idea.

PeterNSteinmetz:
@samuelgarcia BTW, it seems like some testing data of these larger gaps should be included. If you need me to truncate a file to a smaller number of records or to anonymize the header, I can do that.

zm711:
@samuelgarcia, this is also closed by #1425 right?

samuelgarcia:
Oh yes. Thank you.

PR #1419, opened 2024-02-27
Testing: Add skip test for non-existent writable io if nixio is not installed

zm711:
The current error is very hard to understand
```python
file_or_folder = PosixPath('non_existant_file.nix'), args = (), kwargs = {}
ios = [<class 'neo.io.nixio.NixIO'>, <class 'neo.io.nixio_fr.NixIO'>]
io = <class 'neo.io.nixio_fr.NixIO'>

    def get_io(file_or_folder, *args, **kwargs):
        """
        Return a Neo IO instance, guessing the type based on the filename suffix.
        """
        ios = list_candidate_ios(file_or_folder)
        for io in ios:
            try:
                return io(file_or_folder, *args, **kwargs)
            except:
                continue
    
>       raise IOError(f"Could not identify IO for {file_or_folder}")
E       OSError: Could not identify IO for non_existant_file.nix
```

It turns out this error only occurs because nixio is not installed. We should skip this test if nixio is not installed. 

The other option would be to instead give a meaningful warning (ie install nixio to run this test), but not everyone would require nixio, so I think it is a better idea to just skip for those that don't have this.

### DIFF
--- a/neo/test/iotest/test_get_io.py
+++ b/neo/test/iotest/test_get_io.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from tempfile import TemporaryDirectory
+import platform
 from neo.io import get_io, list_candidate_ios, NixIO
 import pytest
 try:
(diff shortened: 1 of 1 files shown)


zm711:
@apdavison, this how been flagged by some people for failing and since the error is not super helpful it might be better to get this into main sooner rather than later. If you get a moment can you look this over?

zm711:
Just tested this on my windows computer and got the error:

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process
```

So I changed to skip this test if we don't have nix or we are on Windows. Windows has a hard time releasing files so it is not worth testing this tiny feature on Windows. I'm also getting other errors with some of the nix tests in general so if we decide to test io/rawio we will need to potentially work on Widows testing side separate. Anyway I added a note in the body of the text on the step which causes the test to fail on Windows.

alejoe91 (approved):

PR #1422, opened 2024-02-27
Paramaterize gap tolerance

PeterNSteinmetz:
This to incorporate the idea of parameterizing the amount of gap tolerated when deciding on sections and using separate parsing when gaps are allowed.

### DIFF
--- a/neo/rawio/neuralynxrawio/ncssections.py
+++ b/neo/rawio/neuralynxrawio/ncssections.py
@@ -110,9 +110,6 @@ class NcsSectionsFactory:
     more complicated.
     """
 
-    _maxGapSampFrac = 0.2  # maximum fraction of a sampling interval between predicted
-    # and actual record timestamps still considered within one section
-
     @staticmethod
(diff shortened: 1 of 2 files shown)


samuelgarcia:
thank you for this.

In the data we have the tolerance is not a fractor of sample interval but several sample and so a fraction of data block interval which can ba 15 samples for instance. The new internal clock for timstamp is not reliable at all from what I see.

In the case of the data I have is "ATLAS" and internally it go to `_buildGivenActualFrequency` so maybe the disinction between "acqType" to switch to  `_buildForMaxGap` or to `_buildGivenActualFrequency` is maybe not the way to go.

Theses two function have many stuff in common. The main differences is gaps are detected from nb_valid != standard block and the other is timestamp diff between data block is out of tolerance.

I propose a refactoring for merging theses two concept in one unique function. This function will accept parameters like gap tolerance. And will call appropriate parameters depending the "acqType" or could be superseed par parameters given in the init.

What do you think ?

zm711:
Closed by #1425.

PR #1423, opened 2024-02-28
Allow for import of openenphys legacy data without .event files

chrishalcrow:
I'm working with OpenEphys Legancy data with no all_channels.event file. Previously, it could not be imported. This allows for it to be imported, leaving the `OpenEphysRawIO` object with `header["event_channels"]=[]`.

Could add a warning that the data has no all_channels.event file?

### DIFF
--- a/neo/rawio/openephysrawio.py
+++ b/neo/rawio/openephysrawio.py
@@ -284,10 +284,12 @@ class OpenEphysRawIO(BaseRawIO):
                 event_filename = f"all_channels_{oe_index + 1}.events"
 
             fullname = os.path.join(self.dirname, event_filename)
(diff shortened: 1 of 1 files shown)


zm711:
@alejoe91, did you want to take a look at this?

alejoe91:
Looks good, thanks @chrishalcrow (and @zm711 for the gentle reminder :) )

zm711 (changes requested):
Could we wait to merge this one? I just received some new files from someone using openephys that have a new naming convention and it seems like there can also be `events` files present, but without actual data. So in my new PR I patched that as well. 

This current patch will only work if someone removes the `events` files. But doesn't address empty events file.

PR #1425, opened 2024-02-28
Neuralynx : new approach gap tolerance

samuelgarcia:
@PeterNSteinmetz : this is a new proposal for dealing with gapTolerance.

I wrote `_buildNcsGeneric()` that shoudl replace `_buildForMaxGap()` and `_buildGivenActualFrequency()`
with a gapTolerance=None. When it is NOne this should replace the old behavior : gap tolerance=0 or tolerance = 0.2 x sample interval but optionally this number can be superseed.

What do you think ?

I still need to handle how the estimated sample rate is computed depending cases.

This should replace #1422 proposal


PeterNSteinmetz:
Yes, this seems to be going a good direction. 

I agree that the question of which frequency you want to use when estimating times will need to be addressed. Sometimes, as in early files, you will have no estimate at all from the header. Other times, a rounding whole sample/s estimate and other times a fractional estimate. 

There is also the issue of computing the actual sampling frequency for each NcsSection. That could be included as another field in the NcsSection when done. An additional issue is how one wants to estimate frequency when there may be small sections between gaps. In that case, it may be best to use an estimate based on the entire file, rather than a small section.

samuelgarcia:
Hi Peter.
I am back on this. I will neeed this PR to be merged very soon because I will have a workshop next week with 50 participants and they will need to install neoto analyzer dattasetthat containss this weird gaps.

If you are happy with this I propose:
  * to keep this new `_buildNcsGeneric` and remove `_buildForMaxGap()` and `_buildGivenActualFrequency()`
  * pust back the sample rate estimation
  *  add a global kwargs in the reader to accept strange gaps like the  file Anthony sended with a tolerance of a few 
      sample

Is this OK ?

PeterNSteinmetz:
@samuelgarcia I think for now and to get things running that sounds good so long as it does not break any tests. 

I will have a look at it then. I would strongly suggest adding some test data for this type of file so there can be tests of it.

samuelgarcia:
Hi Peter.
thanks you.
I just push the changes. Have a look if you have times.

I cannot make tests for this files because I cannot share this file publically unfortunatly.
I made some local test and evrything is OK for this new files with the new  `strict_gap_mode=False`.
I set `strict_gap_mode=True` to keep the previous behavior. Maybe we will make this automatic when we will receive new files with the same  "new clock system".

PeterNSteinmetz:
Not having the test files may have to work for now. However, I would suggest anonymizing the headers and shortening to produce a file that I don't see any objection to. If you get rid of the exact dates of recording there is nothing identifiable left in the file. 

I can do this with some of the files I have from Anthony if you like.

alejoe91 (approved):

samuelgarcia:
Hi Peter.
For the file I have and the one from Anthony, we are not allowed to push then publically even anonymized. They are from the same lab.
Maybe I can ask then to generate a file without any patient just recording noise.

If this is OK for you. I merge this now and then we can improve this later.

PeterNSteinmetz:
Hi Sam, it seems this is the right way to go for now. In terms of test files, I agree that just some recordings by the system should work. The only thing which really matters for this purpose is the gaps being present in the usual way. 

We could also take a few existing files and not only anonymize the headers but replace all the data itself with zeros in the records. At that point they really are not the same files. Another approach would be to generate new files with random data but with the exact same sequence of timestamps in the records, or perhaps with a fixed offset on each record.

samuelgarcia:
Yes you are right I could put some zeros in the same files.
I will try to do that soon.

PR #1428, opened 2024-03-06
TiffIO: Allow bottom left origin for images

### LINKED ISSUE #876: ImageSequences image orientation
Creating an ImageSequence object with the TiffIO currently defines the origin in the upper left corner of the image. This might be unintuitive for some users (including me).
To move the origin to the lower left the images could simply be flipped in the IO using the `PIL.Image.transpose(method=Image.FLIP_TOP_BOTTOM)` or the `numpy.flip(img, axis=-2)` function.
Or in the case that placing the origin into the upper-left corner is the desired behavior, this information should be added to the documentation.
A third option might be to have an optional argument in the IO to determine the origin.

zm711:
Fixes #876.

I added in a note (in the docstring) explaining that upper left is the python default (from PIL and pillow). As a default I left the upper left, but if the user sets a new argument `python_image_origin=False`, then it will do bottom left. I agree that upper left is weird, but it is what the python world expects, so I think the boolean is fair (python standard vs other standard). 

1) I don't love the name I came up with of `python_image_origin` so if someone has a better name let me know. 
2) I only did a boolean, ie 2 choices, but I guess we could do more complicated, but I think all that flipping is dangerous.


zm711:
Meeting discussion: string parsing vs boolean for setting bottom-left vs top-left.

neo/test/iotest/test_tiffio.py
```diff
@@ -35,6 +36,21 @@ def test_read_group_of_tiff_grayscale(self):
         self.assertEqual(blck.segments[0].imagesequences[0].sampling_rate, 1.0 * pq.Hz)
         self.assertEqual(blck.segments[0].imagesequences[0].spatial_scale, 1.0 * pq.micrometer)
 
+        ioclass_bl_origin = TiffIO(
+            directory_path=directory,
+            units="V",
+            sampling_rate=1.0 * pq.Hz,
+            spatial_scale=1.0 * pq.micrometer,
+            python_image_orgin=False,
```
zm711, line 44:
```suggestion
            origin='bottom-left',
```

neo/io/tiffio.py
```diff
@@ -98,13 +130,17 @@ def natural_sort(l):
         list_data_image = []
         for file_name in file_name_list:
             data = np.array(PIL.Image.open(self.filename + "/" + file_name)).astype(np.float32)
+            if self.origin == "bottom-left":
+                data = np.flip(data, axis=1)
```
zm711, line 134: I just need to doublecheck this axis argument.

zm711:
@apdavison I think you're best to review this, but if you think someone else is better just let me know. This is done now so you can review and merge whenever you want (or at our next meeting--no rush).

zm711:
@apdavison double check this before 0.14

zm711:
@apdavison now that we've fixed the ci this is ready to review before the meeting next week :)

PR #1429, opened 2024-03-06
Update Examples in Gallery

### LINKED ISSUE #605: Use sphinx-gallery

### LINKED ISSUE #1264: Example folder needs updating: ChannelIndex is not working
I am new to neo.
[https://github.com/NeuralEnsemble/python-neo/blob/master/examples/generated_data.py](https://github.com/NeuralEnsemble/python-neo/blob/master/examples/generated_data.py) is written four years ago. `ChannelIndex` is gone but examples are not refreshed.
Hopefully `examples` folder can be refreshed.
Thanks!

zm711:
I need to run my testing on read the docs, because some docs build locally fine, but are failing on our docs.  So I need to trigger the build.

Two edits:
I realized we only build docs on main and not on PRs so I can't check some of my solutions until after merge.
Fixes #1264 (by deleting generated_data and replacing it with the updated working tetrode example)
Fixes #605 (by changing names to `plot_*` so that sphinx-gallery actually works + ignore failures so that it does ruin the building if an example fails).


zm711:
Files have to be called `plot_*` to be executed by the gallery [see here](https://sphinx-gallery.github.io/stable/configuration.html)

zm711:
TODO: 
generated_data + maybe extract out the plotting there into its own function.

zm711:
Actually in the doc/source/scripts there was actually a perfect example to replace generated_data (which is super out of date). I moved that to here and got rid of generated data. I think this should get basically all the examples ready to go.

samuelgarcia (approved):

doc/source/conf.py
```diff
@@ -252,6 +252,7 @@
 )
 
 sphinx_gallery_conf = {
+    'only_warn_on_example_error': True, # helps with debugging broken examples
```
samuelgarcia, line 255: Maybe we could set it to False when noe "dev" in version

PR #1430, opened 2024-03-08
Open Ephys: read events from all streams

### LINKED ISSUE #1367: Can't read TTL events from multiple streams in an Open Ephys file
**Describe the bug**
It is impossible to load TTL events from multiple streams in Open Ephys

**To Reproduce**

1) Run Open Ephys v0.6.6
2) Add a Neuropixels source and an Intan RHD source, connected by a Merger, followed by a Record Node to your signal chain
3) Record a brief recording
4) Run the following python script:
```
from neo.rawio.openephysbinaryrawio import explore_folder

explore_folder(folder) # where folder is wherever you saved the data in step 3
```

**Expected behaviour**
The returned data structure contains entries for every TTL event stream in the recording.

**Actual behaviour**
The returned data structure contains entries only for MessageCenter messages and the TTL event stream for whichever stream comes last lexicographically (in my case, the Intan RHD a.k.a. Rhythm FPGA data)

**Environment:**
 - OS: Windows 10
 - Python version: 3.9.0
 - Neo version: 0.12.0
 - NumPy version: 1.26.2

**Additional context**
This appears to be due to the way TTL events are stored in Open Ephys binary format, as a subfolder named 'TTL' under the corresponding event stream folder, which causes each newly read TTL stream to clobber the one before it in the folder structure dictionary created around lines 547-559 of openephysbinaryrawio.py

An Open Ephys session folder with the offending folder structure can be found [here](https://drive.google.com/file/d/1F8ThEgE0LYJ33sXID_WY8gpLbdrfOeqj/view?usp=drive_link). Note that there are no actual events to be loaded in that recording (get_events should return an empty array). If you want one with actual events I have one I can share, but it is ~25 GB, so I don't want to upload it to Drive if I don't have to.

This affects downstream packages (e.g. SpikeInterface) that depend on Neo. It is also prevents properly synchronising datastreams from multiple different recording systems recorded with Open Ephys, because only one sync channel is loaded.

alejoe91:
Fixes #1367

@johnmbarrett could you test it on your side?

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -549,7 +549,7 @@ def explore_folder(dirname, experiment_names=None):
                 for info in rec_structure["continuous"]:
                     # when multi Record Node the stream name also contains
                     # the node name to make it unique
-                    oe_stream_name = Path(info["folder_name"]).name  # remove trailing slash
+                    oe_stream_name = info["folder_name"].split("/")[0]  # remove trailing slash
                     if len(node_name) > 0:
                         stream_name = node_name + "#" + oe_stream_name
                     else:
@@ -580,8 +580,13 @@ def explore_folder(dirname, experiment_names=None):
             if (root / "events").exists() and len(rec_structure["events"]) > 0:
                 recording["streams"]["events"] = {}
                 for info in rec_structure["events"]:
-                    oe_stream_name = Path(info["folder_name"]).name  # remove trailing slash
-                    stream_name = node_name + "#" + oe_stream_name
+                    # when multi Record Node the stream name also contains
+                    # the node name to make it unique
+                    oe_stream_name = info["folder_name"].split("/")[0]  # remove trailing slash
+                    if len(node_name) > 0:
+                        stream_name = node_name + "#" + oe_stream_name
+                    else:
+                        stream_name = oe_stream_name
 
                     event_stream = info.copy()
                     for name in _possible_event_stream_names:


neo/rawio/openephysbinaryrawio.py
```diff
@@ -580,8 +580,13 @@ def explore_folder(dirname, experiment_names=None):
             if (root / "events").exists() and len(rec_structure["events"]) > 0:
                 recording["streams"]["events"] = {}
                 for info in rec_structure["events"]:
-                    oe_stream_name = Path(info["folder_name"]).name  # remove trailing slash
-                    stream_name = node_name + "#" + oe_stream_name
+                    # when multi Record Node the stream name also contains
+                    # the node name to make it unique
+                    oe_stream_name = info["folder_name"].split("/")[0]  # remove trailing slash
+                    if len(node_name) > 0:
+                        stream_name = node_name + "#" + oe_stream_name
```
zm711, line 587: It's always a #?  I have a feeling something is going to change in the future (because why would people be consistent :)

alejoe91: This is purely a naming on the NEO side ;) it's always been with the `#`

zm711: Cool cool. Sounds good. :)

johnmbarrett:
Seems to work insofar as I can load event times from all streams of my test file. There are a bunch of other problems with the durations and timestamps, but I think they're unrelated to this so I'll open new issues for those.

johnmbarrett:
One thing I will note is that the Neuropixels AP and LFP event channels end up with the same name and ID in the resulting header structure inside the `OpenEphysBinaryRawIO` object, which means that one of them will be invisible to downstream packages (in particular spikeinterface) that access event channels by ID rather than a numeric index. Should that be addressed here or should I raise an issue with spikeinterface (once this change is incorporated into a neo release)? It's not catastrophic for my purposes since they're the same events and should have the same timestamps after accounting for sampling rate, but just thought I'd point it out.

johnmbarrett:
Test dataset for the behaviour described in my last comment is available here: https://drive.google.com/file/d/1Hr3-GFOpOkNc1hgXdVr5HTunchQm4enw/view?usp=drive_link

alejoe91:
> One thing I will note is that the Neuropixels AP and LFP event channels end up with the same name and ID in the resulting header structure inside the `OpenEphysBinaryRawIO` object, which means that one of them will be invisible to downstream packages (in particular spikeinterface) that access event channels by ID rather than a numeric index. Should that be addressed here or should I raise an issue with spikeinterface (once this change is incorporated into a neo release)? It's not catastrophic for my purposes since they're the same events and should have the same timestamps after accounting for sampling rate, but just thought I'd point it out.

How would you change the implementation to account for this? Maybe we can merge this one and you can make a PR to fix the other issue?

johnmbarrett:
> How would you change the implementation to account for this? Maybe we can merge this one and you can make a PR to fix the other issue?

Just looking through the code, when the `signal_streams` list is set up each channel seems to be given a unique ID, but when setting up the equivalent `event_streams` list the channel name is just passed in twice (on line 187 for openephysbinaryrawio.py), once as the name and once as the ID. Seems like you could use stream name rather than the channel name (assuming one channel per stream for events, which I think is true since different TTL lines get assigned different labels in the event stream rather than different channels), or make up some identifier that's unique by construction. Or leave as is and document somewhere that event channels are not guaranteed to have unique IDs, then it's an issue for downstream packages.

zm711 (approved):
At least from my side this looks good and stops the clobbering. I think improvements could be a separate PR, no?

samuelgarcia (approved):

PR #1435, opened 2024-03-14
AxonRawIO: Change Strings Parsing for additional formats.

### LINKED ISSUE #1434: Cannot open some .abf files
**Describe the bug**
I am trying the code below to open .abf files, for some it works greatly, while for others it is not working and I am getting error which is also given below. Using Pyabf works, but ofcourse Pyabf is slow and I really like Neo. 

**To Reproduce**
Function I used is as follows:
def load_abf_file(file_path)
        reader = neo.io.AxonIO(filename=file_path)
        block = reader.read_block(signal_group_mode='split-all')
        seg = block.segments[0]
        analogsignal = seg.analogsignals[0]
        data = analogsignal.magnitude.flatten()
        sampling_rate = analogsignal.sampling_rate.magnitude
        return data, sampling_rate

Error I am getting is:

line 173, in on_test_settings_clicked
    data, sampling_rate = self.load_abf_file(file_path)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   line 184, in load_abf_file
    reader = neo.io.AxonIO(filename=file_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shankardutt/anaconda3/envs/SD_Apps/lib/python3.11/site-packages/neo/io/axonio.py", line 44, in __init__
    BaseFromRaw.__init__(self, filename)
  File "/Users/shankardutt/anaconda3/envs/SD_Apps/lib/python3.11/site-packages/neo/io/basefromrawio.py", line 77, in __init__
    self.parse_header()
  File "/Users/shankardutt/anaconda3/envs/SD_Apps/lib/python3.11/site-packages/neo/rawio/baserawio.py", line 185, in parse_header
    self._parse_header()
  File "/Users/shankardutt/anaconda3/envs/SD_Apps/lib/python3.11/site-packages/neo/rawio/axonrawio.py", line 64, in _parse_header
    info = self._axon_info = parse_axon_soup(self.filename)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/shankardutt/anaconda3/envs/SD_Apps/lib/python3.11/site-packages/neo/rawio/axonrawio.py", line 452, in parse_axon_soup
    assert goodstart != -1, "This file does not contain clampex, axoscope or clampfit in the header"

You can download the file here: https://anu365-my.sharepoint.com/:u:/g/personal/u6591731_anu_edu_au/Ed8PI3vnWiJDrxienTkPQjQBT_vddFVcRwisoyO_rZTDuQ?e=mFFKZY 

**Expected behaviour**
It should load the file

**Environment:**
 - OS: macOS
 - Python version: 3.10
 - Neo version: 0.13
 - NumPy version: 1.26.0

zm711:
Fixes #1434.

Still need to be tested by user in PR #1434, but this uses the string parsing strategy that pyABF came up with. License is an MIT https://github.com/swharden/pyABF/blob/main/LICENSE, and docs for his logic here: https://swharden.com/pyabf/abf2-file-format/.

Basically the issue in #1434, the device is not returned (so no `[b"AXENGN", b"clampex", b"Clampex", b"EDR3", b"CLAMPEX", b"axoscope", b"AxoScope", b"Clampfit"`]. I tested the test file locally and it seemed to run when I changed the parsing as in this PR. 

the strategy in short is to parse based \xx0\xx0 rather than trying to find the device. The string format then actually follows the correct indices as given by ADC info so we don't need to subtract one.

@shankardutt can you test from this PR with your data. I tested it (but not extensively) and it seemed to work.

### DIFF
--- a/neo/rawio/axonrawio.py
+++ b/neo/rawio/axonrawio.py
@@ -20,7 +20,11 @@ presented in pyABF MIT License Copyright (c) 2018 Scott W Harden
 written by Scott Harden. His unofficial documentation for the formats
 is here:
 https://swharden.com/pyabf/abf2-file-format/
-
+strings section:
+[uModifierNameIndex, uCreatorNameIndex, uProtocolPathIndex, lFileComment, lADCCChannelNames, lADCUnitsIndex
+lDACChannelNameIndex, lDACUnitIndex, lDACFilePath, nLeakSubtractADC]
+['', 'Clampex', '', 'C:/path/protocol.pro', 'some comment', 'IN 0', 'mV', 'IN 1', 'mV', 'Cmd 0', 'pA', 
+'Cmd 1', 'pA', 'Cmd 2', 'mV', 'Cmd 3', 'mV']
 
 Information on abf 1 and 2 formats is available here:
 http://www.moleculardevices.com/pages/software/developer_info.html
@@ -472,15 +476,17 @@ def parse_axon_soup(filename):
             # not very robust.
             f.seek(sections["StringsSection"]["uBlockIndex"] * BLOCKSIZE)
             big_string = f.read(sections["StringsSection"]["uBytes"])
-
             # this idea comes from pyABF https://github.com/swharden/pyABF
             # previously we searched for clampex, Clampex etc, but this was
             # brittle. pyABF believes that looking for the \x00\x00 is more
             # robust. We find these values, replace mu->u, then split into
             # a set of strings
             indexed_string = big_string[big_string.rfind(b'\x00\x00'):]
+            # replace mu -> u for easy display
             indexed_string = indexed_string.replace(b'\xb5', b'\x75')
-            indexed_string = indexed_string.split(b'\x00')
+            # we need to remove one of the \x00 to have the indices be
+            # the correct order
+            indexed_string = indexed_string.split(b'\x00')[1:]
             strings = indexed_string
 
             # ADC sections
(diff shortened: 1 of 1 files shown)


shankardutt:
@zm711 This has worked nicely for my data. Thanks a lot for all your work! Really appreciate this :) 

Sorry for my late reply (being in Australia - doesn't help :))

zm711:
No worries. :) As long as everything looks reasonable. Would you be willing to share `test.abf` for our testing repo so that we can make sure that any updates also work for your type of files?

zm711:
@samuelgarcia, you wrote the initial so when you have time to read over this I think you're probably the right person to review. It's only a couple line change.

neo/rawio/axonrawio.py
```diff
@@ -489,8 +501,8 @@ def parse_axon_soup(filename):
                         ADCInfo[key] = val[0]
                     else:
                         ADCInfo[key] = np.array(val)
-                ADCInfo["ADCChNames"] = strings[ADCInfo["lADCChannelNameIndex"] - 1]
-                ADCInfo["ADCChUnits"] = strings[ADCInfo["lADCUnitsIndex"] - 1]
+                ADCInfo["ADCChNames"] = strings[ADCInfo["lADCChannelNameIndex"]]
+                ADCInfo["ADCChUnits"] = strings[ADCInfo["lADCUnitsIndex"]]
```
samuelgarcia, line 505: Why do we remove this *-1* ?
I am not sure to follow the logic why the new string parsing change the one or zero based indexing.

zm711: Hey Sam. It's because the string actually has an id at the beginning `[uModifierNameIndex, uCreatorNameIndex, uProtocolPathIndex, lFileComment, lADCCChannelNames, lADCUnitsIndex
lDACChannelNameIndex, lDACUnitIndex, lDACFilePath, nLeakSubtractADC]`

Our previous parsing started at `uCreatorNameIndex` which meant we had to subtract 1 to get the correct indexing. The new indexing starts at the appropriate index `uModifernameIndex` which means that don't need to do the shifting.

samuelgarcia (approved):

samuelgarcia:
ready to merge on side.

samuelgarcia (approved):

PR #1440, opened 2024-03-19
NeuralynxIO: support List[str] for filename parameter

NxNiki:
Add a feature to allow multiple files as filename. 

Now user can create a list of filenames and assign it to the argument `filename` of `neo.io.NeuralynxIO(filename=session_data)` to read the specific files. 

If there are multiple files for a single channel, they should be consistent for all channels.

### DIFF
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -69,11 +69,11 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             Default: False
         """
 
-        if filename:
+        if filename is not None:
             warnings.warn('Deprecated and will be removed. Please use `include_filenames` instead')
             include_filenames = [filename]
 
-        if exclude_filename:
+        if exclude_filename is not None:
             warnings.warn('Deprecated and will be removed. Please use `exclude_filenames` instead')
             exclude_filenames = exclude_filename
 
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -30,7 +30,6 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
     def __init__(
         self,
         dirname="",
-        filename="",
         use_cache=False,
         cache_path="same_as_resource",
         include_filename=None,
@@ -44,9 +43,6 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         ----------
         dirname : str
             Directory containing data files
-        filename : str
-            Name of a single ncs, nse, nev, or ntt file to include in dataset. Will be ignored,
-            if dirname is provided.
         use_cache : bool, optional
             Cache results of initial file scans for faster loading in subsequent runs.
             Default: False
@@ -63,20 +59,17 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             shifted to begin at t_start = 0*pq.second.
             Default: False
         """
+
         NeuralynxRawIO.__init__(
-            self,
-            dirname=dirname,
-            filename=filename,
-            use_cache=use_cache,
-            cache_path=cache_path,
-            include_filename=include_filename,
-            exclude_filename=exclude_filename,
+            self, dirname=dirname,
+            include_filenames=include_filename,
+            exclude_filenames=exclude_filename,
             keep_original_times=keep_original_times,
+            use_cache=use_cache,
+            cache_path=cache_path
         )
 
-        if self.rawmode == "one-file":
-            BaseFromRaw.__init__(self, filename)
-        elif self.rawmode == "one-dir":
+        if self.rawmode == "one-dir":
             BaseFromRaw.__init__(self, dirname)
         elif self.rawmode == "multiple-files":
             BaseFromRaw.__init__(self, include_filename=include_filename)
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -9,6 +9,7 @@ Supported: Read
 Author: Julia Sprenger, Carlos Canova
 """
 
+import warnings
 from neo.io.basefromrawio import BaseFromRaw
 from neo.rawio.neuralynxrawio.neuralynxrawio import NeuralynxRawIO
 
@@ -67,6 +68,14 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             shifted to begin at t_start = 0*pq.second.
             Default: False
         """
+
+        if filename:
+            warnings.warn('Deprecated and will be removed. Please use `include_filenames` instead')
+            include_filenames = [filename]
+
+        if exclude_filename:
+            warnings.warn('Deprecated and will be removed. Please use `include_filenames` instead')
+            exclude_filenames = exclude_filename
 
         NeuralynxRawIO.__init__(
             self,
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -49,6 +49,8 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         cache_path : str, optional
             Folder path to use for cache files.
             Default: 'same_as_resource'
+        exclude_filename: None,
+            Deprecated and will be removed. Please use `exclude_filenames` instead
         include_filenames: str or list
             Filename or list of filenames to be included. This can be absolute path or path relative to dirname.
         exclude_filenames: str or list
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -33,6 +33,7 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         filename="",
         use_cache=False,
         cache_path="same_as_resource",
+        include_filename=None,
         exclude_filename=None,
         keep_original_times=False,
     ):
@@ -52,6 +53,8 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         cache_path : str, optional
             Folder path to use for cache files.
             Default: 'same_as_resource'
+        include_filename: str or list
+            Filename or list of filenames to be included. This can be absolute path or path relative to dirname.
         exclude_filename: str or list
             Filename or list of filenames to be excluded. Expects base filenames without
             directory path.
@@ -66,6 +69,7 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             filename=filename,
             use_cache=use_cache,
             cache_path=cache_path,
+            include_filename=include_filename,
             exclude_filename=exclude_filename,
             keep_original_times=keep_original_times,
         )
@@ -73,3 +77,6 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             BaseFromRaw.__init__(self, filename)
         elif self.rawmode == "one-dir":
             BaseFromRaw.__init__(self, dirname)
+        elif self.rawmode == "multiple-files":
+            BaseFromRaw.__init__(self, include_filename=include_filename)
+
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -74,7 +74,7 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
             include_filenames = [filename]
 
         if exclude_filename:
-            warnings.warn('Deprecated and will be removed. Please use `include_filenames` instead')
+            warnings.warn('Deprecated and will be removed. Please use `exclude_filenames` instead')
             exclude_filenames = exclude_filename
 
         NeuralynxRawIO.__init__(
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -32,8 +32,8 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         dirname="",
         use_cache=False,
         cache_path="same_as_resource",
-        include_filename=None,
-        exclude_filename=None,
+        include_filenames=None,
+        exclude_filenames=None,
         keep_original_times=False,
     ):
         """
@@ -49,9 +49,9 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         cache_path : str, optional
             Folder path to use for cache files.
             Default: 'same_as_resource'
-        include_filename: str or list
+        include_filenames: str or list
             Filename or list of filenames to be included. This can be absolute path or path relative to dirname.
-        exclude_filename: str or list
+        exclude_filenames: str or list
             Filename or list of filenames to be excluded. Expects base filenames without
             directory path.
         keep_original_times : bool
@@ -62,8 +62,8 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
 
         NeuralynxRawIO.__init__(
             self, dirname=dirname,
-            include_filenames=include_filename,
-            exclude_filenames=exclude_filename,
+            include_filenames=include_filenames,
+            exclude_filenames=exclude_filenames,
             keep_original_times=keep_original_times,
             use_cache=use_cache,
             cache_path=cache_path
@@ -72,5 +72,5 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         if self.rawmode == "one-dir":
             BaseFromRaw.__init__(self, dirname)
         elif self.rawmode == "multiple-files":
-            BaseFromRaw.__init__(self, include_filename=include_filename)
+            BaseFromRaw.__init__(self, include_filenames=include_filenames)
 
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -45,6 +45,10 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         ----------
         dirname : str
             Directory containing data files
+        filename : str
+            Deprecated and will be removed. Please use `include_filenames` instead
+            Name of a single ncs, nse, nev, or ntt file to include in dataset. Will be ignored,
+            if dirname is provided.
         use_cache : bool, optional
             Cache results of initial file scans for faster loading in subsequent runs.
             Default: False
--- a/neo/io/neuralynxio.py
+++ b/neo/io/neuralynxio.py
@@ -29,12 +29,14 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
 
     def __init__(
         self,
-        dirname="",
+        dirname,
         use_cache=False,
         cache_path="same_as_resource",
         include_filenames=None,
         exclude_filenames=None,
         keep_original_times=False,
+        filename=None,
+        exclude_filename=None,
     ):
         """
         Initialise IO instance
@@ -63,7 +65,8 @@ class NeuralynxIO(NeuralynxRawIO, BaseFromRaw):
         """
 
         NeuralynxRawIO.__init__(
-            self, dirname=dirname,
+            self,
+            dirname=dirname,
             include_filenames=include_filenames,
             exclude_filenames=exclude_filenames,
             keep_original_times=keep_original_times,
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -151,11 +151,11 @@ class NeuralynxRawIO(BaseRawIO):
 
         if filename:
             include_filenames = filename
-            raise DeprecationWarning("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+            warnings.warn("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
 
         if exclude_filename:
             exclude_filenames = exclude_filename
-            raise DeprecationWarning("`exclude_filename` is deprecated and will be removed. Please use `exclude_filenames` instead")
+            warnings.warn("`exclude_filename` is deprecated and will be removed. Please use `exclude_filenames` instead")
 
         if include_filenames is None:
             include_filenames = []
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -136,13 +136,23 @@ class NeuralynxRawIO(BaseRawIO):
 
     def __init__(
             self,
-            dirname="",
+            dirname,
             include_filenames=None,
             exclude_filenames=None,
             keep_original_times=False,
             strict_gap_mode=True,
+            filename=None,
+            exclude_filename=None,
             **kargs
     ):
+
+        if filename:
+            include_filenames = filename
+            raise DeprecationWarning("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+
+        if exclude_filename:
+            exclude_filenames = exclude_filename
+            raise DeprecationWarning("`exclude_filename` is deprecated and will be removed. Please use `exclude_filenames` instead")
 
         if include_filenames is None:
             include_filenames = []
@@ -158,9 +168,6 @@ class NeuralynxRawIO(BaseRawIO):
             include_filepath = {os.path.dirname(f) for f in include_filenames}
             if len(include_filepath) > 1:
                 raise ValueError("Files in include_filename must be in a single path!")
-
-        if (include_filenames is None) and (dirname == ""):
-            raise ValueError("One of dirname or include_filenames must be provided.")
 
         if dirname and include_filenames:
             dirname = os.path.join(dirname, os.path.dirname(include_filenames[0]))
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -167,7 +167,7 @@ class NeuralynxRawIO(BaseRawIO):
         self.dirname = dirname
         self.filename = filename
         self.include_files = include_filename
-        self.exclude_filename = exclude_filename
+        self.exclude_filename = [os.path.basename(f) for f in exclude_filename] if exclude_filename else None
         self.keep_original_times = keep_original_times
         self.strict_gap_mode = strict_gap_mode
         BaseRawIO.__init__(self, **kargs)
@@ -227,8 +227,8 @@ class NeuralynxRawIO(BaseRawIO):
         file_basenames = [os.path.basename(f) for f in filenames]
         if self.exclude_filename is not None:
             for excl_file in self.exclude_filename:
-                if os.path.basename(excl_file) in file_basenames:
-                    filenames.remove(os.path.join(self.dirname, os.path.basename(excl_file)))
+                if excl_file in file_basenames:
+                    filenames.remove(os.path.join(self.dirname, excl_file))
 
         stream_props = {}  # {(sampling_rate, n_samples, t_start): {stream_id: [filenames]}
 
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -138,9 +138,12 @@ class NeuralynxRawIO(BaseRawIO):
         if dirname != "":
             self.dirname = dirname
             self.rawmode = "one-dir"
-        elif filename != "":
+        elif filename != "" and isinstance(filename, str):
             self.filename = filename
             self.rawmode = "one-file"
+        elif isinstance(filename, list):
+            self.filename = filename
+            self.rawmode = "multiple-files"
         else:
             raise ValueError("One of dirname or filename must be provided.")
 
@@ -184,17 +187,20 @@ class NeuralynxRawIO(BaseRawIO):
 
         if self.rawmode == "one-dir":
             filenames = sorted(os.listdir(self.dirname))
-            dirname = self.dirname
+            filenames = [os.path.join(self.dirname, f) for f in filenames]
         else:
-            if not os.path.isfile(self.filename):
-                raise ValueError(
-                    f"Provided Filename is not a file: "
-                    f"{self.filename}. If you want to provide a "
-                    f"directory use the `dirname` keyword"
-                )
+            if self.rawmode == "one-file":
+                filenames = [self.filename]
+            else:
+                filenames = sorted(self.filename)
 
-            dirname, fname = os.path.split(self.filename)
-            filenames = [fname]
+            for filename in filenames:
+                if not os.path.isfile(filename):
+                    raise ValueError(
+                        f"Provided Filename is not a file: "
+                        f"{filename}. If you want to provide a "
+                        f"directory use the `dirname` keyword"
+                    )
 
         if not isinstance(self.exclude_filename, (list, set, np.ndarray)):
             self.exclude_filename = [self.exclude_filename]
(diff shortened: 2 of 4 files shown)


zm711:
I guess one question I would have overall is why overload the filename argument? Unfortunately I don't think we would change it since it is a core part, but why not add another argument include_filenames to mirror the exclude_filenames. I'll ping @samuelgarcia here so he can decide if he is okay with the overloading or not.

samuelgarcia:
Maybe w could use filenames to avoid confusion ? even if it makes the signature a bit heavy...

zm711:
I think having `filename` and `filenames` would be super confusing and misused a lot. Would you be okay with `include_filenames` then it would still be `dirname='path'`, `filename=None`, and `include_filenames=['file0', 'file1']`?

Or are you suggesting switch filename to filenames and have some broken code?

NxNiki:
> I think having `filename` and `filenames` would be super confusing and misused a lot. Would you be okay with `include_filenames` then it would still be `dirname='path'`, `filename=None`, and `include_filenames=['file0', 'file1']`?
> 
> Or are you suggesting switch filename to filenames and have some broken code?

I don't feel it is necessary to create separate variables for single and multiple input files in the arguments. Maybe we can use `filenames` as the argument and create `self.filename = filenames` if `filenames` is a string and do `self.filename = filenames[0]` if `len(filenames)==1`?

zm711:
@NxNiki,

I agree that's the better overall solution, but changing the argument name could break users' code. So my question for Sam was whether he was okay to do a breaking change at the api level or if he wanted to work around the current api to protect the old way of doing things.

NxNiki:
> @NxNiki,
> 
> I agree that's the better overall solution, but changing the argument name could break users' code. So my question for Sam was whether he was okay to do a breaking change at the api level or if he wanted to work around the current api to protect the old way of doing things.

Maybe add a deprecate warning?

apdavison:
I think this is perhaps an opportunity to formalize IO constructor arguments across all IO classes, before we get to the 1.0 release.

I think it will be confusing to have some IO classes that accept a list of things as the first argument, and others only a single thing. To solve the current issue, I support @zm711's suggestion of adding `include_filenames`, but then let's also remove the `filename` argument (or rather, add a deprecation warning for now, remove it in v1.0)

NxNiki:
> I think this is perhaps an opportunity to formalize IO constructor arguments across all IO classes, before we get to the 1.0 release.
> 
> I think it will be confusing to have some IO classes that accept a list of things as the first argument, and others only a single thing. To solve the current issue, I support @zm711's suggestion of adding `include_filenames`, but then let's also remove the `filename` argument (or rather, add a deprecation warning for now, remove it in v1.0)

Yeah, that makes sense. Shall we ignore `dirname` and `filename` when using `include_filenames` or add files to `filename`/files in `dirname`? The later option seems logical but complex...

apdavison:
> Shall we ignore `dirname` and `filename` when using `include_filenames` or add files to `filename`/files in `dirname`? The later option seems logical, but the logic is kind of complex and the use case may be rare in practice.

I don't use NeuralynxIO very much, so I will defer to those who do, but would suggest:
1) If `include_filenames` is not empty, then `filename` must be empty (otherwise raise exception), and
2) If `dirname` is not empty, then the contents of `include_filenames` should be interpreted as paths relative to `dirname`.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -132,26 +133,56 @@ class NeuralynxRawIO(BaseRawIO):
...
     ):
 
+        if (include_filename is not None) and (filename != ""):
+            raise ValueError("filename and include_filenames cannot be both assigned")
+        if (include_filename is None) and (filename == "") and (dirname == ""):
+            raise ValueError("One of dirname or filename or include_files must be provided.")
+
+        if dirname != "" and (include_filename is not None):
+            include_filename = [os.path.join(dirname, f) for f in include_filename]
+            dirname = ""
```
zm711, line 153: I see why you changed dirname to an empty string for the boolean check below, but for the include_filename it should still return the boolean, so why not just split the boolean below

```python
if dirname != "":
    if include_filename is not None:
          self.rawmode = 'multiple-files'
    else:
          self.rawmode='one-dir'
```

for line 160. I wouldn't delete the dirname info since it is actually the root of the current NeuralynxRawIO instance.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -132,26 +133,56 @@ class NeuralynxRawIO(BaseRawIO):
...
         self.strict_gap_mode = strict_gap_mode
-        self.exclude_filename = exclude_filename
         BaseRawIO.__init__(self, **kargs)
 
     def _source_name(self):
         if self.rawmode == "one-file":
             return self.filename
+        elif self.rawmode == "multiple-files":
+            dirname = {os.path.dirname(x) for x in self.filename}
+            if len(dirname) > 1:
```
zm711, line 180: this should not be possible.... because we make the include_file relative to the dirname so they have to all have the same dirname. In this case if you fix the boolean check above then all you would need to do is:

```python
if self.rawmode == 'one-file':
   return self.filename
else:
    return self.dirname
```

zm711 (commented):
Couple comments about the current implementation.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: Thanks for pointing this out! I meant to check the `dirname` in `include_filename` (forgot to update this after it is added). I was thinking the user may put files in different directories in `include_filename`. Say ['dirname/exp1/file1', 'dirname/exp2/file2'].

zm711: Neo doesn't really support multi-dir in that way. There needs to be a root dir or a root file for each IO. There could be subfolders from the root, that could be explored, but not completely different root directories. So the source should return either a dirname or a filename.

NxNiki: OK, will just extract `dirname` from the first file path, assuming all files are in the same directory.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: Yes, that makes sense!

NxNiki:
Hi, I am trying to run pytest, but it seems the environment cannot be correctly set up. Neither setup.py nor the pyproject.toml works for me.

zm711:
@NxNiki,

I routinely set up new environments building neo from source. What is the exact error you're getting. Note that we are still trying to work on making sure all tests work on Windows so testing may not be perfect on Windows machines (outside of core tests).

NxNiki:
> @NxNiki,
> 
> I routinely set up new environments building neo from source. What is the exact error you're getting. Note that we are still trying to work on making sure all tests work on Windows so testing may not be perfect on Windows machines (outside of core tests).

I am running the code on a Mac (M1 chip). The error when I use poetry to install the package is:
```
poetry install
[tool.poetry] section not found in /Users/XinNiuAdmin/Documents/python-neo/pyproject.toml.
```

I can use pip to install the packages with `pip install .` But running pytest has the following error:
```
======================================================================================================= test session starts =======================================================================================================
platform darwin -- Python 3.9.6, pytest-8.1.1, pluggy-1.4.0
rootdir: /Users/XinNiuAdmin/Documents/python-neo
configfile: pyproject.toml
collected 1389 items / 1 error                                                                                                                                                                                                    

============================================================================================================= ERRORS ==============================================================================================================
_________________________________________________________________________________________ ERROR collecting neo/test/iotest/test_tiffio.py _________________________________________________________________________________________
ImportError while importing test module '/Users/XinNiuAdmin/Documents/python-neo/neo/test/iotest/test_tiffio.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
neo/test/iotest/test_tiffio.py:3: in <module>
    from PIL import Image
E   ModuleNotFoundError: No module named 'PIL'
===================================================================================================== short test summary info =====================================================================================================
ERROR neo/test/iotest/test_tiffio.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================================================================== 1 error in 1.01s =========================================================================================================
```

I guess this is also due to the incorrect `pyproject.toml`?

zm711:
@NxNiki,

If you want to run full tests you should do

```
pip install -e .[test]
```

You are missing a package. This is fine if you don't want to download a bunch of extra packages, but then rather than run pytest at the root you'll need to run individual tests instead. We don't have a poetry section in our code and I don't use poetry so I'm not sure how to install with poetry.

zm711:
@NxNiki,

I would probably avoid fixing typos in this PR for other files since the focus on this PR should be getting your changes to neuralynx to pass the testing. We appreciate typo fixes, but those are better in a dedicated PR (that way we can follow a flow of one PR does one thing).

NxNiki:
> @NxNiki,
> 
> I would probably avoid fixing typos in this PR for other files since the focus on this PR should be getting your changes to neuralynx to pass the testing. We appreciate typo fixes, but those are better in a dedicated PR (that way we can follow a flow of one PR does one thing).

Yeah, that makes sense. Will revert the fix topo change.

zm711:
@samuelgarcia I think this is ready for you to decide on.

zm711:
@NxNiki,

So the Neo maintenance team met today and discussed that we want to move ahead with regularizing the constructors for IOs. That would mean for this io we would want just to accept a `dirname`. We would want to deprecate the use of `filename` and so if someone wanted a single file they would need to use `dirname='my_data/folder', include_filenames=['one_file']`. We would also want to deprecate the `include_filename` and `exclude_filename` in favor of `include_filenames` and `exclude_filenames`. Would you be willing to incorporate those changes to this PR.

zm711:
Thanks we really appreciate it!

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,58 @@ class NeuralynxRawIO(BaseRawIO):
...
+            **kargs
     ):
 
-        if dirname != "":
-            self.dirname = dirname
-            self.rawmode = "one-dir"
-        elif filename != "":
-            self.filename = filename
-            self.rawmode = "one-file"
+        if include_filenames is None:
```
zm711, line 147: here we would check that filename is None. If it is not none we raise an error and mention the deprecation again, so that old users know how to fix for the API change.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,58 @@ class NeuralynxRawIO(BaseRawIO):
...
         ("sample_rate", "uint32"),
         ("nb_valid", "uint32"),
-        ("samples", "int16", (NcsSection._RECORD_SIZE)),
+        ("samples", "int16", NcsSection._RECORD_SIZE),
     ]
 
     def __init__(
-        self, dirname="", filename="", exclude_filename=None, keep_original_times=False, strict_gap_mode=True, **kargs
+            self,
+            dirname="",
```
zm711, line 139: keep filename, but you can set it to None. 
same for exclude_filename. Keep it but set to None. If someone uses it we raise an error.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,58 @@ class NeuralynxRawIO(BaseRawIO):
...
+        elif not isinstance(exclude_filenames, (list, set, np.ndarray)):
+            exclude_filenames = {exclude_filenames}
+
+        if include_filenames:
+            include_filepath = {os.path.dirname(f) for f in include_filenames}
+            if len(include_filepath) > 1:
+                raise ValueError("Files in include_filename must be in a single path!")
+
+        if (include_filenames is None) and (dirname == ""):
+            raise ValueError("One of dirname or include_filenames must be provided.")
```
zm711, line 163: dirname is always required. include_filenames is optional.

zm711, line 163: Basically the flow should be:

Check for deprecated arguments `filename` and `include_filename`: warn user about deprecation and explain that we now only accept dirname and that instead of include_filename we use include_filenames

 check for dirname. If dirname doesn't exist raise an error.

Then we check for include and exclude (those are optional).

zm711 (commented):
Hey @NxNiki,

couple more comments as we work on this. Let me know if anything is unclear.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: Shall we just raise a deprecation warning and assign `filename` to `include_filenames`? And do similar for `execlude_filename`?

neo/rawio/neuralynxrawio/neuralynxrawio.py
alejoe91: Yes, since we're at it let's do the full refactor

neo/rawio/neuralynxrawio/neuralynxrawio.py
zm711: Yeah you could do that too!

NxNiki:
Hi, Sorry for the delay. I think this PR is ready to review. Let me know if there are additional comments.

neo/io/neuralynxio.py
```diff
@@ -52,24 +56,39 @@ def __init__(
...
+        exclude_filenames: str or list
             Filename or list of filenames to be excluded. Expects base filenames without
             directory path.
         keep_original_times : bool
             Preserve original time stamps as in data files. By default datasets are
             shifted to begin at t_start = 0*pq.second.
             Default: False
         """
+
+        if filename:
```
zm711, line 72: It's actually a bit better to say 
```
if filename is not None
```
 Since we are explicitly making it None and so the user would have to actively change this.

neo/io/neuralynxio.py
```diff
@@ -52,24 +56,39 @@ def __init__(
...
+            warnings.warn('Deprecated and will be removed. Please use `include_filenames` instead')
+            include_filenames = [filename]
+
+        if exclude_filename:
```
zm711, line 76:
```suggestion
        if exclude_filename is not None:
```

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
-        if dirname != "":
-            self.dirname = dirname
-            self.rawmode = "one-dir"
-        elif filename != "":
-            self.filename = filename
-            self.rawmode = "one-file"
+        if not dirname:
+            raise ValueError("`dirname` cannot be empty.")
+
+        if filename:
```
zm711, line 152:
```suggestion
        if filename is not None:
```

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+            include_filenames = filename
```
zm711, line 153: filename implies singular so we would expect the user to only give one filename. In this case we would want to convert that single file name to a list.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+            warnings.warn("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+
+        if exclude_filename:
```
zm711, line 156:
```suggestion
        if exclude_filename is not None:
```

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+            exclude_filenames = exclude_filename
+            warnings.warn("`exclude_filename` is deprecated and will be removed. Please use `exclude_filenames` instead")
+
+        if include_filenames is None:
+            include_filenames = []
+        elif not isinstance(include_filenames, (list, set, np.ndarray)):
+            include_filenames = [include_filenames]
```
zm711, line 163: I don't know if we should allow this case. include_filenames should be an Iterable. I think I tuple would be more common than a set. I would probably say just set it equal and then let it error later if the user does it wrong. So I would do

```python
if include_filenames is None:
    include_filenames = []
```

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+
+        if exclude_filenames is None:
+            exclude_filenames = set()
```
zm711, line 166: Why are we using a set? Are you worried that the user will put in the same file twice? That seems like a user error that we don't need to fix.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
-            self.rawmode = "one-file"
+        if not dirname:
+            raise ValueError("`dirname` cannot be empty.")
+
+        if filename:
+            include_filenames = filename
+            warnings.warn("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+
+        if exclude_filename:
+            exclude_filenames = exclude_filename
```
zm711, line 157: I would probably do something like:

```
if isinstance(exclude_filename, str):
    exclude_filenames = [exclude_filename]

File: neo/rawio/neuralynxrawio/neuralynxrawio.py [hunk 14]
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+
+        if exclude_filenames is None:
+            exclude_filenames = set()
+        elif not isinstance(exclude_filenames, (list, set, np.ndarray)):
+            exclude_filenames = {exclude_filenames}
+
+        if include_filenames:
+            include_filepath = {os.path.dirname(f) for f in include_filenames}
+            if len(include_filepath) > 1:
+                raise ValueError("Files in include_filename must be in a single path!")
```
zm711, line 173: This check isn't necessary because we force the user to give us the dirname they want us to look in. We should do that check later.
```suggestion
```

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+
+        if include_filenames:
+            self.rawmode = 'multiple-files'
         else:
-            raise ValueError("One of dirname or filename must be provided.")
+            self.rawmode = "one-dir"
 
+        self.dirname = dirname
+        self.include_filenames = include_filenames
+        self.execlude_filenames = exclude_filenames
```
zm711, line 189:
```suggestion
        self.exclude_filenames = exclude_filenames
```
just a typo :)

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,68 @@ class NeuralynxRawIO(BaseRawIO):
...
+            include_filepath = {os.path.dirname(f) for f in include_filenames}
+            if len(include_filepath) > 1:
+                raise ValueError("Files in include_filename must be in a single path!")
+
+        if dirname and include_filenames:
+            dirname = os.path.join(dirname, os.path.dirname(include_filenames[0]))
+            include_filenames = [os.path.basename(f) for f in include_filenames]
+
+        if exclude_filenames:
+            exclude_filenames = {os.path.join(dirname, os.path.basename(f)) for f in exclude_filenames}
```
zm711, line 180:
```suggestion
```

Also probably not necessary we can do this check later.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -183,33 +222,24 @@ def _parse_header(self):
         event_annotations = []
 
         if self.rawmode == "one-dir":
-            filenames = sorted(os.listdir(self.dirname))
-            dirname = self.dirname
+            filenames = sorted([os.path.join(self.dirname, f) for f in os.listdir(self.dirname)
```
zm711, line 225: This has become very complicated. I would actually keep the original logic and just use `listdir` here.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -183,33 +222,24 @@ def _parse_header(self):
...
+                                if os.path.isfile(os.path.join(self.dirname, f)) and not f.startswith('.')])
         else:
-            if not os.path.isfile(self.filename):
+            filenames = [os.path.join(self.dirname, f) for f in self.include_filenames]
+
+        filenames = [f for f in filenames if f not in self.execlude_filenames]
```
zm711, line 230:
```suggestion
        filenames = [f for f in filenames if f not in self.exclude_filenames]
```
typo fix.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -183,33 +222,24 @@ def _parse_header(self):
...
-        # remove files that were explicitly excluded
-        if self.exclude_filename is not None:
-            for excl_file in self.exclude_filename:
-                if excl_file in filenames:
-                    filenames.remove(excl_file)
-
         stream_props = {}  # {(sampling_rate, n_samples, t_start): {stream_id: [filenames]}
 
         for filename in filenames:
-            filename = os.path.join(dirname, filename)
```
zm711, line 211: Bring this line back into the code. Just to keep it more readable.

zm711 (commented):
Another round of comments @NxNiki. 

It is very close. I think we would want to simplify the logic a bit in a few places where I indicated. You want to try those? and then I'll do another read. Just ping me again when you're ready.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: This is updated. But as `include_filenames` supports both single and multiple files. So it should make not difference.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: This is minor, but a set is faster to check if an element is in it than a list. Do you prefer to use list here?

neo/rawio/neuralynxrawio/neuralynxrawio.py
zm711: I'm wondering if we should make `include_filenames` be an iterable. Such that filename needs to be converted. But I could be convinced otherwise. I'm of the school of thought that we should be more limited in what we accept so that we can reduce our checks on functions, but others on the team like being as accommodating as possible so we could accept both. Let's think about this a bit.

neo/rawio/neuralynxrawio/neuralynxrawio.py
zm711: Are you sure? At what scale?
```python
%timeit "a" in {"a", "b", "c"}
23.5 ns ± 0.245 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

%timeit "a" in ["a", "b", "c"]
20.3 ns ± 0.201 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

NxNiki: In Python, a set is implemented with a hashtable, whereas a list is a dynamic array. The time complexity of membership testing should be O(1) and O(n), respectively. The difference is negligible when the list size is small (as in your test above). Here is a related post on StackOverflow: https://stackoverflow.com/questions/8929284/what-makes-sets-faster-than-lists.

I am open to both list and set as we may not have a huge number of files to read.

NxNiki:
Hi @zm711, thanks for giving valuable suggestions. The PR is ready for another round of review.

neo/rawio/neuralynxrawio/neuralynxrawio.py
zm711: I think that is a good point for speed in the case of a large list of files. I'm not sure of the limit for neuralynx files in a folder but I would guess that it is <10 or 15 files (correct me if I'm wrong). Not to mention even if I increase the size of items to 7 for set it was 30 ns for each item and for list the last item took 70 ns (and 20ns for the first) so for our purpose of running this once it is pretty negligible. I like a list better unless I'm explicitly forcing items to only be included once and hopefully if someone really wants to exclude a lot of items they instead switch to include instead (which you proposed ! )

zm711:
I'll give it another full read this weekend!

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -128,32 +131,59 @@ class NeuralynxRawIO(BaseRawIO):
...
+            if isinstance(exclude_filename, str):
+                exclude_filenames = [exclude_filename]
+            else:
+                exclude_filenames = exclude_filename
+            warnings.warn("`exclude_filename` is deprecated and will be removed. Please use `exclude_filenames` instead")
+
+        if include_filenames is None:
+            include_filenames = []
+        elif isinstance(include_filenames, str):
+            include_filenames = [include_filenames]
```
zm711, line 166: This is the one I'm still a little torn on. I would prevent to force include_filenames to be a list, but I think it is okay if we accept a string in the case of only wanting one file.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -183,33 +213,25 @@ def _parse_header(self):
...
-                if excl_file in filenames:
-                    filenames.remove(excl_file)
-
         stream_props = {}  # {(sampling_rate, n_samples, t_start): {stream_id: [filenames]}
 
-        for filename in filenames:
-            filename = os.path.join(dirname, filename)
-
+        for filename in full_filenames:
+            filename = os.path.join(self.dirname, filename)
```
zm711, line 234:
```suggestion
        for filename in full_filenames:
```

Since you did this in L221 you shouldn't need to do this again right?

zm711 (commented):
One last suggestion and one last discussion point, but then I think we are basically there.

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: Yes indeed! Thanks for pointing this out!

neo/rawio/neuralynxrawio/neuralynxrawio.py
NxNiki: Under the hood, `include_filenames` has to be a list, or the iteration at line 220 would loop over each string character.

zm711: Sorry I wasn't clear. I was saying why should we accept a string at all (so delete line 165 and 166). Since our new argument is filename`s` shouldn't we just always make the users give a list (or any iterable)? Why should we accept a single file rather than a list of a single file?

NxNiki: I asked on Chatgpt and it seems quite common to allow for flexible input types in this case. I am open to both ways though.

Here is the answer from chatgpt:

In Python, when designing an I/O interface to read a list of files, it's generally user-friendly to allow for flexible input types. Here are the considerations for both approaches you mentioned:

Allowing a string or list as input:

Pros: More convenient for users who often deal with a single file. It reduces the overhead of wrapping a filename in a list, making the function easier to use in scripts or interactive sessions where only one file is processed.
Cons: The function implementation needs to handle different types of input, checking whether the input is a string or a list. This can make the function slightly more complex.
Strictly requiring a list of strings:

Pros: Consistency in input type can simplify the function's logic, as it always expects a list. This can make the code simpler and potentially less error-prone when dealing with multiple files.
Cons: Less convenient for cases where only a single file is involved, as the user must wrap the filename in a list even if it's just one.
Given these considerations, a common practice is to design the interface to accept both strings and lists. This approach enhances usability without significantly complicating the implementation.

zm711: @apdavison did you have opinions on this? I prefer to be more restrictive, but I know other parts of Neo's API is relatively permissive. I think you prefer more flexibility too @samuelgarcia, so I'm fine with being outvoted here.

zm711:
@NxNiki, we have a conference that a bunch of us are attending next week, but we will try to get final review done the week after that!

NxNiki:
> @NxNiki, we have a conference that a bunch of us are attending next week, but we will try to get final review done the week after that!

Thanks for the update. I hope you have a nice trip!

zm711:
@apdavison did you want to check out the argument reorganization on this one before we have our meeting?

zm711:
@apdavison, just wanted to ping you again for giving this one a read!

NxNiki:
> @apdavison, just wanted to ping you again for giving this one a read!

Hi, any updates? Let me know if there are additional comments or concerns.

neo/rawio/neuralynxrawio/neuralynxrawio.py
apdavison: I think it's a tough call, but I can't see that accepting a string will cause major problems, so I'm happy to go with the current interface.

apdavison:
Many thanks @NxNiki for all your work on this, and sorry for the delayed response from my side.

NxNiki:
> Many thanks @NxNiki for all your work on this, and sorry for the delayed response from my side.

NP! This is a great learning opportunity for me. Thanks for the discussions and comments!

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -183,33 +213,24 @@ def _parse_header(self):
         event_annotations = []
 
         if self.rawmode == "one-dir":
-            filenames = sorted(os.listdir(self.dirname))
-            dirname = self.dirname
+            filenames = os.listdir(self.dirname)
```
samuelgarcia, line 216: I am realising only now that this little has an bi big impact because the file are unsorted and so channel are not ordered

samuelgarcia: See this https://github.com/SpikeInterface/spikeinterface/issues/3248

zm711: You think this is Anthony's issue? This was just merged yesterday, so I guess we could confirm that the issue showed up, but makes sense that could be the problem. You plan to just open the PR?

NxNiki: Oh, sorry for missing this issue. Do we also want to sort `include_filenames`? One edge case would be the user using `os.listdir` to collect the files and pass it to `include_filenames`.

samuelgarcia: No wory. @Antho2422 did a quick fix.
For include_filenames, I think we do not need the sorter() because it makes that the user can providde a list in the desired order. no ?

PR #1446, opened 2024-03-27
BiocamIO: support sparse event based recordings

mahlzahn:
This PR includes some code extracted from the [3Brain documentation](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/biocam/documentation_brw_4.x_bxr_3.x_bcmp_1.x_in_brainwave_5.x_v1.1.3.pdf) and added with kind permission by 3Brain.

I wrote:
> I am contributing to the inclusion of reading recordings from 3brain
> with python-neo, probeinterface and spikeinterface. May I ask you the
> permission to
> 1. include the example code in python-neo under the BSD 3-Clause License
> https://github.com/NeuralEnsemble/python-neo 
> 2. upload the latest file documentation to
> https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/biocam

and got the answer:
> Regarding your request for permission, yes, please, you can upload your files.


neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
     for key in rf:
         if key[:5] == "Well_":
             return rf[key]["Raw"][nch * t0 : nch * t1].reshape((t1 - t0, nch), order="C")
+
+
+def readHDF5t_brw4_sparse(rf, t0, t1, nch):
+    useSyntheticNoise = True
```
zm711, line 262: I'm not quite sure what the benefit of this is? If it is always true then why do you need the if statement. You're not actually checking anything...

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+    numFrames = t1 - t0
+    for key in rf:
+        if key[:5] == "Well_":
+            wellID = key
+            break
+    # initialize an empty (fill with zeros) data collection
+    data = np.zeros((nch, numFrames), dtype=np.int16)
+    # fill the data collection with Gaussian noise if requested
+    if useSyntheticNoise:
+        generateSyntheticNoise(rf, data, wellID, startFrame, numFrames, stdDev=noiseStdDev)
```
zm711, line 274: I don't know what others think, but I'm personally not a huge fan of mutating the variable in place (even though I do it sometimes myself). I think it is harder to bug.

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+        for i in range(tocStartIdx - 1, 0, -1):
+            previousPosition = noiseToc[i]
+            if previousPosition < noiseStartPosition:
+                noiseEndPosition = noiseStartPosition
+                noiseStartPosition = previousPosition
+                break
+    # obtain the noise info at the start position
+    noiseChIdxs = rf[wellID]["NoiseChIdxs"][noiseStartPosition:noiseEndPosition]
+    noiseMean = rf[wellID]["NoiseMean"][noiseStartPosition:noiseEndPosition]
+    if stdDev is None:
```
zm711, line 346: I don't see this being exposed? If the user will never be allowed to do this why do we need this check?

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+                if j >= startFrame + numFrames:
+                    break
+                if j >= startFrame:
+                    data[chIdx][j - startFrame] = int.from_bytes(
+                            binaryData[rangeDataPos:rangeDataPos + 2], byteorder="little", signed=True)
+                rangeDataPos += 2
+            pos += (toExclusive - fromInclusive) * 2
+
+
+def generateSyntheticNoise(rf, data, wellID, startFrame, numFrames, stdDev=None):
```
zm711, line 320: This seems like a private function?

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+    data = np.zeros((nch, numFrames), dtype=np.int16)
+    # fill the data collection with Gaussian noise if requested
+    if useSyntheticNoise:
+        generateSyntheticNoise(rf, data, wellID, startFrame, numFrames, stdDev=noiseStdDev)
+    # fill the data collection with the decoded event based sparse raw data
+    decodeEventBasedRawData(rf, data, wellID, startFrame, numFrames)
+    return data.T
+
+
+def decodeEventBasedRawData(rf, data, wellID, startFrame, numFrames):
```
zm711, line 280: Also private right? Or is this one public? I would argue that for public functions we should aim for PEP8 compliance (snake_case and not camelCase).

neo/rawio/biocamrawio.py
```diff
@@ -197,25 +197,32 @@ def open_biocam_file_header(filename):
...
+                if "Raw" in rf[key]:
+                    if len(rf[key]["Raw"]) % num_channels:
+                        raise RuntimeError(f"Length of raw data array is not multiple of channel number in {key}")
+                    if num_frames != len(rf[key]["Raw"]) // num_channels:
+                        raise RuntimeError(f"Estimated number of frames from TOC does not match length of raw data array in {key}")
                 break
-        try:
-            num_channels_x = num_channels_y = int(np.sqrt(num_channels))
-        except NameError:
+        if not wellID:
```
zm711, line 213: In most of this code you just use `key` why do you need to set `wellID=key`? To be honest if `wellID` is the better name I would prefer just to say for wellID in rf and then just use wellID wherever key was appearing.

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+    meanCollection = []
+    stdDevCollection = []
+    for i in range(1, noiseLength):
+        noiseInfo[noiseChIdxs[i]] = [noiseMean[i], noiseStdDev[i]]
+        meanCollection.append(noiseMean[i])
+        stdDevCollection.append(noiseStdDev[i])
+    # calculate the median mean and standard deviation of all channels to be used for
+    # invalid channels
+    dataMean = np.median(meanCollection)
+    dataStdDev = np.median(stdDevCollection)
```
zm711, line 361: This is a bit hard to parse (for the future-- the median of the mean) why not `median_data_mean` so that in the future we don't do the double take? I'm assuming that's just how the code you're porting was written. So this is less important. Just made me do a quick doubletake.

zm711 (commented):
Thanks @mahlzahn. Just a few questions/comments before Alessio does the final read through.

neo/rawio/biocamrawio.py
```diff
@@ -197,25 +197,32 @@ def open_biocam_file_header(filename):
...
             raise RuntimeError(f"Cannot determine structure of the MEA plate with {num_channels} channels")
         channels = 1 + np.concatenate(np.transpose(np.meshgrid(range(num_channels_x), range(num_channels_y))))
 
         gain = scale_factor * (max_uv - min_uv) / (max_digital - min_digital)
         offset = min_uv
-        read_function = readHDF5t_brw4
+        if "Raw" in rf[wellID]:
+            read_function = readHDF5t_brw4
+        elif "EventsBasedSparseRaw" in rf[wellID]:
+            read_function = readHDF5t_brw4_sparse
```
alejoe91, line 225: We should probably add a warning here to warn that this is not good practice to fill in the gaps with random synthetic noise.

neo/rawio/biocamrawio.py
```diff
@@ -249,3 +256,115 @@ def readHDF5t_brw4(rf, t0, t1, nch):
...
+        generateSyntheticNoise(rf, data, wellID, startFrame, numFrames, stdDev=noiseStdDev)
+    # fill the data collection with the decoded event based sparse raw data
+    decodeEventBasedRawData(rf, data, wellID, startFrame, numFrames)
+    return data.T
+
+
+def decodeEventBasedRawData(rf, data, wellID, startFrame, numFrames):
+    # Source: Documentation by 3Brain
+    # https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/biocam/documentation_brw_4.x_bxr_3.x_bcmp_1.x_in_brainwave_5.x_v1.1.3.pdf
+    # collect the TOCs
```
alejoe91, line 283: @mahlzahn 

Although this is in the 3Brain docs, we think that the coould really use a more pythonic re-write! Would you have time to give it a go?

neo/rawio/biocamrawio.py
alejoe91: the `use_synthetic_noise` (default `False`) should be exposed as an argument

alejoe91 (changes requested):
Thansk @mahlzahn 

See comments!

neo/rawio/biocamrawio.py
BptGrm: 3Brain's code example is meant as a simple demo script, not to be used as an API, which is why a lot is done explicitely like this. I'm not sure how far up the chain this argument would need to go, I haven't personnaly used a reader that had such options before.

neo/rawio/biocamrawio.py
BptGrm: Couldn't non-filled data interfere with spike detection algorithms based on peak detection ? Since 3Brain data is stored unsigned, filling the gaps with zeros means they will be read as -2048 µV instead of being at the baseline.

neo/rawio/biocamrawio.py
BptGrm: I agree, especially with a name as generic as `data` floating around in the namespace.
From my own fork ([for reference](https://github.com/NeuralEnsemble/python-neo/compare/master...BptGrm:python-neo:master#diff-59e99a2d55ebdd8b4501b6cbc598b3f22bd4e7c3ad64d13172ee93f7edf0c8d1R312-R316)) : 
```
    if synth:
        arr = GenerateSyntheticNoise_arr(rf, wellID, t0, t1, numFrames, nch)

    else:
        arr = np.zeros((numFrames, nch))
```

neo/rawio/biocamrawio.py
BptGrm: For the sake of readability on my own, I declared both this and the main parsing function at the base level of the file then wrote a bare minimum function like the others before : 

```
def readHDF5t_brw4_event_based(rf, t0, t1, nch):
    for key in rf:
        if key[:5] == "Well_":
            return DecodeEventBasedRawData_arr(rf, key, t0, t1, nch)
```

neo/rawio/biocamrawio.py
BptGrm: Also curious about this, I can't find it in the original 3Brain docs.

neo/rawio/biocamrawio.py
zm711: I think the best solution would be to fill with a value that when scaled returns 0 volts/ microvolts. So I think this is a good point that we should keep in mind.

neo/rawio/biocamrawio.py
zm711: I agree that I like something like this better (although see note for zeros below as well).

neo/rawio/biocamrawio.py
zm711: From a neo perspective especially for something like this we would want the user to have a say (although like Alessio said a default of False should be used because Neo is not really meant to create data our goal is really just to supply the underlying data).

zm711:
@mahlzahn, just wanted to ping you to see if you have the bandwidth to work on our comments?

b-grimaud:
Since I can't directly contribute to the PR, I forked it and make some updates [here](https://github.com/b-grimaud/python-neo/tree/sparse). Opening a separate PR might be a bit messy ?

The main improvement is exposing a `use_synthetic_noise` argument, as well as `true_zeroes`. The latter, when `False`, makes sure that zeroes are actually read as zeroes once the recording is converted from unsigned to signed.

Regarding some of the other comments, I'm still not sure what `if stdDev is None:` ([here](https://github.com/mahlzahn/python-neo/blob/cc6ae8d58c123d552578b00095485acb80774655/neo/rawio/biocamrawio.py#L346)) is meant for.

Also, if we indexed the data as `[j - startFrame, chIdx]` instead of `[chIdx][j - startFrame]` and intialized the array as `np.zeros((numFrames, nch))`, I don't think we would need to transpose when returning the filled array ? Not sure how much overhead it actually adds, the bit-by-bit parsing is slow enough as is.

zm711:
I'm honestly neutral on keeping with this PR or opening a new one. If @mahlzahn doesn't have the bandwidth then a new PR that will be worked on is an okay number. Do you want to comment on if you'll have time to work on this soon @mahlzahn or whether @b-grimaud should take over?

mahlzahn:
Hi all, sorry for my silence. I indeed didn’t have much time recently. I now added @b-grimaud as collaborator to my python-neo repository. Maybe you want to push your work to that branch?

zm711:
Thanks all, whatever works the best for you two! Life gets busy we totally understand! We will leave this on the top of the review pile for now and if you both get too busy then we can retag for future review.

b-grimaud:
Thanks for the update ! It looks like my [initial commit](https://github.com/NeuralEnsemble/python-neo/pull/1446/commits/1d84c9f74594b1486ccc424d62be8d557547fff0) was retroactively included in the thread already.

neo/rawio/biocamrawio.py
zm711: @b-grimaud any interest in helping re-write this portion in a pythonic way?

neo/rawio/biocamrawio.py
mahlzahn: I agree that default should return something scaled to 0 volts. But I suggest to change it from the current value of 2048 to something like the mean of the known reads. And in particular, value 2048 fails for channels used as trackers (usually channel 1-1 for calibration events) because this one is only zeroes and sometimes 4095 when there is a an event.

b-grimaud: I'm not exactly sure how the 3Brain compression algorithm handles the calibration channel or if it's treated differently, but in the current configuration and with default parameters channel 1-1 reads as something like `[2048],[2048],[2048],[2048],[4095],[4095],[4095],[2048],[2048],[2048],[2048]`, and once signed as `[   0],[   0],[   0],[   0],[2047],[2047],[2047],[   0],[   0],[   0],[   0],`. I don't know if any pipelines retain that channel downstream for spike sorting, but even if they do it's scaled to the rest of the channels.

Seeing as the 12 bits signed conversion effectively applies a flat -2048 to everything, I think having something recording-independent might be more reliable.

hornauerp:
I just stumbled across this PR, as I wanted to run Kilosort on data from collaborators recorded on 6 well 3Brain plates. I have never worked with their chips, so I am not familiar with the file structure or the relevant neo io functions. 

Not sure if anyone is still working on this, but using the code from this PR the `read_biocam` function works just fine and displays the correct number of channels and duration, but when I try to actually retrieve the traces, I get an out of bounds error: `IndexError: index 1040 is out of bounds for axis 0 with size 1024`. 

Also, in general, I was a bit confused why reading the file only gives one recording, although all six wells were recorded? After a quick look at the code it seems to iterate over all wellIDs but then only stores information for one. Am I missing something or should the wellID be passed as an argument?

b-grimaud:
@hornauerp I think that so far everything was built with single-well MEAs in mind. Since they added 6-wells (and now 96-wells, I believe) MEAs, 3Brain changed the inner structure of their file format so that recordings on standard chips are compatible. That's why the read function is simply iterating over `wellID`.

I don't know if there is a standard for multi-MEA recordings. It would probably help to deposit an exemple of 6-well data on GIN so that the read function can be properly tested.

zm711:
Absolutely agree with @b-grimaud here. the only way we can adapt is with test data. So if you'd like to get involved you could make a PR with the fix and provide data or you could share a small test file (we need < 10MB for bandwidth reasons) that we can put into our CI. 

That being said our main reviewer for biocam is out of the office at the moment, so our work on biocam will be a bit slower unless you or @b-grimaud wants to give it a stab at update the IO/RawIO.

zm711:
I've switched this to 0.15.0 for now. If no one has time for these fixes, I can move the milestone to future. Just let us know :)

b-grimaud:
@zm711 I think it depends on the scope of the PR. I'll need to check but as far as I know compressed data from single well MEAs can be read with the code as is. I think some people had comments on it but I'm not quite sure what needs to be changed. 

As far as the multiwell issue goes, this would affect both compressed and raw data (unless BrainWave doesn't allow raw recordings for multiwell plates) and would probably fit in another PR.

zm711:
I think the scope of the changes are basically 1) a cosmetic rewrite. We are all python people so we were hoping to push forward with as much PEP8 as possible (except line lengths where we prefer 120). We use to have a required lint for PEP8, but now we just enforce it ourselves.  So we would like this to be written with PEP8. 2) There a couple things that allow we understand this is from the company's implementation we would prefer to expose to the user (for example that hardcode  booleans and std stuff should either not be there or should be controlled by users. 

So the work to finish this PR will be quite a bit actually.

b-grimaud:
I switched the camel case from 3Brain's original example to snake case, with the only exception being HDF5 keys that obviously have to stay as they are.

Two extra arguments are exposed as described [above](https://github.com/NeuralEnsemble/python-neo/pull/1446#issuecomment-2158930345).

For now I've disabled the custom noise argument, which was originally added by @mahlzahn. I guess the intent would be for the user to specify a custom standard deviation value for the synthetic noise ?

zm711:
Thanks @b-grimaud. Do you have a moment to fix the conflicts. I glanced at them, but it would take me a moment to doublecheck and fix them appropriately. Then we can run an actual round of the test suite. :) Then we can review.

b-grimaud:
I think there might be an issue with write access to the repo ? I can see the conflicts in the browser but can't change anything, and I can't see them or fix them in VS Code or GitHub desktop.

alejoe91:
@b-grimaud you have to pull upstream from command line:
```
git remote add upstream https://github.com/NeuralEnsemble/python-neo.git
git pull upstream master
```

Then you should see the conflicts locally, you should resolve them, commit and push again :)

b-grimaud:
@alejoe91 Thought I had the remote set up correctly, turns out I didn't, thanks !

The initial conflicts are solved for now, but something else then came up  : checking that the size of the array matches the amount of channels doesn't work with sparse data.

Their compression algorithm works both spatially and temporally, i.e. not all chunks contain data from every channel and every time point. We could theoretically retrieve the number of channels for each chunk and then compare but since everything is stored in indivual headers for each channel we'd essentially be reading the file at this point. 

Especially since it's closed source, I think it would be difficult to implement ahead of the actual read function.

zm711:
Great we will try to review this soon but at least glancing over I appreciate the pythonic rewrite :). Let's see what happens with testing.

zm711:
@alejoe91 did you want to give this a read this week or are you too busy?

alejoe91 (approved):

neo/rawio/biocamrawio.py
```diff
@@ -302,5 +325,120 @@ def readHDF5t_101_i(rf, t0, t1, nch):
...
+    start_frame = t0
+    num_frames = t1 - t0
+    for well_ID in rf:
+        if well_ID.startswith("Well_"):
+            break
+    # initialize an empty (fill with zeros) data collection
+    data = np.zeros((nch, num_frames), dtype=np.int16)
+    if not true_zeroes:
+        # Will read as 0s after 12 bits signed conversion
+        data.fill(2048)
```
alejoe91, line 344: Should this be the default?

b-grimaud: We talked about it [here](https://github.com/NeuralEnsemble/python-neo/pull/1446#discussion_r1572718189) : the idea was that, after unsigned-to-signed conversion, the "filler" data would read as 0.

I guess whether or not it should be the default depends on whether we consider the unsigned-to-signed conversion necessary. I don't think it is explicitely specified anywhere in the docs, but in order to get the back the raw signal as it was recorded we have to go through this step.

b-grimaud: NaNs make sense, but I'm not sure how sorters would react to non-contiguous data

alejoe91: NaN is probably the best. Maybe the best option would be have a `fill_nans` preprocessor in spikeinterface. Does that sound good?

neo/rawio/biocamrawio.py
```diff
@@ -302,5 +325,120 @@ def readHDF5t_101_i(rf, t0, t1, nch):
...
+def readHDF5t_brw4_sparse(rf, t0, t1, nch, true_zeroes=False, use_synthetic_noise=False):
+
+    # noise_std = None
+    start_frame = t0
+    num_frames = t1 - t0
+    for well_ID in rf:
+        if well_ID.startswith("Well_"):
+            break
+    # initialize an empty (fill with zeros) data collection
+    data = np.zeros((nch, num_frames), dtype=np.int16)
```
alejoe91, line 341: @mahlzahn @b-grimaud another quick question: shouldn't this be `unit16`? And the `int.from_bytes` should be `signed=False`?

neo/rawio/biocamrawio.py
alejoe91: actually, the problem with NaNs is that it cannot be int16...

neo/rawio/biocamrawio.py
alejoe91: Yes, because the stream is defined as uint..but I think it's inconsistent and it might lead to problema later

alejoe91 (approved):

alejoe91:
Thanks for working on this @mahlzahn @b-grimaud 

Ready to merge on my end @zm711

Issue #1401, opened 2024-02-13
TDT read issue when 'StoreName' key is missing from header


weiglszonja:
**Describe the bug**

I'm having issues with reading TDT files (not sure where to check the TDT version, but these files are from 2014), where 
'StoreName' key is missing from one of the entries in `info`:

https://github.com/NeuralEnsemble/python-neo/blob/c96ec936049a125d6bedfa26b67514f927bb1ab3/neo/rawio/tdtrawio.py#L553

Here is a snippet how the `infos` look like for one the problematic files:

```
[
    OrderedDict({'NumPoints': b'32', 'DataFormat': b'0', 'SampleFreq': b'24414.062500'}), 
    ...
    OrderedDict({'StoreName': b'Conx', 'HeadName': b'Conx', 'Enabled': b'1', 'CircType': b'5', 'NumChan': b'20', 'StrobeMode': b'0', 'TankEvType': b'33025', 'NumPoints': b'2048', 'DataFormat': b'2', 'SampleFreq': b'24414.062500'}), 
    ...
    OrderedDict({'StoreName': b'LFPx', 'HeadName': b'LFPx', 'Enabled': b'1', 'CircType': b'5', 'NumChan': b'4', 'StrobeMode': b'0', 'TankEvType': b'33025', 'NumPoints': b'64', 'DataFormat': b'0', 'SampleFreq': b'1017.252625'})]
]
```
I'm not sure what signal the first entry should correspond to, the neural signals have the 'Conx' stream name.
Is there a way to fix this?

**To Reproduce**
I can share one of these files privately.

**Additional context**

```pytb
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevd.py", line 1534, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/weian/catalystneuro/turner-lab-to-nwb/check_datalist.py", line 11, in <module>
    stream_names, stream_ids = TdtRecordingExtractor.get_streams(folder_path=problematic_files[0])
[...]

PR #1447, opened 2024-03-27
Fix reading TDT files where Tbk header contains incomplete channel group info

weiglszonja:
Fix https://github.com/NeuralEnsemble/python-neo/issues/1401

For some TDT files (from 2014 as pointed out in the issue), the channel groups info that is extracted from the Tbk file doesn't contain all expected field names. I think we should allow reading these files (they are currently not readable as they fail at `read_tbk()`) but should issue a warning.

### DIFF
--- a/neo/rawio/tdtrawio.py
+++ b/neo/rawio/tdtrawio.py
@@ -557,8 +557,16 @@ def read_tbk(tbk_filename):
         infos.append(info)
 
     # and put into numpy
+    tbk_field_names = [tbk_field_type[0] for tbk_field_type in tbk_field_types]
     info_channel_groups = np.zeros(len(infos), dtype=tbk_field_types)
     for i, info in enumerate(infos):
+        missing_keys = set(tbk_field_names) - set(info.keys())
+        if missing_keys:
+            warnings.warn(
+                f"Could not find all channel group info in tbk file, the missing keys are {missing_keys}.\n"
+                "The reader will continue with channel groups that could be parsed."
+            )
+            continue
         for k, dt in tbk_field_types:
             v = np.dtype(dt).type(info[k])
             info_channel_groups[i][k] = v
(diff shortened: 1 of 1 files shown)


zm711 (commented):
If we are going to warn about missing keys is there a reason not to also include which `info` is being skipped? Is it a super large structure? Do I as the user care about the `missing_keys` or do I care more about the channel group that is being skipped `info`? (I also don't know tdt, but based upon the issue it seems like the missing_keys aren't super helpful info...)

weiglszonja:
> If we are going to warn about missing keys is there a reason not to also include which `info` is being skipped? Is it a super large structure? Do I as the user care about the `missing_keys` or do I care more about the channel group that is being skipped `info`? (I also don't know tdt, but based upon the issue it seems like the missing_keys aren't super helpful info...)

Thank you for the feedback. I agree the group itself is more important, how about something like:
```
/Users/weian/catalystneuro/python-neo/neo/rawio/tdtrawio.py:565: UserWarning: The tbk file contains incomplete channel group info for group [('TankEvType', b'257'), ('NumPoints', b'0'), ('SampleFreq', b'0')]. This channel group will be skipped.
```

zm711:
Yeah I think for this being more verbose is better. Thanks. As someone who has used tdt would seeing that warning be useful to you?

weiglszonja:
Yeah I think for this being more verbose is better. Thanks. As someone who has used tdt would seeing that warning be useful to you?

Yes; unfortunately I have no information why this group is missing those fields, but allowing to read these files and checking the data is very useful.

zm711:
Just noting here that I think this is okay short-term solution, but long-term we may want to see if we still parse the file even if it is missing a subsection in the `info`. What do you guys think @samuelgarcia @alejoe91?

alejoe91:
Yeah let's see if we get more issues with this. Fine to merge this for now!

Issue #1441, opened 2024-03-20
Read channel gain from .rec file in SpikeGadgetsRawIO.


gkBCCN:
In the doc string for this class, it says:
```
Notes
-----
This file format has multiple version. New versions include the gain for scaling.
The current implementation does not contain this feature because we don't have
files to test this. So now the gain is "hardcoded" to 1, and so units are
not handled correctly.
```
It should be possible to read these from the header and a .rec file was recently added to test this.

zm711:
Also, any interest in opening the PR for this?

gkBCCN:
Do you mean once I have a fix?  Or preemptively?

zm711:
Once you have a fix. One of us will work on this when we have time, but if you have experience with the files/format it would also be great if you wanted to try to make the PR yourself. Up to you :)

gkBCCN:
Gladly.  I'm also making changes to the probeinterface and spikeinterface packages in order to support SpikeGadgets .rec files, so maybe it makes sense for me to do the same for the neo extractor.

PR #1448, opened 2024-03-28
Read gain scaling from .rec header.

gkBCCN:
Updates spikegadgetsrawio.py:_parse_header() to read the channel (ntrode) gain scaling from the header file elements.

### DIFF
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -55,10 +55,10 @@ class SpikeGadgetsRawIO(BaseRawIO):
 
         Notes
         -----
-        This file format has multiple version. New versions include the gain for scaling.
-        The current implementation does not contain this feature because we don't have
-        files to test this. So now the gain is "hardcoded" to 1, and so units are
-        not handled correctly.
+        This file format has multiple versions:
+            - Newer versions include the gain for scaling to microvolts [uV].
+            - If the scaling is not found in the header, it will be "hardcoded" to 1,
+              in which case the units are not handled correctly.
 
         Examples
         --------
@@ -173,12 +173,17 @@ class SpikeGadgetsRawIO(BaseRawIO):
 
             chan_ind = 0
             for trode in sconf:
+                if "spikeScalingToUv" in trode.attrib:
+                    gain = float(trode.attrib["spikeScalingToUv"])
+                    units = "uV"
+                else:
+                    gain = 1  # revert to hardcoded gain
+                    units = ""
+
                 for schan in trode:
                     name = "trode" + trode.attrib["id"] + "chan" + schan.attrib["hwChan"]
                     chan_id = schan.attrib["hwChan"]
-                    # TODO LATER : handle gain correctly according the file version
-                    units = ""
-                    gain = 1.0
+
                     offset = 0.0
                     signal_channels.append(
                         (name, chan_id, self._sampling_rate, "int16", units, gain, offset, stream_id)
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -57,7 +57,7 @@ class SpikeGadgetsRawIO(BaseRawIO):
         -----
         This file format has multiple versions:
             - Newer versions include the gain for scaling to microvolts [uV].
-            - If the scaling is not found in the header, it will be "hardcoded" to 1,
+            - If the scaling is not found in the header, the gain will be "hardcoded" to 1,
               in which case the units are not handled correctly.
 
         Examples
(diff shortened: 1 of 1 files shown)


alejoe91:
Fixes #1441

@gkBCCN if you write "fixes" + issue number it will close it automatically as the PR is merged ;) Better this way than in the PR title

neo/rawio/spikegadgetsrawio.py
```diff
@@ -55,10 +55,10 @@ def __init__(self, filename="", selected_streams=None):
...
         Notes
         -----
-        This file format has multiple version. New versions include the gain for scaling.
-        The current implementation does not contain this feature because we don't have
-        files to test this. So now the gain is "hardcoded" to 1, and so units are
-        not handled correctly.
+        This file format has multiple versions:
+            - Newer versions include the gain for scaling to microvolts [uV].
+            - If the scaling is not found in the header, it will be "hardcoded" to 1,
+              in which case the units are not handled correctly.
```
zm711, line 61: I'm wondering while we are fixing this if we want to make this message a bit clear. For newer users of Neo it might not click that `get_analogsignal_chunk` will still work fine, but that the units would not be accurate for `rescale_signal_raw_to_float`? What do you think @alejoe91?

neo/rawio/spikegadgetsrawio.py
```diff
@@ -55,10 +55,10 @@ def __init__(self, filename="", selected_streams=None):
 
         Notes
         -----
-        This file format has multiple version. New versions include the gain for scaling.
-        The current implementation does not contain this feature because we don't have
-        files to test this. So now the gain is "hardcoded" to 1, and so units are
-        not handled correctly.
+        This file format has multiple versions:
+            - Newer versions include the gain for scaling to microvolts [uV].
+            - If the scaling is not found in the header, it will be "hardcoded" to 1,
```
zm711, line 60:
```suggestion
            - If the scaling is not found in the header, the gain will be "hardcoded" to 1,
```
Let's be super explicit here about what is hard coded to 1 :)

zm711 (commented):
Let's think about how to update the notes for this section really quick to try to make this as clear as possible for the end-users.

neo/rawio/spikegadgetsrawio.py
zm711: Or again we could think about a warning since the end-user might not know whether or not their file has the scaling. For example how would they know whether or not their gain is correct?

gkBCCN: I agree that the user may not understand the implications based on this description;  I followed the language in the existing docstring.  I'm not so sure that this change will help much, though.  Actually your comment about `get_analogsignal_chunk` and `rescale_signal_raw_to_float` is more helpful.  How about adding that comment in a sentence after the bullet points?

gkBCCN: A warning that the data has not been scaled (and therefore has no units) is also a good idea!

zm711: My other idea would be that we add an attribute like `self.is_scaleable` and then in the note we say check this attribute to know if the data has units or not. I'm not sure how SpikeInterface determines whether an extractor has units or not so we can wait for Alessio to weigh in to make sure this is smoothest for users just using neo but also people using spikeinterface.

alejoe91: A warning would be great! And also an attribute like @zm711 suggested please

gkBCCN: Thanks for the feedback @zm711  and @alejoe91. 
So what I should add is:
- note in the docstring about functions that will (`get_analogsignal_chunk`) / will not (`rescale_signal_raw_to_float`) work
- warning is raised in the block if "spikeScalingToUv" is not found and set self.is_scaleable to False (otherwise set to True)

Is there a specific format for raising warnings in Neo?  Is the attribute self.is_scaleable uniform across these rawio readers?

zm711: > Is there a specific format for raising warnings in Neo?

No. It's been a bit inconsistent. originally we were using a logger, but for me I'm fine with just `warnings.warn`. If @alejoe91 feels strongly for something else though he can definitely comment.

> Is the attribute self.is_scaleable uniform across these rawio readers?

No. It would just be for this rawio (in fact we would avoid it for others). The ideal solution would be to know the gain in all files, but if that is really not possible we need a workaround. So `is_scaleable` is a workaround that we will just patch in spikeinterface for the extractor for this io-only.

> note in the docstring about functions that will (get_analogsignal_chunk) / will not (rescale_signal_raw_to_float) work

To be precise it will work but it will not actually scale. So values will not become float voltages. (ie function should still work but shouldn't be used to get "real values".

gkBCCN: Great, @zm711. I implemented the above to the best of my understanding.

alejoe91:
@zm711 @gkBCCN looks good to me!

gkBCCN:
@zm711 @alejoe91 Thank you!

PR #1456, opened 2024-04-05
Sphinx-Gallery Touchups

zm711:
We were missing igor2 for dependencies in our docs. This PR adds it and will let me test the action.


zm711:
We were missing igor2 for the docs so this PR was just suppose to add it, but now it is saying it can't find a dependency. Any ideas?

zm711:
Figured it out. When our conda env cache was invalidated by a dependency change it was using the `environment_testing.yaml` which was missing python as a dependency. So when it was doing the `--prune` it was installing python 3.12 instead of 3.9. We will need to update that when we change testing for IO testing.

zm711:
Just to make this clear:

1) We were missing igor2 from the docs section of the pyproject.toml. *I added it*
2) When our env-cache becomes invalidated and is regenerates it adds in the latest python rather than oldest which we use for io testing. *I added python 3.9 to the yaml*
3) sphinx-gallery display output on top of the tutorial unless appropriate sections are included in the tutorial *I added section headers and now it displays correctly*

zm711:
@alejoe91-- if you have the bandwidth to give this a quick look over that'd be great (if not, that's cool too). I just don't want everyone's PRs to have failed actions.

alejoe91 (approved):

zm711:
Thanks Alessio :)

PR #1457, opened 2024-04-06
IntanRawIO: Update support of `rhs` files

zm711:
This will be adding support for `RHX` rhs files with a separate header for `rhs` files. We haven't officially update RHS support since RHS 1.0 and we are on 3.x so this may take a bit of work. We will add tests files soon so that we can properly test this.

TODOS
* [x] ~~Add support for supply and aux channels in RHS~~ Turns out that there isn't any from newest docs!
* [x] Add support for `one-file-per-signal`
* [x] Add support for `one-file-per-channel`
* [x] Add test files for new formats
* [x] Add access to digital channels like we did for rhd
* [x]  Try to understand the stim channels to see if we need to make updates
* [x] Add channel_number_dict for one-file-per-signal
* [x] Add analog-out in the one-file-per-channel-dict function

Neuroconv provenance https://github.com/catalystneuro/neuroconv/issues/789

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -112,7 +112,7 @@ class IntanRawIO(BaseRawIO):
                 header_size,
                 self._block_size,
                 channel_number_dict,
-            ) = read_rhs(self.filename)
+            ) = read_rhs(self.filename, self.file_format)
 
 
         # 3 possibilities for rhd files, one combines the header and the data in the same file with suffix `rhd` while
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -481,8 +481,8 @@ def read_rhs(filename, file_format: str):
         channels_by_type = {k: [] for k in [0, 3, 4, 5, 6]}
         if not file_format == "header-attached":
             # data_dtype for rhs is complicated. There is not 1, 2 (supply and aux),
-            # but there are dc-amp (10) and stim (11). we make timestamps (7)
-            data_dtype = {k: [] for k in [0, 3, 4, 5, 6, 7, 10, 11]}
+            # but there are dc-amp (10) and stim (11). we make timestamps (15)
+            data_dtype = {k: [] for k in [0, 3, 4, 5, 6, 10, 11, 15]}
         for g in range(global_info["nb_signal_group"]):
             group_info = read_variable_header(f, rhs_signal_group_header)
 
@@ -568,7 +568,7 @@ def read_rhs(filename, file_format: str):
             if file_format == "header-attached":
                 data_dtype += [(name + "_STIM", "uint16", BLOCK_SIZE)]
             else:
-                data_dtype[11] == "uint16"
+                data_dtype[11] = "uint16"
     else:
         warnings.warn("Stim not implemented for `one-file-per-channel` due to lack of test files")
 
@@ -952,7 +952,7 @@ one_file_per_signal_filenames_rhs = [
     "supply.dat",
     "analogin.dat",
     "analogout.dat",
-     "digitalin.dat",
+    "digitalin.dat",
     "digitalout.dat",
 ]
 
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -98,12 +98,12 @@ class IntanRawIO(BaseRawIO):
 
         if self.filename.endswith(".rhs"):
             if filename.name == "info.rhs":
-                if any((filename.parent / file).exists() for file in one_file_per_signal_filenames):
+                if any((filename.parent / file).exists() for file in one_file_per_signal_filenames_rhs):
                     self.file_format = "one-file-per-signal"
-                    raw_file_paths_dict = create_one_file_per_signal_dict(dirname=filename.parent, rhs=True)
+                    raw_file_paths_dict = create_one_file_per_signal_dict_rhs(dirname=filename.parent)
                 else:
                     self.file_format = "one-file-per-channel"
-                    raw_file_paths_dict = create_one_file_per_channel_dict(dirname=filename.parent, rhs=True)
+                    raw_file_paths_dict = create_one_file_per_channel_dict_rhs(dirname=filename.parent)
             else:
                 self.file_format = "header-attached"
 
@@ -122,13 +122,13 @@ class IntanRawIO(BaseRawIO):
         elif self.filename.endswith(".rhd"):
             if filename.name == "info.rhd":
                 # first we have one-file-per-signal which is where one neo stream/file is saved as .dat files
-                if any((filename.parent / file).exists() for file in one_file_per_signal_filenames):
+                if any((filename.parent / file).exists() for file in one_file_per_signal_filenames_rhd):
                     self.file_format = "one-file-per-signal"
-                    raw_file_paths_dict = create_one_file_per_signal_dict(dirname=filename.parent)
+                    raw_file_paths_dict = create_one_file_per_signal_dict_rhd(dirname=filename.parent)
                 # then there is one-file-per-channel where each channel in a neo stream is in its own .dat file
                 else:
                     self.file_format = "one-file-per-channel"
-                    raw_file_paths_dict = create_one_file_per_channel_dict(dirname=filename.parent)
+                    raw_file_paths_dict = create_one_file_per_channel_dict_rhd(dirname=filename.parent)
             # finally the format with the header-attached to the binary file as one giant file
             else:
                 self.file_format = "header-attached"
@@ -175,7 +175,8 @@ class IntanRawIO(BaseRawIO):
         # check timestamp continuity
         if self.file_format == "header-attached":
             timestamp = self._raw_data["timestamp"].flatten()
-        # timestamps are always the last stream
+
+        # timestamps are always 6 for rhd and 7 for rhs
         elif self.file_format == "one-file-per-signal":
             time_stream_index = max(self._raw_data.keys())
             timestamp = self._raw_data[time_stream_index]
@@ -213,7 +214,7 @@ class IntanRawIO(BaseRawIO):
         signal_streams = np.zeros(stream_ids.size, dtype=_signal_stream_dtype)
         signal_streams["id"] = stream_ids
         for stream_index, stream_id in enumerate(stream_ids):
-            if self.filename.endswith('.rhd'):
+            if self.filename.endswith(".rhd"):
                 signal_streams["name"][stream_index] = stream_type_to_name_rhd.get(int(stream_id), "")
             else:
                 signal_streams["name"][stream_index] = stream_type_to_name_rhs.get(int(stream_id), "")
@@ -455,14 +456,14 @@ rhs_signal_channel_header = [
     ("electrode_impedance_phase", "float32"),
 ]
 
-stream_type_to_name_rhs= {
+stream_type_to_name_rhs = {
     0: "RHS2000 amplifier channel",
     3: "USB board ADC input channel",
     4: "USB board ADC output channel",
     5: "USB board digital input channel",
     6: "USB board digital output channel",
     10: "DC Amplifier channel",
-    11: "Stim channel"
+    11: "Stim channel",
 }
 
 
@@ -501,8 +502,8 @@ def read_rhs(filename, file_format: str):
     if file_format == "header-attached":
         data_dtype = [("timestamp", "int32", BLOCK_SIZE)]
     else:
-        data_dtype[7] = "int32"
-        channel_number_dict[7] = 1
+        data_dtype[15] = "int32"
+        channel_number_dict[15] = 1
 
     # 0: RHS2000 amplifier channel.
     for chan_info in channels_by_type[0]:
@@ -546,7 +547,7 @@ def read_rhs(filename, file_format: str):
     # so let's skip for now and can be given on request
 
     if file_format != "one-file-per-channel":
-        channel_number_dict[11] = channel_number_dict[0] # should be one stim / amplifier channel
+        channel_number_dict[11] = channel_number_dict[0]  # should be one stim / amplifier channel
         for chan_info in channels_by_type[0]:
             name = chan_info["native_channel_name"]
             chan_info_stim = dict(chan_info)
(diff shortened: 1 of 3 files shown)


neo/rawio/intanrawio.py
```diff
@@ -514,23 +566,46 @@ def read_rhs(filename):
...
 
     # 5: Digital input channel.
     # 6: Digital output channel.
     for sig_type in [5, 6]:
-        # at the moment theses channel are not in sig channel list
-        # but they are in the raw memamp
         if len(channels_by_type[sig_type]) > 0:
-            name = {5: "DIGITAL-IN", 6: "DIGITAL-OUT"}[sig_type]
-            data_dtype += [(name, "uint16", BLOCK_SIZE)]
+            name = {4: "DIGITAL-IN", 5: "DIGITAL-OUT"}[sig_type]
```
zm711, line 578: Bad copy-paste :( set to 5, 6 for rhs

neo/rawio/intanrawio.py
```diff
@@ -838,19 +940,41 @@ def create_one_file_per_signal_dict(dirname):
...
+    ----------
+    dirname: pathlib.Path
+        The folder to explore
+    rhs: bool, default: False
+        Whether this is an rhd or an rhs file
+    """
+    # if rhs we have an extra stream to add
+    if rhs:
+        possible_raw_file_prefixes.insert(4, "board-ANALOG-OUT")
+
```
zm711, line 962: add 'board-ANALOG-OUT'

neo/rawio/intanrawio.py
```diff
@@ -483,22 +522,35 @@ def read_rhs(filename):
...
-        # with bits, they do not fit the gain/offset rawio strategy
-        chan_info_stim["units"] = ""
-        chan_info_stim["gain"] = 1.0
-        chan_info_stim["offset"] = 0.0
-        chan_info_stim["signal_type"] = 11  # put it in another group
-        chan_info_stim["dtype"] = "uint16"
-        ordered_channels.append(chan_info_stim)
-        data_dtype += [(name + "_STIM", "uint16", BLOCK_SIZE)]
+    if file_format != "one-file-per-channel":
+        for chan_info in channels_by_type[0]:
```
zm711, line 533: add channel_number_dict here for one-file-per-signal

neo/rawio/intanrawio.py
```diff
@@ -434,14 +453,29 @@ def read_variable_header(f, header):
     ("electrode_impedance_phase", "float32"),
 ]
 
+stream_type_to_name_rhs= {
```
zm711, line 456: @h-mayorquin,

this is more accurate because the stream_ids change between rhs and rhd so we should translate correctly. `0` is still always amplifier but rhs has extra analog so we could force it to have a random number (like 12) if you think it will be really disruptive on the neuro-conv side...

h-mayorquin: I don't understand,  shoulnd't we have a one dict for rhd and another for rhs if they don't have consistent stream names?

zm711: Yes I was saying we SHOULD have two different. My concern was that the old behavior was to use the RHD dict for the rhs files and we were actually messing up the streams. I was just worried that someone may have written an Interface on Neuroconv expecting this behavior so I wanted to double check!

h-mayorquin: Ok, I agree.

Concerning neuroconv I don't think we did.

neo/test/iotest/test_intanio.py
```diff
@@ -19,6 +19,8 @@ class TestIntanIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091536/info.rhs",
+        "intan/intan_fps_rhs_test_240329_091637/info.rhs",
```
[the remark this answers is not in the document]
h-mayorquin: Did you add new files to gin? I am planning to give another review to this.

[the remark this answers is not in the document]
zm711: No I typed them backwards, but we are missing the fps files from gin. Your PR only ad the fpc files. So I was trying to set-up gin again so that you'd wouldn't have to add the last folder set.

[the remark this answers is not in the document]
h-mayorquin: Let's do this. I also wanted to ask you, could you generrate a small dataset for the option when the "classical" format is partitioned in time?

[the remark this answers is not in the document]
zm711: Sure. I'll slack it to you in a couple minutes. Is three minutes enough? (ie 3 files).

neo/test/rawiotest/test_intanrawio.py
```diff
@@ -16,6 +16,8 @@ class TestIntanRawIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091536/info.rhs",
+        "intan/intan_fps_rhs_test_240329_091637/info.rhs",
```
zm711, line 20:
```suggestion
        "intan/intan_fpc_rhs_test_240329_091537/info.rhs",
        "intan/intan_fps_rhs_test_240329_091636/info.rhs",
```
oops needed to fix this too!

neo/test/iotest/test_intanio.py
```diff
@@ -19,6 +19,8 @@ class TestIntanIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091537/info.rhs",
```
zm711, line 22:
```suggestion
        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
```

neo/test/rawiotest/test_intanrawio.py
```diff
@@ -16,6 +16,8 @@ class TestIntanRawIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091537/info.rhs",
```
zm711, line 19:
```suggestion
        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
```

zm711 (commented):
Bad-copy past of the test files on my part. I think this should fix the testing portion at last.

neo/rawio/intanrawio.py
```diff
@@ -95,10 +96,25 @@ def _parse_header(self):
             raise FileNotFoundError(f"{filename} does not exist")
 
         if self.filename.endswith(".rhs"):
-            self.file_format = "header-attached"
-            self._global_info, self._ordered_channels, data_dtype, header_size, self._block_size = read_rhs(
-                self.filename
-            )
+            if filename.name == "info.rhs":
+                if any((filename.parent / file).exists for file in one_file_per_signal_filenames):
```
zm711, line 100:
```suggestion
                if any((filename.parent / file).exists() for file in one_file_per_signal_filenames):
```

neo/rawio/intanrawio.py
```diff
@@ -514,23 +580,46 @@ def read_rhs(filename):
             chan_info["offset"] = -32768 * 0.0003125
             chan_info["dtype"] = "uint16"
             ordered_channels.append(chan_info)
-            data_dtype += [(name, "uint16", BLOCK_SIZE)]
+            if file_format == "header-attached":
+                data_dtype += [(name, "uint16", BLOCK_SIZE)]
+            else:
+                data_dtype[sig_type] = "unit16"
```
zm711, line 586:
```suggestion
                data_dtype[sig_type] = "uint16"
```

neo/rawio/intanrawio.py
```diff
@@ -808,8 +897,11 @@ def read_rhd(filename, file_format: str):
     return global_info, ordered_channels, data_dtype, header_size, BLOCK_SIZE, channel_number_dict
 
 
-###########################
-# RHD Zone for Binary Files
+##########################################################################
+# RHX Zone for Binary Files
```
h-mayorquin, line 901: I don't understand this comment here. The verb gives in 902 seems to be missing something.

neo/rawio/intanrawio.py
```diff
@@ -822,13 +914,37 @@ def read_rhd(filename, file_format: str):
 ]
 
 
-def create_one_file_per_signal_dict(dirname):
-    """Function for One File Per Signal Type"""
+def create_one_file_per_signal_dict(dirname, rhs: bool = False):
```
h-mayorquin, line 917: Personal preference so your milleagea might vary, I rather have two functions here instead of mixing the two of them with keywords. I prefer code duplication rather than cyclomatic complexity and given that these are almost dictionaries / configurations I don't think that duplication is a cognitive burden.

neo/rawio/intanrawio.py
```diff
@@ -838,19 +954,41 @@ def create_one_file_per_signal_dict(dirname):
     "amp",
     "aux",
     "vdd",
-    "board-ANALOG",
+    "board-ANALOG-IN",
```
h-mayorquin, line 957: How comes?

h-mayorquin (commented):
This in general looks good for me. We are replicating the structure that we did for https://github.com/NeuralEnsemble/python-neo/pull/1402. I would go further for the aproach of segregating the two options (as in the suggestion of having two `create_one_file_per_signal_dict` insead of one) but these are already in the realm of weak personal preference.

The tests are failing, let's see what's going on and I will read it again but I think this mostly works.

neo/rawio/intanrawio.py
zm711: Because for rhs there is also "board-ANALOG-OUT", but if I split them into separate function it wouldn't matter.

neo/rawio/intanrawio.py
zm711: I think that might be fair for this. Mainly because we update the rhd a bit more often than the rhs. So splitting so that we can update the one without breaking the other is a good idea. I would say let's keep this in draft and I will work on that.

zm711:
The test is failing due to timestamp issue.... That means the test file could be corrupt. Not sure how since this is a test file generated from their code. Since it is a binary file I can just read it myself and see if I can figure out what is going. I think for me what I would say is:

1) let me fully separate out rhs and rhd for this. I agree with you that trying to squish them together might be trying to hack too much. Some duplication is probably a good idea so we can adapt and adjust one without needing to worry about the other.
2) Figure out why timestamps are messed up (they really shouldn't be).

neo/rawio/intanrawio.py
zm711: Yeah it's not super clear. I was trying to split the difference between the word "zone" above from the original code and updating it with new wording. What about:

This section of code lists the possible file names in the formats "one-file-per-signal" and "one-file-per-channel" which are optional formats within the rhx intan recording software. They are organized as a series
of headerless binary files (*.dat) with set names.

neo/rawio/intanrawio.py
h-mayorquin: Ok, thanks for the explanation!

h-mayorquin:
> The test is failing due to timestamp issue.... That means the test file could be corrupt. Not sure how since this is a test file generated from Intan's simulator. Since it is a binary file I can just read it myself and see if I can figure out what is going on. I think for me what I would say is:
> 
> 1. let me fully separate out rhs and rhd for this. I agree with you that trying to squish them together might be trying to hack too much. Some duplication is probably a good idea so we can adapt and adjust one without needing to worry about the other.
> 2. Figure out why timestamps are messed up (they really shouldn't be).

Sounds good, let me know if you need to upload something again.

neo/test/iotest/test_intanio.py
```diff
@@ -19,6 +19,8 @@ class TestIntanIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
+        "intan/intan_fps_rhs_test_240329_091636/info.rhs",
```
zm711, line 23:
```suggestion
        "intan/intan_fps_rhs_test_240329_091536/info.rhs",
```

neo/test/rawiotest/test_intanrawio.py
```diff
@@ -16,6 +16,8 @@ class TestIntanRawIO(
         "intan/intan_rhd_test_1.rhd",
         "intan/intan_fpc_test_231117_052630/info.rhd",
         "intan/intan_fps_test_231117_052500/info.rhd",
+        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
+        "intan/intan_fps_rhs_test_240329_091636/info.rhs",
```
zm711, line 20:
```suggestion
        "intan/intan_fps_rhs_test_240329_091536/info.rhs",
```

zm711 (commented):
The number of times I've mistyped these test files is outrageous....

neo/rawio/intanrawio.py
```diff
@@ -159,7 +175,8 @@ def _parse_header(self):
         # check timestamp continuity
         if self.file_format == "header-attached":
             timestamp = self._raw_data["timestamp"].flatten()
-        # timestamps are always the last stream
+
+        # timestamps are always 6 for rhd and 7 for rhs
```
zm711, line 179:
```suggestion
        # timestamps are always last stream for headerless binary files
```

neo/rawio/intanrawio.py
```diff
@@ -195,9 +212,16 @@ def _parse_header(self):
 
         stream_ids = np.unique(signal_channels["stream_id"])
         signal_streams = np.zeros(stream_ids.size, dtype=_signal_stream_dtype)
-        signal_streams["id"] = stream_ids
+
+        # we need to sort the data because the string of 10 is mis-sorted.
+        stream_ids_sorted = sorted([int(stream_id) for stream_id in stream_ids])
+        signal_streams["id"] = [str(stream_id) for stream_id in stream_ids_sorted]
+
         for stream_index, stream_id in enumerate(stream_ids):
```
zm711, line 220:
```suggestion
        for stream_index, stream_id in enumerate(stream_ids_sorted):
```

zm711:
@samuelgarcia @alejoe91 

This is actually ready for review now. Tests are passing (barring the retest happening right now which should pass). With this PR we now support all formats of Intan (rhd and rhs, traditional + headerless binary formats). If you think it would be beneficial for this release feel free to review now. Otherwise we can wait. @h-mayorquin said he would do another read-over later. So we can either merge sooner and then fix up or wait and tidy more now.

I would prefer squashing this one. I had some fights with typos in the paths to the test files, so there are a lot of junk commits.

alejoe91:
Thanks @zm711 

This looks good to me! @h-mayorquin if you want to take another look, otherwise we can squash and merge and include this in the release :)

h-mayorquin:
I took a second look, it looks good to me.

zm711:
Thanks @h-mayorquin!

PR #1464, opened 2024-04-29
Add 1016 to spikeglx

alejoe91:
Requested via slack


zm711:
What's the difference between 1015 and 1016 for NP? (ie, would it be better to test against or do your current tests already cover it's broad features?)

alejoe91:
From @jenniferColonell: "it's a short NHP probe with the linear site configuration, 960 sites, 1.0-like gain and referencing."

jenniferColonell:
1016 is really a twin of 1015. It has a different part number because it's sterilized.

zm711 (approved):
Thanks @jenniferColonell. Sounds good!

PR #1466, opened 2024-05-01
Neuroscope: fix scale of signals.

PeterNSteinmetz:
Need to multiply not divide by 1000 for mV scale.

### DIFF
--- a/neo/rawio/neuroscoperawio.py
+++ b/neo/rawio/neuroscoperawio.py
@@ -92,9 +92,9 @@ class NeuroScopeRawIO(BaseRawIO):
             for xml_rc in xml_chx:
                 channel_group[int(xml_rc.text)] = grp_index
 
+        sig_dtype = "int16"
         if nbits == 16:
-            sig_dtype = "int16"
-            gain = voltage_range / (2**16) / amplification / 1000.0
+            gain = voltage_range * 1000 / (2**nbits) / amplification
             # ~ elif nbits==32:
             # Not sure if it is int or float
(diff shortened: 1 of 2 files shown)


zm711:
Do you have some documentation or some example traces? This seems reasonable, but it would be nice to know where the scale is coming from.

PeterNSteinmetz:
The only thing this corrects is the scaling factor applied to convert to mV from the raw samples. There is nothing in the .dat files. The corresponding .xml files contain the voltage range, number of bits for each sample, and amplification. There is nothing per se in those files which says it is mV. 

I can tell you however that with the correction the other way, as before this change, the recordings are being shown as in the 100 nV range, which makes no sense for neurophysiological signals.

This is a common mistake and I have done it myself many times - dividing instead of multiplying by the amplification. 

This change also includes a comment which clearly explains the derivation of the scale.

zm711:
Since @samuelgarcia wrote this I put him to review it. My only suggestion would be to move the comment into the actual class rather than in the test so that future us see it there :)

PeterNSteinmetz:
OK. I added a comment in the source and clarified that this is the proper scale to convert a raw sample value to mV.

samuelgarcia:
Oups.
Yes it was a mistake. Thanks for the catch.

PeterNSteinmetz:
OK. Can we get this reviewed and merged? I don't like keeping these branches open unnecessarily on my fork. Thanks.

samuelgarcia (approved):

PR #1473, opened 2024-05-10
Testing: Run IO-tests on multiple python versions

zm711:
Since python is deprecating alot of standard library stuff and we have some more legacy IOs we should probably test against multiple python versions. This tests against 
3.9, 3.10, and 3.11. We can add 3.12 to see what will break later for the IOs.

### DIFF
--- a/.github/workflows/io-test.yml
+++ b/.github/workflows/io-test.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.11']
     defaults:
       # by default run in bash mode (required for conda usage)
       run:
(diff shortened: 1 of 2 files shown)


.github/workflows/io-test.yml
```diff
@@ -13,12 +13,12 @@ concurrency:  # Cancel previous workflows on the same pull request
...
 jobs:
   build-and-test:
-    name: Test on (${{ inputs.os }})
+    name: Test on (${{ inputs.os }}) (${{ matrix.python-version}})
     runs-on: ${{ inputs.os }}
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', ]
+        python-version: ['3.9', '3.10', '3.11']
```
samuelgarcia, line 21:
```suggestion
        python-version: ['3.9', '3.11']
```

samuelgarcia (approved):

zm711:
To be more environmentally friendly we will do 3.9 and 3.11 and hope to add/switch to 3.12 in the future.

General strategy would be do first and last python testing for the conda io-testing suite.

PR #1476, opened 2024-05-13
Fix digital streams in Intan for the `one-file-per-channel` format in rhd.

h-mayorquin:
For the `header-attached` and the `one-file-per-signal` the digital channels are interleaved  and need a bitwise operators to separate the channels but this is not the case for the `one-file-per-channel` where each digital channel is separated and is already uint16 with 0 and 1s.

This PR patches this behavior so the digital channels can be correctly retrieved for the case of the `one-file-per-channel`.  Unfortunatlely, we don't have a multiple digital channels in the test examples on git to test for this behavior. Any idea on how to test for this @zm711 ?

I tested this for data that I have and it works.

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -598,7 +598,7 @@ def read_rhs(filename, file_format: str):
     for sig_type in [5, 6]:
         if file_format in ["header-attached", "one-file-per-signal"]:
             if len(channels_by_type[sig_type]) > 0:
-                name = {4: "DIGITAL-IN", 5: "DIGITAL-OUT"}[sig_type]
+                name = {5: "DIGITAL-IN", 6: "DIGITAL-OUT"}[sig_type]
                 chan_info = channels_by_type[sig_type][0]
                 # So currently until we have get_digitalsignal_chunk we need to do a tiny hack to
                 # make this memory map work correctly. So since our digital data is not organized
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -537,6 +537,7 @@ def read_rhs(filename, file_format: str):
         channel_number_dict[10] = channel_number_dict[0]
         for chan_info in channels_by_type[0]:
             chan_info_dc = dict(chan_info)
+            name = chan_info["native_channel_name"]
             chan_info_dc["native_channel_name"] = name + "_DC"
             chan_info_dc["sampling_rate"] = sr
             chan_info_dc["units"] = "mV"
@@ -546,7 +547,6 @@ def read_rhs(filename, file_format: str):
             chan_info_dc["dtype"] = "uint16"
             ordered_channels.append(chan_info_dc)
             if file_format == "header-attached":
-                name = chan_info["native_channel_name"]
                 data_dtype += [(name + "_DC", "uint16", BLOCK_SIZE)]
             else:
                 data_dtype[10] = "uint16"
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -52,9 +52,9 @@ class IntanRawIO(BaseRawIO):
     check for the file extension and will gather the header information based on the
     extension. Additionally it functions with RHS v 1.0 and RHD 1.0, 1.1, 1.2, 1.3, 2.0,
     3.0, and 3.1 files.
-    
+
     * The reader can handle three file formats 'header-attached', 'one-file-per-signal' and
-    'one-file-per-channel'. 
+    'one-file-per-channel'.
 
     * Intan files contain amplifier channels labeled 'A', 'B' 'C' or 'D'
     depending on the port in which they were recorded along with the following
@@ -514,7 +514,6 @@ def read_rhs(filename, file_format: str):
 
     # 0: RHS2000 amplifier channel.
     for chan_info in channels_by_type[0]:
-        name = chan_info["native_channel_name"]
         chan_info["sampling_rate"] = sr
         chan_info["units"] = "uV"
         chan_info["gain"] = 0.195
@@ -528,6 +527,7 @@ def read_rhs(filename, file_format: str):
             chan_info["dtype"] = "int16"
         ordered_channels.append(chan_info)
         if file_format == "header-attached":
+            name = chan_info["native_channel_name"]
             data_dtype += [(name, "uint16", BLOCK_SIZE)]
         else:
             data_dtype[0] = "int16"
@@ -536,7 +536,6 @@ def read_rhs(filename, file_format: str):
         # if we have dc amp we need to grab the correct number of channels
         channel_number_dict[10] = channel_number_dict[0]
         for chan_info in channels_by_type[0]:
-            name = chan_info["native_channel_name"]
             chan_info_dc = dict(chan_info)
             chan_info_dc["native_channel_name"] = name + "_DC"
             chan_info_dc["sampling_rate"] = sr
@@ -547,6 +546,7 @@ def read_rhs(filename, file_format: str):
             chan_info_dc["dtype"] = "uint16"
             ordered_channels.append(chan_info_dc)
             if file_format == "header-attached":
+                name = chan_info["native_channel_name"]
                 data_dtype += [(name + "_DC", "uint16", BLOCK_SIZE)]
             else:
                 data_dtype[10] = "uint16"
@@ -556,7 +556,6 @@ def read_rhs(filename, file_format: str):
     if file_format != "one-file-per-channel":
         channel_number_dict[11] = channel_number_dict[0]  # should be one stim / amplifier channel
         for chan_info in channels_by_type[0]:
-            name = chan_info["native_channel_name"]
             chan_info_stim = dict(chan_info)
             chan_info_stim["native_channel_name"] = name + "_STIM"
             chan_info_stim["sampling_rate"] = sr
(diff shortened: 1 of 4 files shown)


zm711:
I make a simulated data set for this condition, but maybe we just do it for one of the four formats rhd-fpc? Since the rhs will follow the same convention. Then we can add that to gin make sure that works. Just to make sure we are not making any assumptions maybe I'll do like digital in 03 and digital in 08 so that we have a spread and make sure that still works?

zm711:
Also the exact same change should be done for rhs. so maybe I should just make examples of both that way you can do both in this PR.

h-mayorquin:
> I make a simulated data set for this condition, but maybe we just do it for one of the four formats rhd-fpc? Since the rhs will follow the same convention. Then we can add that to gin make sure that works. Just to make sure we are not making any assumptions maybe I'll do like digital in 03 and digital in 08 so that we have a spread and make sure that still works?

Yeah, that would be great. We can make for both or we can substitute the current ones for 'one-file-per-channel' to have this complexity instead of what we have currently.

h-mayorquin:
Data is on pull request:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/133

neo/test/rawiotest/test_intanrawio.py
```diff
@@ -12,12 +12,14 @@ class TestIntanRawIO(
...
-        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
-        "intan/intan_fps_rhs_test_240329_091536/info.rhs",
+        "intan/intan_rhs_test_1.rhs",  # Format header-attached 
+        "intan/intan_rhd_test_1.rhd",  # Format header attach
+        "intan/intan_fpc_test_231117_052630/info.rhd",  # Format one-file-per-channel
+        "intan/intan_fps_test_231117_052500/info.rhd",  # Format one file per signal
+        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",  # Format one-file-per-channel
+        "intan/intan_fps_rhs_test_240329_091536/info.rhs",   # Format one-file-per-signal
+        # "intan/rhd_fpc_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-channel
+        # "intan/rhd_fps_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-signal
```
zm711, line 22:
```suggestion
        "intan/rhd_fpc_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-channel
        "intan/rhd_fps_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-signal
```
Should be ready for these now :).

Also `test_intanio.py` should have the files added

h-mayorquin: Yes, I was waiting for the merge.

zm711: And add to `test_intanio` .

zm711:
Good luck with those tests. :) I swear every time I think the fpc will be easy it takes me like 10 commits to get it all sorted.

neo/test/rawiotest/test_intanrawio.py
h-mayorquin: Done.

h-mayorquin:
> Good luck with those tests. :) I swear every time I think the fpc will be easy it takes me like 10 commits to get it all sorted.

This is a cache problem, can you reset the data caches?

h-mayorquin:
Here:

https://github.com/NeuralEnsemble/python-neo/actions/workflows/caches_cron_job.yml

You should be able to trigger this work manually.

neo/test/rawiotest/test_intanrawio.py
```diff
@@ -12,12 +12,14 @@ class TestIntanRawIO(
...
-        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
-        "intan/intan_fps_rhs_test_240329_091536/info.rhs",
+        # "intan/intan_rhs_test_1.rhs",  # Format header-attached 
+        # "intan/intan_rhd_test_1.rhd",  # Format header attach
+        # "intan/intan_fpc_test_231117_052630/info.rhd",  # Format one-file-per-channel
+        # "intan/intan_fps_test_231117_052500/info.rhd",  # Format one file per signal
+        # "intan/intan_fpc_rhs_test_240329_091637/info.rhs",  # Format one-file-per-channel
+        # "intan/intan_fps_rhs_test_240329_091536/info.rhs",   # Format one-file-per-signal
+        "intan/rhd_fpc_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-channel
+        "intan/rhd_fps_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-signal
```
zm711, line 22:
```suggestion
        "intan/rhs_fpc_multistim_240514_082243/info.rhs",  # Multiple digital channels one-file-per-channel rhs
```

It wasn't a cache issue . The other test was an rhs test file.

neo/test/iotest/test_intanio.py
```diff
@@ -15,12 +15,14 @@ class TestIntanIO(
...
-        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",
-        "intan/intan_fps_rhs_test_240329_091536/info.rhs",
+        "intan/intan_rhs_test_1.rhs",  # Format header-attached 
+        "intan/intan_rhd_test_1.rhd",  # Format header attach
+        "intan/intan_fpc_test_231117_052630/info.rhd",  # Format one-file-per-channel
+        "intan/intan_fps_test_231117_052500/info.rhd",  # Format one file per signal
+        "intan/intan_fpc_rhs_test_240329_091637/info.rhs",  # Format one-file-per-channel
+        "intan/intan_fps_rhs_test_240329_091536/info.rhs",   # Format one-file-per-signal
+        "intan/rhd_fpc_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-channel
+        "intan/rhd_fps_multistim_240514_082044/info.rhd",  # Multiple digital channels one-file-per-signal
```
zm711, line 25:
```suggestion
        "intan/rhs_fpc_multistim_240514_082243/info.rhs",  # Multiple digital channels one-file-per-channel rhs
```
same here

zm711 (commented):
I found out was the issue was. I had made one rhd and one rhs test file since these formats behave a little differently. If we need more fps files I can do it.

Also (just for your information) it would be impossible for the files to have the same timestamp unless someone went into to edit it).

zm711:
Not sure why actual testing is running so slowly.

h-mayorquin:
> Not sure why actual testing is running so slowly.

It is because it is downloading the data as the cache is failing. The commented line on the cache.

zm711:
But shouldn't it have done that during the cache step and not during the pytest step?

h-mayorquin:
> But shouldn't it have done that during the cache step and not during the pytest step?
> 
> You know actions way better than I do so I trust whatever you say for that!

The cache is failing to restore in the 4th step.
Then it goes on without cached data which means all the tests are downloading the data on the spot.

image

h-mayorquin:
The caches for the new data are being created now:

https://github.com/NeuralEnsemble/python-neo/actions/workflows/caches_cron_job.yml

I guess the cron job was later than I expected.

zm711:
I see! That makes sense. I thought it would pause regenerate the cache then test for this type of action. Not skip the cache and download live, but makes sense.

neo/rawio/intanrawio.py
```diff
@@ -53,18 +53,21 @@ class IntanRawIO(BaseRawIO):
...
     1: 'RHD2000 auxiliary input channel',
     2: 'RHD2000 supply voltage channel',
     3: 'USB board ADC input channel',
     4: 'USB board digital input channel',
     5: 'USB board digital output channel'
 
-    * Due to the structure of the digital input and output channels these can be accessed
-    as one long vector, which must be post-processed.
+    * For the "header-attached" and "one-file-per-signal" formats, the structure of the digital input and output channels
+    these can be accessed as one long vector, which must be post-processed.
```
zm711, line 70:
```suggestion
    * For the "header-attached" and "one-file-per-signal" formats, the structure of the digital input and output channels is
    one long vector, which must be post-processed to extract individual digital channel information. See the intantech website for more information on performing this post-processing.
```

neo/rawio/intanrawio.py
```diff
@@ -576,43 +579,50 @@ def read_rhs(filename, file_format: str):
...
+                chan_info["units"] = "TTL"  # arbitrary units TTL for logic
+                chan_info["gain"] = 1.0
+                chan_info["offset"] = 0.0
+                chan_info["dtype"] = "uint16"
+                ordered_channels.append(chan_info)
+                if file_format == "header-attached":
+                    data_dtype += [(name, "uint16", BLOCK_SIZE)]
+                else:
+                    data_dtype[sig_type] = "uint16"
+        elif file_format == "one-file-per-channel":
```
zm711, line 618: Maybe we add a comment over the else here just to say that one-file-per-channel stores actual 1 and 0 s for the digital in and out?

zm711 (commented):
I think the overall PR looks good. A couple of documentation recs.

neo/rawio/intanrawio.py
h-mayorquin: Added

zm711:
@samuelgarcia @alejoe91

let me know if you guys want to take a look. Basically, @h-mayorquin discovered from some neuroconv data that the one-file-per-channel format actually has a different way in how it stores dig-in and dig-out data. So this patch fixes that. It looks good to me and we supplied a set of test files which are passing.

zm711 (approved):

samuelgarcia (approved):

samuelgarcia:
merci!

PR #1478, opened 2024-05-16
[DOC] - Minor update to BlackrockRawIO docstring

TomDonoghue:
This PR is a minor fix / update to the docstring in the BlackrockRawIO object, notably:
- fixing a note that implies only 2.X formats are supported now that 3.0 is also supported
- minor fixes to the parameter listings


zm711 (approved):
Thanks @TomDonoghue 

looks good to me!

PR #1485, opened 2024-06-05
Fix version issue for rhs

zm711:
@alejoe91 could I get a quick review of this?

Problem was that rhs doesn't use packaging.version Version so when we propagated the rhd fixes to rhs we moved Version into rhs. Testing missed this because we don't have any notch filtered rhs test files so the boolean always short circuited. 

Related to https://github.com/SpikeInterface/spikeinterface/issues/2974


alejoe91 (approved):

alejoe91:
LGTM!

zm711:
Thanks!

PR #1486, opened 2024-06-06
Clean-up unused imports and standardized order

zm711:
Inspired by @h-mayorquin and his cleaning up of imports on spikeinterface.

Also we use scipy for 3 functions total and the top import is really slowing down the import time. I moved the import to those functions since I don't know how often people actually use them.

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -18,21 +18,13 @@ the old object.
 """
 
 import logging
-
-try:
-    import scipy.signal
-except ImportError as err:
-    HAVE_SCIPY = False
-else:
-    HAVE_SCIPY = True
+from copy import deepcopy
 
 import numpy as np
 import quantities as pq
 
 from neo.core.baseneo import BaseNeo, MergeError, merge_annotations, intersect_annotations
 from neo.core.dataobject import DataObject
-from copy import copy, deepcopy
-
 from neo.core.basesignal import BaseSignal
 
 logger = logging.getLogger("Neo")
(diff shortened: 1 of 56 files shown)


zm711:
For additional reference we can see that a huge chunk of import time is eaten by importing scipy.signal. It seems about 80ish or so percent of importing neo (through spikeinterface) is coming from the scipy import.

image

neo/core/analogsignal.py
```diff
@@ -612,8 +604,9 @@ def downsample(self, downsampling_factor, **kwargs):
         -----
         For resampling the signal with a fixed number of samples, see `resample` method.
         """
-
-        if not HAVE_SCIPY:
+        try:
+            import scipy.signal
```
zm711, line 608: Import only here so cost if sonly when people want to use the function. Which I assume is rarer than importing the package overall.

neo/io/__init__.py
```diff
@@ -281,7 +281,7 @@
 # if it is not present, use the neurosharectypesio to load files
 try:
     import neuroshare as ns
-except ImportError as err:
+except ModuleNotFoundError as err:
```
zm711, line 284: Seems like this is the more modern error for this? Happy to revert though...

neo/io/tools.py
```diff
@@ -4,12 +4,8 @@
     SPikeTrains
 """
 
-try:
-    from collections.abc import MutableSequence
-except ImportError:
-    from collections import MutableSequence
 
-import numpy as np
+from collections.abc import MutableSequence
```
zm711, line 8: As of 3.7 this is the only way to do this, so since 3.8 is our min this change is fine.

zm711 (commented):
Notes for people to check as we review.

samuelgarcia (approved):

samuelgarcia:
Merci Zach

PR #1490, opened 2024-06-18
Support for NumPy 2.0 in Neo Core

zm711:
following a policy of NEP 29 + 1 year.


zm711:
Looks like we have some work to do :)

apdavison:
note that quantities 0.16.0, with support for NumPy 2.0, has just been released

zm711:
I updated the branch so we should see the neo problems vs the quantities in the testing!

zm711:
I think @JuliaSprenger saw this error before right? 

```
AssertionError: array(99.) * mV == array(99.) * mV
```

The quantities array are the same, but some weird floating point different is messing things up!

zm711:
Let's squash this one when it is ready. A lot of this is me tweaking the action and then messing up on one version. So a lot of these are trash commits.

zm711:
Okay current core failures can be reproduced with the following:

```python
>>> import quantities as pq
>>> import numpy as np
>>> times = np.arange(10, dtype=np.float32)
>>> times.dtype
'float32'
>>> times= times * pq.ms
>>> times.dtype
'float64'
```
and for the other failing test

```python
>>> import quantities as pq
>>> import numpy as np
>>> times = [1,2,3] * pq.s
>>> np.concatenate((times, times))
array([1,2,3,1,2,3]) * dimensionless
```

I need to read more about numpy 2.0 changes to figure out these parts of the tests that have changed. Any ideas @apdavison and @samuelgarcia ?

zm711:
Actually it was a change in default numpy behavior which requires an extra as type in tests. My bad.

.github/workflows/io-test.yml
```diff
@@ -90,9 +91,11 @@ jobs:
       - name: Install neo including dependencies
         # installation with dependencies is only required if no cache was found
         # restore-key hits should result in `cache-hit` == 'false'
+        # in the run we don't do upgrade to prevent forced numpy 2.0. when ready
+        # change to pip install --upgrade -e . instead.
         if: steps.cache-conda-env.outputs.cache-hit != 'true'
         run: |
-          pip install --upgrade -e .
+          pip install -e .
```
zm711, line 98:
```suggestion
          pip install -e .
          pip uninstall numpy
          pip install numpy==1.26.4
```
let's try this?

.github/workflows/io-test.yml
```diff
@@ -90,9 +91,13 @@ jobs:
...
+          pip uninstall numpy
+          pip install numpy==1.26.4
```
zm711, line 100:
```suggestion
          pip uninstall numpy -y
          pip install numpy==1.26.4 -y
```

.github/workflows/io-test.yml
```diff
@@ -90,9 +91,13 @@ jobs:
...
         # installation with dependencies is only required if no cache was found
         # restore-key hits should result in `cache-hit` == 'false'
+        # in the run we don't do upgrade to prevent forced numpy 2.0. when ready
+        # change to pip install --upgrade -e . instead.
         if: steps.cache-conda-env.outputs.cache-hit != 'true'
         run: |
-          pip install --upgrade -e .
+          pip install -e .
+          pip uninstall numpy -y
+          pip install numpy==1.26.4 -y
```
zm711, line 100:
```suggestion
          pip install numpy==1.26.4
```

zm711:
Once #1585 is in main we can merge main into this which fixes the RTD error. Then this ready. 

Note this is only for the core of Neo. I want to do the RawIO/IO in a separate PR.

neo/core/spiketrain.py
```diff
@@ -975,7 +975,7 @@ def _merge_array_annotations(self, others, sorting=None):
 
         omitted_keys_other = [
             key
-            for key in np.unique([key for other in others for key in other.array_annotations])
+            for key in set([key for other in others for key in other.array_annotations])
```
zm711, line 978: only actual change to code base. numpy 2.0 returns this as a `np.str_` thing rather than as the actual string. Using `set` instead of `np.unique` returns the string both >2.0 and < 2.0 for NumPy. Maybe a slight performance hit, but probably not too bad since the annotations should be small either way and this is in a list and not array.

neo/test/coretest/test_spiketrain.py
```diff
@@ -1147,7 +1155,7 @@ def test_merge_multiple(self):
         expected *= time_unit
         sorting = np.argsort(expected)
         expected = expected[sorting]
-        np.testing.assert_array_equal(result.times, expected)
+        np.testing.assert_array_equal(result.times.magnitude, expected.magnitude)
```
zm711, line 1158: This change is necessary because as of numpy 2.0 because numpy assert_array_equal actually checks that the quantities are the same which leads to a units issue. If someone else understands this better please post here.

zm711:
RTD is not building the correct version of neo. I don't have why. Maybe I just need to give it a little time (I had hoped switching to 3.12 would flush any cache, but it still had the wrong Neo and it had other errors). I'll just try to rerun RTD build this weekend and see if time fixes it.

zm711:
Sam and I looked over the RTD failure and we aren't sure why it is happening. I think it should fix itself once this is merged into main since the problem showing up is fixed in main.

neo/test/coretest/test_spiketrain.py
```diff
@@ -313,6 +313,10 @@ def test__create_from_quantity_array(self):
 
     def test__create_from_quantity_array_with_dtype(self):
         times = np.arange(10, dtype="f4") * pq.ms
+        # this step is required for NumPy 2.0 which now casts to float64 in the case either value/array
+        # is float64 even if not necessary
+        # https://numpy.org/devdocs/numpy_2_0_migration_guide.html
+        times = times.astype(dtype="f4")
```
zm711, line 319: Based on discussion with Sam my comment may not have been clear enough. The problem is that 
1* pq.ms is at baseline at `'float64'` and as of numpy 2.0+ everything is up-cast to the bigger dtype even if not necessarily. So in old versions of our test `times` would remain a `float32`, but with new versions of numpy make it into a `'float64'`. I did the astype to keep our original idea explict, but we could think of other rewrites of this test to maintain the dtype test.

zm711:
@apdavison this is ready for review, the RTD is broken with some cache issue, but I fixed the problem on main (and merging main into this branch didn't fix it). so our options are you review and merge ignoring the RTD problem or I open a new PR built off of the current main and make the exact same fixes to make sure all tests pass. What's your preference?

alejoe91 (approved):

samuelgarcia:
thanks for for @zm711 @apdavison and @alejoe91 

I did not folow so much. Does finally we are making a copy of the buffer every time we create a data object from numpy or not ?

PR #1491, opened 2024-06-18
Intan digital channel extraction for header attached and one file per signal format

h-mayorquin:
@zm711 this is a follow-up of  https://github.com/NeuralEnsemble/python-neo/pull/1476 

This makes the extraction of the digital channels work like all the other data in neo (and therefore spikeinterface) so user does not need to do any post-processing. The post-processing was derived from the data sheet, the python [conversion script that intan provides for python](https://intantech.com/downloads.html?tabSelect=Software&yPos=68.57142639160156) and [the matlab version](https://github.com/dattalab/spike_finder/blob/master/read_Intan_RHD2000_file.m) of the same operation .

Let me know what you think.

@luiztauffer

I tested this on Nelson data like this (using this branch and from spikeinterface):

```python
from pathlib import Path

intan_data_folder = Path.home() / "Downloads" 
assert intan_data_folder.is_dir()

file_path = intan_data_folder / "pm231213a_240205_240205_152654.rhd"
assert file_path.is_file()


from spikeinterface.extractors import IntanRecordingExtractor

recording = IntanRecordingExtractor(file_path=file_path, stream_name="USB board digital input channel", all_annotations=True)
recording
```
<div style='border:1px solid #ddd; padding:10px;'><strong>IntanRecordingExtractor: 3 channels - 30.0kHz - 1 segments - 1,800,064 samples - 60.00s (1.00 minutes) - uint16 dtype - 10.30 MiB</strong></div><details style='margin-left: 10px;'>  <summary><strong>Channel IDs</strong></summary><ul>['DIGITAL-IN-12' 'DIGITAL-IN-13' 'DIGITAL-IN-14'] </details><details style='margin-left: 10px;'>  <summary><strong>Annotations</strong></summary><ul><li> <strong> is_filtered </strong>: False</li><li> <strong> name </strong>: USB board digital input channel</li><li> <strong> stream_id </strong>: 4</li><li> <strong> file_origin </strong>: /home/heberto/Downloads/pm231213a_240205_240205_152654.rhd</li><li> <strong> intan_version </strong>: 3.3</li><li> <strong> actual_upper_bandwidth </strong>: 7603.76513671875</li><li> <strong> desired_lower_bandwidth </strong>: 0.10000000149011612</li><li> <strong> note1 </strong>: </li><li> <strong> actual_impedance_test_frequency </strong>: 1000.0</li><li> <strong> notch_filter_mode </strong>: 0</li><li> <strong> num_temp_sensor_channels </strong>: 0</li><li> <strong> desired_upper_bandwidth </strong>: 7500.0</li><li> <strong> desired_impedance_test_frequency </strong>: 1000.0</li><li> <strong> note2 </strong>: </li><li> <strong> nb_signal_group </strong>: 7</li><li> <strong> actual_dsp_cutoff_frequency </strong>: 1.165827989578247</li><li> <strong> eval_board_mode </strong>: 0</li><li> <strong> note3 </strong>: </li><li> <strong> desired_dsp_cutoff_frequency </strong>: 1.0</li><li> <strong> actual_lower_bandwidth </strong>: 0.09452909976243973</li><li> <strong> notch_filter </strong>: False</li><li> <strong> dsp_enabled </strong>: 1</li></ul> </details><details style='margin-left: 10px;'><summary><strong>Channel Properties</strong></summary><ul><details><summary> <strong> gain_to_uV </strong> </summary>[1. 1. 1.]</details><details><summary> <strong> offset_to_uV </strong> </summary>[0. 0. 0.]</details><details><summary> <strong> channel_names </strong> </summary>['DIGITAL-IN-12' 'DIGITAL-IN-13' 'DIGITAL-IN-14']</details><details><summary> <strong> channel_ids </strong> </summary>['DIGITAL-IN-12' 'DIGITAL-IN-13' 'DIGITAL-IN-14']</details><details><summary> <strong> native_order </strong> </summary>[12 13 14]</details><details><summary> <strong> chip_channel_num </strong> </summary>[0 0 0]</details><details><summary> <strong> electrode_impedance_phase </strong> </summary>[0. 0. 0.]</details><details><summary> <strong> spike_scope_digital_trigger_channel </strong> </summary>[0 0 0]</details><details><summary> <strong> electrode_impedance_magnitude </strong> </summary>[0. 0. 0.]</details><details><summary> <strong> spike_scope_digital_edge_polarity </strong> </summary>[0 0 0]</details><details><summary> <strong> spike_scope_trigger_mode </strong> </summary>[1 1 1]</details><details><summary> <strong> board_stream_num </strong> </summary>[0 0 0]</details><details><summary> <strong> custom_order </strong> </summary>[12 13 14]</details><details><summary> <strong> spike_scope_voltage_thresh </strong> </summary>[0 0 0]</details></ul></details>

```
full_traces = recording.get_traces()
partial_traces = recording.get_traces(start_frame=0, end_frame=3_000, channel_ids=["DIGITAL-IN-12", "DIGITAL-IN-14" ])

np.allclose(full_traces[0:3_000, [0, 2]], partial_traces)
```

Can you check if the output makes sense in the light of the experiment you are converting? The signal looks correct but nothing like ecological validation.

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -90,7 +90,10 @@ class IntanRawIO(BaseRawIO):
     Examples
     --------
     >>> import neo.rawio
+    >>> # for a header-attached file
     >>> reader = neo.rawio.IntanRawIO(filename='data.rhd')
+    >>> # for the other formats we point to the info.rhd
+    >>> reader = neo.rawioIntanRawIO(filename='info.rhd')
     >>> reader.parse_header()
     >>> raw_chunk = reader.get_analogsignal_chunk(block_index=0,
                                                   seg_index=0
@@ -113,8 +116,8 @@ class IntanRawIO(BaseRawIO):
         return self.filename
 
     def _parse_header(self):
-        
-        self.filename  = Path(self.filename)
+
+        self.filename = Path(self.filename)
 
         # Input checks
         if not self.filename.is_file():
@@ -201,12 +204,14 @@ class IntanRawIO(BaseRawIO):
                     self._raw_data[stream_index].append(channel_memmap)
 
         # Data Integrity checks
+        # strictness of check is controlled by ignore_integrity_checks
+        # which is set at __init__
         self._assert_timestamp_continuity()
 
         # signals
         signal_channels = []
         self.native_channel_order = {}
-        for c, chan_info in enumerate(self._ordered_channel_info):
+        for chan_info in self._ordered_channel_info:
             name = chan_info["custom_channel_name"]
             channel_id = chan_info["native_channel_name"]
             sig_dtype = chan_info["dtype"]
@@ -558,7 +563,24 @@ class IntanRawIO(BaseRawIO):
                 raise NeoReadWriteError(error_msg)
 
 
+###################################
+# Header reading helper functions
+
+
 def read_qstring(f):
+    """
+    Reads the optional notes included in the Intan RHX software
+
+    Parameters
+    ----------
+    f: BinaryIO
+        The file object
+
+    Returns
+    -------
+    txt: str
+        The string
+    """
     length = np.fromfile(f, dtype="uint32", count=1)[0]
     if length == 0xFFFFFFFF or length == 0:
         return ""
@@ -566,7 +588,22 @@ def read_qstring(f):
     return txt
 
 
-def read_variable_header(f, header):
+def read_variable_header(f, header: list):
+    """
+    Reads the information from the binary file for the header info dict
+
+    Parameters
+    ----------
+    f: BinaryIO
+        The file object
+    header: list[tuple]
+        The list of header sections along with their associated dtype
+
+    Returns
+    -------
+    info: dict
+        The dictionary containing the information contained in the header
+    """
     info = {}
     for field_name, field_type in header:
         if field_type == "QString":
@@ -654,7 +691,11 @@ def read_rhs(filename, file_format: str):
     with open(filename, mode="rb") as f:
         global_info = read_variable_header(f, rhs_global_header)
 
-        # We use signal_type in the header as stream_id in neo
+        # We use signal_type in the header as stream_id in neo with the following
+        # complications: for rhs files in the header-attaached stream_id 0 also
+        # contains information for stream_id 10 and stream_id 11 so we need to
+        # break these up. See notes throughout code base; for timestamps we always
+        # force them to be the last stream_id.
         stream_id_to_channel_info_list = {k: [] for k in [0, 3, 4, 5, 6]}
         if not file_format == "header-attached":
             # data_dtype for rhs is complicated. There is not 1, 2 (supply and aux),
@@ -667,7 +708,11 @@ def read_rhs(filename, file_format: str):
                 for c in range(group_info["channel_num"]):
                     chan_info = read_variable_header(f, rhs_signal_channel_header)
                     if chan_info["signal_type"] in (1, 2):
-                        raise NeoReadWriteError("signal_type of 1 or 2 is not yet implemented in Neo")
+                        error_msg = (
+                            "signal_type of 1 or 2 does not exist for rhs files. If you have an rhs file "
+                            "with these formats open an issue on the python-neo github page"
+                        )
+                        raise NeoReadWriteError(error_msg)
                     if bool(chan_info["channel_enabled"]):
                         stream_id_to_channel_info_list[chan_info["signal_type"]].append(chan_info)
 
@@ -708,6 +753,7 @@ def read_rhs(filename, file_format: str):
 
     if bool(global_info["dc_amplifier_data_saved"]):
         # if we have dc amp we need to grab the correct number of channels
+        # the channel number is the same as the count for amplifier data
         channel_number_dict[10] = channel_number_dict[0]
         for chan_info in stream_id_to_channel_info_list[0]:
             chan_info_dc = dict(chan_info)
@@ -724,6 +770,7 @@ def read_rhs(filename, file_format: str):
                 data_dtype += [(name + "_DC", "uint16", BLOCK_SIZE)]
             else:
                 data_dtype[10] = "uint16"
+
     # I can't seem to get stim files to generate for one-file-per-channel
     # so let's skip for now and can be given on request
 
@@ -749,7 +796,9 @@ def read_rhs(filename, file_format: str):
     else:
         warnings.warn("Stim not implemented for `one-file-per-channel` due to lack of test files")
 
-    # No supply or aux for rhs files (ie no stream 1 and 2)
+    # No supply or aux for rhs files (ie no stream_id 1 and 2)
+    # We have an error above that requests test files to help if the spec is changed
+    # in the future.
 
     # 3: Analog input channel.
     # 4: Analog output channel.
@@ -772,7 +821,9 @@ def read_rhs(filename, file_format: str):
     for stream_id in [5, 6]:
         for chan_info in stream_id_to_channel_info_list[stream_id]:
             chan_info["sampling_rate"] = sr
-            chan_info["units"] = "a.u."  # arbitrary units TTL for logic
+            # arbitrary units are used to indicate that Intan does not
+            # store raw voltages but only the boolean TTL state
+            chan_info["units"] = "a.u." 
             chan_info["gain"] = 1.0
             chan_info["offset"] = 0.0
             chan_info["dtype"] = "uint16"
@@ -1045,7 +1096,9 @@ def read_rhd(filename, file_format: str):
     for stream_id in [4, 5]:
         for chan_info in stream_id_to_channel_info_list[stream_id]:
             chan_info["sampling_rate"] = sr
-            chan_info["units"] = "a.u."  # arbitrary units TTL for logic
+            # arbitrary units are used to indicate that Intan does not
+            # store raw voltages but only the boolean TTL state
+            chan_info["units"] = "a.u." 
             chan_info["gain"] = 1.0
             chan_info["offset"] = 0.0
             chan_info["dtype"] = "uint16"
@@ -1079,7 +1132,7 @@ def read_rhd(filename, file_format: str):
 
 
 ##########################################################################
-# RHX Zone for Binary Files
+# RHX Zone for Binary Files (note this is for version 3+ of Intan software)
 # This section provides all possible headerless binary files in both the rhs and rhd
 # formats.
 
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -205,7 +205,7 @@ class IntanRawIO(BaseRawIO):
 
         # signals
         signal_channels = []
-        self.native_channel_order = dict()
+        self.native_channel_order = {}
         for c, chan_info in enumerate(self._ordered_channel_info):
             name = chan_info["custom_channel_name"]
             channel_id = chan_info["native_channel_name"]
@@ -235,9 +235,9 @@ class IntanRawIO(BaseRawIO):
 
         for stream_index, stream_id in enumerate(stream_ids_sorted):
             if self.filename.endswith(".rhd"):
-                name = stream_id_to_name_rhd.get(int(stream_id), "")
+                name = stream_id_to_stream_name_rhd.get(int(stream_id), "")
             else:
-                name = stream_id_to_name_rhs.get(int(stream_id), "")
+                name = stream_id_to_stream_name_rhs.get(int(stream_id), "")
 
             signal_streams["name"][stream_index] = name
 
@@ -346,10 +346,7 @@ class IntanRawIO(BaseRawIO):
 
             stream_name = self.header["signal_streams"][stream_index]["name"][:]
             stream_is_digital = stream_name in digital_stream_names
-            if stream_is_digital:
-                field_name = name_to_field_name_digital[stream_name]
-            else:
-                field_name = channel_ids[0]
+            field_name = stream_name if stream_is_digital else channel_ids[0]
 
             size = self._raw_data[field_name].size
 
@@ -409,10 +406,8 @@ class IntanRawIO(BaseRawIO):
 
         stream_name = self.header["signal_streams"][stream_index]["name"][:]
         stream_is_digital = stream_name in digital_stream_names
-        if stream_is_digital:
-            field_name = name_to_field_name_digital[stream_name]
-        else:
-            field_name = channel_ids[0]
+
+        field_name = stream_name if stream_is_digital else channel_ids[0]
 
         shape = self._raw_data[field_name].shape
         dtype = self._raw_data[field_name].dtype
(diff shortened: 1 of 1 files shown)


zm711:
Cool I'll look it over soon. What's the output? Just 1 and 0s? right?

h-mayorquin:
> Cool I'll look it over soon. What's the output? Just 1 and 0s? right?

image


Yes, I was torn between that and bools but at the end I decided to use np.int8 as I don't know if traces are returned as bool for any other case so I rather play with that expectation.

zm711:
Cool that makes sense. I know Intan let's you choose, but I think 1 and 0s are probably the better expectation.

neo/rawio/intanrawio.py
```diff
@@ -201,6 +205,7 @@ def _parse_header(self):
 
         # signals
         signal_channels = []
+        self.native_channel_order = dict()
```
zm711, line 208:
```suggestion
        self.native_channel_order = {}
```

I know others like `dict()`, but `{}` is faster. [review here](https://madebyme.today/blog/python-dict-vs-curly-brackets/)

I like {} better, but since the performance is only about 2x faster if you really prefer dict() we can keep it.

neo/rawio/intanrawio.py
```diff
@@ -335,12 +343,21 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
             mask = self.header["signal_channels"]["stream_id"] == stream_id
             signal_channels = self.header["signal_channels"][mask]
             channel_ids = signal_channels["id"]
-            channel_id_0 = channel_ids[0]
-            size = self._raw_data[channel_id_0].size
+
+            stream_name = self.header["signal_streams"][stream_index]["name"][:]
```
zm711, line 347: Remind me the `[:]` I always forget this :)

neo/rawio/intanrawio.py
```diff
@@ -335,12 +343,21 @@ def _get_signal_size(self, block_index, seg_index, stream_index):
...
+            stream_is_digital = stream_name in digital_stream_names
+            if stream_is_digital:
+                field_name = name_to_field_name_digital[stream_name]
```
zm711, line 350: What is a field name? 

I wonder if we will remember this in the future. It is not:
1) Neo terminology (which we are bound to here--block, stream, segment)
2)  Not clear to me as a word in this case

neo/rawio/intanrawio.py
```diff
@@ -591,7 +648,7 @@ def read_variable_header(f, header):
     ("electrode_impedance_phase", "float32"),
 ]
 
-stream_type_to_name_rhs = {
+stream_id_to_name_rhs = {
```
zm711, line 651: Nice renaming!

neo/rawio/intanrawio.py
```diff
@@ -1065,6 +1120,12 @@ def read_rhd(filename, file_format: str):
 # This section provides all possible headerless binary files in both the rhs and rhd
 # formats.
 
+digital_stream_names = ["USB board digital input channel", "USB board digital output channel"]
+name_to_field_name_digital = {
```
zm711, line 1124: Technically we could just check `stream_name in name_to_field_name_digital` and get rid of line 1123, but you are going for clarity over reduced lines I think.

I would still prefer to think if we have a better naming than `name_to_field_name_digital`, but we are using so many terms maybe this is the best.

zm711 (commented):
First set of comments on my side.

neo/rawio/intanrawio.py
h-mayorquin: You mean the why? One is a reference the other gets the data which I need here to have the string version.

neo/rawio/intanrawio.py
h-mayorquin: This is standard terminology for struct and structured arrrays in numpy:

https://numpy.org/doc/stable/glossary.html#term-field

neo/rawio/intanrawio.py
h-mayorquin: Maybe we can go further as there are other references to type instead of id in the code but I wanted to keep the diff small for the first review.

neo/rawio/intanrawio.py
zm711: Okay perfect! You're right. I never use structured arrays, but Neo and SI use them all the time, so I should get more used to the naming. Maybe it should be 

```
stream_name_to_field_name_digital
```

To be as explicit as possible?

neo/rawio/intanrawio.py
h-mayorquin: The name is bad, I agree. Plus, this is bad practice because there is an implicit coupling between the values here and the lines here:

https://github.com/h-mayorquin/python-neo/blob/2af13478edbe92e55a8d308ef7fc8a29220ada40/neo/rawio/intanrawio.py#L799-L801

In other words, a relationship that is implicit. That should not be there.

neo/rawio/intanrawio.py
h-mayorquin: I will change this on the next commit.

neo/rawio/intanrawio.py
h-mayorquin: Good idea. Let's start with this.

neo/rawio/intanrawio.py
h-mayorquin: OK, I decided that for this cases is better to use the stream_name as the field name (as all the data is packed in one field). This avoids us having to define an artificial channel name (as we were doing) and sidesteps the need to define such mapping between stream and the artificial channel name. It also works nicely with the case of one-file-per-signal as the signal is the stream.

neo/rawio/intanrawio.py
h-mayorquin: OK, I went further and changed signal_type for stream_id in most places. Check it out.

neo/rawio/intanrawio.py
h-mayorquin: I decided what I think is a simpler approach. See below.

neo/rawio/intanrawio.py
h-mayorquin: Done.

neo/rawio/intanrawio.py
h-mayorquin: Done.

zm711:
I'll re-check this out tomorrow. I selfishly need to refactor some of my own code that relies on the old way to return the digital streams. So I will load this on my workstation and test for our uses locally too!

h-mayorquin:
OK, thanks, that will be really useful.

samuelgarcia:
I do not have much for reading this in detail.
My first impression is I do not like too much.
You need it now so lets go and merge it now  but...
...but...
you need to swear that when we implement this famous get_digitalsignal_chunk that is never coming you will help me to revert this back.

zm711:
Yep of course. I have a bunch of test data locally and I can create a simulated file for gin to test all of this for now so we will make sure it is all good before merging :)

h-mayorquin:
Hey Sam, I think it is important for users to be able to extract this information directly without having to rely on formatk knowledge to use it. 

If you have some time please comment on what you don't like about it? Maybe there is a way of addressing that.

zm711:
The problem isn't with how we are doing it. The goal is to have analog data and then digital data. But the way to access the digital data isn't built into Neo yet. So Sam is saying that once we have the digital machinery at the BaseRawIO level we should move all the digital-in and out logic there instead of inside the analog machinery where it technically doesn't belong.

samuelgarcia:
Sorry. Thanks a lot Zach for this extacts translation. This was exactly what I meant.

luiztauffer:
@h-mayorquin @zm711 I just tested it, and it seems to be working fine! Thanks for working on this and sorry for taking me a couple of days to try it out
can we merge it and make a release? this will be needed for TTL-based synchronization for every lab using Intan recordings

zm711:
Thanks for testing @luiztauffer. 

I was planning on testing as well last week, but I will do my best to test this tomorrow. I'll bring up release at our next meeting for this.

zm711:
So I tested this and it seems to work for me. The only thing is that for groups that have already hacked the memmap to pull out digital data using the field name 'DIGITAL-IN' will now have to use `USB board digital input channel`. But seems fine. Let me read through the code and then this is good to merge unless you have any ideas to help prevent breaking the code @h-mayorquin ?

neo/rawio/intanrawio.py
```diff
@@ -719,40 +765,25 @@ def read_rhs(filename, file_format: str):
...
-                chan_info["units"] = "TTL"
-                chan_info["gain"] = 1.0
-                chan_info["offset"] = 0.0
-                chan_info["dtype"] = "uint16"
-                ordered_channel_info.append(chan_info)
-                data_dtype[sig_type] = "uint16"
+    for stream_id in [5, 6]:
+        for chan_info in stream_id_to_channel_info_list[stream_id]:
+            chan_info["sampling_rate"] = sr
+            chan_info["units"] = "TTL"  # arbitrary units TTL for logic
```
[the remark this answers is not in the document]
CodyCBakerPhD: Any SI unit would be preferable, including "n.a." or "a.u."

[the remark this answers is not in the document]
zm711: I'm absolutely neutral on this. As long as we don't give  real units like volts/millivolt etc. So if that helps with your guys' schemata then I'm fine with @h-mayorquin changing this to "a.u." if he wants.

neo/rawio/intanrawio.py
```diff
@@ -608,8 +654,8 @@ def read_rhs(filename, file_format: str):
     with open(filename, mode="rb") as f:
         global_info = read_variable_header(f, rhs_global_header)
 
-        # channels_by_type is simpler than data_dtype because 0 contains 0, 10 and 11 internally
-        channels_by_type = {k: [] for k in [0, 3, 4, 5, 6]}
+        # We use signal_type in the header as stream_id in neo
```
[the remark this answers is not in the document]
h-mayorquin: By all means please do, that was an omission of mine caused by ignorance.

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
h-mayorquin: We don't really use units as far as I remember but I think @CodyCBakerPhD idea is a good one.  I will change it.

h-mayorquin:
> So I tested this and it seems to work for me. The only thing is that for groups that have already hacked the memmap to pull out digital data using the field name 'DIGITAL-IN' will now have to use `USB board digital input channel`. But seems fine. Let me read through the code and then this is good to merge unless you have any ideas to help prevent breaking the code @h-mayorquin ?

Can you show me a quick snippet of how they are accesing it? It is from spikeinterface or neo? I never use the neo layer to access data so I need to think on how the hacking looks like. Maybe there is a way.

neo/rawio/intanrawio.py
```diff
@@ -116,9 +116,13 @@ def _parse_header(self):
 
         filename = Path(self.filename)
 
+        # Input checks
         if not filename.exists() or not filename.is_file():
             raise FileNotFoundError(f"{filename} does not exist")
 
+        if not (self.filename.endswith(".rhd") or self.filename.endswith(".rhs")):
+            raise ValueError(f"{self.filename} is not a valid Intan file. Expected .rhd or .rhs extension")
+
```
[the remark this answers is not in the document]
h-mayorquin: We sort of need both I guess, one tests the file exists and the other that is of the correct path. Otherwise a non-existen file that ends in rhs might cause a confusion error down the line, right?

zm711:
Sure. Before we exposed this to the user you could still get the data by doing
```python
raw_digital_data = reader._raw_data['DIGITAL-IN']

```
Technically this is all private so we can change this and it is on the user to adapt, but if people were hacking to get this data they would need to change for this. I'm not saying we have to adapt for this because it was private previously but it was easy to make code to hack this.

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
zm711: Okay. I guess we could also do your boolean and then add a quick exists(), but I think it is also fine to keep them as two separate checks as well.

neo/rawio/intanrawio.py
```diff
@@ -407,14 +424,19 @@ def _get_analogsignal_chunk_header_attached(self, i_start, i_stop, stream_index,
...
+            sigs_chunk = np.zeros((i_stop - i_start, len(channel_ids)), dtype=dtype)
 
+            for chunk_index, channel_id in enumerate(channel_ids):
+                data_chan = self._raw_data[channel_id]
+                if multiple_samples_per_block:
+                    sigs_chunk[:, chunk_index] = data_chan[block_start:block_stop].flatten()[sl0:sl1]
+                else:
+                    sigs_chunk[:, chunk_index] = data_chan[i_start:i_stop]
+        else:  # For digital data the channels come interleaved in a single field so we need to demultiplex
+            digital_raw_data = self._raw_data[field_name].flatten()
```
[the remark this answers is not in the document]
h-mayorquin: image

When you read the numpy structured arrays you read blocks the blocks as columns.  So you need to reshape this as if you were concatenating all the blocks.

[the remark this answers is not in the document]
zm711: I understand the flattening, but why aren't we using the 

`self._raw_data[field_name].flatten()[sl0:sl1]` for the header attached?

h-mayorquin:
Mmm I see, no I rather punish wrong doers for accessing our private thing : P

More seriously, I think we could do something if they were accessing this through the API so we throw a proper deprecation warning. But because they are just accessing the memmap directly there is no way to throw a reasonable deprecation warning I think.

Another option would be to keep two references to the same memmap but then again because there is no deprecation warning that we can throw there is no way to change it slowly. The pain is now or at some other point and we might as well do it now. 

If someone is relying on that I guess they can pin neo to the last version?

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
h-mayorquin: I think I was lazy to implement this because I thought the memory footprint is small? Would you like to give it a go?

We would need to pass this:

```
            digital_raw_data = self._raw_data[field_name][block_start:block_stop].flatten()[sl0:sl1]
```

And then demuliplex only that piece.

h-mayorquin:
I just realize that I don't have the rights to merge master on the web interface.

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
zm711: I see what you mean. So you always load the full and then take what you want. I don't think it is necessary. This vector should always be relatively small compared to the amplifier data, so I don't think we need to worry about memory footprint. We can do that later if needed...

zm711:
Once tests pass I'll add the comment and then I'll merge after they pass the final time :)

neo/rawio/intanrawio.py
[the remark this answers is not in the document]
h-mayorquin: Mmm I think the two checks point the user to two different problems. I think they should remain separate. You mght have pass the wrong file or your might have passed other file from intan that is not rhs or rhd.  I think we should keep it more or less as it is now.

zm711:
Missing a Path :)

h-mayorquin:
Let me fix it.

zm711:
I'll wait for you to have a chance to read the comments I added and then if you're okay with things, then this should be good to go.

zm711 (approved):

h-mayorquin:
Comments look good to me.

zm711:
Do you know why tests decided not to run?

h-mayorquin:
It is Friday? No idea

zm711:
Found the setting and turned it on so I'll update this branch but from now Neo should give people the option to click a button to update!

Issue #1349, opened 2023-11-16
Issue with reading .pl2 files 


Bosswang12138:
**Describe the bug**
I encountered an issue while attempting to read a .pl2 file using the Plexon2RawIO from the neo library. Here's the snippet I used:

`import neo.rawio`
`from pathlib import Path`

`folderPath = Path(r"D:\PROGRAM")`
`filename = Path("16ch1.pl2")`
`fullfilename = folderPath.joinpath(filename)`

`r = neo.rawio.Plexon2RawIO(filename=fullfilename)`
`r.parse_header()`

The demo pl2 file were downloaded from Plexon, [https://plexon.com/wp-content/uploads/2017/08/4chDemoPL2.pl2](url)
I received the following error traceback:
```
File [d:\Anaconda\envs\si_env\lib\site-packages\neo\rawio\baserawio.py:178](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:178), in BaseRawIO.parse_header(self)
    [165](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:165) def parse_header(self):
    [166](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:166)     """
    [167](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:167)     This must parse the file header to get all stuff for fast use later on.
    [168](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:168) 
   (...)
    [176](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:176) 
    [177](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:177)     """
--> [178](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:178)     self._parse_header()
    [179](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/baserawio.py:179)     self._check_stream_signal_channel_characteristics()

File [d:\Anaconda\envs\si_env\lib\site-packages\neo\rawio\plexon2rawio\plexon2rawio.py:239](file:///D:/Anaconda/envs/si_env/lib/site-packages/neo/rawio/plexon2rawio/plexon2rawio.py:239), in Plexon2RawIO._parse_header(self)
[...]

PR #1495, opened 2024-06-22
Patch for issue #1349 Issue with reading .pl2 files

cheydrick:
I was able to replicate issue #1349, and this appears to resolve it.

### DIFF
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -130,6 +130,11 @@ class Plexon2RawIO(BaseRawIO):
 
         # Open the file.
         self.pl2reader.pl2_open_file(self.filename)
+        
+        # Verify the file handle is valid.
+        if self.pl2reader._file_handle.value == 0:
+            self.pl2reader._print_error()
+            raise Exception(f"Opening {self.filename} failed.")
 
     def _source_name(self):
         return self.filename
--- a/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
+++ b/neo/rawio/plexon2rawio/pypl2/pypl2lib.py
@@ -610,7 +610,7 @@ class PyPL2FileReader:
 
         self.pl2_dll.PL2_GetSpikeChannelInfoByName.argtypes = (
             ctypes.c_int,
-            ctypes.c_char * len(channel_name),
+            ctypes.POINTER(ctypes.c_char),
             ctypes.POINTER(PL2SpikeChannelInfo),
         )
 
@@ -744,7 +744,7 @@ class PyPL2FileReader:
 
         self.pl2_dll.PL2_GetSpikeChannelDataByName.argtypes = (
             ctypes.c_int,
-            ctypes.c_char,
+            ctypes.POINTER(ctypes.c_char),
             ctypes.POINTER(ctypes.c_ulonglong),
             ctypes.POINTER(ctypes.c_ulonglong),
             ctypes.POINTER(ctypes.c_ushort),


h-mayorquin:
This patch works for us in neuroconv when reading from mac, windows and linux.

h-mayorquin:
The only strange thing is that the file handle test sometimes fail in a seemigly random way (see https://github.com/NeuralEnsemble/python-neo/pull/1494) but they pass for other versions of python. Not sure what is going on with them.

h-mayorquin:
I confirm that this patch works on neuroconv, we can read in windows, linux and macs with the modifications in this PR.

Neuroconv PR and where the tests are:
https://github.com/catalystneuro/neuroconv/pull/918

cheydrick:
The file handle test fails every time if I run pytest from the project-neo folder (which runs all tests), but if I run pytest only on test_plexon2rawio.py it passes. I tried this in an environment with Python 3.9 and 3.12 - same behavior.

h-mayorquin:
@cheydrick 
check https://github.com/NeuralEnsemble/python-neo/pull/1494, a simple hack of trying to open it couple of times seems to solve the issue.

I have a question regarding the metadata. The datetime objects in plexon, are they in excell format? that is, are they reference to 1990?

cheydrick:
That's pretty wild - I'll take a look at your implementation.

The date/time is returned as a structure based on this: [https://cplusplus.com/reference/ctime/tm/](https://cplusplus.com/reference/ctime/tm/)

cheydrick:
@h-mayorquin It looks like your pull request #1494 implements the same ctypes fixes as this one, and also actually passes tests by fixing the bad-file-handle problem. Should this PR be closed? I don't know what the protocol is here.

h-mayorquin:
I personall think this PR should be merge so your contribution is registered and then we can work the rest on the other.  After all, I got the pointer fixes from here.

Thank you very much, yes, this confirms my hunch that the base of the date is 1900.

Another question, when I look at the output of the `PL2_GetAnalogChannelInfo ` function channel.m_Name and look a the adc signals I get two:
* `WB{x}` (e.g. WB03, WB05, WB06, etc)
* `SPK{x}` (e.g. SPKC01, SPKC01, SPKC01, etc)

both with a sampling rate of 40_000 

and then also
*  `FP{x}`
* `AI{x}`

With a sampling rate of 1000.

I am assuming that the WB is the wideband but what are the SPK, FP and AI stand for? Is there a place where I can check the return of the ddls or the spec of the format so I can assure myself that the we are writing the correct metadata for the users?

cheydrick:
"WB" is wideband.
"SPKC" is a spike-filtered copy of the wideband. It stands for Spike Continuous, which differentiates it from SPK, which is the spike waveform segments.
"FP" is a field-potential-filtered copy of the wideband.
"AI" stands for "Auxiliary Input", which is an auxiliary analog input device.

There is a header called PL2FileReader.h that has the interfaces to the .dll functions here: [https://plexon.com/wp-content/uploads/2017/08/OmniPlex-and-MAP-Offline-SDK-Bundle_0.zip](https://plexon.com/wp-content/uploads/2017/08/OmniPlex-and-MAP-Offline-SDK-Bundle_0.zip). Within that .zip file is another .zip file called "C++ PL2 Offline Files SDK.zip". Within that .zip file is a folder called "SampleCode" that contains this header.

h-mayorquin:
Thanks a bunch this is super useful for us.

Just to confirm, the "SPKC" is literally:
1) You take the wideband.
2) In the sections where there are spikes, remove them from the wideband.

?

cheydrick:
I think I phrased my explanation poorly, sorry about that.

No. The SPKC is a copy of the wideband that has been high-pass filtered. Nothing is removed, just filtered.

samuelgarcia (approved):

h-mayorquin:
We are merging this, thanks a lot for the fix.

samuelgarcia:
Thanks a lot for this fix.

h-mayorquin:
Hey @cheydrick, wondering one thing here. Is the type of filtering in the `SPKC` fixed and something that we can write as annotation or is it different and depends on system or recording configuration at the moment of the experiment? If the latter is the case, is there a way of extracting that information from the format?

cheydrick:
> Hey @cheydrick, wondering one thing here. Is the type of filtering in the `SPKC` fixed and something that we can write as annotation or is it different and depends on system or recording configuration at the moment of the experiment? If the latter is the case, is there a way of extracting that information from the format?

The filtering is adjustable at the time of recording, though I suspect most stick with the default settings. Unfortunately no, the filter details aren't stored in the recorded file.

h-mayorquin:
Thanks for confirming @cheydrick

PR #1496, opened 2024-06-26
Minimal fix of `SpikeGadgetsRawIO` channel ID issue

khl02007:
Only the part from #1303 that fixes the channel id issue from #1215 
Hopefully this can be merged easily

### DIFF
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -20,6 +20,10 @@ The ".rec" file format contains:
 Author: Samuel Garcia
 """
 
+import numpy as np
+
+from xml.etree import ElementTree
+
 from .baserawio import (
     BaseRawIO,
     _signal_channel_dtype,
@@ -27,10 +31,7 @@ from .baserawio import (
     _spike_channel_dtype,
     _event_channel_dtype,
 )
-
-import numpy as np
-
-from xml.etree import ElementTree
+from neo.core import NeoReadWriteError
 
 
 class SpikeGadgetsRawIO(BaseRawIO):
@@ -89,15 +90,13 @@ class SpikeGadgetsRawIO(BaseRawIO):
         0, 32, 64, 96, 1, 33, 65, 97, ..., 128
         See also: https://github.com/NeuralEnsemble/python-neo/issues/1215
         """
-        x = []
-        for k in range(n_channels_per_chip):
-            x.append(
-                [
-                    k + i * n_channels_per_chip
-                    for i in range(int(n_total_channels / n_channels_per_chip))
-                ]
-            )
-        return [item for sublist in x for item in sublist]
+        ephys_channel_ids_list = []
+        for hw_channel in range(n_channels_per_chip):
+            hw_channel_list = [
+                hw_channel + chip * n_channels_per_chip for chip in range(int(n_total_channels / n_channels_per_chip))
+            ]
+            ephys_channel_ids_list.append(hw_channel_list)
+        return [channel for channel_list in ephys_channel_ids_list for channel in channel_list]
 
     def _parse_header(self):
         # parse file until "</Configuration>"
@@ -123,13 +122,13 @@ class SpikeGadgetsRawIO(BaseRawIO):
 
         self._sampling_rate = float(hconf.attrib["samplingRate"])
         num_ephy_channels = int(hconf.attrib["numChannels"])
-        
+
         # check for agreement with number of channels in xml
         sconf_channels = np.sum([len(x) for x in sconf])
         if sconf_channels < num_ephy_channels:
             num_ephy_channels = sconf_channels
         if sconf_channels > num_ephy_channels:
-            raise ValueError(
+            raise NeoReadWriteError(
                 "SpikeGadgets: the number of channels in the spike configuration is larger than the number of channels in the hardware configuration"
             )
 
@@ -137,7 +136,7 @@ class SpikeGadgetsRawIO(BaseRawIO):
             num_chan_per_chip = int(sconf.attrib["chanPerChip"])
         except KeyError:
             num_chan_per_chip = 32  # default value for Intan chips
-            
+
         # explore sub stream and count packet size
         # first bytes is 0x55
         packet_size = 1
@@ -208,9 +207,7 @@ class SpikeGadgetsRawIO(BaseRawIO):
             signal_streams.append((stream_name, stream_id))
             self._mask_channels_bytes[stream_id] = []
 
-            channel_ids = self._produce_ephys_channel_ids(
-                num_ephy_channels, num_chan_per_chip
-            )
+            channel_ids = self._produce_ephys_channel_ids(num_ephy_channels, num_chan_per_chip)
             chan_ind = 0
             self.is_scaleable = "spikeScalingToUv" in sconf[0].attrib
             if not self.is_scaleable:
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -79,6 +79,26 @@ class SpikeGadgetsRawIO(BaseRawIO):
     def _source_name(self):
         return self.filename
 
+    def _produce_ephys_channel_ids(self, n_total_channels, n_channels_per_chip):
+        """Compute the channel ID labels
+        The ephys channels in the .rec file are stored in the following order:
+        hwChan ID of channel 0 of first chip, hwChan ID of channel 0 of second chip, ..., hwChan ID of channel 0 of Nth chip,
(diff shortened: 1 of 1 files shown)


neo/rawio/spikegadgetsrawio.py
```diff
@@ -103,7 +123,21 @@ def _parse_header(self):
...
+            num_ephy_channels = sconf_channels
+        if sconf_channels > num_ephy_channels:
+            raise ValueError(
+                "SpikeGadgets: the number of channels in the spike configuration is larger than the number of channels in the hardware configuration"
+            )
+
+        try:
+            num_chan_per_chip = int(sconf.attrib["chanPerChip"])
+        except KeyError:
+            num_chan_per_chip = 32  # default value for Intan chips
```
zm711, line 139: Intan also has 64 and 16 channel headstages. Where is this 32 coming from?

khl02007: most spikegadgets products use 64 channel chips, but a 64 channel chip is just two 32 channel chips sharing an output pin so this is a reasonable default for most cases.

neo/rawio/spikegadgetsrawio.py
```diff
@@ -79,6 +79,26 @@ def __init__(self, filename="", selected_streams=None):
...
+        """
+        x = []
+        for k in range(n_channels_per_chip):
+            x.append(
+                [
+                    k + i * n_channels_per_chip
+                    for i in range(int(n_total_channels / n_channels_per_chip))
+                ]
+            )
+        return [item for sublist in x for item in sublist]
```
zm711, line 100:
```suggestion
        channel_ids = [k + (i * n_channels_per_chip) for k in range(n_channels_per_chip) for i in range(int(n_total_channels / n_channels_per_chip)]

        return channel_ids
```

This is probably just me so if @alejoe91 takes a quick peek and is fine with the mixed for loop comprehension, but for me I prefer to keep it all list comprehension or nested for loops.  Mainly because I think flattening the nested lists is always super confusing for me. Let's see what Alessio thinks, but otherwise this looks fine to me.

neo/rawio/spikegadgetsrawio.py
```diff
@@ -103,7 +123,21 @@ def _parse_header(self):
...
         self._sampling_rate = float(hconf.attrib["samplingRate"])
         num_ephy_channels = int(hconf.attrib["numChannels"])
-
+        
+        # check for agreement with number of channels in xml
+        sconf_channels = np.sum([len(x) for x in sconf])
+        if sconf_channels < num_ephy_channels:
+            num_ephy_channels = sconf_channels
+        if sconf_channels > num_ephy_channels:
+            raise ValueError(
```
zm711, line 132:
```suggestion
            raise NeoReadWriteError(
```

I think a value error would imply that the user could fix something. Otherwise a NeoReadWriteError is the way to tell the end-user that this is a file that does not fit with the Neo model. We may need to add this to the imports.

zm711: Yes I just check we need to import the error message!

with 
```python
from neo.core import NeoReadWriteError
```

at the top.

zm711 (commented):
Just one error correction to implement and then a discussion on the ephys channels function. I'll also have Alessio be a second pair of eyes, but once the error is fixed I would be fine with this (with a slight preference to tidying up the function either with my comment or switching to better explained for loops).

zm711:
@khl02007,

the only thing preventing this from getting merged from my perspective is using the appropriate error message. Would you like to do that fix?

zm711:
@alejoe91, I've added my cleanup to this PR. So I'm okay with this. Do you know the spikegadgets format well enough to double check the logic of this?

samuelgarcia (approved):

neo/rawio/spikegadgetsrawio.py
```diff
@@ -79,6 +80,24 @@ def __init__(self, filename="", selected_streams=None):
...
+        hwChan ID of channel 1 of first chip, hwChan ID of channel 1 of second chip, ..., hwChan ID of channel 1 of Nth chip,
+        ...
+        So if there are 32 channels per chip and 128 channels (4 chips), then the channel IDs are:
+        0, 32, 64, 96, 1, 33, 65, 97, ..., 128
+        See also: https://github.com/NeuralEnsemble/python-neo/issues/1215
+        """
+        ephys_channel_ids_list = []
+        for hw_channel in range(n_channels_per_chip):
+            hw_channel_list = [
+                hw_channel + chip * n_channels_per_chip for chip in range(int(n_total_channels / n_channels_per_chip))
```
samuelgarcia, line 96: will this int(n_total_channels / n_channels_per_chip) works for any n_total_channels numbers.
Do we need an explicit ceil or  floor ?

samuelgarcia:
Hi Kyu,
Sorry for the super long delay.
this looks good to me.
I made a very small comments
@zm711 i thinkwe can merge.

samuelgarcia (approved):

neo/rawio/spikegadgetsrawio.py
zm711: I think we should be good. Most of the times you'll have increments of 32 channels/chip. So total channels would have to be some multiple of 32. I don't know gadgets well enough, but I guess the problem would be Intan does sell 16 channel headstage so if you were allowed to mix headstage chips then this math wouldn't work. But I think in this case we wait for a user to open an issue so we have mixed headstage data to work with.

khl02007:
@zm711 @samuelgarcia  sorry I forgot about this PR. In general I'm happy to let you guys make any changes to the PR as long as the functionality is preserved (which you seem to have done). Thanks and hopefully we can access spikegadgets rec files via spikeinterface now.

neo/rawio/spikegadgetsrawio.py
```diff
@@ -104,6 +123,20 @@ def _parse_header(self):
         self._sampling_rate = float(hconf.attrib["samplingRate"])
         num_ephy_channels = int(hconf.attrib["numChannels"])
 
+        # check for agreement with number of channels in xml
+        sconf_channels = np.sum([len(x) for x in sconf])
+        if sconf_channels < num_ephy_channels:
+            num_ephy_channels = sconf_channels
+        if sconf_channels > num_ephy_channels:
```
jnjnnjzch, line 130: Sorry to interrupt, but is that necessarry to add a comparing that num_chan_per_chip > num_ephy_channels?
I got a recording recently and found that num_chan_per_chip is ocasionally read as `339134394`.
I manage to extract the xml file and see
` <SpikeConfiguration categories="" chanPerChip="339134394" device="intan">
`
which implies some error...

neo/rawio/spikegadgetsrawio.py
jnjnnjzch: > most spikegadgets products use 64 channel chips, but a 64 channel chip is just two 32 channel chips sharing an output pin so this is a reasonable default for most cases.
Again, sorry to interrupt... Is this still work in signal channel level? I mean in practice,  does it mean there is no need to change num_chan_per_chip manually into 64 even we know we are using a 64 channel chip, but the header_text xml extracts as `sconf.attrib["chanPerChip"] = 32`?

khl02007: @jnjnnjzch if you are using [this product](https://spikegadgets.com/products/msh-modular-datalogger/) then yes, no need to change `num_chan_per_chip`. If you're using some other product, I'm actually not sure what would happen here...

neo/rawio/spikegadgetsrawio.py
khl02007: do you mean the workspace file you use to collect data has `chanPerChip="339134394"`?

neo/rawio/spikegadgetsrawio.py
jnjnnjzch: Thanks! we are exactly using this product! And your opinion is validated again by checking the signal synchronization of the extracted trace. Thanks a lot!

neo/rawio/spikegadgetsrawio.py
jnjnnjzch: No, we are using the workspace file with `chanPerChip="32"`, but when we check the output `.rec` file, it becomes `chanPerChip="339134394"`. We're still trying to figure out what's going on...

MGAMZ: @jnjnnjzch I also encountered a problem very similar to the one you mentioned. The only difference is that chanPerChip is directly set to 0, which caused a division-by-zero error.

The problem still exists even when using Trodes 2.8.1. It's hard to believe that this issue hasn't been fixed yet. This actually leads to potential data unreliability issues.

I am suggesting that python-neo should check and protect against this problem (https://github.com/NeuralEnsemble/python-neo/pull/1830). Given that you encountered this issue two years ago, could you provide some suggestions?

PR #1500, opened 2024-07-01
Make digital traces match memmap dtypes in neo.

h-mayorquin:
@zm711

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -141,7 +141,7 @@ class IntanRawIO(BaseRawIO):
             (
                 self._global_info,
                 self._ordered_channel_info,
-                data_dtype,
+                memmap_data_dtype,
                 header_size,
                 self._block_size,
                 channel_number_dict,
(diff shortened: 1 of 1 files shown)


neo/rawio/intanrawio.py
```diff
@@ -487,7 +487,8 @@ def _get_analogsignal_chunk_one_file_per_signal(self, i_start, i_stop, stream_in
 
     def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_stop):
 
-        output = np.zeros((i_stop - i_start, len(channel_ids)), dtype=np.uint8)
+        dtype = np.uint16 # We fix this to match the memmap dtype
```
h-mayorquin, line 490: This is the only real change on this PR. All the other is a variable re-name to avoid further confusion.

zm711:
This works for me once tests pass. You want it to be draft for more discussion on SI?

h-mayorquin:
Na, man, your call as a neo mantainer.  I have a slight push to got for the performance improvement but I think you are right that this might better (and is easier). We should close the spikeinterface issue once we merge this.

As an aside, I never thought that it makes sense to run a test again when you mark something from draft to ready.

zm711 (approved):

Issue #1413, opened 2024-02-21
Invalid annotation error when trying to read .nwb file


Idavr:
Greetings everyone!
I hope someone here can help with solving a very basic issue.

Currently I am trying to read a .nwb file using NWBIO like this:

from pynwb import NWBFile, NWBHDF5IO
from neo.io import NWBIO

file = '/745606_20231213-probe0.nwb'

reader = NWBIO(file)
data_neo = reader.read()
print(data_neo)

```
/home/idavalik/anaconda3/envs/elephant/lib/python3.8/site-packages/hdmf/utils.py:668: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.2.0 because version 1.8.0 is already loaded.
  return func(args[0], **pargs)
/home/idavalik/anaconda3/envs/elephant/lib/python3.8/site-packages/hdmf/utils.py:668: UserWarning: Ignoring cached namespace 'core' version 2.2.5 because version 2.6.0-alpha is already loaded.
 return func(args[0], **pargs)

Traceback (most recent call last):
  File "/home/idavalik/Documents/elephant/elephant/assembly_detection.py", line 18, in <module>
    data_neo = reader.read()
  File "/home/idavalik/Documents/elephant/elephant/neo/io/baseio.py", line 133, in read
    return self.read_all_blocks(lazy=lazy, **kargs)
  File "/home/idavalik/Documents/elephant/elephant/neo/io/nwbio.py", line 271, in read_all_blocks
    self._read_stimulus_group(lazy)
  File "/home/idavalik/Documents/elephant/elephant/neo/io/nwbio.py", line 390, in _read_stimulus_group
    self._read_timeseries_group("stimulus", lazy)
  File "/home/idavalik/Documents/elephant/elephant/neo/io/nwbio.py", line 351, in _read_timeseries_group
    segment = self._get_segment(block_name, segment_name)
  File "/home/idavalik/Documents/elephant/elephant/neo/io/nwbio.py", line 291, in _get_segment
    block = Block(name=block_name, **self.global_block_metadata)
  File "/home/idavalik/Documents/elephant/elephant/neo/core/block.py", line 96, in __init__
    super().__init__(name=name, description=description, file_origin=file_origin, **annotations)
  File "/home/idavalik/Documents/elephant/elephant/neo/core/container.py", line 204, in __init__
    super().__init__(name=name, description=description, file_origin=file_origin, **annotations)
[...]

PR #1502, opened 2024-07-02
Handle more complex metadata in NWB files

apdavison:
This PR attempts to address the problems in #1413.

It is a work in progress. The first commit lets the example script run through without errors on my machine, but I plan to take more time to check that the Neo data are an accurate representation of the NWB file contents.

### DIFF
--- a/neo/io/nwbio.py
+++ b/neo/io/nwbio.py
@@ -25,6 +25,7 @@ from json.decoder import JSONDecodeError
 import numpy as np
 import quantities as pq
 
+from neo.core.baseneo import _check_annotations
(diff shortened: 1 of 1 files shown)


Moritz-Alexander-Kern:
Hey, 
thanks for looking into this.

Indeed, now I can load the `.nwb` -files from the DANDI archives without errors.

I get some warnings and 542 `neo.SpikeTrain` and 4 `neo.IrregularlySampledSignal` objects.

Unfortunately I don't know what to expect, I was trying to inspect the `.nwb` file using the pynwb package.

See the following `.ipynb` notebook <https://gist.github.com/Moritz-Alexander-Kern/93412334080a1ce021f2df9ec06244c5>

PR #1505, opened 2024-07-08
RawBinarySignalIRawIO: add `dat` extension as possibility

zm711:
`.dat` is super common for raw binary files. If we want a raw binary reader at the neo level we should make sure it is included in possible extensions.


samuelgarcia (approved):

samuelgarcia:
excellent!

PR #1506, opened 2024-07-08
Remove `'all-in-one'` in last remaining IO and tests

### LINKED ISSUE #1197: remove legacy `signal_group_mode` modes
Some IOs still use the `all-in-one` mode, which according to the BaseIO should not exist (anymore).

zm711:
Fixes #1197.

In draft because I need to see which tests fail so I can fix the tests for the removal of this mode.


zm711:
the more I look at this the more it seems like this has just become a feature of the ascii IO. I think we close this PR and close the associated issue and just say that ascii io has the two options.

zm711:
Let's reassess this at another time. Let me close this one.

PR #1509, opened 2024-07-18
Add NeuronexusRawIO/IO

### LINKED ISSUE #1296: Support for Neuronexus (.XDAT) files
Hi. I am new to analyzing ephys data, and our lab currently uses allego neuronexus files (.xdat). In wanting to use spike interface, we reached out to see if there was support on their end for .xdat files. 

I wanted to reach out to see if there was developed support for those files or if it could be done.

According to spikeinterface people, when they reached out to Neuronexus they did not seem interested in collaboration. I know that there are features that allow one to include their own file format. If that could be done by me, I can try; however, if the spikeinterface developers are unable to support .xdat file, I definitely don't think I will be able to make that happen.

zm711:
Fixes #1296.

I'll try to work on this a bit in my free time. Will switch off draft when ready for review.

Also should help for [neuroconv](https://github.com/catalystneuro/neuroconv/issues/702)

* [x] Add tests
* [x] Get file on GIN


zm711:
No rush on this @apdavison, this is a new io. Since Sam will be going on vacation, I think getting some feedback would be nice, but I can also pester him once he is back. There is no rush on this one :)

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+    BaseRawIO,
+    _signal_channel_dtype,
+    _signal_stream_dtype,
+    _spike_channel_dtype,
+    _event_channel_dtype,
+)
+from neo.core import NeoReadWriteError
+
+
+class NeuronexusRawIO(BaseRawIO):
```
zm711, line 50: I guess the company is technically `NeuroNeuxs` maybe I change that?

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+
+    extensions = ["xdat", "json"]
+    rawmode = "one-file"
+
+    def __init__(self, filename: str | Path = ""):
```
h-mayorquin, line 55:
```suggestion
    def __init__(self, filename: str | Path):
```
Why add a default value here?

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+        ----------
+        filename: str | Path, default: ''
+            The filename of the metadata file should end in .xdat.json
+
+        Notes
+        -----
+        * The format involves 3 files:
+            * The *.xdat.json metadata file
+            * The *_data.xdat binary file of all raw data
+            * The *_timestamps.xdat binary file of the timeestamp data
```
h-mayorquin, line 69:
```suggestion
            * The *_timestamps.xdat binary file of the timestamp data
```

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+        From the metadata the other two files are located within the same directory
```
h-mayorquin, line 70:
```suggestion
        From the metadata, the other two files are located within the same directory
```
This is actually a question to you.

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+        if not Path(filename).is_file():
+            raise FileNotFoundError(f"The metadata file {filename} was not found")
+        if Path(filename).suffix != ".json":
+            raise NeoReadWriteError(
+                f"The json metadata file should be given, filename entered is {Path(filename).stem}"
+            )
+        meta_filename = Path(filename)
+        binary_file = meta_filename.parent / (meta_filename.stem.split(".")[0] + "_data.xdat")
+
+        if not binary_file.exists() and not binary_file.is_file():
```
h-mayorquin, line 102: is_file also does the function of exists so there is some redundancy here.

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+            raise FileNotFoundError(f"The data.xdat file {binary_file} was not found. Is it in the same directory?")
+        timestamp_file = meta_filename.parent / (meta_filename.stem.split(".")[0] + "_timestamp.xdat")
+        if not timestamp_file.exists() and not timestamp_file.is_file():
+            raise FileNotFoundError(
+                f"The timestamps.xdat file {timestamp_file} was not found. Is it in the same directory?"
+            )
+
```
h-mayorquin, line 109: Nice and informative checks!

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+    def _parse_header(self):
+
+        # read metadata
+        self.metadata = self.read_metadata(self.filename)
+
+        # Collect information necessary for memory map
+        self._sampling_frequency = self.metadata["status"]["samp_freq"]
+        self._n_samples, self._n_channels = self.metadata["status"]["shape"]
+        # Stored as a simple float32 binary file
```
h-mayorquin, line 126: Dude, where did you get the specification or how did you figured this one out?

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+            timestamp_file,
+            dtype=np.int64,  # this is from the allego sample reader timestamps are np.int64
+            mode="r",
+            offset=0,  # headerless binary file
+        )
+
+        # We can do a quick timestamp check to make sure it is the correct timestamp data for the
+        # given metadata
+        if self._timestamps[0] != self.metadata["status"]["timestamp_range"][0]:
+            raise NeoReadWriteError(f"The metadata indicates a different starting timestamp")
```
h-mayorquin, line 149: It would be good to show which ones.

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # organize the channels
+        signal_channels = []
+        channel_info = self.metadata["sapiens_base"]["biointerface_map"]
+
+        # as per dicussion with the Neo/SpikeInterface teams stream_id will become buffer_id
+        # and because all data is stored in the same buffer stream for the moment all channels
+        # will be in stream_id = 0. In the future this will be split into sub_streams based on
+        # type but for now it will be the end-users responsability for this.
+        stream_id = 0  # hard-coded see note above
```
h-mayorquin, line 159: I think this should be string to not add more confusing to the already confusing stream_id vs stream_index. At least type increases the gap.

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+        with open(fname_metadata, "rb") as read_file:
+            metadata = json.load(read_file)
+
+        return metadata
+
+
+# this is pretty useless right now, but I think after a
+# refactor with sub streams we could adapt this for the sub-streams
+# so let's leave this here for now :)
+stream_id_to_stream_name = {0: "Neuronexus Allego Data"}
```
h-mayorquin, line 275: Yei!

neo/test/iotest/test_neuronexusio.py
```diff
@@ -0,0 +1,21 @@
...
+from neo.test.iotest.common_io_test import BaseTestIO
+
+
+class TestNeuronexusIO(
+    BaseTestIO,
+    unittest.TestCase,
+):
+    ioclass = NeuronexusIO
+    entities_to_download = ["neuronexus"]
+    entities_to_test = ["neuronexus/allego_1/allego_2__uid0701-13-04-49.xdat.json"]
```
h-mayorquin, line 17: is this already in gin? Nice : D

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def _get_signal_t_start(self, block_index, seg_index, stream_index):
+
+        t_start = self.metadata["status"]["t_range"][0]
+        return t_start
+
+    #######################################
+    # Helper Functions
+
+    def read_metadata(self, fname_metadata):
```
h-mayorquin, line 251: I like to add path (hungarian notation) to differentiate between file handles (output of f = open() and paths) but I think name works as well!

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,275 @@
...
+It appears that the metadata['status'] provides most of the information necessary
+for generating the initial memory map (global sampling frequency), n_channels,
+n_samples.
+
+metadata['sapiens_base']['biointerface_map'] provides all the channel specific information
+like channel_names, channel_ids, channel_types.
+
+An additional note on channels. It appears that analog channels are called `pri` or
+`ai0` within the metadata where as digital channels are called `din0` or `dout0`.
+In this first implementation it is up to the user to do the appropriate channel slice
```
h-mayorquin, line 27: What do you mean by this?

h-mayorquin (approved):
Wow, can I approve things? How comes?  Just testing.

Did the first reading, this looks pretty good to me. A bunch of minor comments.

neo/rawio/neuronexusrawio.py
zm711: It's been like that for ages. Every RawIO has a default. There has been a discussion about switching the default to None, but there is always a default.

neo/rawio/neuronexusrawio.py
zm711: There's not really a hard rule here. Growing up I was taught that 4 words is the dividing line between comma and no comma but part of this is stylistic. Do you want to emphasize the separation of the thoughts or not.

neo/rawio/neuronexusrawio.py
zm711: I think this might be residual from an old implementation that I switched. I think the original idea was someone might get a folder (which could exist) and so I added the check for being a file. But not sure anymore why I put both in. Haha.

neo/rawio/neuronexusrawio.py
zm711: Do you want to do an actual suggestion? You want me to compare the timestamp file from the metadata file in the actual error?

neo/test/iotest/test_neuronexusio.py
zm711: Yep. I got it in now that I have gin working on my laptop thanks to you!

neo/rawio/neuronexusrawio.py
zm711: This was a note for myself. I don't know what the naming convention is for neuronexus (like what does pri actually stand for?). So what do we want to expose to the end user. Their `pri` names or their `ai0` names. We could do either?

neo/rawio/neuronexusrawio.py
zm711: I think the parsing is super fast, but one thing I was thinking of was we could allow think about parsing more of header in this function then in spikeinterface or in neuroconv stuff you could run this function separately if you needed separate metadata then we could add a check in the main body that if this is already run we don't need to re-run. But the thought was inchoate (because honestly this function is so small right now). Maybe next time we have a call we can brainstorm on this because I don't think I'm explaining myself super well.

As far as naming, I just didn't want to use the same name so I added the fname tag onto it. But happy to change if it helps in conceptualizing more.

neo/rawio/neuronexusrawio.py
zm711: I need to check with @samuelgarcia, I honestly can't remember if he had any typing rules for the streams or not. I thought both indexes and ids were ints for the streams, but maybe that was just me seeing a pattern was there wasn't one.

neo/rawio/neuronexusrawio.py
zm711: Thanks.!

neo/rawio/neuronexusrawio.py
zm711: Okay that's fair. I'll fix it in all places tomorrow !

neo/rawio/neuronexusrawio.py
h-mayorquin: Any idea why? Seems wrong to me : P

neo/rawio/neuronexusrawio.py
h-mayorquin: So you tried the different types until it worked?

neo/rawio/neuronexusrawio.py
h-mayorquin: I mean showing/displaying the two timestamps in the error assertion.

neo/rawio/neuronexusrawio.py
h-mayorquin: There are definitley some stream_ids that are strings. Plexon that I was just working on for example.

neo/rawio/neuronexusrawio.py
h-mayorquin: Ah, got you, I thought you wer talking about slicing across channels.

neo/rawio/neuronexusrawio.py
zm711: Works for me then. Will change when I change names.

neo/rawio/neuronexusrawio.py
zm711: No that decision was before my time. Maybe at some point it was useful to initialize the class without an actual file for something in the testing suite? But honestly not sure. I thought it was strange to, but I just live with it now. :)

neo/rawio/neuronexusrawio.py
zm711: Yes.... And I looked through comments in some thread where someone thought it was float32. So I iterated a little bit and found things that looked correct. For this to have no gain or offset it basically had to be some type of float.

neo/rawio/neuronexusrawio.py
zm711: That's what I thought you meant but wanted to confirm. Yep I will add that.

neo/rawio/neuronexusrawio.py
zm711: Sorry yeah the plexon issue with unclear channel naming was on my mind. So since we aren't do sub-streaming yet both the analog and digital channels all come out in the same stream (or buffer since I like the buffer term better). So the end-user needs to understand the channel naming scheme if they want to slice out the digital data and only look at the amplifier data.

h-mayorquin: Ah, perfect, that clarifies it all. So in the new terminology you are providing buffer based streams and once someone with more expertise on the format comes along then we can improve it to use "logic" based streams. Right?

neo/rawio/neuronexusrawio.py
h-mayorquin: Yes, not that you mentioned it, I think at some point I removed one and it broke down stuff in the part of neo that I don't use : O (the not raw part)

neo/rawio/neuronexusrawio.py
zm711: Yeah basically. I could technically figure it out, but since this is a buffer I figured I would leave as is for now to see what Sam was thinking of for the stream api to minimize the surface area of fixes. If a user complains then I could just implement something myself.

neo/rawio/neuronexusrawio.py
zm711: Done.

neo/rawio/neuronexusrawio.py
zm711: Done

neo/rawio/neuronexusrawio.py
zm711: Just left for now.

neo/rawio/neuronexusrawio.py
zm711: Fixed.

zm711:
It would be relatively easy. My only concern is that if we have a universal way to do this then we are adding more work to adapt all of these rawios into the new format. But in this case the last six channels are reserved for the non-ephys data. So I could split these off intro logical streams.

zm711:
Or we could merge this as is and then iterate for the logical streams in a new PR. That way if the buffer -> stream api happens before we have time then we fold that in and if this becomes an issue for users we can quickly fix it upon request before the api is added.

zm711:
@samuelgarcia, this one can also take a review in your queue. Just pinging you now that you're back here. We have a couple requests on spikeinterface (and on neuroconv) for this. At your leisure of course.

zm711:
C'est le ping à Sam de la semaine.

h-mayorquin:
ceci n'est pas une ping

zm711:
He's on the thread. He'll see it :P

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,281 @@
...
+        self.header = {}
+        self.header["nb_block"] = 1
+        self.header["nb_segment"] = [1]
+        self.header["signal_streams"] = signal_streams
+        self.header["signal_channels"] = signal_channels
+        self.header["spike_channels"] = spike_channels
+        self.header["event_channels"] = event_channels
+
+        # Add the minimum annotations
+        self._generate_minimal_annotations()
```
samuelgarcia, line 219: Any chance to have a rec_datetime in the metadata dict ?

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,281 @@
...
+        if i_start is None:
+            i_start = 0
+        if i_stop is None:
+            i_stop = self._get_signal_size(block_index, seg_index, stream_index)
+
+        if channel_indexes is None:
+            channel_indexes = slice(None)
+
+        raw_data = self._raw_data
+        return raw_data[i_start:i_stop, channel_indexes]
```
samuelgarcia, line 237: If I remember correctly, double slicing sometimes do not play so well when different kind of slicing

```suggestion
        raw_data = self._raw_data[i_start:i_stop, :]
        if channel_indexes is None:
            raw_data = raw_data[:, channel_indexes]
        return raw_data
```

samuelgarcia:
Salut Zach.
This looks good to me.
I made one or two comments.
Sorry for the long delay.

neo/rawio/neuronexusrawio.py
zm711: I don't remember. Let me check!

neo/rawio/neuronexusrawio.py
zm711: Works for me. I didn't know double-slicing could be bad (I think we do it in other places), so happy to split it!

samuelgarcia: It is a bit fuzzy but some stuff like mixing slice and bool mask in fancy indexing can be strange. Maybe it is not anymore the case. But maybe we have this in many places in neo. SO no worry

zm711: I'm already fixing it :) We were missing one thing.

neo/rawio/neuronexusrawio.py
zm711: Okay so I'm adding to the header like this:

```
self.header['rec_datetime'] = self.metadata['status']['start_time']
```
It is setup like: date and time. Do you want any other info?

zm711:
Thanks Sam. I think this is ready to go now after your feedback!

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,285 @@
...
+    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes):
+
+        if i_start is None:
+            i_start = 0
+        if i_stop is None:
+            i_stop = self._get_signal_size(block_index, seg_index, stream_index)
+
+        raw_data = self._raw_data[i_start:i_stop, :]
+
+        if channel_indexes is None:
```
samuelgarcia, line 237:
```suggestion
        if channel_indexes is not None:
```

samuelgarcia: My bad. Sorry.

samuelgarcia (approved):

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,285 @@
...
+        self.header["nb_segment"] = [1]
+        self.header["signal_streams"] = signal_streams
+        self.header["signal_channels"] = signal_channels
+        self.header["spike_channels"] = spike_channels
+        self.header["event_channels"] = event_channels
+
+        # Add the minimum annotations
+        self._generate_minimal_annotations()
+
+        self.header['rec_datetime'] = self.metadata['status']['start_time']
```
samuelgarcia, line 221: @zm711 could check this instead ?
the rec_datetime must be a real datatime and this must routed inside raw_annotaion at block and segment level for instance like this

```suggestion
        rec_datetime = datetime.datetime(self.metadata['status']['start_time'])
        bl_annotations = self.raw_annotations["blocks"][0]
        seg_annotations = bl_annotations["segments"][0]
        for d in (bl_annotations, seg_annotations):
            d["rec_datetime"] = rec_datetime
```

neo/rawio/neuronexusrawio.py
```diff
@@ -0,0 +1,304 @@
...
+        minute = int(time_info[1])
+        second = int(time_info[2])
+        microsecond = 1000 * 1000 * (float(time_info[2]) - second) # second -> micro is 1000 * 1000
+
+        rec_datetime = datetime.datetime(year, month, day, hour, minute, second, microsecond)
+        bl_annotations = self.raw_annotations["blocks"][0]
+        seg_annotations = bl_annotations["segments"][0]
+        for d in (bl_annotations, seg_annotations):
+            d["rec_datetime"] = rec_datetime
+
```
zm711, line 241: @samuelgarcia so I don't do as much date time stuff you want to look at this. Basically we get this string. I don't know what "T" inside the string or the 4:00 at the end (maybe duration in seconds?) so I did some ugly parsing. If you have a better idea from the parsing let me know!

zm711: I guess my fear with this type of parsing is that if they change how they encode it we end up with giant parsing logic.

h-mayorquin: 2024-07-01T13:04:49.4972245-04:00
date_time_{T}_{timestamps}_{time_zone}

zm711: @h-mayorquin if you have a good system could you actually make a PR off of this one and I'll merge it into this. I'm trash at this parsing and it seems like you have a better idea of how to pull this out!

zm711: We are ready to merge this @h-mayorquin  unless you have a better idea for parsing. Do you want me to hold off until you take a look or good to merge and we can prettify this bit later?

samuelgarcia (approved):

samuelgarcia:
merci

PR #1512, opened 2024-07-25
Added a sorted to the listdir of the files in neuralynx folder to avo…

Antho2422:
This pull request is fixing an issue coming from this : https://github.com/NeuralEnsemble/python-neo/pull/1440

The issue was causing mixed channel ids : https://github.com/SpikeInterface/spikeinterface/issues/3248 when loading data into spikeinterface.

We need to keep the `sorted()` when doing an `os.listdir()` here because the list directory depends on the os. 
This avoids mixing the filenames when reading the headers. 

Thanks,
Anthony


zm711:
Thanks @Antho2422,

We will merge after tests pass!

PR #1513, opened 2024-07-26
buffer description api in neorawio and xarray reference API bridge

samuelgarcia:
Given the idea of @bendichter and this [gist](https://gist.github.com/bendichter/30a9afb34b2178098c99f3b01fe72e75)

This is an implementation of  "buffer_description_api" for reading analogsignal chunk.
This 100% backward compatible.

The idea/goal is:
  * have a simple buffer description for analogsignal for raw binary (np.memmap) or hdf5 cases (maybe more)
  * this description can be exported to json
  * this description can be easily transformed to the xarray/zarr reference external API 
  * make so file cloud ready for streaming chunk of traces

See also [kerchunk](https://github.com/fsspec/kerchunk)


neo/rawio/xarray_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def to_xarray_reference_api(rawio_reader, block_index=0, seg_index=0):
+    """
+    Transform the buffer_description_api into a dict ready for the xarray API 'reference://'
+    """
+    rfs = dict()
+    rfs["version"] = 1
+    rfs["refs"] = dict()
+    rfs["refs"][".zgroup"] = json.dumps(dict(zarr_format=2))
```
bendichter, line 14: You don't need the dumps command anymore

neo/rawio/xarray_utils.py
```diff
@@ -0,0 +1,95 @@
...
+            zattrs = dict(
+                _ARRAY_DIMENSIONS=[f'time_{stream_id}', f'channel_{stream_id}'],
+            )
+            # unique big chunk
+            # TODO : optional split in several small chunks
+            array_size = np.prod(descr["shape"], dtype='int64')
+            chunk_size = array_size * dtype.itemsize
+            rfs["refs"][f"{stream_id}/0.0"] = [str(descr["file_path"]), descr["file_offset"], chunk_size]
+            rfs["refs"][f"{stream_id}/.zarray"] =json.dumps(zarray)
+            rfs["refs"][f"{stream_id}/.zattrs"] =json.dumps(zattrs)
```
bendichter, line 48: Nor here

bendichter:
Looks good!

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+      * the one that can use :func:`np.memmap()` internally
+      * format based on hdf5
+       
+    There are many format that do not handle at all this concept:
+
+      * the one that use an external python package for reading data (edf, ced, plexon2, ...)
+      * the one with a complicated data layout (data block are splited without structure)
+
+    To check if a format have the concept you can check the class attribute flag `has_buffer_description_api` of the
+    rawio class.
```
samuelgarcia, line 309: @zm711 @h-mayorquin.
Have a look to this, this is a very drafty start of documenting the stream and buffer concept.

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
     In [42]: print(ev_times)
     [ 0.0317]
 
+Signal streams and signal buffers
+---------------------------------
+
+For reading analog signal **rawio** has 2 important concepts:
```
zm711, line 287:
```suggestion
For reading analog signals **neo.rawio** has 2 important concepts:
```

I think we want to say neo rawio here no? The signals needs to be plural either way.

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+
+ 1. The **signal_stream** : it is a group of channels that can be read together using :func:`get_analog_signal_chunk()`.
+    This group of channels are guaranteed to have the sample sampling rate, and the same duration per segment.
```
zm711, line 290:
```suggestion
    This group of channels is guaranteed to have the same sampling rate, and the same duration per segment.
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    Most of the time, this group channel is a "logical" group of channel. In hsort there are from the same headstage
```
zm711, line 291:
```suggestion
    Most of the time, this group of channel is a "logical" group of channels. In short they are from the same headstage
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    or from the auxilary board.
```
zm711, line 292:
```suggestion
    or from the same auxiliary board.
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    Optionally, depending the format, a **signal_stream** can be a slice or an entire **signal_buffer**.
```
zm711, line 293:
```suggestion
    Optionally, depending on the format, a **signal_stream** can be a slice of or an entire **signal_buffer**.
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+
+ 2. The **signal_buffer** : it is group channel that share the same data layout in a file. The most simple example
```
zm711, line 295:
```suggestion
 2. The **signal_buffer** : it is group of channels that share the same data layout in a file. The most simple example
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    is channel that can be read by a simple :func:`signals = np.memmap(file, shape=..., dtype=... , offset=...)`.
+    A **signal_buffer** can contain one or several **signal_stream** (very often it is only one).
```
zm711, line 297:
```suggestion
    A **signal_buffer** can contain one or several **signal_stream**'s (very often it is only one).
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    There are two kind of format that handle this concept:
```
zm711, line 298:
```suggestion
    There are two kind of formats that handle this concept:
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+     
+      * the one that can use :func:`np.memmap()` internally
```
zm711, line 300:
```suggestion
      * those which use :func:`np.memmap()` internally
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+      * format based on hdf5
```
zm711, line 301:
```suggestion
      * formats based on hdf5
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+       
+    There are many format that do not handle at all this concept:
```
zm711, line 303:
```suggestion
    There are many formats that do not handle this concept:
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+
+      * the one that use an external python package for reading data (edf, ced, plexon2, ...)
```
zm711, line 305:
```suggestion
      * the ones that use an external python package for reading data (edf, ced, plexon2, ...)
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+      * the one with a complicated data layout (data block are splited without structure)
```
zm711, line 306:
```suggestion
      * the ones with a complicated data layout (e.g. those where the data blocks are split without structure)
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+
+    To check if a format have the concept you can check the class attribute flag `has_buffer_description_api` of the
```
zm711, line 308:
```suggestion
    To check if a format makes use of the buffer api you can check the class attribute flag `has_buffer_description_api` of the
```

zm711 (commented):
Looks pretty good to me.

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
     In [42]: print(ev_times)
     [ 0.0317]
 
+Signal streams and signal buffers
+---------------------------------
+
+For reading analog signals **neo.rawio** has 2 important concepts:
+
+ 1. The **signal_stream** : it is a group of channels that can be read together using :func:`get_analog_signal_chunk()`.
+    This group of channels is guaranteed to have the same sampling rate, and the same duration per segment.
```
h-mayorquin, line 290: Now that we have logical channels, this should be same units as well, right?

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+    Most of the time, this group of channel is a "logical" group of channels. In short they are from the same headstage
+    or from the same auxiliary board.
+    Optionally, depending on the format, a **signal_stream** can be a slice of or an entire **signal_buffer**.
+
+ 2. The **signal_buffer** : it is group of channels that share the same data layout in a file. The most simple example
+    is channel that can be read by a simple :func:`signals = np.memmap(file, shape=..., dtype=... , offset=...)`.
+    A **signal_buffer** can contain one or several **signal_stream**'s (very often it is only one).
+    There are two kind of formats that handle this concept:
+     
+      * those which use :func:`np.memmap()` internally
```
h-mayorquin, line 300:
```suggestion
      * Formats which use :func:`np.memmap()` internally
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+      * formats based on hdf5
```
h-mayorquin, line 301:
```suggestion
      * Formats based on hdf5
```

doc/source/rawio.rst
```diff
@@ -281,6 +281,32 @@ Read event timestamps and times
...
+       
+    There are many formats that do not handle this concept:
+
+      * the ones that use an external python package for reading data (edf, ced, plexon2, ...)
+      * the ones with a complicated data layout (e.g. those where the data blocks are split without structure)
```
h-mayorquin, line 306:
```suggestion
      * the ones with a complicated data layout (e.g. those where the data blocks are split without structure like Intan's single file format)
```

neo/rawio/axonrawio.py
```diff
@@ -115,7 +116,7 @@ def _parse_header(self):
             head_offset = info["sections"]["DataSection"]["uBlockIndex"] * BLOCKSIZE
             totalsize = info["sections"]["DataSection"]["llNumEntries"]
 
-        self._raw_data = np.memmap(self.filename, dtype=sig_dtype, mode="r", shape=(totalsize,), offset=head_offset)
+        # self._raw_data = np.memmap(self.filename, dtype=sig_dtype, mode="r", shape=(totalsize,), offset=head_offset)
```
h-mayorquin, line 119: Should this be commented?

neo/rawio/baserawio.py
```diff
@@ -149,6 +151,9 @@ class BaseRawIO:
 
     rawmode = None  # one key from possible_raw_modes
 
+    # If true then
```
h-mayorquin, line 154: This comment is not clear, if what is True?

neo/rawio/baserawio.py
```diff
@@ -1284,6 +1298,7 @@ def _get_signal_size(self, block_index: int, seg_index: int, stream_index: int):
 
         All channels indexed must have the same size and t_start.
         """
+        # must NOT be implemented if has_buffer_description_api=True
```
h-mayorquin, line 1301: Will be ignored?

neo/rawio/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        if num_channels is None:
+            shape = (size,)
+        else:
+            shape = (size // num_channels, num_channels)
+        return shape
+
+
+def get_memmap_chunk_from_opened_file(fid, num_channels,  start, stop, dtype, file_offset=0):
+    """
+    Utility fonction to get a chunk as a memmap array directly from an opened file.
```
h-mayorquin, line 22: typo fonction

h-mayorquin (commented):
I did a first reading. I think that this PR and the future documentation would benefit greatly if we have the schema of the buffer description somewhere. It does not have to be a formal schema (although that would be great) it could be just a description on the documentation or a python data class with types. Something that I can reference to see what should I expect to fill when I am doing a buffer like this.

doc/source/rawio.rst
zm711: I would argue that intan is split with structure. Each block is organized the exact same way. But I get the point. I think maybe the "split without structure" is actually the unclear portion of this statement.

neo/rawio/axonrawio.py
```diff
@@ -169,7 +174,17 @@ def _parse_header(self):
...
                 length /= fSynchTimeUnit
 
-            self._raw_signals[seg_index] = self._raw_data[pos : pos + length].reshape(-1, nbchannel)
+            # self._raw_signals[seg_index] = self._raw_data[pos : pos + length].reshape(-1, nbchannel)
+
+            self._buffer_descriptions[0][seg_index] = {}
+            self._buffer_descriptions[0][seg_index][buffer_id] = {
+                "type" : "raw",
+                "file_path" : str(self.filename),
+                "dtype" : str(sig_dtype),
```
zm711, line 183: is this for jsonification? rather than store a np.dtype()?

neo/rawio/baserawio.py
```diff
@@ -182,6 +184,15 @@ def __init__(self, use_cache: bool = False, cache_path: str = "same_as_resource"
         self.header = None
         self.is_header_parsed = False
 
+        self._has_buffer_description_api = False
+
+    def has_buffer_description_api(self):
```
zm711, line 189:
```suggestion
    def has_buffer_description_api(self) -> bool:
```
I know you don't like typing, but we make this mistake too often distguishing the function from the attribute. So I want this to say it returns bool so we remember it is the function. I hope this will prevent problems where just saying the function returns truthy.

neo/rawio/baserawio.py
```diff
@@ -1350,6 +1362,149 @@ def _rescale_event_timestamp(self, event_timestamps: np.ndarray, dtype: np.dtype
...
+                    for buffer_id, fid in self._memmap_analogsignal_buffers[block_index][seg_index].items():
+                        fid.close()
+            del self._memmap_analogsignal_buffers
+
+        if hasattr(self, '_hdf5_analogsignal_buffers'):
+            for block_index in self._hdf5_analogsignal_buffers.keys():
+                for seg_index in self._hdf5_analogsignal_buffers[block_index].keys():
+                    for buffer_id, h5_file in self._hdf5_analogsignal_buffers[block_index][seg_index].items():
+                        h5_file.close()
+
```
zm711, line 1507: no del here?

neo/rawio/brainvisionrawio.py
```diff
@@ -78,13 +80,23 @@ def _parse_header(self):
...
+        
+        stream_id = "0"
+        buffer_id = "0"
+        self._buffer_descriptions = {0 :{0 : {}}}
+        self._stream_buffer_slice = {}
+        shape = get_memmap_shape(binary_filename, sig_dtype, num_channels=nb_channel, offset=0)
+        self._buffer_descriptions[0][0][buffer_id] = {
+            "type" : "raw",
+            "file_path" : binary_filename,
+            "dtype" : sig_dtype,
```
zm711, line 92: Here you didn't stringify like above. Why?

zm711 (commented):
couple more comments from me too.

samuelgarcia:
Merci beaucoup @zm711  and @h-mayorquin for the review

doc/source/rawio.rst
samuelgarcia: At the moment not yet. The API do not enforce and ensure this.
See this https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/baserawio.py#L118
Ideally units should be add but some IO are mixing maybe the units in the same stream.

neo/rawio/axonrawio.py
samuelgarcia: You lean removed ?
Done

neo/rawio/baserawio.py
samuelgarcia: No this is done in the BaseRawWithBufferApiIO sub class

neo/rawio/axonrawio.py
samuelgarcia: If a data layout is split with a structure theorically we could handle it in zarr giving the list of chunks.
This could be done a the futur but this will not be in the actual PR.

samuelgarcia: str(sig_dtype) is for jsonification yes.

neo/rawio/baserawio.py
samuelgarcia: oups yes.
tahnks

doc/source/rawio.rst
samuelgarcia: Yes when there is a structure we could handle it in the API but I will it later so intan is not teh best example

neo/rawio/axonrawio.py
zm711: There are some without this safety step. So if we really want to be safe we should do this for all uses of dtype and Path no?

neo/rawio/axonrawio.py
h-mayorquin: Thanks, yeah, we can get stuff from the diff.

doc/source/rawio.rst
samuelgarcia: There are some tricky cases where the units is manually done by the user for some formats.
not sure but maybe spike2 is in that case.
What should we do ? split all groups of same units in separate streams ? this could be an good idea but we need first a deeper check.

samuelgarcia:
@zm711 @h-mayorquin @alejoe91 
This is OK on my side.
test are passing I think we can merge this soon.
we will be able to update stream and buffer on reader per reader basis later.

doc/source/rawio.rst
h-mayorquin: Agree. I think that the concept of stream should include same unit across the stream but this we will need to solve in a case by case basis (the deeper check that you mention).

zm711 (approved):

samuelgarcia:
yeah!!
merci fois pour l'aide.

PR #1514, opened 2024-07-26
Release Notes for `0.13.2` + author update + codemeta.json

zm711:
+ I switched over PRs to 0.13.2 if they qualified. I didn't do the same for issues....


apdavison (approved):
Looks good to me. Would you like to make the release @zm711 ? I've sent you a PyPI invitation.

zm711:
Are we doing the action? It looks like we last did the pypi-action in February, but we did the pypi-test action in May?

apdavison:
I think what I did last time was to run `Release to Test PyPI`, and when that passed I did the release manually, but I see no reason not to use the `Release to PyPI` action. I think I just forgot it existed last time.

PR #1518, opened 2024-08-01
After release

zm711:
@apdavison actions all worked out. Created the new tag, test and auto release. Here is the after release PR. I would have committed it directly but I do have a question.  where do you find the downloadurl for the pythonhosted. The url I found, doesn't seem to look like the current link.


codemeta.json
```diff
@@ -4,15 +4,15 @@
   "license": "https://spdx.org/licenses/BSD-3-Clause",
   "codeRepository": "https://github.com/NeuralEnsemble/python-neo",
   "contIntegration": "https://github.com/NeuralEnsemble/python-neo/actions",
-  "dateModified": "2024-05-07",
-  "downloadUrl": "https://files.pythonhosted.org/packages/8b/75/a9a841dc8f5a4198fe3f7174f096726a8b71444670be3abd6ac168de44d8/neo-0.13.1.tar.gz",
+  "dateModified": "2024-08-01",
+  "downloadUrl": "https://files.pythonhosted.org/packages/8b/75/a9a841dc8f5a4198fe3f7174f096726a8b71444670be3abd6ac168de44d8/neo-0.13.1.tar.gz where?",
```
zm711, line 8: This is the link I can't find for the latest version.

zm711: https://pypi.org/project/neo/#files

This is the link I find with the tar.gzip ?

apdavison: yes, that's where I usually get the download link from to update the codemeta file. You can also find it here: https://pypi.org/pypi/neo/0.13.2/json

PR #1520, opened 2024-08-05
Fix electrode selection in Axona raw recording

### LINKED ISSUE #1454: Electrode selection in Axona raw recording
I'm trying to convert electrophysiological data from Axona to NWB using the code from the Neuroconv [https://github.com/catalystneuro/neuroconv](url).

I'm using neuroconv  version 0.4.9 with Python 3.9 on Windows 11 in a conda environment.

I’m selecting the tetrodes to extract from the `.set` file setting the parameters collectMask_* to 1
In the experiment lab I know they’re recording with Axona on channels 17-32, so I'm setting collectMask_* to 1 for tetrodes 5-8. But after running the conversion and plotting the extracted signals, I don't get the correct channels (I can see it's only noise in the output).
This is the code I'm using for conversion:
```python
from dateutil import tz
from pathlib import Path
from neuroconv.datainterfaces import AxonaRecordingInterface

interface = AxonaRecordingInterface(file_path=".../2509202301.bin", verbose=True, es_key="Ephys")
 
# Extract what metadata we can from the source files
metadata = interface.get_metadata()
# For data provenance we add the time zone information to the conversion
tzinfo = tz.gettz()
session_start_time = metadata["NWBFile"]["session_start_time"]
metadata["NWBFile"].update(session_start_time=session_start_time.replace(tzinfo=tzinfo))
 
# Choose a path for saving the nwb file and run the conversion
path_to_save_nwbfile = ".../2509202301.nwb"
nwbfile_path = f"{path_to_save_nwbfile}"
interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
```

But after running the conversion and plotting the extracted signals, I don't get the correct channels (I can see it's only noise in the output).
This is the plot code:
```python
from pynwb import NWBHDF5IO
import numpy as np
 
fileobject = NWBHDF5IO('2509202301.nwb', 'r')
nwbobject = fileobject.read()
 
xx = nwbobject.acquisition['Ephys'].data
time = np.linspace(0,1000, 100000)

fig, ax = plt.subplots(16,1, sharex='all', sharey='all')
for i in range(16):
    ax[i].plot(time, xx[100000:200000,i])
plt.show()
```
And this is the plot:
image

If instead, I try to extract the first 32 channels (so setting  collectMask_* to 1 for tetrodes 1-8) I'm getting the correct signals in channels 17-32 (here I'm plotting only the last 16 channels).
image

It seems that in any case I'm extracting the first 16 channels if I select 4 tetrodes, or 32 if I'm selecting 8 tetrodes.

See: _Originally posted by @CodyCBakerPhD in https://github.com/catalystneuro/neuroconv/issues/787#issuecomment-2032560528_

> In general, it appears as if upstream control over Axona groups/streams is not exposed as my  minimal attempt to get the RecordingExtractor results in only 16 channels (not 32) and only a single group (well, `None` implying only a single group anyway)
> 
> ```python
> from spikeinterface.extractors import AxonaRecordingExtractor
> 
> recording = AxonaRecordingExtractor(file_path=".../2509202301.bin")
> 
> recording.channel_ids
> > array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15'], dtype='<U64')
> 
> recording.get_channel_groups()
> > None
> ```
 
I can share the sample data via email if needed. Thanks!

letiziasignorelli:
I'm proposing this PR to fix #1454

@zm711


neo/rawio/axonarawio.py
```diff
@@ -629,12 +642,13 @@ def _get_signal_chan_header(self):
...
 
+        first_channel = (active_tetrode_set[0] - 1)*elec_per_tetrode
         sig_channels = []
         for itetr in range(num_active_tetrode):
 
             for ielec in range(elec_per_tetrode):
-                cntr = (itetr * elec_per_tetrode) + ielec
-                ch_name = f"{itetr + 1}{letters[ielec]}"
+                cntr = (itetr * elec_per_tetrode) + ielec + first_channel
+                ch_name = "{}{}".format(itetr + active_tetrode_set[0], letters[ielec])
```
zm711, line 651:
```suggestion
                ch_name = f"{itetr + active_tetrode_set[0]}{letters[ielec]}"
```

we try to keep the codebase on f-strings since they are faster and easier to read :)

neo/rawio/axonarawio.py
```diff
@@ -558,6 +557,20 @@ def get_active_tetrode(self):
                     tetrode_id = int(key.strip("collectMask_"))
                     active_tetrodes.append(tetrode_id)
         return active_tetrodes
+    
+    def get_active_channels(self):
```
zm711, line 561: public or private? what do you think?

neo/rawio/axonarawio.py
```diff
@@ -558,6 +557,20 @@ def get_active_tetrode(self):
...
+        E.g.: [20,21,22,23] for tetrode 6 active.
+        """
+        active_tetrodes = self.get_active_tetrode()
+        active_channels = []
+        
+        for tetrode in active_tetrodes:
+            chan = self._get_channel_from_tetrode(tetrode)
+            active_channels.append(chan)
+        
+        return np.concatenate(active_channels)
```
zm711, line 573: what's the benefit of concat? I need to doublecheck what is `chan` is it a list of all channels from a tetrode. If that is the case I would call it `channels` or `chans` instead so we know it's plural.

zm711 (commented):
This looks pretty good to me. The failing test seems to be a ci issue and not related to this PR so I'll try to figure that out. Let me know what you think about the comments :)

alejoe91:
@zm711 maybe just a random failure, re-triggered to see what happens!

zm711:
I tried retriggering once, but yeah let's see if maybe it was just a timing issue.

neo/rawio/axonarawio.py
h-mayorquin: I advice to make things private by default and move away from it when you have a good reason to.

h-mayorquin:
Gin data for this? : O

neo/rawio/axonarawio.py
zm711: That's how I was leaning too. I don't see why the user would run this function.

letiziasignorelli: Yes, I agree with making them private

h-mayorquin:
Thanks. I did not realize we already had the gin data as I came from the neuroconv trail.

zm711:
@letiziasignorelli just wanted to see if you'd have bandwidth to work on this before the end of the year :) ? Otherwise I'll change the tag to `future` for now so that we just let you get back to it when you want....

letiziasignorelli:
@zm711 I replied to all comments, and from my side I was just waiting for approval :). Is there anything I should still work on/change?

zm711:
Did you forget to push the changes? I don't see any of the changes--(I see your comments, but not a commit where you have the changes). Once they are pushed here then the last thing is just resolving the conflict in the author doc :)

neo/rawio/axonarawio.py
```diff
@@ -377,8 +377,7 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
         if channel_indexes is None:
             channel_indexes = [i for i in range(bin_dict["num_channels"])]
         elif isinstance(channel_indexes, slice):
-            channel_indexes_all = [i for i in range(bin_dict["num_channels"])]
-            channel_indexes = channel_indexes_all[channel_indexes]
+            channel_indexes = self.get_active_channels()
```
zm711, line 380:
```suggestion
            channel_indexes = self._get_active_channels()
```

just going to fix this myself really quick. We need to call the private method since we switched it to private :)

neo/rawio/axonarawio.py
```diff
@@ -377,8 +377,7 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
         if channel_indexes is None:
             channel_indexes = [i for i in range(bin_dict["num_channels"])]
         elif isinstance(channel_indexes, slice):
-            channel_indexes_all = [i for i in range(bin_dict["num_channels"])]
-            channel_indexes = channel_indexes_all[channel_indexes]
+            channel_indexes = self._get_active_channels()
```
zm711, line 380: @h-mayorquin do you want to doublecheck this. I'm just re-reading and I'm thinking of the case of a channel slice so would love another pair of eyes on the logic here.

h-mayorquin: I am confused on why this function does not take any input, I would expect this to use the slice (channel_indexes) that it was passed?

samuelgarcia:
Hi @letiziasignorelli.
Thanks for this.
Sorry the ultra lon delay.
I will read it more carrfully, I am not sure to follow entirely  the logic.
Give me another read.

zm711:
@samuelgarcia just so you know the background (although I really appreciate the extra read :))

In our original implementation we were just giving the channels arbitrary names like `range(0,32)` when in fact the channels have actual names depending on which tetrode they are (so it should be something like A1, A2, A3, A4, B1, B2, B3, B4 instead of 0,1,2,3,4,5,6,7,8.

But I'm just trying to make sure the slicing of channels when some aren't turned on will be correct.

h-mayorquin:
I remember reading it back then. I can take a second look if you pint-point the specific part that you don't feel confident about.

neo/rawio/axonarawio.py
zm711: @h-mayorquin it was this part of the code which you indicated your confusion and I too wasn't completely sure.

h-mayorquin: Seems OK to me. I re-read and it is just that they are getting it from the class attributes

samuelgarcia:
OK lets merge.

zm711:
Thanks @letiziasignorelli and sorry for the delay!

PR #1522, opened 2024-08-14
BiocamRawIO: Prevent loading whole array into memory

zm711:
Iterate through channels to prevent loading the whole array into memory even during a channel slice.

@alejoe91, this was a complete stab in the dark by me. If you have a better solution happy to close this. Since I don't know this format at all I'll let the CI see if I completely messed things up.

### DIFF
--- a/neo/rawio/biocamrawio.py
+++ b/neo/rawio/biocamrawio.py
@@ -122,10 +122,27 @@ class BiocamRawIO(BaseRawIO):
             i_start = 0
         if i_stop is None:
             i_stop = self._num_frames
-        if channel_indexes is None:
-            channel_indexes = slice(None)
+
+
         data = self._read_function(self._filehandle, i_start, i_stop, self._num_channels)
-        return data[:, channel_indexes]
+
+        # older style data return everything
+        if data.shape[1] != 1:
+            if channel_indexes is None:
+                channel_indexes = slice(None)
+            sig_chunk = data[:, channel_indexes]
+        # newer style data returns an initial flat array
+        else:
+            if channel_indexes is None:
+                channel_indexes = [ch for ch in range(self._num_channels)]
+
+            sig_chunk = np.zeros((i_stop-i_start, len(channel_indexes)))
+            # iterate through channels to prevent loading all channels into memory which can cause
+            # memory exhaustion. See https://github.com/SpikeInterface/spikeinterface/issues/3303
+            for index, channel_index in enumerate(channel_indexes):
+                sig_chunk[:, index] = data[i_start+channel_index:i_stop+channel_index:self._num_channels]
+
+        return sig_chunk
 
 
 def open_biocam_file_header(filename):
@@ -230,7 +247,7 @@ def open_biocam_file_header(filename):
             offset=offset,
         )
 
-
+# return the full array for the old datasets
 def readHDF5t_100(rf, t0, t1, nch):
     return rf["3BData/Raw"][t0:t1]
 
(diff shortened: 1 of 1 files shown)


neo/rawio/biocamrawio.py
```diff
@@ -122,10 +122,27 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
+        # newer style data returns an initial flat array
+        else:
+            if channel_indexes is None:
+                channel_indexes = [ch for ch in range(self._num_channels)]
+
+            sig_chunk = np.zeros((i_stop-i_start, len(channel_indexes)))
+            # iterate through channels to prevent loading all channels into memory which can cause
+            # memory exhaustion. See https://github.com/SpikeInterface/spikeinterface/issues/3303
+            for index, channel_index in enumerate(channel_indexes):
+                sig_chunk[:, index] = data[i_start+channel_index:i_stop+channel_index:self._num_channels]
```
alejoe91, line 143: That's a smart solution!! :)

zm711:
Okay @alejoe91 I think this is ready for a final review when you have time. It passed tests when it was on draft so it should still pass now. 

@GrimmSnark, do you want to test installing Neo from this PR and see if this fixes your issue?

neo/rawio/biocamrawio.py
```diff
@@ -122,10 +122,27 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
             i_stop = self._num_frames
-        if channel_indexes is None:
-            channel_indexes = slice(None)
+
+
         data = self._read_function(self._filehandle, i_start, i_stop, self._num_channels)
-        return data[:, channel_indexes]
+
+        # older style data return everything
+        if len(data.shape) > 1:
```
h-mayorquin, line 130: Confused about how do you compare a tuple to > 1.

zm711: It's the len of the tuple. Not the tuple itself.

zm711: In a flat array the len of the shape will be 1 because it will be 

data.shape = (n_samples)

whereas the full array will be 2d

data.shape = (n_samples, n_channels)

neo/rawio/biocamrawio.py
```diff
@@ -122,10 +122,27 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
+            if channel_indexes is None:
+                channel_indexes = slice(None)
+            sig_chunk = data[:, channel_indexes]
+        # newer style data returns an initial flat array
+        else:
+            if channel_indexes is None:
+                channel_indexes = [ch for ch in range(self._num_channels)]
+
+            sig_chunk = np.zeros((i_stop-i_start, len(channel_indexes)))
```
h-mayorquin, line 139: Can channel_indexes be None or slice here, if so, len will fail.

neo/rawio/biocamrawio.py
h-mayorquin: Ah, thanks, I did not see this, my bad. Maybe using ndim https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ndim.html will make this clearear?

zm711: I always forget ndim exists. Thanks that is what I wanted, but for the life of me couldn't remember that. So len(shape) does the same while being more unclear.

h-mayorquin: Yes, same for me. There is size, shape and that other thing that... ehm.

neo/rawio/biocamrawio.py
zm711: It can't be None (because if None it is converted to a list). It shouldn't be a slice based on what Neo allows is. So I think this should be fine.

I wasn't aiming to refactor the IO for this fix, but I think unifying the reading functions despite their different structures would probably be good longer term. But really only an ArrayLike or None should go into channel_indexes and I force it if it is None.

h-mayorquin: Good. Just pointing out something that I stumbled upon a couple of times.

alejoe91:
@zm711 let's merge this one!

Issue #1523, opened 2024-08-15
Plexon signal stream_ids are not cosistent across reads


h-mayorquin:
The problem is here:

https://github.com/NeuralEnsemble/python-neo/blob/315f485eb24c8d17aaebc5d6b56a336840913b66/neo/rawio/plexonrawio.py#L269-L275

We use set to form unique users group, enumerate over them to determine stream_index and use str(stream_index) as a stream_id. But this does not work because iterating over sets and determining sets is not deterministic.

samuelgarcia:
yep. set is dangerous for that. np.unique() could be a better choice.

zm711:
Fixed with a dictionary strategy in #1524. Closing.

PR #1524, opened 2024-08-15
Fix-ate plexon signal streams

h-mayorquin:
This PR solves two things:
1) Should solve https://github.com/NeuralEnsemble/python-neo/issues/1523 by using consistent stream_ids defined by plexon channel prefixes that are stable  (see [here](https://github.com/NeuralEnsemble/python-neo/pull/1495#issuecomment-2184256894)
2) It adds human readable and meaningful stream names. Right now they are called "Signal + {stream_id}" and after this PR, they will be:

```python
            channel_prefix_to_stream_name = {
                "WB": "WB-Wideband",
                "FP": "FPl-Low Pass Filtered ",
                "SP": "SPKC-High Pass Filtered",
                "AI": "AI-Auxiliary Input",
            }

```

Not that this moves forward the poject in https://github.com/SpikeInterface/spikeinterface/issues/3197 that aims to have "logical streams" and relegate "buffer streams" to an internal implementation detail of the raw io API.

### DIFF
--- a/neo/rawio/plexonrawio.py
+++ b/neo/rawio/plexonrawio.py
@@ -255,7 +255,7 @@ class PlexonRawIO(BaseRawIO):
                     0.5 * (2 ** global_header["BitsPerSpikeSample"]) * h["Gain"] * h["PreampGain"]
                 )
             offset = 0.0
-            s   tream_id = "0"
+            stream_id = "0"  # This is overwritten later
             sig_channels.append((name, str(chan_id), sampling_rate, sig_dtype, units, gain, offset, stream_id))
 
         sig_channels = np.array(sig_channels, dtype=_signal_channel_dtype)
--- a/neo/rawio/plexonrawio.py
+++ b/neo/rawio/plexonrawio.py
@@ -146,7 +146,7 @@ class PlexonRawIO(BaseRawIO):
 
             # Update tqdm with the number of bytes processed in this iteration
             if self.progress_bar:
-                progress_bar.update(length)
+                progress_bar.update(length)  # This was clever, Sam : )
 
         if self.progress_bar:
             progress_bar.close()
@@ -231,10 +231,7 @@ class PlexonRawIO(BaseRawIO):
         # signals channels
         sig_channels = []
         all_sig_length = []
-        if self.progress_bar:
-            chan_loop = trange(nb_sig_chan, desc="Parsing signal channels", leave=True)
-        else:
-            chan_loop = range(nb_sig_chan)
+        chan_loop = trange(nb_sig_chan, desc="Parsing signal channels", leave=True, disable=not self.progress_bar)
         for chan_index in chan_loop:
             h = slowChannelHeaders[chan_index]
             name = h["Name"].decode("utf8")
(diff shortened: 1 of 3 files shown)


neo/rawio/plexonrawio.py
```diff
@@ -363,8 +389,8 @@ def _segment_t_start(self, block_index, seg_index):
     def _segment_t_stop(self, block_index, seg_index):
         t_stop = float(self._last_timestamps) / self._global_ssampling_rate
         if hasattr(self, "_signal_length"):
-            for stream_id in self._signal_length:
-                t_stop_sig = self._signal_length[stream_id] / self._sig_sampling_rate[stream_id]
+            for stream_index in self._signal_length:
```
zm711, line 392: My only comment is that you are iterating through the keys, but it is not clear as a developer that this is a dict and not a list. I think iterating through the `keys()` would make it clear the object is a dict. But this is just an  optional comment. Switching it to `stream_index` fits with the current data, so I think that is a great semantic fix.

zm711 (dismissed):
this looks good to me. I'll leave this open until next week to see if @alejoe91 or @samuelgarcia want to take a look before merging. :)

neo/rawio/plexonrawio.py
h-mayorquin: This was already there, I just change the wonrgly named variable stream_id to stream_index.

But I agree with you that it would be better to be explicit and do self._signal_length.keys().

zm711: No I know you were just fixing it! But when it's on the diff it gets spotted :)

h-mayorquin:
OK, the test file that we have is quite limited. I would need this:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/137

zm711:
> OK, the test file that we have is quite limited. I would need this:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/137

Merged @h-mayorquin 

I'll re-run tests so you can confirm your changes work with the new demo file.

zm711:
The rtd failure should also be looked at. (Since plexus is used for the doc build!

h-mayorquin:
Fixed the test and added a new one.

neo/rawio/plexonrawio.py
h-mayorquin: Changed.

neo/rawio/plexonrawio.py
```diff
@@ -264,22 +267,49 @@ def _parse_header(self):
...
 
         else:
-            # detect groups (aka streams)
-            all_sig_length = all_sig_length = np.array(all_sig_length)
-            groups = set(zip(sig_channels["sampling_rate"], all_sig_length))
+            # Detect streams
+            all_sig_length = np.asarray(all_sig_length)
+
+            # names are WB{number}, FPX{number}, SPKCX{number}, AI{number}, etc
+            pattern = r"^\D+"  # Match any non-digit character at the beginning of the string
```
zm711, line 274: Thanks for this. I always find regex so hard to read :P The comment is super helpful.

neo/rawio/plexonrawio.py
```diff
@@ -264,22 +267,49 @@ def _parse_header(self):
...
+            channels_prefixes = np.asarray([re.match(pattern, name).group(0) for name in sig_channels["name"]])
+            buffer_stream_groups = set(zip(channels_prefixes, sig_channels["sampling_rate"], all_sig_length))
+
+            # There are explanations of the streams bassed on channel names
```
zm711, line 278:
```suggestion
            # There are explanations of the streams based on channel names
```

neo/rawio/plexonrawio.py
```diff
@@ -264,22 +267,49 @@ def _parse_header(self):
...
 
             signal_streams = []
             self._signal_length = {}
             self._sig_sampling_rate = {}
-            for stream_index, (sr, length) in enumerate(groups):
-                stream_id = str(stream_index)
+
+            for stream_index, (channel_prefix, sr, length) in enumerate(buffer_stream_groups):
+                # The users of plexon can modify the channel names, is not common but in that case
+                # We assign the channel_prefix both as stream_name and stream_id
```
zm711, line 302: What does this comment mean? 

So they can re-name a channel to not be WB, FB, SP, AI etc... What part can they actually rename? Do we need to protect against that better or could we make this comment clearer to what the renaming actually entails?

h-mayorquin: The can modify the prefix. For example, in the test file that we had before I added the new one, the channel is named V1. 

In that case I am adding V as both the stream name and id. This can't be generalized but the user changed the prefix for some reason so they should know what it means in neo when they see the header and in spikeinterface when they get the assertion that the stream was not found.

neo/rawio/plexonrawio.py
h-mayorquin: Yes. FYI there is this pattern of writing regex expressions like this:

```python
I mean sources for the suggestio to annotate/comment on regex like this:

email_regex = re.compile(r"""
    ^                     # Start of the string
    [a-zA-Z0-9_.+-]+      # Username: allows letters, digits, underscores, dots, pluses, and hyphens
    @                     # The "@" symbol
    [a-zA-Z0-9-]+         # Domain name: allows letters, digits, and hyphens
    (\.[a-zA-Z]{2,})+     # Top-level domain: allows dots followed by 2 or more letters
    $                     # End of the string
""", re.VERBOSE)
```
See the documentation in the docs:
https://docs.python.org/3/library/re.html#re.VERBOSE

Which I advocate for but some people dislike it : P

zm711: It's helpful, but I get people's point. :)

neo/rawio/plexonrawio.py
zm711: Okay I think I see what you mean! Thanks.

zm711 (approved):
This is still good by me. @alejoe91 or @samuelgarcia want to take a quick look as well?

neo/rawio/plexonrawio.py
h-mayorquin: I don't. I think that many programmers fancy themselves as mathematicians and not as the boring manual writters that we are : P

neo/rawio/plexonrawio.py
h-mayorquin: Great! Let me know if there is an improvement to the comment that I could make.

zm711: I guess it would make a little more sense in the comment to clarify channel name vs channel prefix. In the first part you say channel names, then you say we take the channel prefix. So making it clear: can the user modify the prefix, all of the name, the numbering of the channel.

neo/rawio/plexonrawio.py
```diff
@@ -264,22 +267,49 @@ def _parse_header(self):
...
+            }
 
             signal_streams = []
             self._signal_length = {}
             self._sig_sampling_rate = {}
-            for stream_index, (sr, length) in enumerate(groups):
-                stream_id = str(stream_index)
+
+            for stream_index, (channel_prefix, sr, length) in enumerate(buffer_stream_groups):
+                # The users of plexon can modify the channel names, is not common but in that case
```
h-mayorquin, line 301:
```suggestion
                # The users of plexon can modify the prefix of the channel names (e.g. `my_prefix` instead of `WB`). This is not common but in that case
```

h-mayorquin: @zm711

zm711: Yep, that looks perfect!

Issue #1530, opened 2024-08-27
[IgorIO] Not all .pxp files have "xUnits" in the header


apdavison:
**Describe the bug**

Loading blocks from a .pxp file with IgorIO results in a KeyError.

**To Reproduce**

The error occurs with files from [this EBRAINS dataset](https://search.kg.ebrains.eu/instances/f25ec9c9-2348-4a62-a6af-907d389f263d).

```
from neo import io
filename = io.IgorIO(filename = 'sbj4-170614/cell8/original/hbp-00940_ephy-ChIN-STR_sbj4-170614_cell8_original__md_20170614_cell_7_8_ChIN.pxp')
block = filename.read_block()
```

Error:
```
KeyError                                  Traceback (most recent call last)
Cell In[1], line 3
      1 from neo import io
      2 filename = io.IgorIO(filename='sbj4-170614/cell8/original/hbp-00940_ephy-ChIN-STR_sbj4-170614_cell8_original__md_20170614_cell_7_8_ChIN.pxp')
----> 3 block = filename.read_block()

File ~/dev/analysis/neo/neo/io/igorproio.py:76, in IgorIO.read_block(self, lazy)
     73     raise NeoReadWriteError("This IO does not support lazy reading")
     75 block = Block(file_origin=str(self.filename))
---> 76 block.segments.append(self.read_segment(lazy=lazy))
     77 return block

File ~/dev/analysis/neo/neo/io/igorproio.py:96, in IgorIO.read_segment(self, lazy)
     93             signal = self._wave_to_analogsignal(value.wave["wave"], dirpath)
     94             segment.analogsignals.append(signal)
---> 96     pxp.walk(self.filesystem, callback)
     97 else:
     98     segment.analogsignals.append(self.read_analogsignal(lazy=lazy))

File ~/dev/analysis/env-202307/lib/python3.11/site-packages/igor2/packed.py:185, in walk(filesystem, callback, dirpath)
    183 callback(dirpath, key, value)
    184 if isinstance(value, dict):
--> 185     walk(filesystem=value, callback=callback, dirpath=dirpath + [key])

File ~/dev/analysis/env-202307/lib/python3.11/site-packages/igor2/packed.py:183, in walk(filesystem, callback, dirpath)
    181     dirpath = []
    182 for key, value in sorted((_bytes(k), v) for k, v in filesystem.items()):
--> 183     callback(dirpath, key, value)
    184     if isinstance(value, dict):
[...]

PR #1531, opened 2024-08-27
Fix for #1530 [IgorIO and missing "xUnits"]

apdavison:
"xUnits" is only used in Versions 1, 2, 3 of .pxp files; Version 5 uses "dimUnits" - see https://github.com/AFM-analysis/igor2/blob/43fccf51714661fb96372e8119c59e17ce01f683/igor2/binarywave.py#L501


zm711 (approved):
Looks fine as long as tests pass. Do we have  a test file for this we could add to GIN? I'm happy to do it if you can share with me :)

apdavison:
The files I have are around 70 MB, which is probably too big for a test file, no?

zm711 (approved):

PR #1532, opened 2024-08-27
BiocamRawIO: Fix case where slice is given to `get_analogsignal_chunk`

zm711:
@h-mayorquin I was wrong, haha. We give a slice in spikeinterface which broke this :) 

@alejoe91, I don't have a biocam file easily available do you? I could test this, tonight with my personal laptop or we could merge and see what happens with full tests on SpikeInterface?

### DIFF
--- a/neo/rawio/biocamrawio.py
+++ b/neo/rawio/biocamrawio.py
@@ -140,8 +140,13 @@ class BiocamRawIO(BaseRawIO):
         # some careful checks around slicing of None in the case we need to iterate through
         # channels. First check if None. Then check if slice and only if slice check that it is slice(None)
         else:
-            if (channel_indexes is None) or (isinstance(channel_indexes, slice) and channel_indexes == slice(None)):
+            if channel_indexes is None:
                 channel_indexes = [ch for ch in range(self._num_channels)]
+            elif isinstance(channel_indexes, slice):
+                start = channel_indexes.start or 0
+                stop = channel_indexes.stop or self._num_channels
+                step = channel_indexes.step or 1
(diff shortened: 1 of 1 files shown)


zm711:
Give me a second I need to fix for array too. Will ping you when ready.

zm711:
Okay the biocamrawio passed tests now. This is ready for review and we can merge after all tests merge or after one of us tests locally that passing a slice(None) is fine now.

alejoe91 (dismissed):

zm711:
@alejoe91 one question since you and Sam know this better. Would it ever happen in spikeinterface to generate a non-None slice that we need to worry about? I could add something for that if it would actually happen.

alejoe91:
I don't think this is the case, because the segment `channel_indexes` are almost always called by the recording wrapper. However, it is possible to access the segment and call the `get_traces()` directly.

A simple fix could be the following:
```
if channel_indexes is None:
    channel_indexes = [ch for ch in range(self._num_channels)]
    start = channel_indexes.start or 0
    stop = channel_indexes.stop or self._num_channels
    step = channel_indexes.step or 1
    channel_indexes = [ch for ch in range(start, stop, step)]
```

h-mayorquin:
Maybe you should have a test for this here?

zm711:
That's a fair fix. Let me do it!

I thought about that, but then I was worried about what if Neo wasn't actually read for that. So I think it should be a separate PR to add that test in.

alejoe91 (approved):

alejoe91:
Ok let's give this a try!

PR #1536, opened 2024-08-27
Limit quantities < `0.16.0` until we fix neo for view vs copy behavior

zm711:


apdavison:
having just merged this, it now occurs to me that we also need to limit numpy<2.0, otherwise it makes no sense to limit the quantities version.

zm711:
That's true. For testing we don't need to but if we want a final pre-numpy 2.0 release before any bigger refactoring we should limit it and then cut a release.

PR #1539, opened 2024-08-28
Add test that all rawios accept `slice(None)`

### LINKED ISSUE #1533: Add a test for `slice(None)` to `RawIO` testing
We test for `slice(None)` in SpikeInterface which according to the docstring of `get_analogsignal_chunk` should be an acceptable `channel_indexes` type. I changed BiocamRawIO in a non-compliant way, but the tests missed this. We should add some quick test for giving slice(None) to make sure our RawIO's accept this (based on SpikeInterface testing I think most should already be fine)

zm711:
Fixes #1533.

This was found in the SpikeInterface full-test suite where an option is exposed to `prefer_slice` which will attempt to give a slice for `channel_indexes`. The `get_analogsignal_chunk` docstring says that it should accept slices, so we should test that all rawios are actually compliant for this requirement. If any of rawios fail this test then we need to fix them.


samuelgarcia:
merci!

alejoe91 (approved):

PR #1541, opened 2024-08-29
Fix-ate plexon 2 streams

h-mayorquin:
Same as https://github.com/NeuralEnsemble/python-neo/pull/1524 but for Plexon 2.

Also, I am exposing the number of retries as that is non-deterministic in plexon2 and that has been failing in some our CI in neuroconv because the number is too small.

### DIFF
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -55,7 +55,7 @@ class Plexon2RawIO(BaseRawIO):
         If None will find correct dll for architecture and if it does not exist will download it
     reading_attempts: int, default: 15
         Number of attempts to read the file before raising an error
-        This opening process is somehow unreliable and might fail occasionally. Adjust this higher
+        This opening process is somewhat unreliable and might fail occasionally. Adjust this higher
         if you encounter problems in opening the file.
 
     Notes
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -25,8 +25,8 @@ Author: Julia Sprenger
 import pathlib
 import warnings
 import platform
+import re
 
-from collections import namedtuple
 from urllib.request import urlopen
 from datetime import datetime
 
@@ -53,6 +53,10 @@ class Plexon2RawIO(BaseRawIO):
     pl2_dll_file_path: str | Path | None, default: None
         The path to the necessary dll for loading pl2 files
         If None will find correct dll for architecture and if it does not exist will download it
+    reading_attempts: int, default: 15
+        Number of attempts to read the file before raising an error
+        This opening process is somehow unreliable and might fail occasionally. Adjust this higher
+        if you encounter problems in opening the file.
 
     Notes
     -----
@@ -88,7 +92,7 @@ class Plexon2RawIO(BaseRawIO):
     extensions = ["pl2"]
     rawmode = "one-file"
 
-    def __init__(self, filename, pl2_dll_file_path=None):
+    def __init__(self, filename, pl2_dll_file_path=None, reading_attempts=15):
 
         # signals, event and spiking data will be cached
         # cached signal data can be cleared using `clear_analogsignal_cache()()`
@@ -128,7 +132,6 @@ class Plexon2RawIO(BaseRawIO):
 
         self.pl2reader = PyPL2FileReader(pl2_dll_file_path=pl2_dll_file_path)
 
-        reading_attempts = 10
         for attempt in range(reading_attempts):
             self.pl2reader.pl2_open_file(self.filename)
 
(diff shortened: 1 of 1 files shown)


neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -128,7 +132,6 @@ def __init__(self, filename, pl2_dll_file_path=None):
 
         self.pl2reader = PyPL2FileReader(pl2_dll_file_path=pl2_dll_file_path)
 
-        reading_attempts = 10
```
zm711, line 131: Is it possible that you have a good attempt in the middle and bad attempt at the end of the range that then messes up the attempts such that you just have to guess when to end on a "good attempt". Ie is there a smarter way for us to check like

for i in range(num_attempts):
read
if read is good
break

h-mayorquin: That's how it is right now. If the handle is good you exist the loop:

https://github.com/NeuralEnsemble/python-neo/blob/221befab6c1d90567107d84bc58ad233374614b8/neo/rawio/plexon2rawio/plexon2rawio.py#L139-L141

In other words, it is first success out.

zm711: sorry I missed that line. Sorry didn't expand the diff.

h-mayorquin: It is a good suggestion though!

zm711: And so it silently fails? We have yet to see any stochastic failures of our test files so far. It's just so weird for it to fail. Has plexon commented on this?

h-mayorquin: You did at the beginning before we made this patch:

https://github.com/NeuralEnsemble/python-neo/pull/1495#issuecomment-2184193508

The re-attempt mechanism solves most of the case almost all of the time except when it does not. For some reason it has been failing for neuroconv's docstring tests.

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -354,16 +383,12 @@ def _segment_t_stop(self, block_index, seg_index):
         end_time = (
             self.pl2reader.pl2_file_info.m_StartRecordingTime + self.pl2reader.pl2_file_info.m_DurationOfRecording
         )
-        return end_time / self.pl2reader.pl2_file_info.m_TimestampFrequency
+        return float(end_time / self.pl2reader.pl2_file_info.m_TimestampFrequency)
```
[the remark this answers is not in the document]
h-mayorquin: I think we need to asure that the output is a float, the division is internal. I think that division between numpy float scalars is OK.

neo/rawio/plexon2rawio/plexon2rawio.py
zm711: I meant test failures since your PR. It seems like the 10 tries has always worked for us thus far. It seems weird that the flakiness is so flaky, but yeah based on the discussion with plexon there they also seem to be a bit confused by this issue. So I'm fine with just letting the person try as many times as they want.

neo/rawio/plexon2rawio/plexon2rawio.py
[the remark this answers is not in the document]
zm711: Okay cool. I just wasn't sure since the last problem was between an int - numpy.scalar so I wasn't sure if float/numpy.scalar could also cause a problem or not.

neo/rawio/plexon2rawio/plexon2rawio.py
h-mayorquin: Yeah, it is kind of fishy that it always fails in our doctest. But if I discover that that indeed is the case that more attempst solve the problem and that different envs require different attemps I don't want to have to create a new PR every time so better exposed it.

It is unlikely that we find a root case solution if the plexon people also are suprised about it because it is probably a wine dllc related problem.

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -53,6 +53,10 @@ class Plexon2RawIO(BaseRawIO):
     pl2_dll_file_path: str | Path | None, default: None
         The path to the necessary dll for loading pl2 files
         If None will find correct dll for architecture and if it does not exist will download it
+    reading_attempts: int, default: 15
+        Number of attempts to read the file before raising an error
+        This opening process is somehow unreliable and might fail occasionally. Adjust this higher
```
zm711, line 58:
```suggestion
        This opening process is somewhat unreliable and might fail occasionally. Adjust this higher
```
somewhat gives a stronger flavor of unpredictability in success (so it feels like "sometimes"). somehow implies a more mechanistically nuance. So it would mean more like there is a problem in the implementation that we could figure out and fix, but we haven't yet.

In this case I would tend toward somewhat (and somewhat is more commonly used in general).

zm711 (commented):
This looks pretty good to me. @samuelgarcia or @alejoe91  want to take a quick scan of this one?

neo/rawio/plexon2rawio/plexon2rawio.py
h-mayorquin: I was no aware of such nuance. Thanks for that.

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -152,46 +155,72 @@ def _parse_header(self):
...
+            # We are using the channel prefixes as ids
+            # The users of plexon can modify the prefix of the channel names (e.g. `my_prefix` instead of `WB`).
+            # In that case we use the channel prefix both as stream id and name
+            stream_name = stream_id_to_stream_name.get(stream_id, stream_id)
+            signal_streams.append((stream_name, stream_id))
+
         signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
 
+        self.stream_id_samples = {}
+        self.stream_index_to_stream_id = {}
```
samuelgarcia, line 213: Why this ?
`stream_index = header["signall_streams"]["id"][stream_index]` should work no ?

zm711: To be honest I think Heberto and I like the slightly more explicit variable names in order to quickly know where to get things. it's tricky to remember the complicated header structure sometimes so the separate variables just make it a little more explicit for the steps.

zm711:
I think we should merge #1546 first and then incorporate it's bug fix into this PR.

neo/rawio/plexon2rawio/plexon2rawio.py
h-mayorquin: Yeah, I think that could work as well it just did not occur to me. I don't have a strong preference here, Sam. As zach mentions I have a preference for explicit but in this case I just came with this idea first. If you prefer I can eliminate it.

h-mayorquin:
> I think we should merge #1546 first and then incorporate it's bug fix into this PR.

Seems that there were no conflicts.

zm711:
> Seems that there were no conflicts.

Cool. I didn't look too closely to see if you had touch those specific lines or not. So I just wanted to check that before. We can wait for Sam to decide if you should user the header instead of the dict you made and then we can merge after he decides on that.

samuelgarcia (approved):

neo/rawio/plexon2rawio/plexon2rawio.py
samuelgarcia: OK. No worry. Just curious.

samuelgarcia:
This is OK for me. Can I merge ?

zm711 (approved):
Good  by me. Feel free to merge Sam or I will merge this afternoon :)

PR #1544, opened 2024-09-03
simple proposal for buffer_id in rawio

samuelgarcia:
A very simple proposal for spikeglxrawio
@h-mayorquin 
@zm711


neo/test/rawiotest/rawio_compliance.py
```diff
@@ -47,6 +48,12 @@ def header_is_total(reader):
     assert "nb_segment" in h, "`nb_segment`missing in header"
     assert len(h["nb_segment"]) == h["nb_block"]
 
+    assert "signal_buffers" in h, "signal_buffers missing in header"
+    if h["signal_buffers"] is not None:
+        dt = h["signal_buffers"].dtype
+        for k, _ in _signal_buffer_dtype:
+            assert k in dt.fields, f"{k} not in _ignal_buffers.dtype"
```
zm711, line 55:
```suggestion
            assert k in dt.fields, f"{k} not in _signal_buffers.dtype"
```

neo/rawio/spikeglxrawio.py
```diff
@@ -140,15 +141,20 @@ def _parse_header(self):
...
+        signal_buffers = []
         signal_streams = []
         signal_channels = []
         for stream_name in stream_names:
             # take first segment
             info = self.signals_info_dict[0, stream_name]
 
+            buffer_id = stream_name
+            buffer_name = stream_name
+            signal_buffers.append((buffer_name, buffer_id))
```
samuelgarcia, line 153: At the moment the mapping buffer to stream is one to one.
But after this PR we will be able to split buffer into logical stream super easily.

zm711:
In your example the idea would be that we move the sync channel into a new stream right? So it will keep the buffer but you'll assign a new stream ?

h-mayorquin:
Hey, Sam, what I had in mind last time we discussed was something like this that I think should be less concentrate the disruption in one PR and then we can work piece by piece on the formats:

https://github.com/NeuralEnsemble/python-neo/pull/1545

samuelgarcia:
> In your example the idea would be that we move the sync channel into a new stream right? So it will keep the buffer but you'll assign a new stream ?

In this PR the idea to have no change IOs. Only adding this buffer_id stuff in a very neutral way.

The next PR will have some changes for some IOs (plexon, spikeglx, ced) to split actual stream in more streams but with the same buffer.

samuelgarcia:
> Hey, Sam, what I had in mind last time we discussed was something like this that I think should be less concentrate the disruption in one PR and then we can work piece by piece on the formats:

Your proposal is cool and easy but I think I prefer to be brave and explicitly add this features everywhere as a necessary implementation for rawios.

h-mayorquin:
Good. We covered in the discussion today:
The buffer will be undefined if:
1) The stream is divided across multiple files. Example intan's one-file-per-channel
2) There is an API between neo and the data which obfuscates the data layout. Example plexon2

We decide between two approaches:
* Have a strict version where stream and buffer are orthogonal concepts.
* Have a "nested-unless-edge-case" approach where the stream headers contain a reference fo their buffer unless the buffer is hard to define (cases 1 and 2 above).

I am weary of not going with the strict apporach but this will faciliate some work on Sam and I could not find a concrete example where this could cause problems. In such a case "practicallity should beat purity" and we will stick to this.

zm711:
> Good. We covered in the discussion today:
The buffer will be undefined if:

I think what I would add to this  so we don't forget that we didn't completely settle on a rigid buffer definition so that we have some flexibility as we implement it. Buffers ought to have same dtype, sampling rate, but something like same file vs same memmap was left in the air
* plexon 1 for example is one file, but multiple memmaps.
*  Sam mentioned thinking about intan header-attached which is one file, one memmap but with the block structure doesn't quite fit with what he was imagining for a buffer (when you hopped off @h-mayorquin).

Thanks for summary of the discussion @h-mayorquin! I think keeping our documentation up will help us when we go back and think about our choices!

samuelgarcia:
Thanks for this summary of internal discussion @h-mayorquin  and @zm711.

samuelgarcia:
@zm711 @h-mayorquin this is ready on my side.

zm711:
Looks like raw binary signal wasn't done quite right (tests failing). Want to fix that? I'll look over this in parallel.

neo/rawio/axonrawio.py
```diff
@@ -56,6 +56,7 @@
     BaseRawIO,
     _signal_channel_dtype,
     _signal_stream_dtype,
+    _signal_buffer_dtype,
```
zm711, line 59: sneaky. could've had them alphabetized, but that would be too conventional.

neo/rawio/blackrockrawio.py
```diff
@@ -399,7 +401,10 @@ def _parse_header(self):
                         ext_header.append(d)
 
                 if len(ext_header) > 0:
-                    signal_streams.append((f"nsx{nsx_nb}", str(nsx_nb)))
+                    buffer_id = stream_id = str(nsx_nb)
```
zm711, line 404: I'll be honest I don't know blackrock at all. I'll trust you on this one.

neo/rawio/examplerawio.py
```diff
@@ -111,12 +112,18 @@ def _parse_header(self):
         # In short `_parse_header()` can be slow but
         # `_get_analogsignal_chunk()` needs to be as fast as possible
 
-        # create fake signal streams information
+        # create fake signal streams and buffer information
+        # a buffer is a group of channels that are in the same buffer for instance hdf5 or binary : this is optional
```
zm711, line 116: I'm wondering if we want to weird this differently. buffer is not optional in the developer sense because it has to be a value or `""` but is optional in the sense that you can give an `""` string. So maybe we make that clearer. Let me think about this while you fix the tests.

neo/rawio/intanrawio.py
```diff
@@ -226,25 +229,45 @@ def _parse_header(self):
...
         signal_streams["id"] = [str(stream_id) for stream_id in stream_ids_sorted]
-
         for stream_index, stream_id in enumerate(stream_ids_sorted):
             if self.filename.suffix == ".rhd":
                 name = stream_id_to_stream_name_rhd.get(int(stream_id), "")
             else:
                 name = stream_id_to_stream_name_rhs.get(int(stream_id), "")
-
             signal_streams["name"][stream_index] = name
+            # zach I need you help here
```
zm711, line 252: I think we need to say why we don't count these as buffers.

So for header-attached it is because the data is set up in blocks so there is no continous signal
and for one-file-per-channel it is because the stream is split among files.

neo/rawio/openephysrawio.py
```diff
@@ -221,10 +225,13 @@ def _parse_header(self):
...
             stream_ids, order = np.unique(chan_stream_ids, return_index=True)
             stream_ids = stream_ids[np.argsort(order)]
-            signal_streams = [(f"Signals {stream_id}", f"{stream_id}") for stream_id in stream_ids]
-            signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
+            signal_streams = [(f"Signals {stream_id}", f"{stream_id}", "") for stream_id in stream_ids]
         else:
-            signal_streams = np.array([])
+            signal_streams = []
+        signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
+        # no buffer handling in this format because one channel per file
```
zm711, line 232: c'est parfait! That's a perfect explanation of why not.

neo/rawio/tdtrawio.py
```diff
@@ -308,13 +310,16 @@ def _parse_header(self):
...
-                signal_channels.append((chan_name, str(chan_id), sampling_rate, dtype, units, gain, offset, stream_id))
+                buffer_id = ""
+                signal_channels.append((chan_name, str(chan_id), sampling_rate, dtype, units, gain, offset, stream_id, buffer_id))
 
         if missing_sev_channels:
             warnings.warn(f"Could not identify sev files for channels {missing_sev_channels}.")
 
         signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
         signal_channels = np.array(signal_channels, dtype=_signal_channel_dtype)
+        # buffer concept here, data are spread per channel and paquet
```
zm711, line 321: Qu'est-ce que c'est un paquet la? channel are files and the paquet is for folder? or packets of info meaning like internal blocks of data like data packets?

h-mayorquin:
I will take a look once the tests are passing.

neo/rawio/tdtrawio.py
samuelgarcia: Yes this is what I meant

zm711:
Now plexon2 tests are failing. Heberto has done a bunch of updates. He has another PR that we need to review (but I need to check if it is plexon1 or plexon2). Do we want to get his last PR done and then you can finalize this or do you want to finish this first. There are slightly organizational changes occurring that I think will keep leading to test failures. Up to you both from my perspective.

zm711:
I think with the latest boost to plexon2 once you fix conflicts and merge from main I hope the plexon2 won't give us any troubles.

samuelgarcia:
@zm711 : it would be nice to merge this soon.
So I can continue working on #1513.

zm711:
scanning through this it looks good. Once tests pass I'll do a final read through, so let's target to merge this by end of weekend. We can ping @h-mayorquin for a read through too, hopefully today if he has time?

neo/test/rawiotest/test_neuronexusrawio.py
```diff
@@ -12,3 +12,7 @@ class TestNeuroNexusRawIO(
     rawioclass = NeuroNexusRawIO
     entities_to_download = ["neuronexus"]
     entities_to_test = ["neuronexus/allego_1/allego_2__uid0701-13-04-49.xdat.json"]
+
+
+if __name__ == "__main__":
```
zm711, line 17: hahahahah

zm711 (approved):
This works for me. There are a couple of comment touch ups and some clean up we should do as we push forward, but I agree it is too hard to keep up the parallel structure for so long.

neo/rawio/baserawio.py
```diff
@@ -86,9 +86,15 @@
...
 
+_signal_buffer_dtype = [
+    ("name", "U64"),  # not necessarily unique
+    ("id", "U64"),  # must be unique
+]
+
 _signal_stream_dtype = [
     ("name", "U64"),  # not necessarily unique
     ("id", "U64"),  # must be unique
+    ("buffer_id", "U64"), # optional can be ""
```
h-mayorquin, line 97:
```suggestion
    ("buffer_id", "U64"), # should be "" (empty string) when the stream is not nested under a buffer or the buffer is undefined for some reason.
```

neo/rawio/edfrawio.py
```diff
@@ -115,16 +116,18 @@ def _parse_header(self):
...
 
         # convert channel index lists to arrays for indexing
         self.stream_idx_to_chidx = {k: np.array(v) for k, v in self.stream_idx_to_chidx.items()}
 
         signal_channels = np.array(signal_channels, dtype=_signal_channel_dtype)
 
-        signal_streams = [(f"stream ({sr} Hz)", i) for i, sr in enumerate(stream_characteristics)]
+        signal_streams = [(f"stream ({sr} Hz)", i, "") for i, sr in enumerate(stream_characteristics)]
         signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
+        # no buffer handling here
```
h-mayorquin, line 129:
```suggestion
        # TODO: find out buffers here.
```
This can be done, right? this is a single file. I am asking because you might know something that I don't.

neo/rawio/baserawio.py
```diff
@@ -86,9 +86,15 @@
 
 error_header = "Header is not read yet, do parse_header() first"
 
+_signal_buffer_dtype = [
+    ("name", "U64"),  # not necessarily unique
+    ("id", "U64"),  # must be unique
+]
+
```
h-mayorquin, line 93:
```suggestion
# To be left an empty array if the concept of buffer is undefined for a reader.
```

h-mayorquin (approved):
LGMT. I added some comments that I think will be helpful for future developers.

I still think that including the buffer in the data signal_stream might lead to problems in the future and I would prefer the concepts to be decoupled (the schema to be normalized in the database sense and make the signal channel array/table the normalized table).

But, what is important for data extraction is that we can decouple the stream and the buffer and divide the streams "logically" and this PR allows us to do this:

Fine by me.

zm711:
If Sam doesn't respond I'll package up your edits into a commit and push them, then once final tests pass we can merge it. 

Definitely down for another call at some point to discuss your desire for decoupling more !

neo/rawio/edfrawio.py
h-mayorquin: Ah, that makes sense. That's true, feel free to ignore this suggestion @samuelgarcia

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -206,9 +208,11 @@ def _parse_header(self):
             # The users of plexon can modify the prefix of the channel names (e.g. `my_prefix` instead of `WB`).
             # In that case we use the channel prefix both as stream id and name
             stream_name = stream_id_to_stream_name.get(stream_id, stream_id)
-            signal_streams.append((stream_name, stream_id))
-
+            buffer_id = ""
+            signal_streams.append((stream_name, stream_id, buffer_id))
         signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
+        # In plexon buffer is unkown
```
zm711, line 214:
```suggestion
        # In plexon buffer is unknown
```

neo/rawio/examplerawio.py
```diff
@@ -111,12 +112,18 @@ def _parse_header(self):
         # In short `_parse_header()` can be slow but
         # `_get_analogsignal_chunk()` needs to be as fast as possible
 
-        # create fake signal streams information
+        # create fake signal streams and buffer information
```
zm711, line 115:
```suggestion
        # create simulated signal streams and buffer information
```

neo/rawio/examplerawio.py
zm711, line 116:
```suggestion
        # a buffer is a group of channels that are in the same IO buffer for instance hdf5 or binary that can mapped to with np.memmap : this is optional machinery but must be specified as an empty list if not present
```

neo/rawio/examplerawio.py
```diff
@@ -141,7 +148,10 @@ def _parse_header(self):
             # channels inside a "stream" share the same characteristics
             #  (sampling rate/dtype/t_start/units/...)
             stream_id = str(c // 8)
-            signal_channels.append((ch_name, chan_id, sr, dtype, units, gain, offset, stream_id))
+            # buffer_id indicates optionally if channels are in the same buffer
+            # this is optional. In case the buffer concept do not apply then buffer_id must be ""
```
zm711, line 152:
```suggestion
            # this is optional. In case the buffer concept does not apply then buffer_id must be ""
```

zm711 (commented):
adding a couple doc fixes to be added with Heberto's

neo/rawio/examplerawio.py
h-mayorquin: Nice!

samuelgarcia:
merci les amis

PR #1545, opened 2024-09-04
[Showcase] Enchance headers with buffer id at `BaseRawIO`

h-mayorquin:
This PR is not mean to be merged but to add a concrete implementation to suggest in #1544. This PR probably is incomplete and it has errors.

The idea here is that the `BaseRawIO` will add post parsing logic to modify the `signal_streams` and `signal_channels` headers by adding the buffer id if they are not in the new type yet. Same for `signal_channels` where the buffer id will be added as an extra column. In all of those cases the buffer_id is just a carbon copy of the stream_id. 

This approach should enable the following workflow and progression plan:
1) In one sweep all the classess should have the correct dtype for the header avoiding an intermediate state of the library where not all the formats have the same dtype for the headers.
2) We can then approach each format piece-wise and assign the proper buffer ids and separate the streams logically.
3) Once we are done with all the formats we can eliminate the post parsing code here and everything should be clean.

### DIFF
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -86,10 +86,19 @@ possible_raw_modes = [
 
 error_header = "Header is not read yet, do parse_header() first"
 
+_signal_buffer_dtype = [
+    ("name", "U64"),  # not necessarily unique
+    ("id", "U64"),  # must be unique
+]
+
 _signal_stream_dtype = [
     ("name", "U64"),  # not necessarily unique
     ("id", "U64"),  # must be unique
 ]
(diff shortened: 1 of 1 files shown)


h-mayorquin:
OK, so Sam decided to do this in one sweep. Let's close this once we are finished with https://github.com/NeuralEnsemble/python-neo/pull/1544.

zm711:
We said we would close with the merge of #1544. Reopen if needed Heberto.

h-mayorquin:
Thanks for closing!

PR #1547, opened 2024-09-06
Refactor plexon rawio to have same ids as plexon2

h-mayorquin:
After #1541 Plexon2 ended up with the meaningful but unique stream_ids:

```
        stream_id_to_stream_name = {
            "WB": "WB-Wideband",
            "FP": "FPl-Low Pass Filtered",
            "SP": "SPKC-High Pass Filtered",
            "AI": "AI-Auxiliary Input",
        }
```

But in Plexon 1 we are using numbers instead: "1", "2", ... 

This PR unifies them so they have the same stream_ids


neo/rawio/plexonrawio.py
```diff
@@ -258,61 +268,57 @@ def _parse_header(self):
...
+                # The users of plexon can modify the prefix of the channel names (e.g. `my_prefix` instead of `WB`).
+                # In that case we use the channel prefix both as stream id and name
+                stream_name = stream_id_to_stream_name.get(stream_id, stream_id)
                 signal_streams.append((stream_name, stream_id))
 
             signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
 
+            self.stream_id_samples = {}
+            self.stream_id_sampling_frequency = {}
+            self.stream_index_to_stream_id = {}
```
zm711, line 307: the general idea of these attributes was that the equivalents were private before. Why switch to public in general?

neo/rawio/plexonrawio.py
```diff
@@ -258,61 +268,57 @@ def _parse_header(self):
...
+            for stream_index, stream_id in enumerate(signal_streams["id"]):
+                # Keep a mapping from stream_index to stream_id
+                self.stream_index_to_stream_id[stream_index] = stream_id
+
+                # We extract the number of samples for each stream
+                mask = signal_channels["stream_id"] == stream_id
+                signal_num_samples = np.unique(channel_num_samples[mask])
+                assert signal_num_samples.size == 1, "All channels in a stream must have the same number of samples"
```
zm711, line 315: Just a style thing. I need to work on the dev docs, but at Neo we were trying to push for specific errors rather than asserts. So this would be a NeoReadWriteError.

neo/rawio/plexonrawio.py
```diff
@@ -258,61 +268,57 @@ def _parse_header(self):
...
+                self.stream_id_samples[stream_id] = signal_num_samples[0]
+
+                signal_sampling_frequency = np.unique(signal_channels[mask]["sampling_rate"])
+                assert signal_sampling_frequency.size == 1, "All channels in a stream must have the same sampling frequency"
```
zm711, line 319: Same here.

zm711 (commented):
Thanks Heberto one question and then two tiny fixes to make.

neo/rawio/plexonrawio.py
h-mayorquin: Ok, makes sense.

h-mayorquin: Changed.

h-mayorquin:
> https://github.com/NeuralEnsemble/python-neo/actions/runs/10776574774/job/29883638305?pr=1547
> 
> I just want to save this failed attempt with 15 times. This is good to know since I haven't seen the runner fail since you implemented the 10 retries. This is interesting.

Oh, what a pain : /

zm711:
Agreed. I think explicitly setting this is probably a better idea rather than setting trying until success. So I don't know, if we set something big like 100 I don't want to slow down the test suite. Maybe we hive off plexon2 as a conditional test--only when baserawio, toml, or plexon2 changes. Any opinions?

h-mayorquin:
Let's measure how much time it takes.

h-mayorquin:
@zm711 This is ready.

I measured the opening on my computer for a file of around 2 GiB and it took around two milliseconds:

2.46 ms ± 58.6 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)



```python
import platform
from urllib.request import urlopen
import pathlib 

architecture = platform.architecture()[0]
if architecture == "64bit" and platform.system() in ["Windows", "Darwin"]:
    file_name = "PL2FileReader64.dll"
else:  # Apparently wine uses the 32 bit version in linux
    file_name = "PL2FileReader.dll"
pl2_dll_folder = pathlib.Path.home() / ".plexon_dlls_for_neo"
pl2_dll_folder.mkdir(exist_ok=True)
pl2_dll_file_path = pl2_dll_folder / file_name

if not pl2_dll_file_path.exists():
    url = f"https://raw.githubusercontent.com/Neuralensemble/pypl2/master/bin/{file_name}"
    dist = urlopen(url=url)

    with open(pl2_dll_file_path, "wb") as f:
        print(f"dll file does not exist, downloading plexon dll to {pl2_dll_file_path}")
        f.write(dist.read())

from neo.rawio.plexon2rawio.pypl2.pypl2lib import PyPL2FileReader

pl2reader = PyPL2FileReader(pl2_dll_file_path=pl2_dll_file_path)from pathlib import Path

file_path = Path("/home/heberto/Downloads/1_conspecific_vocalization_block_10.16.23.pl2")
assert file_path.exists()   
filename = str(file_path)

# In another cell
%%timeit
pl2reader.pl2_open_file(filename)

```

zm711:
Cool, But seems like you didn't change the open tries for plexon2? I'm just wondering if we should change that (in a new PR?). Let me double check this PR one last time after my morning meeting and then we can merge. I'm just going to update the branch first.

h-mayorquin:
I just changed them.

zm711 (approved):
Looks good to me, note that we increased the attempts because this is failing on CI sometimes likely due to a wine issue for the CI runner.

Since we are meeting tomorrow I'll wait for our meeting to merge.

h-mayorquin:
OK.

Issue #1549, opened 2024-09-09
python-neo fails to load unsorted spike channels from PL2 files


nikhilchandra:
**Describe the bug**
python-neo fails to load spike channels from PL2 (Plexon2) files containing only unsorted units.

**To Reproduce**
Find a PL2 file that includes spike channels containing only unsorted units. One candidate is NC16FPSPKEVT_1m.pl2 included in the ephys_testing_data dataset. Upon loading the file using plexon2rawio, you will find that the 4 spike channels that contain spikes (SPK01, SPK05, SPK09, SPK13) are not loaded because they contain only unsorted spikes.

**Expected behaviour**
I expect channels with unsorted spikes to be included.

**Environment:**
 - OS: Windows
 - Python version: 3.11.9
 - Neo version: 0.14.0dev
 - NumPy version: 1.26.4

**Additional context**
I have a potential fix and will submit it soon.

zm711:
thanks @nikhilchandra! We appreciate your work on this!

PR #1551, opened 2024-09-10
Fixed bug that prevented PL2 spike channels with unsorted spikes from…

nikhilchandra:
… loading properly.

See [Issue #1549](https://github.com/NeuralEnsemble/python-neo/issues/1549)

### DIFF
--- a/neo/rawio/plexon2rawio/plexon2rawio.py
+++ b/neo/rawio/plexon2rawio/plexon2rawio.py
@@ -231,7 +231,7 @@ class Plexon2RawIO(BaseRawIO):
             if not (schannel_info.m_ChannelEnabled and schannel_info.m_ChannelRecordingEnabled):
                 continue
 
-            for channel_unit_id in range(schannel_info.m_NumberOfUnits):
+            for channel_unit_id in range(schannel_info.m_NumberOfUnits+1):
                 unit_name = f"{schannel_info.m_Name.decode()}.{channel_unit_id}"
                 unit_id = f"unit{schannel_info.m_Channel}.{channel_unit_id}"
                 wf_units = schannel_info.m_Units


samuelgarcia:
Hi @nikhilchandra.
thanks a lot for for this.
The fix looks a bit weird, is this one based or zeros based ?
`m_NumberOfUnits` variable name indicate that it is the number of units. So `+1` is strange.
If the `+1` is an extra hidden unsorted units maybe we should add a clear coment in the code for futur reading.
No ?

zm711:
> The fix looks a bit weird, is this one based or zeros based ?
m_NumberOfUnits variable name indicate that it is the number of units. So +1 is strange.
If the +1 is an extra hidden unsorted units maybe we should add a clear coment in the code for futur reading.
No ?

I agree with Sam. If you are 1-indexed then I would expect the range to be `range(1, n+1)`. If 0-indexed then why do you need to add 1. I think either way we need a comment above the range to explain what this means. This has bitten us before where someone with knowledge added a crucial index thing, but someone else tries to figure it out later.

nikhilchandra:
Here is the comment I have written up. Let me know if you have any concerns.

In a PL2 spike channel header, the field "m_NumberOfUnits" denotes the number of units to which spikes detected on that channel have been assigned. It does not account for unsorted spikes, i.e., spikes that have not been assigned to a unit. It is Plexon's convention to assign unsorted spikes to channel_unit_id=0, and sorted spikes to unit_id's 1, 2, 3...etc. Therefore, for a given value of m_NumberOfUnits, there are m_NumberOfUnits+1 channel_unit_ids to consider - 1 unsorted channel_unit_id (0) + the m_NumberOfUnits sorted channel_unit_ids.

zm711:
That looks like a great comment my only question as someone who doesn't look at the plexon code too much other than for review is whether `channel_unit_ids` is a plexon convention or a neo convention. If plexon it would be helpful to translate it to what the neo variable is called within the comment! 

So it seems with our previously implementation we were actually including unsorted spikes, but losing the final sorted spike unit?

nikhilchandra:
"channel_unit_ids" was part of the code before I made my changes, and I tried to use the naming conventions that were already present in the code while writing my comment.

zm711 (approved):
Good by me. @samuelgarcia does this explanation work for you?

zm711:
@nikhilchandra one tiny github trick for you to use. If you type 
fixes #1549, it allows us to autoclose the issue as soon as the PR is merged. It helps with the GitHub issue clean-up for us. Let me know if you have question about that.

nikhilchandra:
@zm711 I see -- where would have been the place to write fixes #1549? In the title?

zm711:
Or in the first post where I wrote "See issue https://github.com/NeuralEnsemble/python-neo/issues/1549" I should have written "fixes https://github.com/NeuralEnsemble/python-neo/issues/1549" instead?

Exactly. So I would recommend if you have  a PR that helps but doesn't close an issue you could write something like 
related to or linked to issue x. If you are fixing an issue you would write in the text box

Fixes issue X. Then it autolinks the issue and closes on merge.

Issue #1550, opened 2024-09-10
BaseTestRawIO's setUp function deletes data needed for subsequent tests


nikhilchandra:
**Describe the bug**
I am putting together a unit test for issue #1546 ([Link](https://github.com/NeuralEnsemble/python-neo/pull/1546)) but when I try to run unittest.main() from test_plexon2rawio.py, the code breaks.

**To Reproduce**

1. I added a unit test called "test_check_enabled_flags" to TestPlexon2RawIO in test_plexon2rawio.py.
2. I ran unittest.main() from test_plexon2rawio.py
3. I think that BaseTestRawIO's setUp() function runs before each unit test. So before test_check_enabled_flags actually ran, setUp() called download_dataset() from neo.utils.dataset. On Line 69, the command "dataset=datalad.api.install(path=local_folder, source=repo)" triggered lazy loading of the data repository to my $HOME folder (files are all 1 KB large)
4. Within download_dataset() there is a subsequent command on line 75 -- dataset.get(remote_path). With remote_path set to "plexon", this triggered the actual download of the plexon files. Here are the contents of the plexon folder after running this command: ![image](https://github.com/user-attachments/assets/f55b130a-9854-4884-8633-f4165aaf851c)
5. My unit test ran fine.
6. BaseTestRawIO has another unit test called test_read_all. Before this ran, setUp() was called again, and hence download_dataset(). This time, since ephy_testing_data had already been downloaded to my $HOME folder, and so a different branch of the if/else statement ran.  On line 66, the command "repo.call_git(["checkout", "--force", "master"]) seemed to wipe out a subset of the files contained within the folder "plexon". Here are the contents after running this command: 
![image](https://github.com/user-attachments/assets/9e21f712-a259-4f99-9d4e-a65d72376ad1)
7. When dataset.get(remote_path) runs for the 2nd time, nothing happens -- the files remain 1 KB.
8. When test_read_all runs, it tries to load 4chDemoPL2.pl2, which is empty. All values contained in the file_info header are zero. Therefore, when the code tries to run compliance.count_element(reader) on Line 114 of common_rawio_test.py, it breaks on a ZeroDivisionError. Here is the error message:

`Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\nikhil\Documents\Projects\python-neo\neo\rawio\baserawio.py", line 463, in segment_t_start
    return self._segment_t_start(block_index, seg_index)
[...]

PR #1553, opened 2024-09-11
Make  `BaseTestRawIO` setup per class and not per test

h-mayorquin:
Related to the discussion in #1550

### DIFF
--- a/neo/test/rawiotest/common_rawio_test.py
+++ b/neo/test/rawiotest/common_rawio_test.py
@@ -52,7 +52,7 @@ class BaseTestRawIO:
     rawioclass = None  # the IOclass to be tested
 
     entities_to_test = []  # list of files to test compliances
-    entities_to_download = []  # when files are at gin
+    entities_to_download = []  # when files a   re at gin
 
     # allow environment to tell avoid using network
     use_network = can_use_network()
--- a/neo/test/rawiotest/common_rawio_test.py
+++ b/neo/test/rawiotest/common_rawio_test.py
@@ -52,7 +52,7 @@ class BaseTestRawIO:
     rawioclass = None  # the IOclass to be tested
 
     entities_to_test = []  # list of files to test compliances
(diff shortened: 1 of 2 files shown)


zm711:
Looks like maxwell testing relies directly on setUp? Do you want to dig into or you want me to? 

```
FAILED neo/test/rawiotest/test_maxwellrawio.py::TestMaxwellRawIO::test_read_all - AttributeError: type object 'BaseTestRawIO' has no attribute 'setUp'
```

h-mayorquin:
Fixed.

neo/test/rawiotest/common_rawio_test.py
```diff
@@ -52,21 +52,22 @@ class BaseTestRawIO:
     rawioclass = None  # the IOclass to be tested
 
     entities_to_test = []  # list of files to test compliances
-    entities_to_download = []  # when files are at gin
+    entities_to_download = []  # when files a   re at gin
```
zm711, line 55:
```suggestion
    entities_to_download = []  # when files are at gin
```
haha itchy space bar finger?

zm711:
@nikhilchandra, want to test to see if this fixes your issues using the testing locally?

h-mayorquin:
I want to add that I think this is a good improvement on its own for tests with multiple cases so the download is not attempted twice anyway.

zm711:
> I want to add that I think this is a good improvement on its own for tests with multiple cases so the download is not attempted twice anyway.

I agree. I was hoping for a magic speed-up. But it looks like we go from 7-8  min to 6-7 min. So a slight (although not statistically tested) in test speed.

h-mayorquin:
1 minute is great : )

nikhilchandra:
Will give it a shot, setting up a Linux box for testing.

nikhilchandra:
@zm711 will give it a shot in a bit, thanks!

zm711 (approved):

PR #1558, opened 2024-09-18
IntanRawIO: Fix memmap order for Intan one-file-per-stream

### LINKED ISSUE #1556: IntanIO problem with one file per stream format
**Which page is the problem on?**
[IntanIO ](https://neo.readthedocs.io/en/latest/iolist.html#neo.io.IntanIO)

**What is the problem?**
IntanIO reads problematic. The function works well with rhd files but is problematic with dat files. Here are the pictures of the result.
from_raw_data
neo_intan_fig1
neo_intan_fig1_1
When reading .dat file from numpy directly(Fig 1), everything goes well. But the IntanIO seems wrongly loaded data. I doubt whether the data type is causing this problem. [.rhd, uint16; .dat, int16].

**Suggestions for fixing the problem**
Could you please provide more information about reading .dat files?

zm711:
Fixes #1556 

Data before fix:
image

Data after fix:

image

Basically, since Intan was designed to be easily read with Matlab (their example work is to use `fread` which is column-major reading) we should make our memmap with `order='F'`.

@h-mayorquin want to look over this quickly and @alejoe91 I think this can be a quick review as long as everyone is agreed. See the linked issue for more information.

Can you test this @Tevin-yue?

NOTE: this is simulated data so it should look like a repeating motif of simulated "spikes".

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -179,6 +179,9 @@ class IntanRawIO(BaseRawIO):
             self._raw_data = np.memmap(self.filename, dtype=memmap_data_dtype, mode="r", offset=header_size)
 
         # for 'one-file-per-signal' we have one memory map / neo stream
+        # based on https://github.com/NeuralEnsemble/python-neo/issues/1556 and the recommendations on the
+        # Intan website the data appears to be ordered column-major ('F') rather than row-major ('C') so
+        # when making our memmaps we should do order='F'
         elif self.file_format == "one-file-per-signal":
             self._raw_data = {}
             for stream_index, (stream_index_key, stream_datatype) in enumerate(memmap_data_dtype.items()):
@@ -188,7 +191,7 @@ class IntanRawIO(BaseRawIO):
                 dtype_size = np.dtype(stream_datatype).itemsize
                 n_samples = size_in_bytes // (dtype_size * num_channels)
                 signal_stream_memmap = np.memmap(
-                    file_path, dtype=stream_datatype, mode="r", shape=(num_channels, n_samples)
+                    file_path, dtype=stream_datatype, mode="r", shape=(num_channels, n_samples), order='F',
                 ).T
                 self._raw_data[stream_index] = signal_stream_memmap
 
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -179,9 +179,7 @@ class IntanRawIO(BaseRawIO):
             self._raw_data = np.memmap(self.filename, dtype=memmap_data_dtype, mode="r", offset=header_size)
 
         # for 'one-file-per-signal' we have one memory map / neo stream
-        # based on https://github.com/NeuralEnsemble/python-neo/issues/1556 and the recommendations on the
-        # Intan website the data appears to be ordered column-major ('F') rather than row-major ('C') so
-        # when making our memmaps we should do order='F'
+        # based on https://github.com/NeuralEnsemble/python-neo/issues/1556 bug in versions 0.13.1, .2, .3
         elif self.file_format == "one-file-per-signal":
             self._raw_data = {}
             for stream_index, (stream_index_key, stream_datatype) in enumerate(memmap_data_dtype.items()):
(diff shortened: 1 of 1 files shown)


neo/rawio/intanrawio.py
```diff
@@ -188,7 +191,7 @@ def _parse_header(self):
                 dtype_size = np.dtype(stream_datatype).itemsize
                 n_samples = size_in_bytes // (dtype_size * num_channels)
                 signal_stream_memmap = np.memmap(
-                    file_path, dtype=stream_datatype, mode="r", shape=(num_channels, n_samples)
+                    file_path, dtype=stream_datatype, mode="r", shape=(num_channels, n_samples), order='F',
```
samuelgarcia, line 194: why not 
```suggestion
                    file_path, dtype=stream_datatype, mode="r", shape=(n_samples, num_channels), order='C',
```

zm711: Good point. Let me simplify. We originally were reimplementing the Matlab function from the Intan docs, so there is also a `.T` of the map that we need to remove.

h-mayorquin:
I don't have your experimental expertise so this is for my learning. How do you know based on your plots (before and after fix) that your fix is right?

zm711:
I don't have your experimental expertise so this is for my learning. How do you know based on your plots (before and after fix) that your fix is right?

The values are really weird on the axis (almost like we are summing spikes ie the samples and channels values are mixed up). And for intan simulated data it should like a regularly spiking (regular in the non FS/RS way) neuron so it should have a nice downward followed by upward deflections on a background of some level of noise. The "bad" graph looks to sinusoidal (again from in appropriate summing of the voltages).

so this meant either we scaled wrong. I double checked and we weren't. Or we loaded the data in wrong (easy to mix up samples and channels when making the memmap). Since we did the documentation from intan (I think this part was me so mea culpa), I did the F-major style but didn't specify. Sam's tip was just flip the sample and channel to allow us to read the C style.

So to summarize I would accept amplifier/wideband data to be noisy with spikes in it. Not weird sinusoidal things.

Or do you have an additional question related to this?

zm711:
Maybe this will help you @h-mayorquin if I plot all channels instead of one channel.

image

You see we have sinusoidal noise (because this is simulated data. In real data it wouldn't be so clean ahah) with a bunch of spikes!

h-mayorquin (approved):
LGTM

Thanks for the diagrams.

I am trying to think on whether we could make a list of heuristics to check when we make a change to avoid doing mistakes like this I wonder.

As I am not a very graphical person I also wrote some code to  explain the nature of the error so it remains here for future reference:

```python
num_channels = 3
num_samples = 2
# This is according to the documention how to the data is layoued A_{channel}(sample)
data = np.asarray(["A1(0)", "A2(0)", "A3(0)", "A1(1)", "A2(1)", "A3(1)"], dtype='U5')
data.tofile("testing_order.bin")


data_memmap = np.memmap(
    "testing_order.bin", 
    dtype='U5', 
    mode='r', 
    shape=(num_channels, num_samples), 
    order='C'
)

print("Normal")
print(data_memmap)
print("Transposed")
print(data_memmap.T)

Normal
[['A1(0)' 'A2(0)']
 ['A3(0)' 'A1(1)']
 ['A2(1)' 'A3(1)']]
Transposed
[['A1(0)' 'A3(0)' 'A2(1)']
 ['A2(0)' 'A1(1)' 'A3(1)']]
```

You can see that the channels are scrambled 

The fix is like this:

```python
num_channels = 3
num_samples = 2
# This is according to the documention how to the data is layoued A_{channel}(sample)
data = np.asarray(["A1(0)", "A2(0)", "A3(0)", "A1(1)", "A2(1)", "A3(1)"], dtype='U5')
data.tofile("testing_order.bin")


data_memmap = np.memmap(
    "testing_order.bin", 
    dtype='U5', 
    mode='r', 
    shape=(num_samples, num_channels), 
    order='C'
)

print(data_memmap)

[['A1(0)' 'A2(0)' 'A3(0)']
 ['A1(1)' 'A2(1)' 'A3(1)']]

```

zm711:
Honestly dude you should propose that to Numpy! They just link to the wikipedia to explain this, but at least with your example you show what is actually happening with Numpy code. That's great!

@alejoe91 or @samuelgarcia. this is good to go now :) Tested and everyone is happy!

alejoe91 (approved):

PR #1561, opened 2024-09-20
Add back in test of file using props_only option.

PeterNSteinmetz:
Uses the existing NoDateHeader test data.

### DIFF
--- a/neo/test/rawiotest/test_neuralynxrawio.py
+++ b/neo/test/rawiotest/test_neuralynxrawio.py
@@ -369,7 +369,7 @@ class TestNlxHeader(TestNeuralynxRawIO, unittest.TestCase):
 
         hdr = NlxHeader(filename, props_only=True)
 
-        self.assertEqual(len(hdr), 11)
+        self.assertEqual(len(hdr), 11) # 9 properties plus channel_ids and channel_names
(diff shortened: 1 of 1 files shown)


neo/test/rawiotest/test_neuralynxrawio.py
```diff
@@ -360,20 +360,18 @@ def test_equality(self):
...
+        with self.assertRaises(IOError):
+            hdr = NlxHeader(filename)
 
-#         hdr = NlxHeader(filename, props_only=True)
+        hdr = NlxHeader(filename, props_only=True)
 
-#         self.assertEqual(len(hdr), 11)
-#         self.assertEqual(hdr['ApplicationName'], 'Pegasus')
-#         self.assertEqual(hdr['FileType'], 'Event')
+        self.assertEqual(len(hdr), 11)
```
zm711, line 372: For a magic test like this could we get a comment explaining why this should be 11. Is it always 11 or because you made the file you know it is 11? Thanks @PeterNSteinmetz !

PeterNSteinmetz: OK, I added a comment about how there are 9 properties in the file plus the channel_ids and channel_names properties added.

zm711 (approved):
Thanks!

PR #1568, opened 2024-09-25
[NeuralynxRawIO] Clean time parsing

PeterNSteinmetz:
At long last, replaced the complicated header dictionaries with more general patterns and checks. There are a few more tweaks on this that will be nice but this is a good chunk. As before, depends on the prior chain with NeuraviewTestAndRefactor.

### DIFF
--- a/neo/rawio/neuralynxrawio/ncssections.py
+++ b/neo/rawio/neuralynxrawio/ncssections.py
@@ -37,6 +37,19 @@ gaps from those introduced by a quirk of the hardware, recording software, or fi
 import math
 import numpy as np
 
+from enum import IntEnum, auto
+
+
+class AcqType(IntEnum):
+    PRE4 = auto()
+    BML = auto()
+    DIGITALLYNX = auto()
+    DIGITALLYNXSX = auto()
+    CHEETAH64 = auto()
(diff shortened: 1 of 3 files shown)


PeterNSteinmetz:
Sadly the delay in review has now caused there to be conflicts between my others use of the latest master python-neo and these changes. I guess no conflicts with python-neo per se yet.

PeterNSteinmetz:
I should also mention that the change to the BaseRawIO._signal_channel_dtype broke a lot of derived classes that I have. I know that is not part of the public interface but thought I should mention it.

zm711:
Hey Peter yeah, sorry the buffer io was something we had planned for quite a spell and needed to get done. Sam was putting all of his effort into that lately. We have one last big buffer based set of changes, but reviewing these PRs is on Sam's todo list. (I remind him to do it :) )

PeterNSteinmetz:
I guess I don't understand the need for buffer_id, even having read the enhancement proposal, but I assume it is important at some level. 

I ended up merging the changes in this PR into my master branch forked from the main master. There weren't many conflicts so hopefully that should not be a problem in the future when these changes here are merged into the main master.

PR #1571, opened 2024-09-26
Add filename property

PeterNSteinmetz:
Adds the OriginalFileName property to the NlxHeader. Also cleans up private members and moves a useful method to NlxHeader from test. 

This is the last in this sequence of commits for a while now. This now provides a much clearer method of parsing the header which works with all extant files (@JuliaSprenger and thus provides that refactoring discussed a while back ;-). It should be more flexible for future possible variations and easier to modify if needed.

If one wanted to squish all of these since the divergence to master there would be much cleaner code to review, however, it would be major changes in one shot.

### DIFF
--- a/neo/rawio/neuralynxrawio/ncssections.py
+++ b/neo/rawio/neuralynxrawio/ncssections.py
@@ -37,6 +37,19 @@ gaps from those introduced by a quirk of the hardware, recording software, or fi
 import math
 import numpy as np
 
+from enum import IntEnum, auto
+
+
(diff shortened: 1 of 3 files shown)


zm711:
Thanks Peter. Sam said he would review this when he has some time. So I assigned him to all of these and he will be the main reviewer.

samuelgarcia:
@zm711 : this is ready to merge on my side.

samuelgarcia (approved):

PR #1572, opened 2024-09-27
Neuronexus datetime parsing

h-mayorquin:
Follow up on this:

#3373

From this [comment](https://github.com/NeuralEnsemble/python-neo/pull/1509#discussion_r1777624427) I did not have enough time to address there

Just a more standard way of parsing datetime strings @zm711

### DIFF
--- a/neo/rawio/neuronexusrawio.py
+++ b/neo/rawio/neuronexusrawio.py
@@ -228,9 +228,9 @@ class NeuroNexusRawIO(BaseRawIO):
         # TODO: remove once minimal version is 3.11
         # Python 3.10 and below do not support the fractional seconds in fromisoformat
         if sys.version_info.minor <= 3.11:
-            datetime_string = re.sub(r"(\.\d{6})\d+", r"\1", datetime_string)        
-        
-        rec_datetime = datetime.datetime.fromisoformat(datetime_string)
+            rec_datetime = datetime.datetime.strptime(datetime_string, "%Y-%m-%dT%H:%M:%S.%f%z")
+        else:
+            rec_datetime = datetime.datetime.fromisoformat(datetime_string)
 
         bl_annotations = self.raw_annotations["blocks"][0]
         seg_annotations = bl_annotations["segments"][0]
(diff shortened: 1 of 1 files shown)


zm711:
Thanks Sam told me something like that existed, but he couldn't remember so he said we could merge it and that you would swoop in and save the day :)

zm711:
Interesting it says it's not isoformat?

h-mayorquin:
No, that's iso format. It seems that support for reading iso datetimes was improved on python 3.11. This works just fine locally for me.

More worrisome I discovered that your test on python 3.11 are not really on python 3.11:

image

zm711:
I think we must've had a bad cache and it kept using the bad cache because it only updates if the cache is wrong in the action. I guess let me remove the cache and we can try again without the cache.

neo/rawio/neuronexusrawio.py
```diff
@@ -219,21 +219,10 @@ def _parse_header(self):
...
-        hour = int(time_info[0][-2:])
-        minute = int(time_info[1])
-        second = int(float(time_info[2]))
-        microsecond = int(1000 * 1000 * (float(time_info[2]) - second))# second -> micro is 1000 * 1000
-
-        rec_datetime = datetime.datetime(year, month, day, hour, minute, second, microsecond)
+        # date comes out as: '2024-07-01T13:04:49.4972245-04:00' so in ISO format
+        datetime_string = self.metadata["status"]["start_time"]
+        rec_datetime = datetime.datetime.fromisoformat(datetime_string)
+
```
zm711, line 225: what if we do
try-except where we cut off the trailing -4:00 which I think was support < 3.11.

zm711:
Cache is now fixed!

neo/rawio/neuronexusrawio.py
```diff
@@ -219,21 +221,17 @@ def _parse_header(self):
...
-        second = int(float(time_info[2]))
-        microsecond = int(1000 * 1000 * (float(time_info[2]) - second))# second -> micro is 1000 * 1000
-
-        rec_datetime = datetime.datetime(year, month, day, hour, minute, second, microsecond)
+        # date comes out as: '2024-07-01T13:04:49.4972245-04:00' so in ISO format
+        datetime_string = self.metadata["status"]["start_time"]
+        
+
+        # TODO: remove once minimal version is 3.11
+        # Python 3.10 and below do not support the fractional seconds in fromisoformat
```
zm711, line 229: great comments thanks!

zm711 (approved):

zm711:
Thanks Heberto!

zm711:
And thanks for the learning experience!

PR #1573, opened 2024-09-27
Drop python 3.8

zm711:
Since 3.8 goes EOL in October we should be good to drop this.


apdavison:
yes, based on our [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) + 1 year policy, we should be on Python 3.9+ now. Also NumPy 1.22+

zm711:
Cool, I didn't change NumPy yet since Numpy 1.21 works with 3.9. But we could drop that in another PR.

PR #1582, opened 2024-10-13
Add python 3.13 to CI

zm711:
Python 3.13 released. See notes [here](https://www.python.org/downloads/release/python-3130/). We should be testing for it now :) 

I think we had mentioned we wanted our CI to be oldest and newest python at some point too. Let's see what happens...

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -47,12 +47,12 @@ jobs:
              numpy-version: '1.25.1'
     steps:
         - name: Set up Python ${{ matrix.python-version }}
-          uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


zm711:
Seems like NumPy 1.26.4 is not ready for python 3.13 just for Windows. I guess we wait. IO tests passed though which is nice!

zm711:
Actually there is no guarantee that 3.13 will work with 1.26 so let me drop that soon. Seems like there's no point to updating to 3.13 until we have the numpy 2.0 stuff solved

https://github.com/numpy/numpy/issues/24810

zm711:
Will be closed by #1490 :)

PR #1580, opened 2024-10-12
0.13.4 Release

zm711:
Feel free to start proofing. Just wanted to get this started in particular I started editing the codemeta json, but then the links will be wrong until I actually do the release, so maybe I fix those in an after release?

I updated all PRs and issues tagged as 0.14.0 back down to 0.13.4.

PR #1583, opened 2024-10-16
Limit dhn_med_py

zm711:
See discussion here:

https://github.com/MEDFormat/MEDPython/issues/7

I discovered this while working on #1580, but I think it is better to get this into main here rather than in the release PR.


samuelgarcia (approved):

samuelgarcia:
merci

PR #1584, opened 2024-10-16
EDFIO: Alleviate EDF single handle problem

### LINKED ISSUE #1557: EDF blocks access when io is opened.
This is coming from here in Spikeinterface:

https://github.com/SpikeInterface/spikeinterface/issues/1228

Here, basically this format can't only be opened once per file-system.

```python
from pathlib import Path

from spikeinterface.extractors import EDFRecordingExtractor
from spikeinterface.core.base import _load_extractor_from_dict

file_path = Path("/home/heberto/neuroconv_testing_data/ephy_testing_data/edf/edf+C.edf")

recorder = EDFRecordingExtractor(file_path=file_path)
recorder_dict = recorder.to_dict()
recorder_from_dict = _load_extractor_from_dict(recorder_dict)


  File "/home/heberto/miniconda3/envs/spikeinterface_env/lib/python3.10/site-packages/neo/rawio/edfrawio.py", line 70, in _parse_header
    self.edf_reader = EdfReader(self.filename)
  File "pyedflib/_extensions/_pyedflib.pyx", line 146, in pyedflib._extensions._pyedflib.CyEdfReader.__init__
  File "pyedflib/_extensions/_pyedflib.pyx", line 209, in pyedflib._extensions._pyedflib.CyEdfReader.open
  File "pyedflib/_extensions/_pyedflib.pyx", line 181, in pyedflib._extensions._pyedflib.CyEdfReader.check_open_ok
OSError: /home/heberto/neuroconv_testing_data/ephy_testing_data/edf/edf+C.edf: file has already been opened
```

h-mayorquin:
This should close #1557

This PR moves the logic of opening the edf reader to the functions that extract the data (`_get_analogsignal_chunk`).

This avoids the problem of someone initializing the io (in a notebook for example) and then having a failure to initialize it again because the handle was left dangling. Moreover, it would allow parallel processing to work better (but not perfect as it might be concurrent access crashes).

The performance costs are minimal in this case because the io is really slow (I think this can be fixed but that's a matter for another PR). For example, I have this mid-sized file 7000 MiB, 276 channels, and around 20 minutes. It takes 10.8 seconds to make a reading of the whole data in master. With the new changes it takes ... 11.0 seconds:

image

The moral is that when the data extraction itself takes most of the time, the `io.open` instruction cost does not count.

As mentioned in #1557 the best way would be to handle this with `mne` library but this should be an improvement over the current state of affair where this io does not play nice with a lot of downstream usage patterns (spike interface re-opens the io to get streams for example which precludes get the streams and then opening the reader again for the extractor).


zm711 (commented):
Sort of annoying, but I think I would prefer symmetry. Either delete the `close` and only have `_close_reader` or make an `open` function that calls the `_open_reader` internally. Does that make sense?

h-mayorquin:
I would rather erase close but that is a public method that I did not add. Would break backwards compatibility:

https://github.com/NeuralEnsemble/python-neo/blob/ade13db1cc5be010825b8ec2a44459793868fd9d/neo/test/rawiotest/test_edfrawio.py#L30-L41

If you are OK with that I can do it. I think that is a good thing, when we fix this the close method will stop making sense. For that reason, I also think that adding an open function would be a bad idea. 

That said, I changed the private method to be `_open_reader` instead of `open`.

h-mayorquin:
But I would like to deprecate the close method in another PR. I don't want this to derail the conversation when Sam reviews it.

neo/rawio/edfrawio.py
```diff
@@ -219,12 +224,13 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
 
         # load data into numpy array buffer
         data = []
+        self._open()
```
zm711, line 227:
```suggestion
        self._open_reader()
```

zm711:
That all works for me we can think about that later.

neo/rawio/edfrawio.py
```diff
@@ -174,6 +173,15 @@ def _parse_header(self):
         for array_key in array_keys:
             array_anno = {array_key: [h[array_key] for h in self.signal_headers]}
         seg_ann["signals"].append({"__array_annotations__": array_anno})
+        
+        # We store the following attributes for rapid access without needing the reader
+        
+        self._t_stop = self.edf_reader.datarecord_duration * self.edf_reader.datarecords_in_file
+        # use sample count of first signal in stream
+        self._stream_index_samples = {stream_index : self.edf_reader.getNSamples()[chidx][0] for stream_index, chidx in self.stream_idx_to_chidx.items()}
```
zm711, line 181: Where is `[0]` coming from? 

this whole block seems like a great idea, but I don't the extra [0] in previous `_get_signal_size`?

neo/rawio/edfrawio.py
```diff
@@ -183,14 +191,11 @@ def _segment_t_start(self, block_index, seg_index):
         return 0.0  # in seconds
 
     def _segment_t_stop(self, block_index, seg_index):
-        t_stop = self.edf_reader.datarecord_duration * self.edf_reader.datarecords_in_file
         # this must return an float scale in second
-        return t_stop
+        return self._t_stop
 
     def _get_signal_size(self, block_index, seg_index, stream_index):
-        chidx = self.stream_idx_to_chidx[stream_index][0]
```
h-mayorquin, line 191: @zm711 here is the zero. The idea is to use the first channel of the stream to get the samples.

neo/rawio/edfrawio.py
h-mayorquin: See https://github.com/NeuralEnsemble/python-neo/pull/1584/files#r1806665680

samuelgarcia (approved):

samuelgarcia:
OK for me.

zm711 (approved):
cool me too.

PR #1585, opened 2024-10-17
PlexonIO: Fix another plexon overflow

zm711:
@h-mayorquin could I have you give this a quick read through. Then we can have Sam merge it later.

Error here:

https://readthedocs.org/projects/neo/builds/25985599/

### DIFF
--- a/neo/rawio/plexonrawio.py
+++ b/neo/rawio/plexonrawio.py
@@ -155,7 +155,9 @@ class PlexonRawIO(BaseRawIO):
         if self.progress_bar:
             progress_bar.close()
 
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
Ok.

I wonder how comes this failed now and not before and why this is being executed on the read the docs : /

zm711:
> I wonder how comes this failed now and not before 

I've been messing with CI stuff so maybe I changed over some cache thing or version thing. Who knows....

> why this is being executed on the read the docs : /

To show off Neo with real data. I suggested simulated in the past and the point was this can work with real data so it should show real data.

Thanks Heberto!

alejoe91 (approved):

PR #1589, opened 2024-10-25
Micromed segments

samuelgarcia:
I though that micromed was mono segment but apparently it can be multi segment.
Here the implementation.

This include a test that need theses two files to be merge in gin  https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/143


zm711:
@samuelgarcia do you think this is ready to go after getting the test files? Or do we want a milestone of 0.15.0 (for after next release).

samuelgarcia:
I think this is ready to go.
Let me check again.

neo/rawio/micromedrawio.py
```diff
@@ -99,25 +99,35 @@ def _parse_header(self):
...
-                "dtype": sig_dtype,
-                "order": "C",
-                "file_offset": 0,
-                "shape": signal_shape,
-            }
+
+            # "TRONCA" zone define segments
+            zname2, pos, length = zones["TRONCA"]
+            f.seek(pos)
+            max_segments = 100
```
zm711, line 106: where does this 100 come from. We probably need a comment for any magic number.

neo/rawio/micromedrawio.py
```diff
@@ -99,25 +99,35 @@ def _parse_header(self):
...
+            self.info_segments = []
+            for i in range(max_segments):
+                seg_start = int(np.frombuffer(f.read(4), dtype="u4")[0])
```
zm711, line 109: Same here. Could we get a comment for why we are reading 4 at a time?

neo/rawio/micromedrawio.py
```diff
@@ -155,6 +163,32 @@ def _parse_header(self):
...
+            self._buffer_descriptions = {0 :{}}
+            for seg_index in range(nb_segment):
+                seg_start, trace_offset = self.info_segments[seg_index]
+                self._t_starts.append(seg_start / self._sampling_rate)
+
+                start = seg_limits[seg_index]
+                stop = seg_limits[seg_index + 1]
+                
+                shape = (stop - start, Num_Chan)
+                file_offset = Data_Start_Offset + start * np.dtype(sig_dtype).itemsize * Num_Chan
```
zm711, line 179: any desire to add parentheses to help the reader with the math for this file_offset calculation?
```suggestion
                file_offset = Data_Start_Offset + ( start * np.dtype(sig_dtype).itemsize * Num_Chan)
```

neo/rawio/micromedrawio.py
```diff
@@ -99,25 +99,35 @@ def _parse_header(self):
...
-                "type": "raw",
-                "file_path": str(self.filename),
-                "dtype": sig_dtype,
-                "order": "C",
-                "file_offset": 0,
-                "shape": signal_shape,
-            }
+
+            # "TRONCA" zone define segments
+            zname2, pos, length = zones["TRONCA"]
```
zm711, line 104: Light recommendation to have better name for `pos` :)

zm711 (commented):
I think we should do a bit of a rewrite of this RawIO in general. Some of the variable naming is hard to follow and this low level seeking is also not super easy to follow. We should add some more comments to what is happening at the steps. But that can always be a future PR.

neo/rawio/micromedrawio.py
samuelgarcia: 4 is 4bytes for "u4"

zm711: Sorry I meant why are you doing 'u4'? why not some other size for reading. How do you know reading u4 gets us what we need instead of some other size?  If we do a comment then we know that either 1) this size is documented or 2) you are doing this empirically.

neo/rawio/micromedrawio.py
zm711: It's not crucial here :) But if any micromed issues come up I will send them your way.

zm711:
To move forward with this I think you need to merge main into this branch to fix the CI/RTD builds. Then we can really see how the tests are doing :)

zm711 (approved):

PR #1594, opened 2024-11-04
OpenEphysBinary: Skip streams if folder doesn't exist

alejoe91:
This has come up several times in the past. If a user deletes a stream subfolder (even if irrelevane, e.g., MessageCenter) the RawIO fails to read the folder.

This PR simply skips missing folders.

@zm711 I'll add a test file to GIN

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -12,6 +12,7 @@ Author: Julia Sprenger, Samuel Garcia, and Alessio Buccino
 import os
 import json
 from pathlib import Path
+from warnings import warn
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


alejoe91:
@zm711 GIN PR here: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/144

neo/test/rawiotest/test_openephysbinaryrawio.py
```diff
@@ -1,4 +1,5 @@
 import unittest
+import pytest
```
zm711, line 2: What's this import for?

zm711 (commented):
Otherwise this looks good to me. Let me check the GIN really quick :)

neo/test/rawiotest/test_openephysbinaryrawio.py
alejoe91, line 2: I found a built in unit test function instead..removing!

alejoe91: removed

zm711 (commented):
Sorry one last question:

Are events and continuous completely separate? Or are they paired in some way? Or can I record just event or just continuous for different experiments? I just want to make sure the skipping stream strategy is best or are there cases where we should raise a hard error. For example what if someone deletes everything will this fail and say empty or would this pass with a bunch of warnings that could be shut off?

alejoe91:
they could be separate, but some have both. On the NEO side of things, they are separate :)

zm711 (approved):
Okay works for me :)

PR #1602, opened 2024-11-28
Remove stimfitio from pyproject

h-mayorquin:
As discussed with @samuelgarcia and @zm711 

this has not been working for a while.

Trying to install it generate this error.

```
pip install neo[stimfitio]==0.13.4
Collecting neo==0.13.4 (from neo[stimfitio]==0.13.4)
  Using cached neo-0.13.4-py3-none-any.whl.metadata (8.9 kB)
Requirement already satisfied: packaging in ./miniconda3/envs/test_env/lib/python3.11/site-packages (from neo==0.13.4->neo[stimfitio]==0.13.4) (24.1)
Requirement already satisfied: numpy<2.0.0,>=1.22.4 in ./miniconda3/envs/test_env/lib/python3.11/site-packages (from neo==0.13.4->neo[stimfitio]==0.13.4) (1.26.4)
Collecting quantities>=0.16.1 (from neo==0.13.4->neo[stimfitio]==0.13.4)
  Using cached quantities-0.16.1-py3-none-any.whl.metadata (8.4 kB)
INFO: pip is looking at multiple versions of neo[stimfitio] to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement stfio; extra == "stimfitio" (from neo[stimfitio]) (from versions: none)
ERROR: No matching distribution found for stfio; extra == "stimfitio"
```

And if you remove the version it goes back to a version of python-neo that did not have that extra and you end up with version 0.5.

This PR removes this.


h-mayorquin:
Mysterious failure of one test, @zm711 . Any hints on this?

apdavison:
this makes sense since stfio cannot be pip-installed, as per the comment below the "all" target.

h-mayorquin:
Would you be OK with removing it, @Antho2422? Right now this is working as text documentation which is a plus but it also breaks some tooling. I was mentioning Sam that uv tree ([a nice and efficient tool to build the dependency tree](https://docs.astral.sh/uv/getting-started/features/#projects)) is unable to work because stfio is not on an index (e.g. pipy) as far as I know. In addition, it also has the strange behavior described above with normal pip where you end up installing a really old version of neo. 

How do you think about it?

zm711 (commented):
Since I don't know stimfitio at all, I feel like we should make sure there is an appropriate error in the io that let's people know what's going on. Have you checked this @h-mayorquin ?

Also I reran tests to see if it's stochastic failures (which we will need to fix--but would be a separate issue).
--It is a random failure. We can track this if it comes up again. This was in filtering code (another section that I haven't explored very deeply).

h-mayorquin:
> ll, I feel like we should make sure there is an appropriate error in the io that let's people know what's going on.

What sort of error do you have in mind?

apdavison:
It is packaged for Debian, MacPorts, Homebrew, Windows, see:

https://github.com/neurodroid/stimfit?tab=readme-ov-file#installation

h-mayorquin:
I think that right now if someone was to use `StimfitIO` and they don't have the library you will get a classical import error. 

https://github.com/NeuralEnsemble/python-neo/blob/685fd3d9d259ca3131a8ad97d09c36bbf6558c00/neo/io/stimfitio.py#L93-L95

This seems good enough for me. You could use importlib or the try-except pattern with imports to get a more specific import error but I kind of feel that's an overkill.

zm711 (approved):
Works for me then :)

What is this even doing in the pyproject then...?

Issue #1437, opened 2024-03-19
Multiple issues with gettings durations for events from Open Ephys binary files


johnmbarrett:
**Describe the bug**
`OpenEphysBinaryRawIO._parse_header` refuses to extract durations when the number of rising edges detected exceeds the number of falling edges, even though there is enough information to compute almost all durations in this case, and also incorrectly assumes that successive events always have alternating polarity.

**To Reproduce**
1) Download this OpenEphys dataset: https://drive.google.com/file/d/1Hr3-GFOpOkNc1hgXdVr5HTunchQm4enw/view?usp=drive_link
2) Unzip the dataset somewhere
3) Run the following code
```
from neo.rawio.openephysbinaryrawio import OpenEphysBinaryRawIO

folder = # fill in the folder you unzipped the data to

io = OpenEphysBinaryRawIO(folder)

io._parse_header()

timestamps, durations, labels = io._get_event_timestamps(0,0,0,None,None)
```

**Expected behaviour**
The correct durations are returned for all events

**Actual behaviour**
`durations` is `None`

**Environment:**
 - OS: Windows 10
 - Python version: 3.9
 - Neo version: https://github.com/NeuralEnsemble/python-neo/pull/1430 (you can't load all events from multistream datasets without this PR)
 - NumPy version: 1.26.2

**Additional context**
For the first problem, the check on line 245 of `openephysbinaryrawio.py` is unnecessarily stringent. It should be less than or equal, not strict equality.

Moreover, the logic in this section doesn't work when events from multiple binary channels are interleaved on the same event stream (which is allowed by the Open Ephys specification and happens for e.g. data from an Intan RHD2000 USB evaluation board), as it incorrectly assumes that events alternate between rising and falling. However, if multiple channels are interleaved, then two rising or two falling events from different channels can follow each other, and so the duration calculation on line 246 is wrong. In the example dataset, this happens for the 26th & 27th events, in which a falling edge on channel 2 follows a falling edge on channel 3. A better approach would be, for each rising edge, to look for the next falling edge with the same absolute state value (and either drop that event from the list or assign it a NaN duration if none is found, rather than Noneing out the entire durations array)

alejoe91:
@johnmbarrett would you have time to make a PR and propose a fix?

vigji:
Reporting here similar issues.

[...]

PR #1603, opened 2024-12-02
OpenEphysBinaryRawIO: Fixing ttl multichan

vigji:
Addressing #1437.

Currently, all the logic in the events parser for OpenEphys data assumes that there is a single stream of events (simply positive or negative `states`), and it breaks down when acquiring multiple digital signals at once (`states` having different absolute values for different channels).

This is a simple fix that looks for rising and falling edges separately for each of the digital channels.

Add simple testing relying on existing test data, which fails before the fix and passes afterwards.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -252,47 +252,47 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                     # 'states' was introduced in OpenEphys v0.6. For previous versions, events used 'channel_states'
                     if "states" in info or "channel_states" in info:
                         states = info["channel_states"] if "channel_states" in info else info["states"]
-                            
+
                         if states.size > 0:
                             timestamps = info["timestamps"]
                             labels = info["labels"]
-                            
+
                             # Identify unique channels based on state values
                             channels = np.unique(np.abs(states))
-                            
+
                             rising_indices = []
                             falling_indices = []
-                            
+
                             for channel in channels:
                                 # Find rising and falling edges for each channel
                                 rising = np.where(states == channel)[0]
                                 falling = np.where(states == -channel)[0]
-                                
+
                                 # Ensure each rising has a corresponding falling
                                 if rising.size > 0 and falling.size > 0:
                                     if rising[0] > falling[0]:
                                         falling = falling[1:]
                                     if rising.size > falling.size:
                                         rising = rising[:-1]
-                                    
+
                                     rising_indices.extend(rising)
                                     falling_indices.extend(falling)
-                            
+
                             rising_indices = np.array(rising_indices)
                             falling_indices = np.array(falling_indices)
-                            
+
                             # Sort the indices to maintain chronological order
                             sorted_order = np.argsort(rising_indices)
                             rising_indices = rising_indices[sorted_order]
                             falling_indices = falling_indices[sorted_order]
-                            
+
                             durations = None
                             if len(rising_indices) == len(falling_indices):
                                 durations = timestamps[falling_indices] - timestamps[rising_indices]
                                 if not self._use_direct_evt_timestamps:
                                     timestamps = timestamps / info["sample_rate"]
                                     durations = durations / info["sample_rate"]
-                            
+
                             info["rising"] = rising_indices
                             info["timestamps"] = timestamps[rising_indices]
                             info["labels"] = labels[rising_indices]
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -263,9 +263,9 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                             rising_indices = []
                             falling_indices = []
 
-                            # all channels are packed into the same `states` array. 
+                            # all channels are packed into the same `states` array.
                             # So the states array includes positive and negative values for each channel:
-                            #  for example channel one rising would be +1 and channel one falling would be -1, 
+                            #  for example channel one rising would be +1 and channel one falling would be -1,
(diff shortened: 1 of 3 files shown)


alejoe91:
Thanks @vigji! Looking great!

vigji:
Great, let me know if there's anything needed before merging!

neo/rawio/openephysbinaryrawio.py
```diff
@@ -253,30 +252,50 @@ def _parse_header(self):
...
 
-                            # infer durations
+                            # Identify unique channels based on state values
+                            channels = np.unique(np.abs(states))
+
+                            rising_indices = []
+                            falling_indices = []
+
+                            for channel in channels:
+                                # Find rising and falling edges for each channel
```
zm711, line 267: Could we get a slightly more explanatory comment (maybe above L266). Based on the code it seems like the rising are encode by +1 channel so if I'm at channel 1 then +1 is rising and -1 is falling and then for channel 2 it would be +2 and -2.  What would the array of states actually look like? So I'm imagining something like:

```
all channels are packed into the same `states` array. So the states array includes positive and negative values for each channel for example channel one rising would be +1 and channel one falling would be -1. So we can check for where each channel appears to get each TTL rise and fall for each channel
```

Or something like this. Since we have to support multiple versions it is helpful in the cases where the format is changed for us to know exactly how the former format was set up.

zm711 (commented):
This looks good to me, although I would ask for one additional comment in the code to explain what version this works for (all or v0.6+ whatever) along with what the structure of the data is.

neo/rawio/openephysbinaryrawio.py
vigji: Ok, should have been fixed now!

zm711 (dismissed):
This works for me, but Alessio knows this way better so I'll let him decide when to merge/do final review.

neo/rawio/openephysbinaryrawio.py
```diff
@@ -253,30 +252,55 @@ def _parse_header(self):
...
+
+                                # Ensure each rising has a corresponding falling
+                                if rising.size > 0 and falling.size > 0:
+                                    if rising[0] > falling[0]:
+                                        falling = falling[1:]
+                                    if rising.size > falling.size:
+                                        rising = rising[:-1]
+
+                                    rising_indices.extend(rising)
+                                    falling_indices.extend(falling)
```
alejoe91, line 284: @vigji maybe it's better to skip the extend in case `rising_indices != falling_indices` here? Otherwise an issue with one channel will result in all durations being lost!

We could also add a warning in case this happens for a certain channel. What do you think?

alejoe91 (changes requested):
See comment

neo/rawio/openephysbinaryrawio.py
vigji: I guess there is also a question of what the edge cases might be. Is there really a scenario where this can break down? The only option is the acquisition board to be missing rising/falling edges, but I am not sure at that point I would still use the data - it is possible that could have happened for other digital channels as well, and pairs of miss-detections could go unnoticed, in principle.

Have you seen or heard of any of those edge scenarios?

zm711: One case could be if I accidentally stop a recording midstimulus (a mouse wakes up or the head-fixation apparatus comes loose). So maybe I want to save data, but I don't have a fall for the last "channel" of stim.  So I agree with Alessio!

alejoe91: That would still work though because [here](https://github.com/iurillilab/python-neo/blob/fixing-ttl-multichan/neo/rawio/openephysbinaryrawio.py#L277-L281) we check if the first falling is before the first rising and viceversa.

vigji: you raise a valid point @zm711 , but addressing that would require completely changing the way the interface represents digital data (currently, timestamp+duration of the ON state). Not sure this is the place for that. I agree that the current representation is not the most general, and I would be in favour of changing it, but I assume there might be a lot of code by now depending on it (or not - the fact that the bug addressed in this PR was still around lets me wondering how many people use this interface for multi-digital channel experiment logging).

@alejoe91 so you suggest to just add a warning, with no changes to the current logic in the end?

zm711: He's saying add those lines on top of your PR to skip a channel if it doesn't fit those conditions.

And fair enough. We have to live with the code we have :)

vigji: Should be gtg now!

alejoe91 (approved):

samuelgarcia:
Hi,
I was totally off for a while.
I read this very very quickly.
I suspect that this `np.where()` can slow down a lot the parse_header for long recording.
Is it a general or a corner case ?
We need to keep in mind that the parse header is done multiple time when multiprocessing on spikeinterface side so we need to be very carfull for this.
Do we have benchmark for this ?

vigji:
Not data but rule of thumbs benchmark:
assuming 10 channels sampling TTL pulses at 100 Hz for 5 hours I generate a random array of integers from 0 to 10, `np.where()` then takes approx. 15 ms on my laptop (MacBook Pro M1 2022). This matches my experience that I never see long times even for fast digital signals recorded for a couple of hours.

Does this make sense?

zm711:
@vigji do we have you in our author list? I'm prepping our documents for new release. Would you like me to add you to the list (and what is the institution you were at/or are at for your Neo contribution?)

vigji:
I don't think so as it was my first contribution! 
Yes I'd like to, my institute is [optional: Center for Neuroscience and Cognitive systems], Istituto Italiano di Tecnologia (IIT), Italy

zm711:
Perfect. I will add you before our next release :) Thanks for the contribution!

Issue #1511, opened 2024-07-24
Spikeglx: parse tcat input 


tabedzki:
**Describe the bug**
I am unable to read in my `lf` files when using SpikeGLXRawIO. The file structure is currently as below. When I had all the `lf` files be `t0` instead of `tcat` it worked flawlessly. 

```
ya008_20240525_g0
├── ya008_20240525_g0_ct_offsets.txt
├── ya008_20240525_g0_fyi.txt
├── ya008_20240525_g0_imec0
│  ├── ya008_20240525_g0_t0.imec0.ap.bin
│  ├── ya008_20240525_g0_t0.imec0.ap.meta
│  ├── ya008_20240525_g0_tcat.imec0.lf.bin
│  └── ya008_20240525_g0_tcat.imec0.lf.meta
├── ya008_20240525_g0_imec1
│  ├── ya008_20240525_g0_t0.imec1.ap.bin
│  ├── ya008_20240525_g0_t0.imec1.ap.meta
│  ├── ya008_20240525_g0_tcat.imec1.lf.bin
│  └── ya008_20240525_g0_tcat.imec1.lf.meta
├── ya008_20240525_g0_t0.nidq.bin
└── ya008_20240525_g0_t0.nidq.meta
```

**To Reproduce**
```
(neuroconv_312) ➜  test2 ipython
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import neo.rawio

In [2]: reader = neo.rawio.SpikeGLXRawIO(dirname='./ya008_20240525_g0')

In [3]: reader.parse_header()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[3], line 1
----> 1 reader.parse_header()

File /scratch/ct5868/.conda/envs/neuroconv_312/lib/python3.12/site-packages/neo/rawio/baserawio.py:189, in BaseRawIO.parse_header(self)
    176 """
    177 Parses the header of the file(s) to allow for faster computations
    178 for all other functions
    179
    180 """
    181 # this must create
    182 # self.header['nb_block']
    183 # self.header['nb_segment']
   (...)
    186 # self.header['spike_channels']
    187 # self.header['event_channels']
--> 189 self._parse_header()
[...]

PR #1608, opened 2024-12-09
Open folders with user altered filenames in SpikeGLX

### LINKED ISSUE #1606: Unable to open folders with user altered filenames in SpikeGLX
I have a use case when there is  multi probe data form SpikeGLX with the following structure:

```
├── probe_1
│   ├── npx.imec.ap.bin
│   └── npx.imec.ap.meta
├── probe_2
│   ├── npx.imec.ap.bin
│   └── npx.imec.ap.meta
└── probe_3
    ├── npx.imec.ap.bin
    └── npx.imec.ap.meta
```

The problem is that the probe index was omitted from the name. The pattern {device}{probe_index} is used as a stream_index so in this case we are unable to open this file and the error in this line is triggered:

https://github.com/NeuralEnsemble/python-neo/blob/25760377b944ce2856766db3e6c90a8612ef69d9/neo/rawio/spikeglxrawio.py#L136-L137

The solution, I think, is to determine the probe index directly from the metadata and use that to tell the streams apart. I have some stubbed test data and a patch that I will push soon.

h-mayorquin:
Should fix #1606

To solve #1606 this does two things:
1) It calculates device (probe) index from the metadata instead of relying on file parsing as we currently do. 
2) When the gate, port and dock are not available in the metadata it parses the filename as named in the metadata (as that was the original one) instead of from the filepath in the system as we are currently doing.

Both things should make the reader more robust. 

I am preparing some data for gin where the filename was changed. I will change this from draft once I add that.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -85,7 +85,7 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
     * This IO reads the entire folder and subfolders locating the `.bin` and `.meta` files
     * Handles gates and triggers as segments (based on the `_gt0`, `_gt1`, `_t0` , `_t1` in filenames)
     * Handles all signals coming from different acquisition cards ("imec0", "imec1", etc) in a typical
-        PXIe chassis setup and also external signal like"nidq". 
+        PXIe chassis setup and also external signal like "nidq". 
     * For imec devices both "ap" and "lf" are extracted so even a one device setup will have several "streams"
 
     Examples
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -82,11 +82,11 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
 
     Notes
     -----
-    * Contrary to other implementations this IO reads the entire folder and subfolders and:
-      deals with several segments based on the `_gt0`, `_gt1`, `_gt2`, etc postfixes
-      deals with all signals coming from different acquisition cards ("imec0", "imec1", etc) in a typical 
-      PXIe chassis and also external signal like"nidq". This is the "device"
-    * For imec device both "ap" and "lf" are extracted so even a one device setup has several "streams"
+    * This IO reads the entire folder and subfolders locating the `.bin` and `.meta` files
+    * Handles gates and triggers as segments (based on the `_gt0`, `_gt1`, `_t0` , `_t1` in filenames)
+    * Handles all signals coming from different acquisition cards ("imec0", "imec1", etc) in a typical
+        PXIe chassis setup and also external signal like"nidq". 
+    * For imec devices both "ap" and "lf" are extracted so even a one device setup will have several "streams"
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


neo/rawio/spikeglxrawio.py
```diff
@@ -82,14 +82,11 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
...
-      deals with all signals "imec0", "imec1" for neuropixel probes and also
-      external signal like"nidq". This is the "device"
-    * For imec device both "ap" and "lf" are extracted so one device have several "streams"
-    * There are several versions depending the neuropixel probe generation (`1.x`/`2.x`/`3.x`)
-    * Here, we assume that the `meta` file has the same structure across all generations.
-    * This IO is developed based on neuropixel generation 2.0, single shank recordings.
+    * This IO reads the entire folder and subfolders locating the `.bin` and `.meta` files
+    * Handles gates and triggers as segments (based on the `_gt0`, `_gt1`, `_t0` , `_t1` in filenames)
+    * Handles all signals coming from different acquisition cards ("imec0", "imec1", etc) in a typical
+        PXIe chassis setup and also external signal like"nidq".
```
alejoe91, line 88:
```suggestion
        PXIe chassis setup and also external signal like "nidq". 
```

h-mayorquin:
OK, I added the tests and this should be ready for review.

neo/rawio/spikeglxrawio.py
h-mayorquin: Thanks, man.

neo/rawio/spikeglxrawio.py
```diff
@@ -82,14 +82,11 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
 
     Notes
     -----
-    * Contrary to other implementations this IO reads the entire folder and subfolders and:
```
h-mayorquin, line 85: Changed this because it is not clear what other implementations this is referring to, probably the ones before but then I don't think it should be here.

neo/rawio/spikeglxrawio.py
```diff
@@ -82,14 +82,11 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
...
-      deals with several segments based on the `_gt0`, `_gt1`, `_gt2`, etc postfixes
-      deals with all signals "imec0", "imec1" for neuropixel probes and also
-      external signal like"nidq". This is the "device"
-    * For imec device both "ap" and "lf" are extracted so one device have several "streams"
-    * There are several versions depending the neuropixel probe generation (`1.x`/`2.x`/`3.x`)
-    * Here, we assume that the `meta` file has the same structure across all generations.
-    * This IO is developed based on neuropixel generation 2.0, single shank recordings.
```
h-mayorquin, line 92: I think this is a note of the first version but we have gradually added support. This supports the shanked version just fine and also NHP and other varieties.

neo/rawio/spikeglxrawio.py
```diff
@@ -82,14 +82,11 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
...
+    * This IO reads the entire folder and subfolders locating the `.bin` and `.meta` files
+    * Handles gates and triggers as segments (based on the `_gt0`, `_gt1`, `_t0` , `_t1` in filenames)
```
h-mayorquin, line 86: Added specific mention to gates and triggers as that is more in line with SpikeGLX documentation.

neo/rawio/spikeglxrawio.py
```diff
@@ -354,23 +348,54 @@ def scan_files(dirname):
...
-        if k0 is None:
-            k0 = -1
-        k1 = info["trigger_num"]
-        if k1 is None:
-            k1 = -1
-        return (k0, k1)
-
-    order_key = list({make_key(info) for info in info_list})
-    order_key = sorted(order_key)
+    # This sets non-integers values before integers
```
h-mayorquin, line 351: So this section is the core of the PR. It uses the same mechanism that we have for gates and triggers before but also for probe index.:

* It creates a set of unique combinations of (gate, trigger) and (probe, port,dock).
* That combination is used to create the index (segment index and probe index)
* The parameters of info are used to assign the segment and probe index with a dict.

The normalize lambda unifies this behavior. Maybe there is a better naming but I could not come with anything, it just follows the convention that Sam had before.

neo/rawio/spikeglxrawio.py
```diff
@@ -354,23 +348,54 @@ def scan_files(dirname):
...
+    probe_tuple_to_probe_index = {key: idx for idx, key in enumerate(sorted_probe_keys)}
+
+    for info in info_list:
+        if info.get("device") == "nidq":
+            info["device_index"] = ""  # TODO: Handle multi nidq case, maybe use meta["typeNiEnabled"]
+        else:
+            info["device_index"] = probe_tuple_to_probe_index[get_probe_tuple(info)]
 
+    # Define stream base on device [imec|nidq], device_index and stream_kind [ap|lf] for imec
+    for info in info_list:
```
h-mayorquin, line 391: This defines the stream here after all the information is available. Previously, it was done inside the `extract_info_function`.

neo/rawio/spikeglxrawio.py
```diff
@@ -488,13 +513,15 @@ def extract_stream_info(meta_file, meta):
     else:
         # NIDQ case
         has_sync_trace = False
-    fname = Path(meta_file).stem
+
+    bin_file_path = meta["fileName"]
```
h-mayorquin, line 517: This is the part that uses the meta["fileName"] (the original one) instead of the file_path in the system (which the user might have changed).

neo/rawio/spikeglxrawio.py
```diff
@@ -563,12 +593,15 @@ def extract_stream_info(meta_file, meta):
     info["trigger_num"] = trigger_num
     info["device"] = device
     info["stream_kind"] = stream_kind
-    info["stream_name"] = stream_name
+    info["device_kind"] = meta.get("typeThis", device.split(".")[0])
```
h-mayorquin, line 596: Get the device kind directly from the metadata. imec or nidq. All non-production probes should have this in their metadata. For production probes this would rely on file parsing as we were doing before:

https://billkarsh.github.io/SpikeGLX/help/parsing/

h-mayorquin (commented):
Added comments to hopefully make the reviewing easier.

samuelgarcia:
good investigatio in the meander of the spikeglx meta file.
I need to read to read it more carfully but this looks ok

alejoe91 (approved):

alejoe91:
This looks good to me! 

Thank you @h-mayorquin for diving into the SpikeGLX meta world!

@samuelgarcia good to merge on my end, let me know if you want to have another go

alejoe91:
Should also fix #1511

zm711:
Just updating branch because I made a bigger CI fix (we weren't correctly testing on 3.12 each time. So just want us to run at least one round of CI will the correct action :))

h-mayorquin:
Can I please get this into the minor release if it is working well? : ) 
@samuelgarcia @alejoe91 @zm711

zm711 (approved):

PR #1609, opened 2024-12-10
Add timing metadata for `SpikeGLXRawIO`

### LINKED ISSUE #1607: Multi-gate and multi-trigger  `t_starts` in SpikeGLX are incorrectly set to 0
As discussed with @samuelgarcia the starting time of the segments in SpikeGLX are set to 0 here:

https://github.com/NeuralEnsemble/python-neo/blob/0ae6e76d2e2c2d9f9a7b42a23b04f393ee8a11cd/neo/rawio/spikeglxrawio.py#L231-L238

But in the metadata there is a `firstFrame` value:

https://billkarsh.github.io/SpikeGLX/Sgl_help/Metadata_30.html

```
firstSample=779283
All acquired samples are assigned an index number; the first acquired during the run is sample 0. This value is the index number of the first sample recorded in this file.
```

This is important so the correct times can be propagated to downstream applications.

We have data on gin that has this metadata so I can submit a patch.

h-mayorquin:
Should close #1607

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -229,11 +229,32 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
         spike_channels = np.array(spike_channels, dtype=_spike_channel_dtype)
 
         # deal with nb_segment and t_start/t_stop per segment
-        self._t_starts = {seg_index: 0.0 for seg_index in range(nb_segment)}
+
+        self._t_starts = {stream_name: {} for stream_name in stream_names}
         self._t_stops = {seg_index: 0.0 for seg_index in range(nb_segment)}
+        self._t_since_recording_time = {stream_name: {} for stream_name in stream_names}
+        self._session_start_time = {stream_name: {} for stream_name in stream_names}
+        self._gate_trigger = {stream_name: {} for stream_name in stream_names}
         for seg_index in range(nb_segment):
             for stream_name in stream_names:
                 info = self.signals_info_dict[seg_index, stream_name]
+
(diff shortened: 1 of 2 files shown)


h-mayorquin:
I had a discussion with Graham Findlay from the Tononi and Cirelli lab and he suggested to make the smallest t_start the reference point (hence, it would be 0). That seemed like a good idea to me but I wanted to check with you guys @alejoe91 @samuelgarcia.

samuelgarcia:
Thanks for this. This is really cool. We should have done this from the beginings.
Maybe we could have an option to force the previous behavior (t_start=0 everywhere) in case some people handle event with this in mind for bawkard compatibility.

forcing the very first t_start to be 0 is convinient sometimes (I did for for some readers).
For long term data conversion the original t_start is better ?
Maybe we could have a secondary option for this ...

alejoe91:
Thanks for updating this @h-mayorquin 

I'm not sure we need the boolean option to force t start at 0...as we're making a new minor release, a small API change as this one is ok. Let's document it properly in the release notes @zm711

samuelgarcia:
> Introduce a boolean flag that allows users to enable the old behavior. However, this flag should also trigger a deprecation warning, informing users that the feature will be removed in the next major release.

Lets forget about it, this useless and painfull work.

> I think this purpose would be better served by having a general function or boolean option at `init` that can be used for any reader. This will be easier to maintain, more ergonomic for the user and I think more useful. What do you think?

Yes. this should be a hack in the main class but then we should not reset the first t_start to 0 then.

zm711 (approved):

PR #1612, opened 2024-12-13
Fix another Plexon overflow

zm711:
@alejoe91 I found another plexon overflow. could you merge this one. This is the problem with the documentation. 

cc @h-mayorquin (this will fix the PR. I found another overflow).


zm711:
Okay so the problem was casting behavior with numpy 2.0. So for docs I limited to < 2.0 because we depend on IOs and we have not updated the IOS yet. This is ready to go when tests pass @alejoe91. Heberto and I did this together to figure out it was a dependency issue. And this reminds me that each individual IO will need additional confirmation to be Numpy 2.0 ready.

alejoe91 (approved):

PR #1613, opened 2024-12-13
Fix overflow of Plexon in numpy 2.0

h-mayorquin:
This should fix the errors in https://github.com/NeuralEnsemble/python-neo/pull/1612.

At least for Plexon.

### DIFF
--- a/neo/rawio/plexonrawio.py
+++ b/neo/rawio/plexonrawio.py
@@ -368,11 +368,11 @@ class PlexonRawIO(BaseRawIO):
             channel_index = np.nonzero(dspChannelHeaders["Channel"] == chan_id)[0][0]
             dsp_channel_headers = dspChannelHeaders[channel_index]
 
-            name = h["Name"].decode("utf8")
+            name = dsp_channel_headers["Name"].decode("utf8")
             _id = f"ch{chan_id}#{unit_id}"
             wf_units = ""
             if global_header["Version"] < 103:
-                wf_gain = 3000.0 / (2048 * h["Gain"] * 1000.0)
+                wf_gain = 3000.0 / (2048 * dsp_channel_headers["Gain"] * 1000.0)
             elif 103 <= global_header["Version"] < 105:
                 wf_gain = global_header["SpikeMaxMagnitudeMV"] / (
                     0.5 * 2.0 ** (global_header["BitsPerSpikeSample"]) * dsp_channel_headers["Gain"] * 1000.0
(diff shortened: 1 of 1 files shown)


h-mayorquin:
@zm711 Ok, this fixes Plexon error with the current build on the docs but then the error goes to some other io.

I think this can be merged after #1612 and then BlackRock can be fixed separately.

zm711:
@h-mayorquin merge conflict from #1612.

h-mayorquin:
Solved.

zm711:
Cool. I'll double check this weekend :)

neo/rawio/plexonrawio.py
```diff
@@ -209,6 +209,10 @@ def _parse_header(self):
                 for index, pos in enumerate(positions):
                     bl_header = data[pos : pos + 16].view(DataBlockHeader)[0]
 
+                    # To avoid overflow errors when doing arithmetic operations on numpy scalars
+                    np_scalar_to_python_scalar = lambda x: x.item() if isinstance(x, np.generic) else x
```
zm711, line 213: Is there a reason not to make this lambda function near the top so it is reusable. Or even define this as a true private function. How are you thinking about using the same lambda function twice being generated in the for loop. It should be relatively low cost although seems a bit of a waste no? We could define the function once and reuse in both the for loops, no?

zm711 (commented):
One strategy question @h-mayorquin :)

neo/rawio/plexonrawio.py
h-mayorquin: Keeping the definition close to the use so people can quickly see what it is for. This takes nanoseconds:

image

Extracting the data from plexon takes minutes. It does not factor.

neo/rawio/plexonrawio.py
```diff
@@ -255,24 +259,29 @@ def _parse_header(self):
...
-            name = h["Name"].decode("utf8")
-            chan_id = h["Channel"]
+            channel_headers = slowChannelHeaders[chan_index]
+            
+            # To avoid overflow errors when doing arithmetic operations on numpy scalars
+            np_scalar_to_python_scalar = lambda x: x.item() if isinstance(x, np.generic) else x
+            channel_headers = {key: np_scalar_to_python_scalar(channel_headers[key]) for key in channel_headers.dtype.names}
+            
+            name = channel_headers["Name"].decode("utf8")
+            chan_id = channel_headers["Channel"]
```
zm711, line 269: Last question.

so below you called it dsp_channel_headers which was really nice. Any interest in changing this to `slow_channel_headers`, just so we make sure we never accidentally overwrite any variable. I think the explicit makes it cleaner and since we are already fixing up the variable naming might be worth the push?

neo/rawio/plexonrawio.py
zm711: I figured you'd have checked. Fair enough. :)

zm711 (approved):

samuelgarcia:
merci

PR #1614, opened 2024-12-13
NumPy 2.0 support for IOs

zm711:
We also need to start seeing which IOs are potentially not 2.0 ready. This is just a draft to start assessing this.

This is much better than I was expecting :) 

```
FAILED neo/test/rawiotest/test_blackrockrawio.py::TestBlackrockRawIO::test_compare_blackrockio_with_matlabloader - KeyError: np.int64(-1)
FAILED neo/test/rawiotest/test_blackrockrawio.py::TestBlackrockRawIO::test_compare_blackrockio_with_matlabloader_v21 - OverflowError: Python integer -1 out of bounds for uint16
FAILED neo/test/rawiotest/test_blackrockrawio.py::TestBlackrockRawIO::test_read_all - KeyError: np.int64(-1)
FAILED neo/test/rawiotest/test_plexonrawio.py::TestPlexonRawIO::test_read_all - OverflowError: Python integer 4294967296 out of bounds for uint16
```

### DIFF
--- a/.github/workflows/io-test.yml
+++ b/.github/workflows/io-test.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.12']
+        python-version: ['3.9', '3.12', '3.13']
     defaults:
       # by default run in bash mode (required for conda usage)
       run:
--- a/environment_testing.yml
+++ b/environment_testing.yml
@@ -4,6 +4,3 @@ channels:
 dependencies:
   - datalad
   - pip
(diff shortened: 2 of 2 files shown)


zm711:
@samuelgarcia you can see the errors here for blackrock :)

zm711:
I almost forgot about the med format. The Dark Horse Devs said they would help work on it when ready see issue [here](https://github.com/MEDFormat/MEDPython/issues/7). Do we want to ping them for NumPy 2.0 or do we want to just have a branch to test if we want to fix it ourselves. I think they are doing some more major API changes. So maybe we actually want to have a separate install option for users? Not sure.

Issue #1618, opened 2025-01-06
CI Failures--Wrong Python Version + NWB failures


zm711:
**Describe the bug**
Our python 3.12 test is currently running on 3.9. I need to try to fix this. I'll have to look into this today or tomorrow.

The main issue is the following:

```
self = <neo.test.iotest.test_nwbio.TestNWBIO testMethod=test_roundtrip>

    def test_roundtrip(self):
    
        annotations = {"session_start_time": datetime.now()}
        # Define Neo blocks
        bl0 = Block(name="First block", **annotations)
        bl1 = Block(name="Second block", **annotations)
        bl2 = Block(name="Third block", **annotations)
        original_blocks = [bl0, bl1, bl2]
    
        num_seg = 4  # number of segments
        num_chan = 3  # number of channels
    
        for blk in original_blocks:
    
            for ind in range(num_seg):  # number of Segments
                seg = Segment(index=ind)
                blk.segments.append(seg)
    
            for seg in blk.segments:  # AnalogSignal objects
    
                # 3 Neo AnalogSignals
                a = AnalogSignal(
                    name="Signal_a %s" % (seg.name),
                    signal=np.random.randn(44, num_chan) * pq.nA,
                    sampling_rate=10 * pq.kHz,
                    t_start=50 * pq.ms,
                )
                b = AnalogSignal(
                    name="Signal_b %s" % (seg.name),
                    signal=np.random.randn(64, num_chan) * pq.mV,
                    sampling_rate=8 * pq.kHz,
                    t_start=40 * pq.ms,
                )
                c = AnalogSignal(
                    name="Signal_c %s" % (seg.name),
                    signal=np.random.randn(33, num_chan) * pq.uA,
                    sampling_rate=10 * pq.kHz,
                    t_start=120 * pq.ms,
                )
                # 2 Neo IrregularlySampledSignals
                d = IrregularlySampledSignal(np.arange(7.0) * pq.ms, np.random.randn(7, num_chan) * pq.mV)
    
                # 2 Neo SpikeTrains
                train = SpikeTrain(times=[1, 2, 3] * pq.s, t_start=1.0, t_stop=10.0)
                train2 = SpikeTrain(times=[4, 5, 6] * pq.s, t_stop=10.0)
                # todo: add waveforms
    
                # 1 Neo Event
                evt = Event(name="Event", times=np.arange(0, 30, 10) * pq.ms, labels=np.array(["ev0", "ev1", "ev2"]))
    
[...]

PR #1619, opened 2025-01-06
Fix NWB parsing from pynwb/hmdf

zm711:
## Error
```
self = <[AttributeError("'TimeSeries' object has no attribute '_AbstractContainer__name'") raised in repr()] TimeSeries object at 0x7f032efad6a0>
kwargs = {'name': 'First block : segment0 Signal_a None 0'}
name = 'First block : segment0 Signal_a None 0'

    @docval({'name': 'name', 'type': str, 'doc': 'the name of this container'})
    def __init__(self, **kwargs):
        name = getargs('name', kwargs)
        if ('/' in name or ':' in name) and not self._in_construct_mode:
>           raise ValueError(f"name '{name}' cannot contain a '/' or ':'")
E           ValueError: name 'First block : segment0 Signal_a None 0' cannot contain a '/' or ':'
```
from here
```
/usr/share/miniconda/envs/neo-test-env/lib/python3.9/site-packages/hdmf/utils.py:668: in func_call
    return func(args[0], **pargs)
```
## Fix

Fixes part of #1618. It says that hmdf no longer accepts symbol `:` in text. In this case we just remove the offending :'s. This should be fine for old version too. Not quite as clean to read, but seems okay to me.

### DIFF
--- a/neo/io/nwbio.py
+++ b/neo/io/nwbio.py
@@ -495,7 +495,7 @@ class NWBIO(BaseIO):
             if segment.block is not block:
                 raise TypeError(f"segment.block must be block it is {segment.block}")
             if not segment.name:
-                segment.name = f"{block.name} : segment{i}"
+                segment.name = f"{block.name}  segment{i}"
             self._write_segment(self._nwbfile, segment, electrodes)
         self.blocks_written += 1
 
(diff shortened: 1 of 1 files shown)


zm711:
@apdavison, This should fix nwb parsing and is needed to fix the CI. I'll tag @samuelgarcia since he is so active in case he wants to take a look too, but this is a super simple PR. :)

apdavison:
ok for now. I'd prefer to use a different separator, maybe "-" or ";", but I'll revisit that when I overhaul annotation handling in NWBIO.

zm711:
That's absolutely fair. At the time I was worried they could update to forbid any symbol, but I think they banned : and \ for path processing issues so I would lean more toward "-" for a future PR.

PR #1624, opened 2025-01-16
`OpenEphysBinaryRawIO`: Separate Neural and Non-Neural Data into Distinct Streams

h-mayorquin:
The continuous.dat file in the Open Ephys Binary format often contains both neural and non-neural (ADC) channels. For detailed documentation, refer to the [Open Ephys GUI User Manual](https://open-ephys.github.io/gui-docs/User-Manual/Recording-data/Binary-format.html#continuous).

An example dataset (provided by @rly) illustrating this structure is available [here](https://github.com/calderast/jdb_to_nwb/tree/main/tests/test_data/raw_ephys/2022-07-25_15-30-00). Additionally, a related discussion can be found in the NeuroConv repository: [Issue #1023](https://github.com/catalystneuro/neuroconv/issues/1023).

This pull request modifies the reader to separate non-neural channels into their own stream. This approach allows users to clearly distinguish between neural and non-neural data and will enable programmatic separation in downstream libraries like SpikeInterface and Neuroconv.

**Implementation Details:**

To maintain backward-compatibility of stream_ids, the non-neural data's stream_id is defined as the neural data's stream_id plus the total number of available streams. For example, if the neural data has a stream_id of "1" and there are three streams in total, the non-neural data will be assigned a stream_id of "4".

The stream_name for non-neural data is derived by appending _ADC to the neural data's stream_name. For instance, if the neural data's stream_name is Data, the non-neural data's stream_name will be DATA_ADC.

I welcome better suggestions for conventions of naming and any other type of feedback from people who are more familiar with this format @alejoe91

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -121,8 +121,9 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
         # signals zone
         # create signals channel map: several channel per stream
         signal_channels = []
+        
         for stream_index, stream_name in enumerate(sig_stream_names):
-            # stream_index is the index in vector sytream names
+            # stream_index is the index in vector stream names
             stream_id = str(stream_index)
             buffer_id = stream_id
             info = self._sig_streams[0][0][stream_index]
@@ -132,11 +133,23 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                 if "SYNC" in chan_id and not self.load_sync_channel:
                     # the channel is removed from stream but not the buffer
                     stream_id = ""
+
                 if chan_info["units"] == "":
                     # in some cases for some OE version the unit is "", but the gain is to "uV"
                     units = "uV"
                 else:
                     units = chan_info["units"]
+                
+                if "ADC" in chan_id:
+                    # These are non-neural channels and their stream should be separated
+                    # We defined their stream_id as the stream_index of neural data plus the number of neural streams
+                    # This is to not break backwards compatbility with the stream_id numbering
+                    stream_id = str(stream_index + len(sig_stream_names))
+                    # For ADC channels multiplying by the bit_volts whe un units are not provided converts to Volts
+                    units = "V" if units == "" else units
+                    
+                gain = chan_info["bit_volts"]
+                offset = 0.0
                 new_channels.append(
                     (
                         chan_info["channel_name"],
@@ -144,24 +157,41 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                         float(info["sample_rate"]),
                         info["dtype"],
                         units,
-                        chan_info["bit_volts"],
-                        0.0,
+                        gain,
+                        offset,
                         stream_id,
                         buffer_id,
                     )
                 )
             signal_channels.extend(new_channels)
+        
         signal_channels = np.array(signal_channels, dtype=_signal_channel_dtype)
 
         signal_streams = []
         signal_buffers = []
-        for stream_index, stream_name in enumerate(sig_stream_names):
-            stream_id = str(stream_index)
-            buffer_id = str(stream_index)
-            signal_buffers.append((stream_name, buffer_id))
+        
+        unique_streams_ids = np.unique(signal_channels["stream_id"])
+        for stream_id in unique_streams_ids:
+            # Handle special case of Synch channel having stream_id empty
+            if stream_id == "":
+                continue
+            stream_index = int(stream_id)
+            # Neural signal
+            if stream_index < len(sig_stream_names):
+                stream_name = sig_stream_names[stream_index]
+                buffer_id = stream_id
+                # We add the buffers here as both the neural and the ADC channels are in the same buffer
+                signal_buffers.append((stream_name, buffer_id))
+            else: # This names the ADC streams
+                neural_stream_index = stream_index - len(sig_stream_names)
+                neural_stream_name = sig_stream_names[neural_stream_index]
+                stream_name = f"{neural_stream_name}_ADC"
+                buffer_id = str(neural_stream_index)
             signal_streams.append((stream_name, stream_id, buffer_id))
+        
         signal_streams = np.array(signal_streams, dtype=_signal_stream_dtype)
         signal_buffers = np.array(signal_buffers, dtype=_signal_buffer_dtype)
+
 
         # create memmap for signals
         self._buffer_descriptions = {}
@@ -431,6 +461,8 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
         return group_id
 
     def _get_signal_t_start(self, block_index, seg_index, stream_index):
+        if stream_index >= len(self._sig_streams[block_index][seg_index]):
+            stream_index = stream_index - len(self._sig_streams[block_index][seg_index])    
         t_start = self._sig_streams[block_index][seg_index][stream_index]["t_start"]
         return t_start
 


h-mayorquin:
Testing data is here:

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/150

Once the data is merged I can add a test

alejoe91:
@h-mayorquin can you fix Ryan's last name here? https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/150#issuecomment-5561

Se we can merge and add a test here

neo/rawio/openephysbinaryrawio.py
```diff
@@ -431,6 +461,8 @@ def _channels_to_group_id(self, channel_indexes):
         return group_id
 
     def _get_signal_t_start(self, block_index, seg_index, stream_index):
+        if stream_index >= len(self._sig_streams[block_index][seg_index]):
+            stream_index = stream_index - len(self._sig_streams[block_index][seg_index])    
         t_start = self._sig_streams[block_index][seg_index][stream_index]["t_start"]
```
samuelgarcia, line 466: I understand correctly the ptach `stream_index` is not anymore a stream_index but an internal variable which is not a direct mapping to stream_index.
`self._sig_streams[block_index][seg_index][stream_index]["t_start"]`

If this is the case I would change this internal variable to something else to avoid mistale when reading the entire code.

Maybe I am mistunderstanding or not clear at all but this writting is a bit weird to me.

h-mayorquin: You are correct. This is a good idea.

h-mayorquin: Changed.

neo/rawio/openephysbinaryrawio.py
```diff
@@ -132,37 +133,66 @@ def _parse_header(self):
...
                     units = "uV"
                 else:
                     units = chan_info["units"]
+                
+                if "ADC" in chan_id:
+                    # These are non-neural channels and their stream should be separated
+                    # We defined their stream_id as the stream_index of neural data plus the number of neural streams
+                    # This is to not break backwards compatbility with the stream_id numbering
+                    stream_id = str(stream_index + len(sig_stream_names))
+                    # For ADC channels multiplying by the bit_volts whe un units are not provided converts to Volts
```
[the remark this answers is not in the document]
h-mayorquin: Thanks.

h-mayorquin:
I corrected Ryan's last name (my bad) on the test data repo.

Question: why do we have stream_id of the synch channel as the empty string? It is loaded with the neural data of the continuous event with a flag, shouldn't they belong to the same stream?

alejoe91:
Gin is merged!

alejoe91:
@h-mayorquin so just missing the test with the new test file and this is ready for final review, right?

h-mayorquin:
This is ready for final review now.

neo/rawio/openephysbinaryrawio.py
```diff
@@ -191,12 +223,30 @@ def _parse_header(self):
...
+                        else:
+                            self._stream_buffer_slice[stream_id] = None
                     else:
-                        self._stream_buffer_slice[stream_id] = None
-
+                        # For ADC channels, we remove the last channel which is the Synch channel
+                        stream_id_neural = stream_id
+                        stream_id_non_neural = str(int(stream_id) + self._num_of_signal_streams)
+                        
+                        # Note this implementation assumes that the neural channels come before the non-neural channels
```
h-mayorquin, line 241: Here guys, these are the strongest assumptions that the reader makes. I wanted to check with @alejoe91 if he knows about this:
* Is it true that the synch trace will always be the last channel even if you have non-neural channels? I could not find this in the documentation.
* Are the neural channels and non-neural always in order and not interleaved? 
 
 If the first case is not True, we can extract the specific index from the structure.oebin. That's a bigger change but I think doable.
 
 For the second case, if it is not true, I am afraid we will have to error somehow as the current data model does not support non-contigious streams across the buffer with the buffer slice mechanism (@samuelgarcia )

h-mayorquin: I am adding error messages for this advising people to open an issue if they have interleaved neural and non-neural or synch trace that is not at the end.

alejoe91:
@h-mayorquin some tests are failing. This is because the array annotations have changed now, since also they need to be split across streams

h-mayorquin:
> @h-mayorquin some tests are failing. This is because the array annotations have changed now, since also they need to be split across streams

I fixed them now. The array annotations were not divided for the stream.

neo/rawio/openephysbinaryrawio.py
```diff
@@ -132,34 +135,62 @@ def _parse_header(self):
...
+
+                if "ADC" in chan_id:
+                    # These are non-neural channels and their stream should be separated
+                    # We defined their stream_id as the stream_index of neural data plus the number of neural streams
+                    # This is to not break backwards compatbility with the stream_id numbering
+                    stream_id = str(stream_index + len(sig_stream_names))
+                    # For ADC channels multiplying by the bit_volts when units are not provided converts to Volts
+                    units = "V" if units == "" else units
+
+                gain = chan_info["bit_volts"]
```
zm711, line 153: is this a safe float? or do we need to convert it?

neo/rawio/openephysbinaryrawio.py
```diff
@@ -192,10 +223,43 @@ def _parse_header(self):
                             "SYNC channel is not present in the recording. " "Set load_sync_channel to False"
                         )
 
-                    if has_sync_trace and not self.load_sync_channel:
-                        self._stream_buffer_slice[stream_id] = slice(None, -1)
+                    # Check if ADC and non-ADC channels are contiguous
+                    is_adc = ["ADC" in ch["channel_name"] for ch in info["channels"]]
+                    first_adc = is_adc.index(True) if True in is_adc else len(is_adc)
+                    if any(not x for x in is_adc[first_adc:]):
```
zm711, line 229: slight request for a better variable name than x. Not really blocking, but I know you dislike when d is used to mean dict. :)

neo/rawio/openephysbinaryrawio.py
```diff
@@ -121,8 +123,9 @@ def _parse_header(self):
         # signals zone
         # create signals channel map: several channel per stream
         signal_channels = []
+
         for stream_index, stream_name in enumerate(sig_stream_names):
-            # stream_index is the index in vector sytream names
+            # stream_index is the index in vector stream names
```
zm711, line 128: Nice :)

neo/rawio/openephysbinaryrawio.py
```diff
@@ -132,34 +135,62 @@ def _parse_header(self):
...
                     units = "uV"
                 else:
                     units = chan_info["units"]
+
+                if "ADC" in chan_id:
+                    # These are non-neural channels and their stream should be separated
+                    # We defined their stream_id as the stream_index of neural data plus the number of neural streams
+                    # This is to not break backwards compatbility with the stream_id numbering
+                    stream_id = str(stream_index + len(sig_stream_names))
+                    # For ADC channels multiplying by the bit_volts when units are not provided converts to Volts
```
zm711, line 150: What do you mean here? 

the way it is written sounds like if units are provided then we SHOULD NOT worry about using the gain,? So is the gain 1 in this case and our code is fine?

neo/rawio/openephysbinaryrawio.py
```diff
@@ -192,10 +223,43 @@ def _parse_header(self):
...
+                        )
+
+                    # Find sync channel and verify it's the last channel
+                    sync_index = next(
+                        (index for index, ch in enumerate(info["channels"]) if ch["channel_name"].endswith("_SYNC")),
+                        None,
+                    )
+                    if sync_index is not None and sync_index != num_channels - 1:
+                        raise ValueError(
+                            "SYNC channel must be the last channel in the buffer. Open an issue in python-neo to request this feature."
```
zm711, line 242: we will need sample data to implement this? 

maybe adding something like this to the comment? Someone might request but prewarning them that we will only do it with data might be helpful.

zm711 (commented):
Just a few comments. I know Alessio and Sam provide better feedback for openephys so take what you like and ignore what you don't :)

neo/rawio/openephysbinaryrawio.py
h-mayorquin: What do you mean safe? as in numpy?

neo/rawio/openephysbinaryrawio.py
h-mayorquin: Fair point. Thanks for not letting me get away with laziness.

neo/rawio/openephysbinaryrawio.py
h-mayorquin: Yes, that's not very clear without context. More context: When the units are not provided the units of ADC are Volts. This is in opposition to the neural channels where if the units are not provided the units are microvolts.

https://open-ephys.github.io/gui-docs/User-Manual/Recording-data/Binary-format.html#continuous

But the units are always the units of the data AFTER you use the gain. Only the unitless(`units==""`) case differs between neural and non-neural channels. Microvolts for the former, volts for the latter.

neo/rawio/openephysbinaryrawio.py
h-mayorquin: I think it is fine to request data on the issue. I don't want to create higher barriers for someone reporting to us that the case indeed exists.

neo/rawio/openephysbinaryrawio.py
zm711: Yes. I mean is it a numpy scalar (I should have written that better :) ) or is it a python float?

neo/rawio/openephysbinaryrawio.py
zm711: Okay, yeah that makes sense in the context (and is the same with intan so ADC being volts and neural being microvolts seems common enough.

neo/rawio/openephysbinaryrawio.py
zm711: Fair enough!

neo/rawio/openephysbinaryrawio.py
h-mayorquin: For floats, there is no difference I think right? We only have to be concerned about integers. I think np.float is the same thing as float in python (from memory).

neo/rawio/openephysbinaryrawio.py
h-mayorquin: Re-wrote the whole thing to separate unit determination from stream determination and I think it reads better now.

neo/rawio/openephysbinaryrawio.py
zm711: So I looked it up mostly from stackoverflow discussion
https://stackoverflow.com/questions/40726490/overflow-error-in-pythons-numpy-exp-function
and reddit
https://www.reddit.com/r/learnpython/comments/feakn2/numpy_overflow_help_needed/?rdt=48979
And seems like it can overflow, but honestly the issue of float overflow seems to be more benign and casting to python seems to cost precision that honestly shouldn't really matter right? But float precision is so OS dependent that I think requiring that level of precision can't really be expected from floats. Appreciate the casting though, but in general your intuition/memory of this being an int thing seems to be true.

neo/rawio/openephysbinaryrawio.py
zm711: Thanks!

neo/rawio/openephysbinaryrawio.py
zm711: Yeah I agree it makes it clearer to me at least!

zm711 (commented):
This is good for me, but I'll leave this to @alejoe91 and @samuelgarcia .

neo/rawio/openephysbinaryrawio.py
```diff
@@ -192,10 +224,44 @@ def _parse_header(self):
...
-                        self._stream_buffer_slice[stream_id] = slice(None, -1)
+                    # Check if ADC and non-ADC channels are contiguous
+                    is_channel_adc = ["ADC" in ch["channel_name"] for ch in info["channels"]]
+                    first_adc_index = is_channel_adc.index(True) if any(is_channel_adc) else len(is_channel_adc)
+                    non_adc_channels_after_adc_channels = [not flag for flag in is_channel_adc[first_adc_index:]]
+                    if any(non_adc_channels_after_adc_channels):
+                        raise ValueError(
+                            "Interleaved ADC and non-ADC channels are not supported. "
+                            "ADC channels must be contiguous. Open an issue in python-neo to request this feature."
+                        )
```
alejoe91, line 235: I think this would be easier to read:
```
                if any(is_channel_adc):
                    first_adc_index = is_channel_adc.index(True)
                    non_adc_channels_after_adc_channels = [not is_adc for is_adc in is_channel_adc[first_adc_index:]]
                    if any(non_adc_channels_after_adc_channels):
                        raise ValueError(
                            "Interleaved ADC and non-ADC channels are not supported. "
                            "ADC channels must be contiguous. Open an issue in python-neo to request this feature."
                        )
```
I found it a bit hard to understand the logic if no ADC channels are present (then `non_adc_channels_after_adc_channels` would be an empty list)

neo/rawio/openephysbinaryrawio.py
```diff
@@ -192,10 +224,44 @@ def _parse_header(self):
...
+                        (index for index, ch in enumerate(info["channels"]) if ch["channel_name"].endswith("_SYNC")),
+                        None,
+                    )
+                    if sync_index is not None and sync_index != num_channels - 1:
+                        raise ValueError(
+                            "SYNC channel must be the last channel in the buffer. Open an issue in python-neo to request this feature."
+                        )
+
+                    num_neural_channels = sum(1 for ch_info in info["channels"] if "ADC" not in ch_info["channel_name"])
+                    num_non_neural_channels = sum(1 for ch_info in info["channels"] if "ADC" in ch_info["channel_name"])
```
alejoe91, line 248:
```suggestion
                    num_neural_channels = sum("ADC" not in ch_info["channel_name"] for ch_info in info["channels"])
                    num_non_neural_channels = sum("ADC" in ch_info["channel_name"] for ch_info in info["channels"])
```
easier to read?

neo/rawio/openephysbinaryrawio.py
```diff
@@ -375,17 +441,33 @@ def _parse_header(self):
...
 
-                    for k in ("identifier", "history", "source_processor_index", "recorded_processor_index"):
-                        if k in info["channels"][0]:
-                            values = np.array([chan_info[k] for chan_info in info["channels"]])
                             if has_sync_trace:
                                 values = values[:-1]
-                            sig_ann["__array_annotations__"][k] = values
+
+                            num_neural_channels = sum(
+                                1 for ch_info in info["channels"] if "ADC" not in ch_info["channel_name"]
```
alejoe91, line 463:
```suggestion
                                "ADC" not in ch_info["channel_name"] for ch_info in info["channels"]
```

alejoe91 (changes requested):
Thanks @h-mayorquin 

I just had a couple of suggestions to clarify some logic. Let me know what you think!

neo/rawio/openephysbinaryrawio.py
h-mayorquin: Yes, this is better.

neo/rawio/openephysbinaryrawio.py
h-mayorquin: It is easier to read than mine. My fault. But now that I think about it, it would be easier to get a list and then use length.

h-mayorquin: I just defined first the list and then used len().

Trying to reduce number of lines is a foolish endeavor. Un gusto culposo.

alejoe91 (approved):

zm711 (approved):
merging

Issue #1628, opened 2025-01-17
Clean-up CI files


zm711:
With #1627, we piloted not using caching anymore. Especially with the speed of creating these envs now (currently the slowest step is installing wine, which takes ~ 1min).  If we are fine with not caching our conda env after a maintenance cycle then it might be worth doing the following:

Remove environment_testing.yml--> this file only has datalad and pip with the conda-forge channel in it. Only nice for updating a cache which we no longer have.

Clean up the io-testing.yml where currently I've only commented out the cache code. I think it looks messy so maybe we want to delete the commented out code ?

PR #1633, opened 2025-01-24
Stop Cron Job for Creating Conda Env

zm711:
Just checked our actions and realized we were creating a cache already. To be environmentally friendly let's not do this cron job while we decide whether to permanently remove the conda env cache.

for @samuelgarcia, but @alejoe91 since you'll see this first (most likely) feel free to make sure I didn't mess it up.

We should keep this in mind for #1628.

### DIFF
--- a/.github/workflows/caches_cron_job.yml
+++ b/.github/workflows/caches_cron_job.yml
@@ -10,57 +10,57 @@ on:
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


zm711:
@alejoe91 can you scan this really quick. I just want to be more environmentally friendly while we decide on whether to fully remove the caching :)

PR #1634, opened 2025-01-24
Do Igor demo from g-node

zm711:
The file is labeled as win, which worked locally for me, but I have a windows so first need to make sure it works on the CI. Then we can tweak more if we want to.

### DIFF
--- a/examples/plot_igorio.py
+++ b/examples/plot_igorio.py
@@ -1,41 +1,39 @@
 """
-IgorProIO Demo (BROKEN)
+IgorProIO Demo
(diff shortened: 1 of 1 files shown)


zm711:
Only issue is that it is kind of a boring graph.....
https://neo--1634.org.readthedocs.build/en/1634/examples/plot_igorio.html#sphx-glr-examples-plot-igorio-py

alejoe91:
What about these data? https://github.com/wking/igor/tree/master/test/data

PR #1643, opened 2025-02-11
Update author list

zm711:
@alejoe91 quick merge if you have a moment : )


apdavison:
maybe update codemeta.json at the same time?

zm711:
I intentionally didn't because I always forget if we want it synced up to the release or to main.... For the future I can always update them at the same time? Or do I need to fiddle with other parts of the codemeta.json? As part of my release workflow I doublecheck the codemeta and the author list. But happy to change the workflow to make sure it is appropriately synced up.

apdavison:
That's a good point, perhaps making the update at release time is best.

PR #1645, opened 2025-02-28
Fix: In openephysbinaryrawio.py

remi-pr:
falling and rising indices of events must be arrays of integers, even if empty, otherwise it fails when trying to use them to compute durations. This issue is only triggered when trying to access recordings without any events. The small change was enough to fix the issue in the dataset I was exploring but there might be other places that need updating.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -292,15 +292,15 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                                     rising_indices.extend(rising)
                                     falling_indices.extend(falling)
 
-                            rising_indices = np.array(rising_indices)
-                            falling_indices = np.array(falling_indices)
+                            rising_indices = np.array(rising_indices, dtype=np.intp)
+                            falling_indices = np.array(falling_indices, dtype=np.intp)
 
                             # Sort the indices to maintain chronological order
                             sorted_order = np.argsort(rising_indices)
                             rising_indices = rising_indices[sorted_order]
                             falling_indices = falling_indices[sorted_order]
 
-                            durations = None
+                            # durations = None
                             # if len(rising_indices) == len(falling_indices):
                             durations = timestamps[falling_indices] - timestamps[rising_indices]
                             if not self._use_direct_evt_timestamps:
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -292,8 +292,8 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                                     rising_indices.extend(rising)
                                     falling_indices.extend(falling)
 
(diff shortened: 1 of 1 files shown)


neo/rawio/openephysbinaryrawio.py
```diff
@@ -292,15 +292,15 @@ def _parse_header(self):
...
+                            falling_indices = np.array(falling_indices, dtype=np.intp)
 
                             # Sort the indices to maintain chronological order
                             sorted_order = np.argsort(rising_indices)
                             rising_indices = rising_indices[sorted_order]
                             falling_indices = falling_indices[sorted_order]
 
-                            durations = None
+                            # durations = None
                             # if len(rising_indices) == len(falling_indices):
```
alejoe91, line 304:
```suggestion
```
Let's just go ahead and remove these

neo/rawio/openephysbinaryrawio.py
```diff
@@ -292,16 +292,14 @@ def _parse_header(self):
                                     rising_indices.extend(rising)
                                     falling_indices.extend(falling)
 
-                            rising_indices = np.array(rising_indices)
-                            falling_indices = np.array(falling_indices)
+                            rising_indices = np.array(rising_indices, dtype=np.intp)
```
alejoe91, line 295: @remi-pr coudl you explain why did you use `np.intp` instead of another int type (like np.int64)?

remi-pr:
Just a habit I took a while ago. Numpy intp is just an alias to int32/int64 depending on the platform so it does not affect the code efficiency but when I see it, I know it is for indices and not for specific data that would need a specific type (like uint8 for images or int16 for ephys). But I am not particularly attached to it in this case, I could change if you prefer.

neo/rawio/openephysbinaryrawio.py
remi-pr: Just a habit I took a while ago. Numpy intp is just an alias to int32/int64 depending on the platform so it does not affect the code efficiency but when I see it, I know it is for indices and not for specific data that would need a specific type (like uint8 for images or int16 for ephys). But I am not particularly attached to it in this case, I could change if you prefer.

alejoe91: Not sure, I would do int64 for clarity. @samuelgarcia @zm711 what do you think?

zm711: I think for me since I don't use it I would also be confused by it. Not that it is wrong, it's just I think all the maintainers tend to specify 32 or 64. So I would also tend toward int64 for our own readability.

remi-pr: Works for me. Switched to int64

alejoe91 (approved):

remi-pr:
I don't think I am in the author list, that would be nice. You can list me as being affiliated with [Aquineuro](https://aquineuro.eu/) Thanks!

PR #1650, opened 2025-03-05
Fix tdt default case dtype

h-mayorquin:
@samuelgarcia @alejoe91 
This is from this error in SpikeInterface:

https://github.com/SpikeInterface/spikeinterface/issues/3025

The thing is that for streams (`StoreName`) that have no matching data for the channel we assign the dtype as `int` and then transform it to string when we cast to the signal_channel_dtype for the struct.

This leads to the following error:

```python
_dtype = "int"
_dtype_as_string = str(_dtype)

np.dtype(_dtype_as_string)
TypeError: data type "<class 'int'>" not understood
```

The solution is to use `"int"` as a placeholder instead. 

I also changed some values that were used as numpy scalar as they will probably introduce bugs as they have done in the past.


samuelgarcia (approved):

samuelgarcia:
Thanks for this fix.

PR #1652, opened 2025-03-06
Add utility method to get timestamps on Intan base

h-mayorquin:
Hey @zm711 

I realize that we don't this. I would like to have a method like this exposed. I am fine with private and I can keep track of if if necessary but I think public is fine.

What do you think?

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -531,13 +531,14 @@ class IntanRawIO(BaseRawIO):
 
         return output
     
-    def get_intan_sample_indices(self, i_start=None, i_stop=None):
+    def get_intan_timestamps(self, i_start=None, i_stop=None):
         """
         Retrieves the sample indices from the Intan raw data within a specified range.
         
-        This function extracts the sample index timestamps from Intan files, which represent 
-        relative time points in sample units (not absolute time). These indices can be 
-        particularly useful when working with recordings that have discontinuities.
+        Note that sample indices are called timestamps in the Intan format but they are 
+        in fact just sample indices. This function extracts the sample index timestamps 
+        from Intan files, which represent  relative time points in sample units (not absolute time).
+        These indices can be  particularly useful when working with recordings that have discontinuities.
         
         Parameters
         ----------
@@ -597,7 +598,7 @@ class IntanRawIO(BaseRawIO):
             The error message includes a table detailing the discontinuities found.
         """
         # check timestamp continuity
-        timestamps = self.get_intan_sample_indices()
+        timestamps = self.get_intan_timestamps()
 
         discontinuous_timestamps = np.diff(timestamps) != 1
         timestamps_are_not_contiguous = np.any(discontinuous_timestamps)
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -597,7 +597,7 @@ class IntanRawIO(BaseRawIO):
             The error message includes a table detailing the discontinuities found.
         """
         # check timestamp continuity
-        timestamps = self.get_intan_timestamps()
+        timestamps = self.get_intan_sample_indices()
 
         discontinuous_timestamps = np.diff(timestamps) != 1
         timestamps_are_not_contiguous = np.any(discontinuous_timestamps)
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -554,8 +554,7 @@ class IntanRawIO(BaseRawIO):
             
         Notes
         -----
-        - Sample indices can be converted to seconds by dividing by the sampling rate.
-        - These are relative timestamps (sample indices) and not absolute time.
+        - Sample indices can be converted to seconds by dividing by the sampling rate of the amplifier stream.
         - The function automatically handles different file formats:
         * header-attached: Timestamps are extracted directly from the timestamp field
         * one-file-per-signal: Timestamps are read from the timestamp stream
@@ -574,11 +573,9 @@ class IntanRawIO(BaseRawIO):
         elif self.file_format == "one-file-per-channel":
             timestamps = self._raw_data["timestamp"][0]
 
-        # Apply flattening if needed
-        if timestamps.ndim > 1:
-            timestamps = timestamps.flatten()
+        # TODO if possible ensure that timestamps memmaps are always of correct shape to avoid memory copy here.
+        timestamps = timestamps.flatten() if timestamps.ndim > 1 else timestamps
         
-        # Apply range slicing
         if i_stop is None:
             return timestamps[i_start:]
         else:
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -530,6 +530,26 @@ class IntanRawIO(BaseRawIO):
             output[:, channel_index] = demultiplex_data[i_start:i_stop].flatten()
 
         return output
+    
+    def get_intan_timestamps(self):
+        """
+        
+        Retrieves the timestamps from the Intan raw data based on the file format.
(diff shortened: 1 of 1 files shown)


zm711:
Why do we need it? The timestamps in intan are just sample indices? They are good for checking skipping which I always throw away. Are you thinking that someone might want the timestamps exposed in the case they did ignore_integrity_checks?

h-mayorquin:
>  Are you thinking that someone might want the timestamps exposed in the case they did ignore_integrity_checks?

Exactly.

zm711:
Okay sounds good. Then I'll review this tomorrow when I get to lab!

h-mayorquin:
Yes, the plan is to load the timestamps on SpikeInterface for the cases when there are discontinuities on the timestamps.

zm711:
Do we have a mechanism to make it clear that these aren't "time"stamps but rather sample-stamps. I don't want us to start getting time vector stuff mixed up with this sample stuff.

h-mayorquin:
We can just multiply by the sampling rate to get timestamps and those will reflect the gaps. Right?

zm711:
Yeah I guess this is my own annoyance with the lack of consistency. For me timestamps are real timestamps (like 9am on March 7 2025), but in this case the conversion you're doing is to say I'm 10seconds into my recording with 0 time at the start of the recording. My concern for scientists are if they don't use a system that aligns stimuli and they record timestamps for their stimuli assuming they can link it up they might ruin one experiment not realizing that. So for my own mental schema it is

timestamps -- actual clock time
"sample"stamps/indices -- machine time
time vector -- sample time converted to the seconds domain or timestamps converted to a local time domain

All of this to say I agree with you in that we can convert the Intan "timestamps" into a time vector, but they having nothing to do with absolute time.

neo/rawio/intanrawio.py
```diff
@@ -555,16 +575,9 @@ def _assert_timestamp_continuity(self):
         * **one-file-per-channel:** Timestamps are retrieved from the first channel of the last stream.
         """
```
zm711, line 576: Do these notes make sense any more for the private method? Maybe you should move this to the get timestamps method now in a notes section.

neo/rawio/intanrawio.py
```diff
@@ -530,6 +530,26 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
             output[:, channel_index] = demultiplex_data[i_start:i_stop].flatten()
 
         return output
+    
+    def get_intan_timestamps(self):
+        """
+        
+        Retrieves the timestamps from the Intan raw data based on the file format.
```
zm711, line 537:
```suggestion
        Retrieves the timestamps (which are in sample indices) from the Intan raw data.
        
```
I would say we basically add the note from below in a Notes section, but as a user of the function I shouldn't care in the simple docstring about file format I should just care about the timestamps no? 

For me I want to make it clear to the user that these are in samples and not seconds. So I want to highlight specifically that it is sample time in the main one-liner.

neo/rawio/intanrawio.py
```diff
@@ -530,6 +530,26 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
...
+        Returns
+        -------
+        timestamps: ndarray
+            The flattened array of timestamps.
+        """
+        if self.file_format == "header-attached":
+            timestamps = self._raw_data["timestamp"]
+
+        # timestamps are always last stream for headerless binary files
```
zm711, line 546:
```suggestion
```
This comment means nothing now that you made the dictionary format :)

neo/rawio/intanrawio.py
```diff
@@ -530,6 +530,26 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
...
+    
+    def get_intan_timestamps(self):
+        """
+        
+        Retrieves the timestamps from the Intan raw data based on the file format.
+        Returns
+        -------
+        timestamps: ndarray
+            The flattened array of timestamps.
+        """
```
zm711, line 542: I would add a notes section here:

1 line to explain that samples can be converted to time through the sampling rate and maybe 1 line to say that this is relative time and not absolute time stamps
and then the 1-3 lines to explain that we handle each file format automatically (1 line if you think it is enough to just say we do it automatically or 3 lines if you want to explain how we do each format).

zm711 (commented):
I'm fine with this being public overall. I can imagine spikeinterface or neuroconv wanting to call this to load this for nwb etc. So I think if we know that you might want to call this separately for user-metadata then I think it is fair that we call this public. It is also a separate stream of data we don't support so I think having a user-facing way to get this data is important.

neo/rawio/intanrawio.py
h-mayorquin: Thanks. I will remove it.

neo/rawio/intanrawio.py
h-mayorquin: No, they should be moved.

neo/rawio/intanrawio.py
h-mayorquin: Addressed.

neo/rawio/intanrawio.py
```diff
@@ -530,6 +530,56 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
             output[:, channel_index] = demultiplex_data[i_start:i_stop].flatten()
 
         return output
+    
+    def get_intan_sample_indices(self, i_start=None, i_stop=None):
```
zm711, line 534: One tiny naming thing. Why do need intan in the name if this is already the intan io? Could we just do `get_sample_indices`? I guess the other thing about the timestamp name is it fits with the intan file naming convention so if I'm a random intan user would I expect to get data from the timestamp file from a function that is called `get_timestamps` or from a function that is called `get_sample_indices`. What do you end-uers would prefer-- the real values coming out or the company's naming convention? I lean slightly toward the company's naming with a good docstring (as you've done), but I could be convinced by your current naming.

h-mayorquin: I have a slight preference to not propagate bad nanung decisions by the company. If you want to call timestamps though I think it would be a good idea to keep the intan to emphasize that they are "intan timestamps" (i.e. they are not timestamps).  I think you should make the call here. I really don't use neo by itself I will wrap this on Spikeinterface so it really does not matter to me. Just to confirm, is the name that you prefer:

`get_timestamps`

zm711: I guess I'm fine with `get_intan_timestamps` to emphasize that they are not real timestamps. I think from a Neo level we have to work under the assumption that users will interact with their rawio in a slightly rawio specific way so our main functions need to fit with the neo hierarchy, but auxiliary functions are more rawio specific. 

Would you be less annoyed with `get_intan_timestamps`?

h-mayorquin: Both work me really. This is a design decision of the neo team. I was not trying to argue against it just saying how I feel.

> I think from a Neo level we have to work under the assumption that users will interact with their rawio in a slightly rawio specific way so our main functions need to fit with the neo hierarchy, but auxiliary functions are more rawio specific

This seems like a good reasoning to me.

So, what do you prefer?

zm711: I would say let's do `get_intan_timestamps` . I like your idea of emphasizing that this is the company's version of timestamps and not the standard definition of timestamps.

h-mayorquin: Let's move forward with this.

h-mayorquin: Changed.

zm711 (approved):

PR #1383, opened 2024-01-29
Extract events from spike glx

manimoh:
Trying to fix [this](https://github.com/NeuralEnsemble/python-neo/issues/1365?notification_referrer_id=NT_kwDOAHcV2LI4ODAzMzY1MTk4Ojc4MDQzNzY#issuecomment-1910319253)

PR #1657, opened 2025-03-16
WIP, imec sync channel processing

guptadivyansh:
Slowly working towards https://github.com/SpikeInterface/spikeinterface/issues/2234

Trying to reuse the nidq digital channel event processing for handling sync edges. Will then align edges and change sampling rate/time vector as necessary

Also, I'm not sure if any of the current test files have simultaneously recorded sync signals in both imec and nidq streams.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -300,21 +300,47 @@ class SpikeGLXRawIO(BaseRawWithBufferApiIO):
             channel = dig_ch[event_channel_index]
             ch_idx = 7 - int(channel[2:])  # They are in the reverse order
             this_stream = event_data[:, ch_idx]
-            this_rising = np.where(np.diff(this_stream) == 1)[0] + 1
-            this_falling = (
-                np.where(np.diff(this_stream) == 255)[0] + 1
-            )  # because the data is in unsigned 8 bit, -1 = 255!
-            if len(this_rising) > 0:
-                timestamps.extend(this_rising)
-                labels.extend([f"{channel} ON"] * len(this_rising))
-            if len(this_falling) > 0:
-                timestamps.extend(this_falling)
-                labels.extend([f"{channel} OFF"] * len(this_falling))
+            timestamps, durations, labels = self._find_events_in_channel(this_stream, channel)
         timestamps = np.asarray(timestamps)
         if len(labels) == 0:
             labels = np.asarray(labels, dtype="U1")
         else:
             labels = np.asarray(labels)
+
+        return timestamps, durations, labels
+    
+    def _get_sync_events(self, stream_index):
+        #find sync events in the 'SY0' channel of imec streams
+        channel = 'SY0'
+        sync_data = self.get_analogsignal_chunk(channel_names = [channel], stream_index = stream_index)
+        #uint16 word to uint8 bytes to bits
+        sync_data_uint8 = sync_data.view(np.uint8)
+        unpacked_sync_data = np.unpackbits(sync_data_uint8, axis=1)
+        #sync line is the 6th bit (so 1st bit because reversed order)
+        #https://billkarsh.github.io/SpikeGLX/Sgl_help/Metadata_30.html
+        sync_line = unpacked_sync_data[:,1]
+        raise NotImplementedError("Work in progress")
+        return self._find_events_in_channel(sync_line, "SY0")
+
+    def _find_events_in_channel(self, channel_data, channel_name):
+
+        timestamps, durations, labels = [], None, []
+
+        this_rising = np.where(np.diff(channel_data) == 1)[0] + 1
+        this_falling = (
+            np.where(np.diff(channel_data) == 255)[0] + 1
+        )  # because the data is in unsigned 8 bit, -1 = 255!
+        if len(this_rising) > 0:
+            timestamps.extend(this_rising)
+            labels.extend([f"{channel_name} ON"] * len(this_rising))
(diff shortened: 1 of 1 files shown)


manimoh:
Hi @guptadivyansh 
Thank you so much for working on this issue!
This is a real-world (although really big) [dataset](https://datasets.datalad.org/?dir=/labs/mvdm/OdorSequence/sourcedata/raw/M541/M541-2024-08-31_g0) that has simultaneously recorded sync signals from 2 imec device and 1 nidq device. If you think it is too cumbersome to handle, I can try clipping it and trying to submit a pull-request for much more sensibly sized  dataset on [here](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx)

guptadivyansh:
Hi Manish,

For now I'm testing with some data that I recorded myself. It's about
2-3GB, but still too big to add to the testing suite. Once I have something
working, we can think about how to add a worthwhile test case that includes
at least a few seconds of data.

zm711:
Thanks so much for working on this. Please ping Alessio and Sam to review when you're ready!

guptadivyansh:
I think this should expose all that we need from neo to solve https://github.com/SpikeInterface/spikeinterface/issues/2234 that is, the timestamps of sync edges of the correct channels in any given streams. Actually aligning the edges across streams can/should(?) be done in spikeinterface, likely [here](https://github.com/SpikeInterface/spikeinterface/blob/38d82d4f8ce1430f5e3375a907e5c3c557a78711/src/spikeinterface/extractors/neuropixels_utils.py#L107). 

I also noticed that the nidq digital channel processing introduced in https://github.com/NeuralEnsemble/python-neo/pull/1383 is hardcoded for the first segment. I can fix it in a separate PR. Once that is done, we should be able to get sync events from all segments too.

neo/rawio/spikeglxrawio.py
```diff
@@ -292,33 +293,113 @@ def _event_count(self, event_channel_idx, block_index=None, seg_index=None):
         return timestamps.size
 
     def _get_event_timestamps(self, block_index, seg_index, event_channel_index, t_start=None, t_stop=None):
+        #TODO: fix seg_index usage, currently hardcoded for first segment
         timestamps, durations, labels = [], None, []
         info = self.signals_info_dict[0, "nidq"]  # There are no events that are not in the nidq stream
```
guptadivyansh, line 298: currently the event detector for nidq digital channels ignores the seg_index. This needs fixing for detecting sync events across segments. 

Should still work for multi-segment imec streams though

neo/rawio/spikeglxrawio.py
```diff
@@ -292,33 +293,113 @@ def _event_count(self, event_channel_idx, block_index=None, seg_index=None):
...
+            niChanType = int(meta['syncNiChanType'])
+            niChan = int(meta['syncNiChan'])
+            if niChanType == 0: #digital channel          
+                timestamps, _, labels = self._get_event_timestamps(0, seg_index, niChan)
+            elif niChanType == 1: #analog channel
+                niThresh = float(meta['syncNiThresh']) #volts
+                sync_line = self.get_analogsignal_chunk(channel_names = [f'XA{niChan}'],
+                                                        stream_index = stream_index, 
+                                                        seg_index = seg_index)
+                #Does this need to be scaled by channel gain before threshold?
```
guptadivyansh, line 366: Help!

guptadivyansh:
@alejoe91 , @samuelgarcia , would be grateful if you could have a look!

guptadivyansh:
There's a messy testing notebook [here](https://github.com/guptadivyansh/python-neo/blob/divdev/neo/test/rawiotest/divdev_neo.ipynb) that uses some of my data but I still need to look into which (if any) of the current testing datasets have simultaneously recorded sync lines for more than a couple of seconds.

Ideally, we want a good mix of multiple probes, and sync signal in nidq digital and analog channels. Most of the neural channels in the test imec streams can be dropped to keep a managable filesize. (otherwise we'll need a ~23 MB test file per second per probe)

neo/rawio/spikeglxrawio.py
```diff
@@ -292,33 +293,113 @@ def _event_count(self, event_channel_idx, block_index=None, seg_index=None):
...
     def _get_event_timestamps(self, block_index, seg_index, event_channel_index, t_start=None, t_stop=None):
+        #TODO: fix seg_index usage, currently hardcoded for first segment
         timestamps, durations, labels = [], None, []
         info = self.signals_info_dict[0, "nidq"]  # There are no events that are not in the nidq stream
         dig_ch = info["digital_channels"]
         if len(dig_ch) > 0:
             event_data = self._events_memmap
-            channel = dig_ch[event_channel_index]
+            channel = dig_ch[event_channel_index] # 'XD0', 'XD1', etc.
             ch_idx = 7 - int(channel[2:])  # They are in the reverse order
```
manimoh, line 303: This calculation of channel index is also something that I had hard-coded and will work only when there are exactly 8 digital channels saved (which was the cases for our setup, but isn't generally true); but not when you save fewer channels or have I/O devices that have more than 8 digital channels.

h-mayorquin: Thanks for pointing this out, this is something to fix.

h-mayorquin:
I can take a look at this but first we should determine if there is available data on the testing set that contains the cases that you want to cover in your scope:

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx

PR #1660, opened 2025-03-20
Add stim demultiplex rhs intan

### LINKED ISSUE #1649: Intan. Decode stim stream for rhs devices
@zm711

I discussed this today with @TommasoLambresa, and he mentioned that the stim channel in the RHS is also encoded. I think we should do something similar to what we did before in #1491 with the multiplexed digital channels, making `get_analogsignal_chunk` return the decoded signal so no further post-processing is necessary at user side.

From the manual:

> Stimulation current is stored in the lower 9 bits. Units: ‘Stim Step Size’ as set in the header file. To convert to current in amps, multiply the lowest 8 bits by the ‘Stim Step Size’. Then, for each 16-bit unsigned integer entry, if the 9th bit is 1 (signifying a negative current), multiply the magnitude by -1.  
> Bit 16 (the MSB) indicates that a compliance limit was reached. Bit 15 is 1 if charge recovery is activated. Bit 14 is 1 if amplifier settling is activated. Bits 10-13 are unused and always zero.

Image

https://intantech.com/files/Intan_RHS2000_data_file_formats.pdf

Similar encoding schemes are in effect for the header-only, file-per-channel, and file-per-stream options.

The group led by @TommasoLambresa might be able to provide some testing data for this.

h-mayorquin:
Should fix #1649

We will be needing test data but this work for reading the data for me:

```python
from spikeinterface.extractors import IntanRecordingExtractor

recording = IntanRecordingExtractor(file_path, stream_name="Stim channel")

recording.get_traces()
```

If you can give it a try @TommasoLambresa and @FrancescoNegri and check that the outputs matches what you expect it would be great.

@zm711


zm711:
> No idea why the docs are failing on this one.

We've been having a lot of failures of gin lately. it's intermittent (3.9, 1.26 testing failed too because of that). I'm rerunning now :)

neo/rawio/intanrawio.py
```diff
@@ -505,21 +523,57 @@ def _get_analogsignal_chunk_one_file_per_signal(self, i_start, i_stop, stream_in
...
         return output
 
     def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_stop):
-
+        """
+        Demultiplex digital data by extracting individual channel values from packed 16-bit format.
+        
+        According to the Intan format, digital input/output data is stored with all 16 channels
+        encoded bit-by-bit in each 16-bit word. This method extracts the specified digital channels
+        from the packed format into separate boolean arrays.
```
zm711, line 549: These are uint16 arrays not boolean.  Just don't want people to be confused about the dtype. The idea is boolean but the dtype is not.... unles syou want a different wording.
```suggestion
        from the packed format into separate uint16 arrays of 0 and 1.
```

neo/rawio/intanrawio.py
```diff
@@ -505,21 +523,57 @@ def _get_analogsignal_chunk_one_file_per_signal(self, i_start, i_stop, stream_in
...
+        i_start : int
+            Starting sample index for demultiplexing.
+        i_stop : int
+            Ending sample index for demultiplexing (exclusive).
+            
+        Returns
+        -------
+        ndarray
+            Demultiplexed digital data with shape (i_stop-i_start, len(channel_ids)), 
+            containing boolean values for each requested channel.
```
zm711, line 567:
```suggestion
            containing 0 or 1 values for each requested channel.
```

? again maybe we make sure we know what the actual values are

neo/rawio/intanrawio.py
```diff
@@ -530,6 +584,61 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
...
+        Bit structure of each 16-bit stimulation word:
+        - Bits 0-7: Current magnitude
+        - Bit 8: Sign bit (1 = negative current)
+        - Bits 9-13: Unused (always zero)
+        - Bit 14: Amplifier settle flag (1 = activated)
+        - Bit 15: Charge recovery flag (1 = activated) 
+        - Bit 16 (MSB): Compliance limit flag (1 = limit reached)
+        
+        The actual current value in amperes is obtained by multiplying the
+        output by the 'stim_step_size' parameter from the file header.
```
zm711, line 623:
```suggestion
        output by the 'stim_step_size' parameter from the file header. These scaled values can be 
        obtained with the `rescale_signal_raw_to_float` function.
```

I know we deal with this at the SI level, but I think we should make this clear at the Neo level that the 
amps are only obtained if we rescale. So this requires two steps 1) get data which we now demultiplex 2) rescale the data which the user must do in a second function

neo/rawio/intanrawio.py
```diff
@@ -530,6 +584,61 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
...
+        data = raw_stim_data[i_start:i_stop]
+        
+        # Extract current value (bits 0-8)
+        magnitude = np.bitwise_and(data, 0xFF)
+        sign_bit = np.bitwise_and(np.right_shift(data, 8), 0x01)  # Shift right by 8 bits to get the sign bit
+        
+        # Apply sign to current values
+        current = np.where(sign_bit == 1, -magnitude, magnitude)
+        
+        # Note: If needed, other flag bits could be extracted as follows:
```
zm711, line 635: Yeah this is interesting.... I guess we can see what users are interested in.

neo/rawio/intanrawio.py
```diff
@@ -462,7 +463,18 @@ def _get_analogsignal_chunk_header_attached(self, i_start, i_stop, stream_index,
...
+            # For stim data, we need to extract the raw data first, then demultiplex it
+            stim_data = np.zeros((i_stop - i_start, len(channel_ids)), dtype=dtype)
+            for chunk_index, channel_id in enumerate(channel_ids):
+                data_chan = self._raw_data[channel_id]
+                if multiple_samples_per_block:
+                    stim_data[:, chunk_index] = data_chan[block_start:block_stop].flatten()[sl0:sl1]
+                else:
+                    stim_data[:, chunk_index] = data_chan[i_start:i_stop]
+            # Now demultiplex the stim data
+            sigs_chunk = self._decode_current_from_stim_data(stim_data, 0, stim_data.shape[0])
```
zm711, line 476: What if the user wants the stim data only for a subset of channels because they only ran a subset of amplifier channels? How would that work? (ie why hardcode 0 here?)

zm711 (commented):
First round of comments.

neo/rawio/intanrawio.py
zm711: This isn't clear what I'm point at :) I mean your comments from line 635 down.

neo/rawio/intanrawio.py
h-mayorquin: As far as I understand the data we are passing here for decoding of the current already was reduce to the required channels and time slice above (see the lines before this). 

Can you double check with me if this is the case? I was hoping we could discuss it together before I refactor this a bit as the process is common to all other signals and stim.

neo/rawio/intanrawio.py
h-mayorquin: Sure.

neo/rawio/intanrawio.py
zm711: I was assuming this function would be exactly like the demultiplex of digital data which still takes in the channels. but let re-read this.

h-mayorquin: More like, the header case first extracts the data for the specific channels that were requested and then you pass it for decoding.

zm711: Okay I agree with this except the one file per channel. Since I couldn't generate stim data I don't know where or how this is stored for that format. Unless we get test files I would argue we remove that option. If we remove that option for now then the demultiplex doesn't need a start/stop since you've already calculated it above. Are you confident how one file per channel works for stim?

TommasoLambresa:
>If you can give it a try @TommasoLambresa and @FrancescoNegri and check that the outputs matches what you expect it would >be great.

@h-mayorquin, @zm711 
We're still here, sorry for the delay. We moved on this branch pointing to this PR, but running the lines that you sent us **nothing** changes compared with the previous neo version. We tried to debug, but it never enters in the function that you have modified: when `spikeinterface` calls the `neorawio` `parse_header` function, we see that the funtion `read_rhs` is called. In this function there is no reference to the function modified in this PR, so we're asking: how can we test the changes?      
We tried also to print some random text but nothing happens.

zm711:
Is your neo install editable?

TommasoLambresa:
Yes, we installed neo editable! @zm711

TommasoLambresa:
By trying these lines, it raised this error:

```
ValueError                                Traceback (most recent call last)
Cell In[6], line 5
      3 reader = IntanRawIO("Z:/SNN/K-SNN/R24-21/02_stim_240621_132910.rhs")
      4 reader.parse_header()
----> 5 sig_chunk = reader.get_analogsignal_chunk(stream_index=11)
      7 import matplotlib.pyplot as plt
      8 plt.plot(sig_chunk.T)

File ~\Repositories\test_NEOIntan\mayorquin-neo\neo\rawio\baserawio.py:805, in BaseRawIO.get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes, channel_names, channel_ids, prefer_slice)
    799     error_message = (
    800         "get_analogsignal_chunk can't be called on a file with no signal streams or channels."
    801         "Double check that your file contains signal streams and channels."
    802     )
    803     raise AttributeError(error_message)
--> 805 stream_index = self._get_stream_index_from_arg(stream_index)
    806 channel_indexes = self._get_channel_indexes(stream_index, channel_indexes, channel_names, channel_ids)
    808 # some check on channel_indexes

File ~\Repositories\test_NEOIntan\mayorquin-neo\neo\rawio\baserawio.py:653, in BaseRawIO._get_stream_index_from_arg(self, stream_index_arg)
    651 else:
    652     if stream_index_arg < 0 or stream_index_arg >= self.header["signal_streams"].size:
--> 653         raise ValueError(f"stream_index must be between 0 and {self.header['signal_streams'].size}")
    654     stream_index = stream_index_arg
    655 return stream_index

ValueError: stream_index must be between 0 and 3
```

As you can see the functions are called from my local repository, which contains the `neo` updated version. It seems it doesn't recognize the file as containing a stim stream, but I'm sure it's wrong!

zm711:
```python
from neo.rawio import IntanRawIO

reader = IntanRawIO(path/to/data)
reader.parse_header()
signal_streams = reader.header['signal_streams']
stream_ids = signtal_streams['id'].tolist()
sig_chunk = reader.get_analogsignal_chunk(stream_index=11)

File: neo/rawio/intanrawio.py (hunk 5 above)
Comment (h-mayorquin) [reply neo/rawio/intanrawio.py:476]: The documentation on the rhs manual clearly indicates is the same format:

image

Is there something I am misreading?

**Comment by h-mayorquin** [comment]
Hi, @TommasoLambresa 

Using this branc, this is what I Get for the file that you shared with me:

```python
from spikeinterface.extractors import IntanRecordingExtractor
from matplotlib import pyplot as plt


recording = IntanRecordingExtractor(file_path, stream_name="Stim channel")

traces = recording.get_traces(return_scaled=True)
times = recording.get_times()
num_channels = traces.shape[1]
for i in range(num_channels):
    channel_traces = traces[:, i]
    if channel_traces.sum() > 0:
        plt.plot(times, channel_traces, label=f"Channel {i}")
        
plt.legend()
plt.xlabel("Time (s)")
plt.ylabel("Amperes")

```

TommasoLambresa:
Hi @h-mayorquin 
The same goes for me, but unfortunately it is not the correct conversion. This is basically the same trace got without the binary conversion (it is just scaled accordingly to some gain and offset). I'm looking what goes wrong in this conversion function that you wrote. The correctly decoded traces contains biphasic anodic-cathodic stimuli of $\pm 60  \mu A$. Debugging within your function it seems the conversion is not made properly. I'm trying to understand what's wrong.

h-mayorquin:
> Hi @h-mayorquin The same goes for me, but unfortunately it is not the correct conversion. This is basically the same trace got without the binary conversion (it is just scaled accordingly to some gain and offset). I'm looking what goes wrong in this conversion function that you wrote. The correctly decoded traces contains biphasic anodic-cathodic stimuli of ± 60 μ A . Debugging within your function it seems the conversion is not made properly. I'm trying to understand what's wrong.

Thanks a lot, this is very helpful and precisely the reason why I was waiting for someone to implement this. Two questions:

1) Can you reproduce the figure that I have? (this is to see that we can get you the code to make debugging faster in the future)
2) Can you share the code that you use for decoding with me? Maybe that way I can spot the difference and what I did wrong.

TommasoLambresa:
>1. Can you reproduce the figure that I have? (this is to see that we can get you the code to make debugging faster in the future)

Yes, I have the same result:
```python
from spikeinterface.extractors import IntanRecordingExtractor
from matplotlib import pyplot as plt


recording = IntanRecordingExtractor("C:/Users/t.lambresa/Downloads/newintanFile 1.rhs", stream_name="Stim channel")

traces = recording.get_traces(return_scaled=True)
times = recording.get_times()
num_channels = traces.shape[1]
for i in range(num_channels):
    channel_traces = traces[:, i]
    if channel_traces.sum() > 0:
        plt.plot(times, channel_traces, label=f"Channel {i}")
        
plt.legend()
plt.xlabel("Time (s)")
plt.ylabel("Amperes")
```

image
---
>2. Can you share the code that you use for decoding with me? Maybe that way I can spot the difference and what I did wrong.
Sure here is the code @h-mayorquin:
```python
def decode_stim_data(orig_data, stim_step_size):
    """Converts the stimulation data from the Intan RHS format to micro-amps.
    Parameters
    ----------
    stim_data : np.ndarray (dtype=uint16)
        One-dimensional array of STIM samples read from the .rhs file.
    stim_step_size : float
        'Stim Step Size' value read from the header (in amperes per step).
    Returns
    -------
    stim_curr : np.ndarray (float)
        Current in micro-amps (positive or negative).
    """
    stim_data = orig_data.copy()
    compliance_limit_data = stim_data >= 2**15
    stim_data = stim_data - (compliance_limit_data * 2**15)
    charge_recovery_data = stim_data >= 2**14
    stim_data = stim_data - (charge_recovery_data * 2**14)
    amp_settle_data = stim_data >= 2**13
    stim_data = stim_data - (amp_settle_data * 2**13)
    stim_polarity = stim_data >= 2**8
    stim_data = stim_data - (stim_polarity * 2**8)
    stim_polarity = 1 - 2 * stim_polarity  # Convert (0 = pos, 1 = neg) to +/-1
    stim_data = stim_data * stim_polarity
    # Convert to amps
    stim_curr = stim_data * stim_step_size
    return stim_curr, compliance_limit_data, charge_recovery_data, amp_settle_data
```

TommasoLambresa:
Yes, now it works :)

zm711:
@h-mayorquin wasn't our whole discussion for spikeinterface that we weren't going to allow people to do this through spikeinterface (without going through the scale mechanism)?

TommasoLambresa:
I think the issue could be that the signal is casted as `uint` but it is `int` actually. And the exadecimal to take the first 8 bits is `0x7F10`, taking into account that the first 8 bits codify the magnitude. By using `0xFF` you consider only the last 8 bits if I'm not wrong. @h-mayorquin

neo/rawio/intanrawio.py
```diff
@@ -857,8 +967,8 @@ def read_rhs(filename, file_format: str):
             chan_info_stim["sampling_rate"] = sr
             # stim channel are complicated because they are coded
             # with bits, they do not fit the gain/offset rawio strategy
-            chan_info_stim["units"] = ""
-            chan_info_stim["gain"] = 1.0
+            chan_info_stim["units"] = "A"  # Amps
+            chan_info_stim["gain"] = global_info["stim_step_size"]
```
zm711, line 971: We already skip this in the header so it will fail anyway. See line 962.

TommasoLambresa:
No sorry, you're right! I'm terrible too...

zm711:
We will get this all sorted :) 

Like I said in my note, spikeinterface is moving toward limiting non-voltage scaling in the `get_traces` function. So there will be a workaround to do what you want. That is why I was having you try to test in neo first. We allow for any units for scaling. Then Heberto will make sure it scales correctly in neuroconv. But spikeinterface will require an extra few lines of code to scale correctly. I'll wait until Heberto posts his fix then I would recommend you both test in neo first :)

h-mayorquin:
@TommasoLambresa your hunch was right

The problem with my implementation is that I received the data on uint16 and multiplying by the current sign did not set the right negative values. 

I pushed a fix which maybe will make some tests fail (because of dtype mismatch) but should get the correct result. As per what @zm711 was saying, this would be the correct code to get the data in units in SpikeInterface:

```python
from spikeinterface.extractors import IntanRecordingExtractor
from spikeinterface.preprocessing import scale
from matplotlib import pyplot as plt


recording = IntanRecordingExtractor(file_path, stream_name="Stim channel", all_annotations=True)
gain_to_physical_unit = recording.get_property("gain_to_physical_unit")
traces_unscaled = recording.get_traces()
recording_scaled = scale(recording, gain=gain_to_physical_unit)

traces = recording_scaled.get_traces()
times = recording_scaled.get_times()  
num_channels = traces.shape[1]
for i in range(num_channels):
    channel_traces = traces[:, i]
    if np.abs(channel_traces).sum() > 0:
        plt.plot(times, channel_traces, label=f"Channel {i}")
        
plt.legend()
plt.xlabel("Time (s)")
plt.ylabel("Amperes")
```

You will need an updated main branch from spikeinterface as well though.

neo/rawio/intanrawio.py
```diff
@@ -631,6 +631,9 @@ def _decode_current_from_stim_data(self, raw_stim_data, i_start, i_stop):
         sign_bit = np.bitwise_and(np.right_shift(data, 8), 0x01)  # Shift right by 8 bits to get the sign bit
         
         # Apply sign to current values
+        # We need to cast to int16 to handle negative values correctly
```
h-mayorquin, line 634: @zm711 @TommasoLambresa

neo/rawio/intanrawio.py
h-mayorquin: I see.

h-mayorquin:
> I don't think that will break our tests. It will break extractors test on spikeinterface instead. -- how do you want to deal with the si tests?
> 
> This method of storage is annoying for the neo object model. Maybe Heberto and I (+ Sam if he is free should have a call).

Which tests do you think will be broken? 
I was talking about the memamp dtype not matching the corresponding channel dtype but I realize that it is decoupled so that's OK.

zm711:
I thought this would fail:
```python
            dtype = rec.get_dtype()


            for segment_index in range(num_seg):
                num_samples = rec.get_num_samples(segment_index=segment_index)


                full_traces = rec.get_traces(segment_index=segment_index)
                assert full_traces.shape == (
                    num_samples,
                    num_chans,
                ), f"{full_traces.shape} != {(num_samples, num_chans)}"
                assert full_traces.dtype == dtype, f"{full_traces.dtype} != {dtype=}"
```
because the rec will start the dtype as the uint16 and the full_trace will be converted to int16

h-mayorquin:
> I thought this would fail:
> 
> ```python
>             dtype = rec.get_dtype()
> 
> 
>             for segment_index in range(num_seg):
>                 num_samples = rec.get_num_samples(segment_index=segment_index)
> 
> 
>                 full_traces = rec.get_traces(segment_index=segment_index)
>                 assert full_traces.shape == (
>                     num_samples,
>                     num_chans,
>                 ), f"{full_traces.shape} != {(num_samples, num_chans)}"
>                 assert full_traces.dtype == dtype, f"{full_traces.dtype} != {dtype=}"
> ```
> 
> because the rec will start the dtype as the uint16 and the full_trace will be converted to int16

Right, OK, so we had the same on mind, but now, the memmap dtype and the channel dtype are decoupled so we can set the former to uint16 and the latter to int16.

zm711:
Okay so I want to summarize where we are right now:

1) Finalize if we are okay with changing the dtype from the memmap dtype (uint-> int). These seems basically required because we need negative values for the final unscaled stim value. Let me take the weekend to think about this.
2) Finalize what to do with one-file-per-channel. I would say unless we get a test file for it that we keep it as unsupported until a user requests it. (ie I would comment out/delete your function) because as of right now we don't even parse the header at the moment.

h-mayorquin:
> Okay so I want to summarize where we are right now:
> 
> 1. Finalize if we are okay with changing the dtype from the memmap dtype (uint-> int). This seems basically required because we need negative values for the final unscaled stim value. Let me take the weekend to think about this.

We don't need to change the memmap dtype just the channel dtype.

zm711:
> We don't need to change the memmap dtype just the channel dtype.

That's what I meant. The memmap dtype is uint and the `sig_chunk` is int.

h-mayorquin:
> > We don't need to change the memmap dtype just the channel dtype.
> 
> That's what I meant. The memmap dtype is uint and the `sig_chunk` is int.

Why would that be a problem? It seems to me that the memmap dtype is an internal detail and sig_chunk dtype is ,correctly, the channel dtype as per the channel streams header.

zm711:
That's why I want to take the weekend to re-read carefully about doing it this way. I want to see where everything happens at each step I think you already have a mental model, but I want to get one too.

h-mayorquin:
Makes sense @zm711. Thanks for thinking carefully about this.

zm711:
So I also want to make sure we have no chance for overflow. So really our uint16 storage is actually uint8 storage with the other 8 bits giving various information. So switched to signed from uint8 should be fully encapsulated by int16 right? Realistically a user shouldn't hit that but I don't want a tough to find overflow bug from our conversion. Does my logic make sense @h-mayorquin ?

neo/rawio/intanrawio.py
```diff
@@ -531,6 +579,65 @@ def _demultiplex_digital_data(self, raw_digital_data, channel_ids, i_start, i_st
...
+        # Get the relevant portion of the data
+        data = raw_stim_data[i_start:i_stop]
+
+        # Extract current value (bits 0-8)
+        magnitude = np.bitwise_and(data, 0xFF)
+        sign_bit = np.bitwise_and(np.right_shift(data, 8), 0x01)  # Shift right by 8 bits to get the sign bit
+
+        # Apply sign to current values
+        # We need to cast to int16 to handle negative values correctly
+        # The max value of 8 bits is 255 so the casting is safe as there are non-negative values
```
h-mayorquin, line 629: @zm711 here a comment regarding your question.

zm711: Sorry I totally missed that. I think this should be good to go once we either get a one-file-per-channel test file set or comment out that stim demultiplex code.

h-mayorquin:
Here is the gin PR:

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/151

zm711:
gin file is merged and can be added to testing.

h-mayorquin:
I added a test, here is the data:
image

And I am testing that the first pulse matches the description that the authors gave us to ensure that the decoding is done properly

image

zm711:
We may have discovered an issue with nixio by updating the io... let me look into that for a moment before we merge this.

zm711:
looks like setuptools just did a release five hours ago. I think that new release is breaking our nixio install. I'll see if I can limit setup tools for now, but I'll be away from my computer for a minute so I'll have to debug this more tomorrow.

neo/rawio/intanrawio.py
```diff
@@ -479,6 +485,8 @@ def _get_analogsignal_chunk_one_file_per_channel(self, i_start, i_stop, stream_i
 
         stream_name = self.header["signal_streams"][stream_index]["name"][:]
         signal_data_memmap_list = self._raw_data[stream_name]
+        stream_is_stim = stream_name == "Stim channel"
```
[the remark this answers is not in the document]
h-mayorquin: I think that the playing it safe here should be an error. Shouldn't it?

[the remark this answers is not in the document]
zm711: But if we aren't writing these into possible channels how will this stream be exposed to the user? To be safe I'm fine with raising an error though!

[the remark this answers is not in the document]
h-mayorquin: So what happens when someone requests this stream from `SpikeInterface` and calls `get_traces`?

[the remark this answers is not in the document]
zm711: They won't have that option. We don't add this into `ordered_channel_info` and we build the `signal_streams` based on the `stream_ids` of the channels. Since we don't load the channel info for the stim in that format the signal_stream won't be in the header. So they will get an error in spikeinterface saying that stream doesn't exist.

https://github.com/NeuralEnsemble/python-neo/blob/07fa5b32218b7b5515f03a29429a80b674efac4a/neo/rawio/intanrawio.py#L224-L244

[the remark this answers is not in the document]
h-mayorquin: But that won't tell them why.

If I understand you are afraid of returning wrong data. But when a user tries to load a stream and it does not work there is a chance they just won't report this to us (which is when I guess you want to tell them, give us data!) and move away from the library.

I want to throw an error letting them know why and what to do so we can push this feature out as soon as possible. I don't want to leave this as "it just does not work".

From that perspective, I would rather would rather return (possible) wrong data with a warning than a silent lack of the feature without a reachable explanation.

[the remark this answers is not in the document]
h-mayorquin: My desire to put features out would be less strong if neo had a quicker release possible but this is not the case at all so the cycle from people finding out an error to fix is REALLY SLOW. People turning away from our tools becomes a possibility in that scenario. Waiting months or years for something to be reported and then fixed is too much.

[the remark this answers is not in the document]
zm711: Just wanted your opinion on whether you think this is not enough of a warning. :) To close the loop and finalize this PR?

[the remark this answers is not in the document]
h-mayorquin: It is OK to me. I will throw an error in neuroconv and that would fit our needs.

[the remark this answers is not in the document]
zm711: > My desire to put features out would be less strong if neo had a quicker release possible but this is not the case at all so the cycle from people finding out an error to fix is REALLY SLOW. People turning away from our tools becomes a possibility in that scenario. Waiting months or years for something to be reported and then fixed is too much.

I missed this comment. I think this is a fair concern. Releases are actually pretty easy. I would be happy to propose something like a quarterly release schedule at Neo so we don't run into that. It takes me maybe an hour to prep a release. We can discuss this at the meeting too and if I need to push forward with doing more releases to be helpful that is minimal burden on my side.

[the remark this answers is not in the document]
h-mayorquin: Yes, a faster release cycle would be great.

The last time we found an error with the one-file-per-channel format was because we released a feature and a user reported it to us.

That said, there are of course costs that are harder to assess:

* If we are cautious, we might have users who see the lack of a feature and turn away from the library, thinking we don't have enough features or support—and without reporting it.
* If we are brash, we might have users who get incorrect data and turn away from the library, thinking it's buggy—again, without reporting it.

I recognize it's a hard call.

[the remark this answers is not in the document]
zm711: No, I take your point. You and I have been working on intan for ages together and their documentation has always been perfect. So what I propose instead is we fix the header reading such that it reads the header as a `one-file-per-channel` the same as the other formats but at the end of loading the header add a warning that this is based on docs and we need test files to verify. If that sounds acceptable to you I can push those changes instead.

I think Neo in the past has been less cautious than I'm being now and relied on users reporting bugs so I take your point!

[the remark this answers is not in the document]
h-mayorquin: > So what I propose instead is we fix the header reading such that it reads the header as a one-file-per-channel the same as the other formats but at the end of loading the header add a warning that this is based on docs and we need test files to verify. If that sounds acceptable to you I can push those changes instead.

All right.

zm711:
@h-mayorquin could you scan through my changes really quick just to make sure I didn't do anything stupid!

zm711:
@h-mayorquin I added in the dc only test folder, because it has stim file tests and it tests the fact that the end user can choose not to save normal amp files this will allow to test one-file-per-channel for dc amp and for stim file. The other test is nice but not necessary to get this merged :)

neo/rawio/intanrawio.py
```diff
@@ -793,8 +900,18 @@ def read_rhs(filename, file_format: str):
...
         channel_number_dict["DC Amplifier channel"] = channel_number_dict["RHS2000 amplifier channel"]
-        channel_number_dict["Stim channel"] = channel_number_dict["RHS2000 amplifier channel"]
+        if file_format != "one-file-per-channel":
+            channel_number_dict["Stim channel"] = channel_number_dict["RHS2000 amplifier channel"]
+            raw_file_paths_dict = create_one_file_per_signal_dict_rhs(dirname=filename.parent)
+        else:
+            raw_file_paths_dict = create_one_file_per_channel_dict_rhs(dirname=filename.parent)
+            channel_number_dict["Stim channel"] = len(raw_file_paths_dict["Stim channel"])
+            # but the user can shut off the normal amplifier and only save dc amplifier
+            channel_number_dict["RHS2000 amplifier channel"] = len(raw_file_paths_dict["RHS2000 amplifier channel"])
```
zm711, line 914: @h-mayorquin does this make sense or can you think of a better way of doing this?

neo/rawio/intanrawio.py
```diff
@@ -808,24 +925,25 @@ def read_rhs(filename, file_format: str):
...
-        else:
-            chan_info["dtype"] = "int16"
-        ordered_channel_info.append(chan_info)
-        if file_format == "header-attached":
-            name = chan_info["native_channel_name"]
-            memmap_data_dtype += [(name, "uint16", BLOCK_SIZE)]
-        else:
-            memmap_data_dtype["RHS2000 amplifier channel"] = "int16"
+    if file_format != "one-file-per-channel" or channel_number_dict["RHS2000 amplifier channel"] > 0:
+        for chan_info in stream_name_to_channel_info_list["RHS2000 amplifier channel"]:
```
zm711, line 929: This is the only change here (the diff is wrong). Basically if the user shut off saving of files in the file-per-channel we need to skip this section. Does this make sense.

neo/rawio/intanrawio.py
```diff
@@ -847,27 +965,31 @@ def read_rhs(filename, file_format: str):
...
-    # so let's skip for now and can be given on request
-    if file_format != "one-file-per-channel":
-        for chan_info in stream_name_to_channel_info_list["RHS2000 amplifier channel"]:
+    # so ideally at some point we need test data to confirm this is true
+    # based on what Heberto and I read in the docs
+    for chan_info in stream_name_to_channel_info_list["RHS2000 amplifier channel"]:
+        # we see which stim were activated
+        if file_format != "one-file-per-channel" or any(
+            [chan_info["native_channel_name"] in stim_file.stem for stim_file in raw_file_paths_dict["Stim channel"]]
+        ):
```
zm711, line 974: For this we check if these exist and if they do we analyze. This was my hacky way of doing this (since the user can choose individual channels to stim (ie stim files only exist for a subset of amp channels). Check this for me.

zm711 (commented):
@h-mayorquin I marked three comments of what I want you to glance out before we merge this. The fact that tests are passing means that my assumptions of the data org appear to be correct, but if you can think of a clearer/better way to do what I've done feel free to tweak. Once you glance over my additions this is ready for merge.

neo/rawio/intanrawio.py
h-mayorquin: I don't understand the context. Why are we doing this? I re-read 

```
        # if using the `header-attached` or `one-file-per-signal` formats
        # the amplifier data is stored in the same place in the header so no matter what the DC amp
        # uses the same info as the RHS amp
```

There is something strange about it and makes me confused.

neo/rawio/intanrawio.py
zm711: @h-mayorquin you want to push your last change with our patch here? Then I merged. We can add the rhs channel test either now or in a new PR.

neo/rawio/intanrawio.py
zm711: Also fine if you want to do your reorg for this if statement before we merge :)

h-mayorquin: Na, same thing, this is working, I am afraid of working something and I want to be sure we have some sort of support on the next release.

Issue #1656, opened 2025-03-15
SpikeGLXRawIO incorrectly handles folder per probe organization.


grahamfindlay:
**Describe the bug**
SpikeGLX may write data using "folder per probe" organization. Here is the description, from the SpikeGLX User Manual:
>Note that there is an option on the `Save Tab` called `Folder per probe`.
If this is set, there is still a run folder for each g-index `run-name_gN`.
However, inside that there is also a subfolder for each probe that contains
all the t-indices for that g-index and that probe. A probe subfolder is
named like `run-name_gN/run-name_gN_imecM`.

For example:
```
- run_dir/ (example: 3-1-2021/)
  - gate_dir_g0/ (example: 3-1-2021_g0/)
  ...
  - gate_dir_gN/
    - probe_dir_imec0/ (example: 3-1-2021_g0_imec0/)
    ...
    - probe_dir_imecN/
      - trigger_file_gN_imecN_t0.lf.bin
      - trigger_file_gN_imecN_t0.lf.meta
      - trigger_file_gN_imecN_t0.ap.bin
      - trigger_file_gN_imecN_t0.ap.meta
      ...
      - trigger_file_gN_imecN_tN.lf.bin
      - trigger_file_gN_imecN_tN.lf.meta
      - trigger_file_gN_imecN_tN.ap.bin
      - trigger_file_gN_imecN_tN.ap.meta
```

This issue applies when using the folder-per-probe layout and reading probes with a device index greater than 0. For example, trying to read the following file:
```
3-1-2021/3-1-2021_g0/3-1-2021_g0_imec1/3-1-2021_g0_imec1.ap.bin
```
will result in the following error:
```
AssertionError: stream_id imec1.ap is not in [np.str_('imec0.ap'), np.str_('imec0.lf')]
```
It should be checking for membership in `[np.str_('imec1.ap'), np.str_('imec1.lf')]`, as that is the only probe in the directory. 

The incorrect stream IDs parsed for this directory are generated in  `neo.rawio.spikeglxrawio.scan_files()`, on line 405.
``` 
stream_name = f"{device_kind}{device_index}{stream_kind}"
```
This creates the string `{imec}{0}{.ap}` because `device_index` is `0`, when it should be `1`. 

The reason that `device_index` is `0` is because of lines 388-398 in `neo.rawio.spikeglxrawio.scan_files()`:
```
    # TODO: handle one box case
[...]

PR #1661, opened 2025-03-20
Fix spikeglx stream for isolated folder from multi probe case.

h-mayorquin:
Should fix #1661
See discussion here: #1656
Added a test as @zm711 requested.

This is restoring some behavior that we lost after #1608

Plus, reading the metadata history changes here:
https://billkarsh.github.io/SpikeGLX/help/parsing/

Calculating `device_index` through the original file name is robust up till the phase 3A whereas the current `probe_index` calculation might fail for some phases. So I think this is a good improvement overall and simplifies the code.

### DIFF
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -378,6 +378,7 @@ def scan_files(dirname):
 
 
     for info in info_list:
+        # we have two main device types `imec` and `nidq` that the user has control of enabling
         if info.get("device_kind") == "imec":
             info["device_index"] = info["device"].split("imec")[-1]
         else:
--- a/neo/rawio/spikeglxrawio.py
+++ b/neo/rawio/spikeglxrawio.py
@@ -378,13 +378,14 @@ def scan_files(dirname):
 
 
     for info in info_list:
-        # we have two main device types `imec` and `nidq` that the user has control of enabling
+        # device_kind is imec, nidq
         if info.get("device_kind") == "imec":
             info["device_index"] = info["device"].split("imec")[-1]
         else:
             info["device_index"] = ""  # TODO: Handle multi nidq case, maybe use meta["typeNiEnabled"]
 
-    # Define stream base on device [imec|nidq], device_index and stream_kind [ap|lf] for imec
+    # Define stream base on device_kind [imec|nidq], device_index and stream_kind [ap|lf] for imec
+    # Stream format is "{device_kind}{device_index}.{stream_kind}"
     for info in info_list:
         device_kind = info["device_kind"]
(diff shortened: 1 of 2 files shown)


zm711:
@guptadivyansh, so this was a loading regression that occurred. So we also want to fix this. Depending on which PR gets merged first you two might end up with some conflicts. So just want you to take a look at the tweak we are making here as well.

neo/rawio/spikeglxrawio.py
```diff
@@ -376,26 +376,12 @@ def get_segment_tuple(info):
...
-    # TODO: handle one box case
-    info_list_imec = [info for info in info_list if info.get("device") != "nidq"]
-    unique_probe_tuples = {get_probe_tuple(info) for info in info_list_imec}
-    sorted_probe_keys = sorted(unique_probe_tuples)
-    probe_tuple_to_probe_index = {key: idx for idx, key in enumerate(sorted_probe_keys)}
 
     for info in info_list:
-        if info.get("device") == "nidq":
-            info["device_index"] = ""  # TODO: Handle multi nidq case, maybe use meta["typeNiEnabled"]
+        if info.get("device_kind") == "imec":
```
zm711, line 381: why switch back to imec. Just so I can better understand this for the future.

h-mayorquin: I just switch the order of the if here and the diff is not working as it should.

But overall, instead of calculating the stream_index by considering all the possible configurations of probe, dock, slot we are just parsing it from the file name that is written on the metadata.

zm711: Okay I'm just showing off my total ignorance of this IO. I see how you flipped it now.

zm711: maybe a comment saying the options are nidq or imec above the if-else? In case other devs work on this in the future??

h-mayorquin: Feel free to suggest a line that would make it easier for you as you are in a better position to judge what's helpful.

I will revisit this once we are adding support for one box data:

https://www.neuropixels.org/_files/ugd/328966_9d776efdcff74658a3724690610d4b41.pdf

zm711, line 381:
```suggestion
        # we have two main device types `imec` and `nidq` that the user has control of enabling
        if info.get("device_kind") == "imec":
```
maybe something like this?

grahamfindlay:
@zm711 it works for me!

zm711:
Great fixing the CI tomorrow-ish and then we can merge after we retest :)

zm711 (approved):
Two separate users confirm this fixes the regression. Merging.

PR #1663, opened 2025-03-24
Cleanup import to use `importlib.util.find_spec`

zm711:
Worked on this for spikeinterface so also doing this here. As of python 3.8 this is desired way to verify packages. Also as of python 3.10 they completely removed sequence from collections so I think we should be fine to remove that try-except.


h-mayorquin:
+1 on this.

PR #1667, opened 2025-03-25
Fix ci testing-datalad install for python 3.12 + git-annex

zm711:
I noticed our 3.12 testing is just skipping rather than running. I'll track down what is happening here.

### DIFF
--- a/neo/test/rawiotest/common_rawio_test.py
+++ b/neo/test/rawiotest/common_rawio_test.py
@@ -19,6 +19,8 @@ __test__ = False
 
 import logging
 import unittest
+import importlib.util
+import warnings
 
 from neo.utils.datasets import download_dataset, get_local_testing_data_folder, default_testing_repo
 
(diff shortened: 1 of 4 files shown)


zm711:
So it turns out our datalad install was failing but that just led to tests being skipped rather than warning us. I've added a warning for datalad installation failure so we can see it on the  CI. I also added in the appropriate git-annex install based on what Heberto got working for spikeinterface CI. So as soon as tests pass we should be good to go with this.

neo/test/rawiotest/common_rawio_test.py
```diff
@@ -19,19 +19,23 @@
...
-try:
-    import datalad
-
+datalad_spec = importlib.util.find_spec("datalad")
+if datalad_spec is not None:
     HAVE_DATALAD = True
-except:
+else:
     HAVE_DATALAD = False
+    # pytest skip doesn't explain why we are skipping.
```
h-mayorquin, line 36: Consider throwing an error when you run this on the CI:

os.environ.get('GITHUB_ACTIONS') == 'true'

zm711: That's a great idea!

zm711: Done.

alejoe91 (approved):

PR #1668, opened 2025-03-26
Separate sync as its own stream in `OpenEphysBinaryRawIO`

h-mayorquin:
With love for @chrishalcrow 

As discussed with @alejoe91 and @samuelgarcia 

This PR:
* Keeps backward compatibility for `load_sync_trace=True` but throws a deprecation warning (when?)
* Adds the sync channels as their own stream so they can be loaded by downstream libraries. 
* Adds a tests.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -72,6 +72,12 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                 experiment_names = [experiment_names]
         self.experiment_names = experiment_names
         self.load_sync_channel = load_sync_channel
+        if load_sync_channel:
+            warn(
+                "The load_sync_channel=True option is deprecated and will be removed in a future version. "
+                "Use load_sync_channel=False instead, which will add sync channels as separate streams.",
+                DeprecationWarning, stacklevel=2
+            )
         self.folder_structure = None
         self._use_direct_evt_timestamps = None
 
@@ -123,7 +129,7 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
         # signals zone
         # create signals channel map: several channel per stream
         signal_channels = []
-
+        sync_stream_id_to_buffer = {}
         for stream_index, stream_name in enumerate(sig_stream_names):
             # stream_index is the index in vector stream names
             stream_id = str(stream_index)
@@ -141,8 +147,9 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
 
                 # Special cases for stream
                 if "SYNC" in chan_id and not self.load_sync_channel:
-                    # the channel is removed from stream but not the buffer
-                    stream_id = ""
+                    # Every stream sync channel is added as its own stream
+                    stream_id = chan_id
+                    sync_stream_id_to_buffer[stream_id] = buffer_id
 
                 if "ADC" in chan_id:
                     # These are non-neural channels and their stream should be separated
@@ -174,12 +181,18 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
         signal_buffers = []
 
         unique_streams_ids = np.unique(signal_channels["stream_id"])
+        
         for stream_id in unique_streams_ids:
-            # Handle special case of Synch channel having stream_id empty
-            if stream_id == "":
+            
+            # Handle sync channel on a special way
+            if "SYNC" in stream_id:
+                # This is a sync channel and should not be added to the signal streams
+                sync_stream_buffer = sync_stream_id_to_buffer[stream_id]
+                signal_streams.append((stream_id, stream_id, sync_stream_buffer))
                 continue
-            stream_index = int(stream_id)
+            
             # Neural signal
+            stream_index = int(stream_id)
             if stream_index < self._num_of_signal_streams:
                 stream_name = sig_stream_names[stream_index]
                 buffer_id = stream_id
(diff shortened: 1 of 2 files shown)


chrishalcrow:
Hello, great! Just for reference: the main application of this is to be able to let NeuroConv deal with sync channels in a nice way. And so that loading sync channels in spikeinterface doesn't mess up the probe object.

One possible problem: with this PR in using `read_openephys` with `spikeinterface` you'd now have to specify your steam. E.g. if you just do a vanilla

```
rec = si.read_openephys('path/to/recording')
>>> ValueError: This reader have several streams: 
Names: [np.str_('Record Node 109#Neuropix-PXI-103.ProbeA'), np.str_('CH_SYNC')]
IDs: [np.str_('0'), np.str_('CH_SYNC')]. 
Specify it from the options above with the 'stream_name' or 'stream_id' arguments
```

whereas before, there was only one stream, so it just used that. So this will break code for people using openephys with just neural and sync streams (e.g. my lab!). Is there a simple way (on the spikeinterface side?) to retain the current default behavior?

guptadivyansh:
This would also have implications for https://github.com/NeuralEnsemble/python-neo/pull/1657. In that PR I'm extracting sync edges as events, rather than exposing the entire sync stream that is being done here.

I don't have a preference for one or the other but I do think the handling/API should be consistent across SGLX and OpenEphys.

I can make changes to the other SGLX PR if thus extra sync stream has already been agreed upon.

zm711:
I think users of this IO would need to provide more feedback to us. I don't use this so I also don't really know the best solution. I think Sam and Heberto have talked the most about this. So maybe Heberto you could summarize more here why you think the sync should be this way.

h-mayorquin:
> This would also have implications for https://github.com/NeuralEnsemble/python-neo/pull/1657. In that PR I'm extracting sync edges from SpikeGLX recordings as events, rather than exposing the entire sync stream that is being done here.

It is already exposed there as well. Just as an extra channel when `load_sync_channel=True`.

> I don't have a preference for one or the other but I do think the handling/API should be consistent across SGLX and OpenEphys.

I think it is fine if we do both. People can extract the synch channel as the TTL analog stream and also as rising and falling event times if they prefer it that way. 

> whereas before, there was only one stream, so it just used that. So this will break code for people using openephys with just neural and sync streams (e.g. my lab!). Is there a simple way (on the spikeinterface side?) to retain the current default behavior?

That's a good point. We can definitley add code in SpikeInterface to keep this from failing by checking that if there are only two streams and one of them has sync on their name then load the non-sync channel by default.

>

zm711:
I'm going to leave Sam or Alessio to make final decisions on this since they know this IO better than I do. I put in the request so they can get to this and contribute to the decisions.

alejoe91 (approved):

neo/rawio/openephysbinaryrawio.py
```diff
@@ -141,8 +147,9 @@ def _parse_header(self):
 
                 # Special cases for stream
                 if "SYNC" in chan_id and not self.load_sync_channel:
-                    # the channel is removed from stream but not the buffer
-                    stream_id = ""
+                    # Every stream sync channel is added as its own stream
+                    stream_id = chan_id
```
alejoe91, line 151:
```suggestion
                    stream_id = f"{stream_name}#{chan_id}"
```
@h-mayorquin if you have multiple probes, then we probably need the stream name as well because the `chan_id` could be the same, right?

h-mayorquin: Yes, this makes sense. 
Do we have an example like that in our testing suite? I think you are more familiar with what we have there.

h-mayorquin:
Fixed. How should we deprecated the `load_sync_channel` behavior? What is your deprecation policy here? 

Removing that will simplify the code.

alejoe91:
I would keep it for this release and change it in the next one :)

h-mayorquin:
OK. So I will add a deprecation warning for the next release.

h-mayorquin:
Added the deprecation.

h-mayorquin:
I also did the equivalent PR for spikeglx #1683

neo/rawio/openephysbinaryrawio.py
```diff
@@ -174,12 +181,21 @@ def _parse_header(self):
...
+        # And for users to understand
         for stream_id in unique_streams_ids:
-            # Handle special case of Synch channel having stream_id empty
-            if stream_id == "":
+            
+            # Handle sync channel on a special way
+            if "SYNC" in stream_id:
+                # This is a sync channel and should not be added to the signal streams
+                buffer_id = sync_stream_id_to_buffer_id[stream_id]
+                stream_name = stream_id
```
chrishalcrow, line 193: As suggested here (https://github.com/SpikeInterface/spikeinterface/pull/3839#issuecomment-2785810190), we should keep track of the record node that the sync channel came from. Something like this?
```suggestion
                stream_name = sig_stream_names[stream_index].split('#')[0] + '#' + stream_id
```

chrishalcrow: Ok, this works, but there could be a better way
```suggestion
                stream_index = int(stream_id.split('-')[-1])
                stream_name = sig_stream_names[stream_index].split('#')[0] + '#' + stream_id
```

alejoe91: @chrishalcrow this would only keep the record node and not the OE stream name though, right?

h-mayorquin: I changed this already, can you test it @chrishalcrow?

h-mayorquin: Changed this to use `{stream_name}SYNC` for both id and name.

chrishalcrow: Tested - all works for the "real world" cases I'm trying!

alejoe91 (approved):

alejoe91 (approved):

PR #1669, opened 2025-03-26
BlackrockRawIO: Fix int overflow

zm711:
Discovered as a warning in our CI. Should fix just to be safe
```
neo/test/rawiotest/test_blackrockrawio.py::TestBlackrockRawIO::test_read_all
  /home/runner/work/python-neo/python-neo/neo/rawio/blackrockrawio.py:287: RuntimeWarning: overflow encountered in scalar multiply
    _id = f"Unit {1000 * channel_id + unit_id}"
```

channel_id is an uint16 so we convert to a python int. This should prevent overflows.


zm711:
With this PR we no longer have the warning on CI : )

samuelgarcia (approved):

PR #1670, opened 2025-03-26
Drop support for numpy 1.22 and 1.23

zm711:
We follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) + 1 year.

html copied below but basically we are over on 1.22 and we just hit the 1 year point for 1.23

<html>
<body>

Jun 23, 2023 | 3.9+ | 1.22+

-- | -- | --
Jan 01, 2024 | 3.9+ | 1.23+

</body>
</html>


pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 
 dependencies = [
     "packaging",
-    "numpy>=1.22.4",
+    "numpy>=1.24.4",
```
h-mayorquin, line 27: This is gonna break downstream stuff I think. You should qualify this by python versions. But maybe I am wrong.

zm711: We are relatively permissive (although I guess spikeinterface still allows 1.20 which has been deprecated for ages...). We are consistent in our policy that the dreprecations will occur any time after NEP29 + 1 year we just are inconsistent in remembering to apply it right at 1 year. This has been consistent for many versions of numpy. I agree we may need to adjust due to the the python 3.12/3.13 split and the numpy <2, >2 split, I think those struggles are for later versions. We can discuss this as a group in our April meeting and decide if we need to change our policy. That's why I did draft mode so that no one would rush to merge it :)

zm711: @h-mayorquin, we discussed this and we will continue our policy, but if end-users report issues we are happy to be flexible (we would add numpy back in if this breaks an important feature!). We would prefer to move forward and follow our policy since it has been consistent.

h-mayorquin: THanks for double checking. I stand corrected 

That's great, it seems that they are not following their own advice in the NEP.

Let's see how this goes:

image

zm711: I think the issue is we will never know what exact combos people want to use so we will merge this into this release and if people complain excessively (/devs complain Neo is ruining their CIs) then we would be open to adjust. Alessio and Sam agreed that SI basically just hadn't set rules yet so having the fence provided by Neo was fine. But we are happy to also hear from the neuro-conv side since you also are user facing.

zm711:
@alejoe91 / @samuelgarcia / @apdavison this is ready for merge whenever you want to take a look. Feel free to add to my discussion with Heberto if you want to mention anything else.

alejoe91 (approved):

PR #1676, opened 2025-04-01
IntanRawIO: Improve intan logic for "one-file-per-channel" in the rhs format

h-mayorquin:
Should come after #1660

PR #1677, opened 2025-04-01
Add another tests for one file per channel intan rhs

h-mayorquin:
OK, @zm711.

I won't have time to fix the file name on gin this week but at least this way you can merge knowing that there is a test for file ordering in the one file per channel case:

#1676


neo/test/rawiotest/test_intanrawio.py
```diff
@@ -84,11 +85,84 @@ def test_correct_reading_one_file_per_channel(self):
...
         channel_names = [path.name[4:-4] for path in amplifier_file_paths]
 
+        amplifier_stream_index = 0
         for channel_name, amplifier_file_path in zip(channel_names, amplifier_file_paths):
             data_raw = np.fromfile(amplifier_file_path, dtype=np.int16).squeeze()
-            data_from_neo = intan_reader.get_analogsignal_chunk(channel_ids=[channel_name], stream_index=0).squeeze()
+            data_from_neo = intan_reader.get_analogsignal_chunk(channel_ids=[channel_name], stream_index=amplifier_stream_index).squeeze()
             np.testing.assert_allclose(data_raw, data_from_neo)
 
+    def test_correct_reading_one_file_per_channel_rhs_stim(self):
```
h-mayorquin, line 94: Duplication?

h-mayorquin: @zm711

zm711: You're right one of the tests got duplicated. Weird I honestly thought it was a diff issue, but it must have been that the tests got reorganized so it didn't connected that they were the same? Not sure. But yeah we should clean that. I doesn't hurt anything, but it's untidy!

Issue #1678, opened 2025-04-02
The stream names of neuralynx are hard to use and affected by numpy > 2.1 changes


h-mayorquin:
I’m looking at the testing data in the gin repo ("neuralynx" / "Cheetah_v6.4.1dev" / "original_data"), and the stream names are as follows:

![Image](https://github.com/user-attachments/assets/40e91a0b-873b-4c43-905f-b5ef2411fe6e)

I’m not very familiar with this reader, so I’m not sure whether these names can be improved to something more meaningful.

Why is this a problem? Users rely on the stream name to fetch information, and these names are hard to type and remember. More importantly, with recent changes in how NumPy represents scalar values, some previously working code is now broken.

PR #1679, opened 2025-04-02
Remove numpy scalar from stream name representation for neuralynx

h-mayorquin:
This is part of https://github.com/NeuralEnsemble/python-neo/issues/1678 but does not solve it completly.


zm711 (approved):
Just for full explanation in the PR this is due to the fact that numpy has changed how it represents strs such that they appear as `np.str_` which are much harder to parse. Converting to a python values (floats or ints) will allow the header representation of this data to work regardless of the version of numpy.

PR #1683, opened 2025-04-06
Separate sync as its own stream in `SpikeGLXRawIO`

h-mayorquin:
Same as #1668. Should come after it.


samuelgarcia:
I am OK with this.

Maybe we should have `load_sync_channel=None` being the semantic not trigerring warning before the removal of this option.

Do we need a spikeinterface change or not ?

neo/rawio/spikeglxrawio.py
```diff
@@ -109,6 +112,12 @@ def __init__(self, dirname="", load_sync_channel=False, load_channel_location=Fa
         BaseRawWithBufferApiIO.__init__(self)
         self.dirname = dirname
         self.load_sync_channel = load_sync_channel
+        if load_sync_channel:
+            warn(
+                "The load_sync_channel=True option is deprecated and will be removed in version 0.15 \n"
+                "The sync channel is now loaded as a separate stream by default and should be accessed as such. ",
```
zm711, line 118: This is still a little unclear if you're going to still give me the sync channel as the last channel or not during the deprecation period. 
```suggestion
                "The sync channel is now loaded as a separate stream by default and should be accessed as such. ",
```

neo/rawio/spikeglxrawio.py
```diff
@@ -177,25 +198,33 @@ def _parse_header(self):
...
-                    signal_channels = signal_channels[:-1] + [last_chan]
+                    # the last channel is removed from the stream but not from the buffer
                     self._stream_buffer_slice[stream_id] = slice(0, -1)
+                    
+                    # Add a buffer slice for the sync channel
+                    sync_stream_id = f"{stream_name}-SYNC"
+                    self._stream_buffer_slice[sync_stream_id] = slice(-1, None)
+                    
                 if self.load_sync_channel and not info["has_sync_trace"]:
                     raise ValueError("SYNC channel is not present in the recording. " "Set load_sync_channel to False")
```
zm711, line 220: You want to deal with this later?

zm711 (commented):
One question but then this is good by me?

for me I don't see the need of the None mechanism. People have been using this for a while so if they say True we should warn them that this is changing and the default is False so no warning.

neo/rawio/spikeglxrawio.py
h-mayorquin: Yes, we remove this code when we deprecate, right? Or what do you mean?

zm711: I'm wondering if instead we just turn this to a warning and just do nothing rather than raise a value error? Since we will remove this anyway in the future.

h-mayorquin: I rather just remove everything when we remove the argument as that will be two warnings. 

Just to keep everything the way it was until removal.

zm711: Okay. I'm okay with that. I'll ping Sam on Monday and if he tells me he absolutely wants to reread it then I'll pressure him to get it read, and if not then we can merge then. The plexon2 tests are sometimes succeeding so I might just run the suite a few times to try to get to a green check mark :)

zm711 (approved):

samuelgarcia (approved):

h-mayorquin:
Thanks for the review.

PR #1685, opened 2025-04-08
Blackrock make sampling rate a class attribute.

h-mayorquin:
Make this available without the need to parse the header.

https://github.com/SpikeInterface/spikeinterface/pull/3843

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -129,7 +129,7 @@ class BlackrockRawIO(BaseRawIO):
     extensions = ["ns" + str(_) for _ in range(1, 7)]
     extensions.extend(["nev", "sif", "ccf"])  # 'sif', 'ccf' not yet supported
     rawmode = "multi-file"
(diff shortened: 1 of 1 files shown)


neo/rawio/blackrockrawio.py
```diff
@@ -130,6 +130,10 @@ class BlackrockRawIO(BaseRawIO):
     extensions.extend(["nev", "sif", "ccf"])  # 'sif', 'ccf' not yet supported
     rawmode = "multi-file"
 
+    # We need to document the origin of this value
+    main_sampling_rate = 30000.0
```
zm711, line 134: Yeah I'm not sure about this. Black rock has so many versions. I would want a link to docs or something to indicate this is the required. Most software gives you some options.

h-mayorquin: Note that this was already hardcoded. This is only to make it accessible without parsing.

alejoe91:
This is ok for me

PR #1691, opened 2025-04-23
Fix linux datalad-git-annex install

zm711:
kitenet is forbidden access now so we need to switch.


zm711:
Changing the action for IO tests tested coretest but not IO tests. That's hilarious. Let me tweak one io to trigger tests and then I will fix that action in another PR later.

zm711:
Again a thank you to @h-mayorquin for what I'm quite sure will be the solution to our testing needs :)

alejoe91 (approved):

Issue #1693, opened 2025-04-24
Issue opening .rhd files


OlivierPeron:
Hi, 

I was using the function `read_intan()` to read an rhd file (from spikeinterface), with neo version= 0.13, but I have now updated the package and the function doesn't work (version 0.14.1). It seems to look for several intan files, such as the time.dat file, which shouldn't be necessary to read the info.rhd file. Also all stream_index are being loaded, even_tho a `stream_id `or `stream_name` is given. 

I am running :
```
intan_info = se.read_intan(paths['rhd'], stream_id = '0')
```

and I got the following error, due to trying to check all streams :

```
intan_info = se.read_intan(paths['rhd'], stream_id = '0')
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\spikeinterface\extractors\neoextractors\intan.py", line 58, in __init__
    NeoBaseRecordingExtractor.__init__(
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 191, in __init__
    _NeoBaseExtractor.__init__(self, block_index, **neo_kwargs)
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 27, in __init__
    self.neo_reader = self.get_neo_io_reader(self.NeoRawIOClass, **neo_kwargs)
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\spikeinterface\extractors\neoextractors\neobaseextractor.py", line 69, in get_neo_io_reader
    neo_reader.parse_header()
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\neo\rawio\baserawio.py", line 211, in parse_header
    self._parse_header()
  File "C:\Users\DeepLabCut\Anaconda3\envs\spykeline\lib\site-packages\neo\rawio\intanrawio.py", line 297, in _parse_header
    tmp.append(raw_data[0].size)
IndexError: list index out of range
``` 

with :
``` 
{'RHD2000 amplifier channel': [], 'RHD2000 auxiliary input channel': [], 'USB board digital input channel': [], 'timestamp': [memmap([       0,        1,        2, ..., 12559741, 12559742, 12559743])]}
```

zm711:
[...]

PR #1694, opened 2025-04-24
Better error intan when missing files in the "one-file-per-signal" type.

h-mayorquin:
From the discussion on #1693, @zm711 

image

### DIFF
--- a/neo/rawio/intanrawio.py
+++ b/neo/rawio/intanrawio.py
@@ -1363,14 +1363,53 @@ def create_one_file_per_signal_dict_rhd(dirname):
     -------
     raw_files_paths_dict: dict
         A dict of all the file paths
+
+    Raises
+    ------
+    NeoReadWriteError
+        If any required files are missing
     """
 
     raw_file_paths_dict = {}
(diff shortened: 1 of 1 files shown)


h-mayorquin:
Maybe there are some missing ones that are acceptable. But I think we can start here and be explicit which is better.

zm711:
Look at line 1185 for the check I'm referring to. We see that the signal is enable. Then we could use that information for the formats I think. Maybe you could try an implementation with that instead?

h-mayorquin:
> For one-file-per-signal that are certain combination that have to show up together but you could shut off dig in /dig out, aux channels. So I think don't think we have a realistic core. 

Yeah, this is the case that I want to cover. one-file-per-channel seems too complicated at the moment but might do it later.

>  I think we should pull the info from the header and whatever the header says is enabled we raise an error no? We do that when we read the header. Most things have a boolean check to see if that signal is enabled?

Yes, this is the way. Thanks.

h-mayorquin:
Thanks for the reminder. Most likely will need to wait for the next version.

PR #1697, opened 2025-05-02
Expand heka stimulus reconstruction option

easy-electrophysiology:
Allow 'False' to be passed down such that no stimulus is loaded.


easy-electrophysiology:
I've done it again, sorry! I should just stop been lazy and acutally open a PR for the HEKA loader 😅

zm711:
Always nice to see a headfake about us getting access to the HEKA loader :)

Issue #1696, opened 2025-05-01
ZeroDivisionError - Unable to read a Micromed file


theairbend3r:
**Describe the bug**
Get a `ZeroDivisionError` when trying to read a micromed `.trc` file via `MicromedIO(file_path).read()`. The same file can be read with Wonambi (another python package) as `wonambi.Dataset(file_path)`.

**To Reproduce**
The error traceback is as follows.

```
Traceback (most recent call last):
  File ".venv/lib/python3.11/site-packages/marimo/_runtime/executor.py", line 141, in execute_cell
    exec(cell.body, glbls)
    ....

    trc = MicromedIO(file_path).read()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/neo/io/micromedio.py", line 13, in __init__
    BaseFromRaw.__init__(self, filename)
  File ".venv/lib/python3.11/site-packages/neo/io/basefromrawio.py", line 77, in __init__
    self.parse_header()
  File ".venv/lib/python3.11/site-packages/neo/rawio/baserawio.py", line 211, in parse_header
    self._parse_header()
  File ".venv/lib/python3.11/site-packages/neo/rawio/micromedrawio.py", line 176, in _parse_header
    self._t_starts.append(seg_start / self._sampling_rate)
                          ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
ZeroDivisionError: float division by zero
```

**Expected behaviour**
The file should have been read. Based on the error, the `ZeroDivisionError` maybe caused by `self._sampling_rate` being 0 when it is read in the [MicromedRawIO](https://github.com/NeuralEnsemble/python-neo/blob/1dfda7a9ded08653fe5c1a02020edfc71ecf984d/neo/rawio/micromedrawio.py#L176) file as `self._sampling_rate = float(np.unique(signal_channels["sampling_rate"])[0])`.

But, I am able to access the sampling frequency in Wonambi as follows (it comes out to be 512 btw). 
```
trc = wonambi.Dataset(file_path)
trc.header["s_freq"]
```

**Environment:**
 - OS: Linux
 - Python 3.11
 - Neo 0.15.0.dev0 (the latest main branch)
 - NumPy 2.2.5

zm711:
Two questions @theairbend3r:

does this happen for multiple files or just this one file? 
[...]

PR #1701, opened 2025-05-07
Update micromed for samping issue

zm711:
See #1696.

Testing to see if there is something wrong with class based approach @samuelgarcia maybe we could discuss this.

### DIFF
--- a/neo/rawio/micromedrawio.py
+++ b/neo/rawio/micromedrawio.py
@@ -146,8 +146,13 @@ class MicromedRawIO(BaseRawWithBufferApiIO):
                 gain = factor
                 offset = -logical_ground * factor
 
-                f.seek(8, 1)
-                (sampling_rate,) = f.read_f("H")
+                # this skips the filtering info done with the machine
+                f.seek(8, 1) 
+                
+                #(sampling_rate,) = f.read_f("H")
+                # sampling_rate is actually the rate_coefficient which is multipled by the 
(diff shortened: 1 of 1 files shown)


zm711:
@theairbend3r
could you test this PR? I just want to see what happens if we try removing the class approach we used before (ie your suggestion and what wonambi was doing). Our tests are passing for micromed so this could be a fix we make if it works for you :)

theairbend3r:
Tried this out but I still get the same ZeroDivisionError unfortunately

zm711:
Okay, I was hoping this would be the easy fix, but when I have time I'll try to read wonami more to see if they are doing some header parsing that we skip that could better account for different versions of the code.

theairbend3r:
Sounds good! Thanks for all your efforts, really appreciate it! :)

JuliaSprenger:
@zm711 Potentially this request will provide test files for debugging here: https://github.com/wonambi-python/wonambi/issues/42

zm711:
@theairbend3r,

just wanted to give you a small update. We discussed this at our maintenance meeting and our hope is to try to work on this for the next version of neo. As you can see from the link we put out the feelers to wonambi to see if they might have a test file that can reproduce the error so we can fix it.

theairbend3r:
Appreciate the update @zm711! Hopefully, the test files help! 🤞

PR #1708, opened 2025-05-14
Fix time blackrock

luiztauffer:
a quick fix for the rec_datetime

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -1808,7 +1808,7 @@ class BlackrockRawIO(BaseRawIO):
                 hour=self.__nev_basic_header["hour"],
                 minute=self.__nev_basic_header["minute"],
                 second=self.__nev_basic_header["second"],
-                microsecond=self.__nev_basic_header["millisecond"],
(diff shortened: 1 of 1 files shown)


luiztauffer:
@zm711 @h-mayorquin found this other error, small but important. We were getting the wrong start recording time (in the miliseconds scale)

zm711:
@luiztauffer could you post an image just of the times before and after? Since I don't know the format it is hard for me to know when to do the factor of 1000 or not?

h-mayorquin:
@zm711 

image

Here is the spec for the timestamps where you can see that the last two bytes are milliseconds as per the windows convention and we read here:

https://github.com/NeuralEnsemble/python-neo/blob/5df340368f264be757719ced01b3f96a49f3506d/neo/rawio/blackrockrawio.py#L1131-L1161

And you can see that the datetime object requieres microseconds here:

https://docs.python.org/3/library/datetime.html#datetime.datetime

h-mayorquin (approved):
LGTM

zm711 (approved):
Cool thanks I didn't realize datetime needed the microseconds. Sorry read too fast :)

PR #1709, opened 2025-05-14
Update IO tests for Python 3.13

zm711:
I think we should really test python 3.13 since 3.14 is almost out. We still need to test the oldest python with 1.26.

Or we add in other tests to be safe.

### DIFF
--- a/.github/workflows/io-test.yml
+++ b/.github/workflows/io-test.yml
@@ -18,8 +18,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.12']
-        numpy-version: ['1.26', '2.0']
+        python-version: ['3.9', '3.13']
+        numpy-version: ['1.26', '2.2']
+        exclude:
+           - python-version: '3.9'
+             numpy-version: '2.2'
+           - python-version: '3.13'
+             numpy-version: '1.26'
     defaults:
       # by default run in bash mode (required for conda usage)
       run:


.github/workflows/io-test.yml
```diff
@@ -18,8 +18,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.12']
-        numpy-version: ['1.26', '2.0']
+        python-version: ['3.9', '3.13']
+        numpy-version: ['1.26', '2.2']
```
zm711, line 22:
```suggestion
        numpy-version: ['1.26', '2.1']
```

zm711:
numpy 2.2 failed. So we will try numpy 2.1

zm711:
Failed on 2.1. I guess we keep the ci as is for now :)

.github/workflows/io-test.yml
```diff
@@ -18,8 +18,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ['3.9', '3.12']
-        numpy-version: ['1.26', '2.0']
+        python-version: ['3.9', '3.13']
+        numpy-version: ['1.26', '2.1']
```
zm711, line 22:
```suggestion
        numpy-version: ['1.26', '2.3.1']
```

.github/workflows/io-test.yml
```diff
@@ -18,8 +18,13 @@ jobs:
...
+        exclude:
+           - python-version: '3.9'
+             numpy-version: '2.1'
```
zm711, line 25:
```suggestion
             numpy-version: '2.3.1'
```

zm711:
Double check Windows testing?

zm711:
Add extra numpy version for core.

zm711:
> Double check Windows testing?

Actually after reading we don't test IOs on Windows yet. So that should be a future PR.

> Add extra numpy version for core.

Added.

Issue #1707, opened 2025-05-14
Plexon2 tests failing again with greater than 25 attempts to open files


zm711:
**Describe the bug**
```
FAILED neo/test/iotest/test_plexon2io.py::TestPlexon2RawIO::test_read_all - OSError: Opening /home/runner/ephy_testing_data/plexon/NC16FPSPKEVT_1m.pl2 failed after 25 attempts.
FAILED neo/test/iotest/test_plexon2io.py::TestPlexon2IO::test_assert_read_neo_object_is_compliant - OSError: Opening /home/runner/ephy_testing_data/plexon/NC16FPSPKEVT_1m.pl2 failed after 25 attempts.
FAILED neo/test/iotest/test_plexon2io.py::TestPlexon2IO::test_create_group_across_segment - OSError: Opening /home/runner/ephy_testing_data/plexon/NC16FPSPKEVT_1m.pl2 failed after 25 attempts.
FAILED neo/test/iotest/test_plexon2io.py::TestPlexon2IO::test_read_with_lazy_is_compliant - OSError: Opening /home/runner/ephy_testing_data/plexon/NC16FPSPKEVT_1m.pl2 failed after 25 attempts.
```
copied since the action log will be deleted eventually.
https://github.com/NeuralEnsemble/python-neo/actions/runs/15005917152/job/42207865879
log link for now.

**Expected behaviour**
For our CI to pass :P 

**Environment:**
 - OS: Linux + Wine
 - Python version 3.9 and 3.12
 - Neo version: main
 - NumPy version: 1.26 and 2.0

**Additional context**

This is a know flakiness of running plexon2 through wine on linux. We don't know why but the file fails to open. Normally it works within 25 tries, but suddenly 25 tries is not enough. We can wait a little bit of time and then try increasing the attempts to see if 35 is better, but we don't want the test suite to run forever since the longest part of our suite is installing wine anyway.

@h-mayorquin any thoughts?

h-mayorquin:
I have no idea. It has not failed for us in a while but it is a really flaky test : /

zm711:
Seems like it worked based on Luiz's most recent PR. Who knows.....

h-mayorquin:
Mmm yeah, how strange that is consistently failing for you guys here : /

zm711:
Just to follow up you're using wine? And you're just running our test suite to get it to pass? Or are you running plexon specific testing and can get it to pass?

zm711:
[...]

PR #1710, opened 2025-05-14
Resolve plexon2rawio symlink if necessary

cheydrick:
This resolves (for me) the "...failed after 25 attempts" problem when running tests. It was recently noted in #1707.


zm711:
Seems like for the file that you all contributed is still misbehaving....  I guess we could try with `strict=True` to see if the symlink is not resolving?

neo/rawio/plexon2rawio/plexon2rawio.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, filename, pl2_dll_file_path=None, reading_attempts=25):
 
         # note that this filename is used in self._source_name
         self.filename = pathlib.Path(filename)
+        if (self.filename.is_symlink()):
```
zm711, line 106:
```suggestion
        if self.filename.is_symlink():
```
I think this is an okay change regardless so just one code style tweak. Python typically won't do the () around boolean expressions :)

PR #1711, opened 2025-05-14
NeuroNexusIO: Add in appropriate streams to go with buffer API

zm711:
I first implemented this before the buffer api and since this is one buffer for all streams we left them all packed together. Now that we have the buffer api we should divide these streams into
ephys data (called analog in allego docs)
din
dout
aux

~~No rush on this so we can wait until we figure out plexon2 tests.~~

### DIFF
--- a/neo/rawio/neuronexusrawio.py
+++ b/neo/rawio/neuronexusrawio.py
@@ -132,10 +132,11 @@ class NeuroNexusRawIO(BaseRawWithBufferApiIO):
         self._n_samples, self._n_channels = self.metadata["status"]["shape"]
         # Stored as a simple float32 binary file
         BINARY_DTYPE = "float32"
+        TIMESTAMP_DTYPE = "int64" # this is from the allego sample reader timestamps are np.int64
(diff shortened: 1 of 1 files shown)


neo/rawio/neuronexusrawio.py
```diff
@@ -297,4 +300,9 @@ def _get_analogsignal_buffer_description(self, block_index, seg_index, buffer_id
 # this is pretty useless right now, but I think after a
 # refactor with sub streams we could adapt this for the sub-streams
 # so let's leave this here for now :)
-stream_id_to_stream_name = {"0": "Neuronexus Allego Data"}
+stream_id_to_stream_name = {
+    "0": "Neuronexus Allego Analog (pri) Data",
```
zm711, line 304:
```suggestion
    "0": "NeuroNexus Allego Analog (pri) Data",
```

samuelgarcia:
OK for me.
Don't you want to keep the inner stream  ids instead of numbers ("di", "do", "ai"...) ?

PR #1727, opened 2025-06-06
Fix integer overflow error

### LINKED ISSUE #1721: Difference in channel names: python's neo vs matlab's fieldtrip
**Describe the bug**
Difference in channel names when the same file is loaded in `python/neo` vs `matlab/fieldtrip`.

**To Reproduce**

In python + neo:
```
reader = neo.io.MicromedIO("path/to/file.trc")
reader.parse_header()
reader.header["signal_channels"]
```

In matlab + fieldtrip:
```
addpath(path/to/fieldtrip/toolbox)
cfg = [];
cfg.filename = "path/to/file.trc";
output = ft_read_header(cfg.filename);
output.label
```

**Expected behaviour**
Both ways should output the same channel names. Everything between neo and fieldtrip matches except the last few channel names. Neo repeats existing channel names again while fieldtrip has new channel names. 

Last few channels example:
```
matlab/fieldtrip

..
   {'121+' }
    {'122+' }
    {'123+' }
    {'emg2+'}
    {'emg1+'}
    {'orb+' }
    {'ECG+' }
    {'AH+'  }
```

```
python/neo repeats C24 - C31 again.

..
"C24"
"C25"
"C26"
"C27"
"C28"
"C29"
"C30"
"C31"
```

This creates an issue because if, for e.g., I want to use the channel `C24`, I cannot do it because I don't know which of the two is the real channel `C24`.

**Environment:**
 - OS: Linux
 - Python version: 3.11
 - Neo version: latest github >= 0.14.1
 - NumPy version: >=2.2.6

**Additional context**
Here is `fieldtrip's` relevant file to read micromed .trc data: [read_micromed_trc.m](https://github.com/fieldtrip/fieldtrip/blob/master/fileio/private/read_micromed_trc.m).

In neo's [docs](https://neo.readthedocs.io/en/latest/rawio.html), it says channel_names may not be unique. 

> There are 3 ways to select a subset of channels: by index (0 based), by id or by name. By index is unambiguous 0 to n-1 (inclusive), whereas for some IOs channel_names (and sometimes channel_ids) are not guaranteed to be unique. In such cases, using names or ids may raise an error.

Do you know why this is the case? and why in matlab the channel_names don't repeat for the same file?

theairbend3r:
Force `code[c]` which is currently a `uint16 (or u2)` into a `int` to prevent integer overflow for the operation --  `f.seek(pos + int(code[c]) * 128 + 2, 0)`.

Fixes #1721

### DIFF
--- a/neo/rawio/micromedrawio.py
+++ b/neo/rawio/micromedrawio.py
@@ -51,7 +51,6 @@ class MicromedRawIO(BaseRawWithBufferApiIO):
         self.filename = filename
 
     def _parse_header(self):
-
         with open(self.filename, "rb") as fid:
             f = StructFile(fid)
 
@@ -133,7 +132,11 @@ class MicromedRawIO(BaseRawWithBufferApiIO):
             sig_grounds = []
             for c in range(Num_Chan):
                 zname2, pos, length = zones["LABCOD"]
-                f.seek(pos + code[c] * 128 + 2, 0)
+                # Force code[c] which is currently a uint16 (or u2) into a int to prevent integer overflow
+                # for the following operation -- code[c] * 128 + 2.
+                # An integer overflow below may have side - effects including but not limited
+                # to having repeated channel names.
+                f.seek(pos + int(code[c]) * 128 + 2, 0)
 
                 chan_name = f.read(6).strip(b"\x00").decode("ascii")
                 ground = f.read(6).strip(b"\x00").decode("ascii")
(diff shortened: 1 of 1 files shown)


neo/rawio/micromedrawio.py
```diff
@@ -133,7 +132,11 @@ def _parse_header(self):
             sig_grounds = []
             for c in range(Num_Chan):
                 zname2, pos, length = zones["LABCOD"]
-                f.seek(pos + code[c] * 128 + 2, 0)
+                # Force code[c] which is currently a uint16 (or u2) into a int to prevent integer overflow
+                # for the following operation -- code[c] * 128 + 2.
+                # An integer overflow below may have side - effects including but not limited
+                # to having repeated channel names.
+                f.seek(pos + int(code[c]) * 128 + 2, 0)
```
zm711, line 139: Great comment. I've been meaning to fix this overflow too! Let's make sure tests pass then we can merge :)

zm711:
Once we merge if you'd like to be on our author list could you let me know the name you want to appear and your affiliation.

PR #1731, opened 2025-06-16
NeuralynxIO: Error from no closing time, turned into a warning

eduardosand:
I have some recordings that were spliced from a larger recording. AS such, there is no timeclosed parameter or anything. I don't need to do downstream tasks, so it'd be nice to change the raise error for not having that parameter in the header, to something more like a warning.

### DIFF
--- a/neo/rawio/neuralynxrawio/nlxheader.py
+++ b/neo/rawio/neuralynxrawio/nlxheader.py
@@ -3,6 +3,7 @@ from packaging.version import Version
 import os
 import re
 from collections import OrderedDict
+import warnings
 
 
 class NlxHeader(OrderedDict):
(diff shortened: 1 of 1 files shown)


PeterNSteinmetz:
Yes, it is fine to not have a closing time. Indeed, most of the early versions of the neuralynx files did not have this field in the header at all.

samuelgarcia:
Hi I am OK to not raise an error for closing but I think the new refactoring of @PeterNSteinmetz already has this behavior.
See https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/neuralynxrawio/nlxheader.py#L335
In short I think this PR is not relevant anymore.

@eduardosand can you test it ? And so we can close this PR.

eduardosand:
Excuse the delay. I tested the most recent version on PyPI (0.14.2) and it works. Thanks!

PR #1735, opened 2025-06-26
Adopt the NumFOCUS Code of Conduct

JuliaSprenger:
As decided in the last Neo Core Meeting, this MR replaces the current CoC with the NumFOCUS CoC. As recommended by NumFOCUS the MR does not copy the CoC, but rather links to the NumFOCUS document.

Furthermore this implies that Neo is now employing the NumFOCUS workflow for reporting CoC violations and NumFOCUS will advise Neo maintainers on how to proceed in case of a CoC violation.

The CoC also applies for events organized by the Neo Team, where the CoC will be introduced in the beginning of the event and a dedicated CoC contact person will be assigned.

Once this MR is merged I we will also announce the CoC change on the community mailing list.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,5 +1,7 @@
 # [The NumFOCUS Code of Conduct](https://numfocus.org/code-of-conduct)
 
+Neo subscribes to the NumFOCUS Code of Conduct.
+
 ## The Short Version
 Be kind to others. Do not insult or put down others. Behave professionally. Remember that harassment and sexist, racist, or exclusionary jokes are not appropriate for NumFOCUS.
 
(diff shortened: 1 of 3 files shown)


apdavison:
Hi Julia,

Thanks for starting this. I think this needs more context, I don't think we yet mention in the docs that we're a NumFocus affiliated project, so people might wonder why the CoC talks about NumFocus but not Neo. Could you update the text in `doc/source/governance.rst`, and perhaps also add a paragraph at the start of `doc/source/contributing.rst`?

zm711:
Just a reminder we are doing the next release Monday or Tuesday of next week. So if we want the new code of conduct we should try to get that done today/tomorrow ideally or we can discuss on Monday.

apdavison:
My apologies for committing directly to your branch Julia, but I can't attend the maintainers' meeting on Monday, so I wanted to get these changes in before the release.

PR #1738, opened 2025-06-30
Update for MaxWell systems

cmccrimmon:
MaxWell channel-electrode mappings may not be bijective (see https://github.com/NeuralEnsemble/python-neo/pull/1703). Throw a warning rather than an error if this is the case and address this correctly downstream in spikeinterface modules.


samuelgarcia:
Hi.
Unfortunatly, this is not possible, we need to ensure that the ids are unique.
What we can do is to use the channel name in maxwell (which can be none unique).
And also we could have a hook in maxwell channel_ids generation a check of unicity and in case it is not unique make a transformation of the id to make it unique (like adding the index of the channel)
What do you think ?

h-mayorquin:
Using names should be OK, right? Then people can use the "use_names_as_ids" in SpikeInterface 

https://github.com/SpikeInterface/spikeinterface/blob/fb72ed2af963e257a2a12b1d4280f622c259c7c3/src/spikeinterface/extractors/neoextractors/maxwell.py#L33-L35

They names are also propagated as a property that can be used in SpikeInterface to fix things downstream.

cmccrimmon:
> Using names should be OK, right? Then people can use the "use_names_as_ids" in SpikeInterface
> 
> https://github.com/SpikeInterface/spikeinterface/blob/fb72ed2af963e257a2a12b1d4280f622c259c7c3/src/spikeinterface/extractors/neoextractors/maxwell.py#L33-L35
> 
> They names are also propagated as a property that can be used in SpikeInterface to fix things downstream.

Yes, using names should be fine as these are propagated and can be subsequently dealt in SpikeInterface. A singular enforcement of uniqueness should be implemented here then (throw out any duplicated names after the first) as required.

PR #1740, opened 2025-07-02
 Blackrock add block validation

h-mayorquin:
This is a PR I made when working together @luiztauffer to figure out how to open a malformed file. According to the blackrock spec heach data block has a flag indicating the start of the data block:

image

This PR adds a check and provides detailed debugging information including file offset, block index, expected vs actual values, and corruption indicators.

I also improved the  missing file error message which now specifically list which NSX files are missing and did other minor improvements like using f-string formatting for filename
construction, clearer variable naming (replacing ambiguous spec with spec_version and
dh with packet_header), etc.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -383,7 +383,7 @@ class BlackrockRawIO(BaseRawIO):
             for nsx_nb in self.nsx_to_load:
                 self.__match_nsx_and_nev_segment_ids(nsx_nb)
 
-        self.nsx_data = {}
+        self.nsx_datas = {}
         self.sig_sampling_rates = {}
         if len(self.nsx_to_load) > 0:
             for nsx_nb in self.nsx_to_load:
@@ -399,7 +399,7 @@ class BlackrockRawIO(BaseRawIO):
                     _data_reader_fun = self.__nsx_data_reader["3.0-ptp"]
                 else:
                     _data_reader_fun = self.__nsx_data_reader[spec_version]
-                self.nsx_data[nsx_nb] = _data_reader_fun(nsx_nb)
+                self.nsx_datas[nsx_nb] = _data_reader_fun(nsx_nb)
 
                 sr = float(self.main_sampling_rate / basic_header["period"])
                 self.sig_sampling_rates[nsx_nb] = sr
@@ -446,7 +446,7 @@ class BlackrockRawIO(BaseRawIO):
                     signal_channels.append((ch_name, ch_id, sr, sig_dtype, units, gain, offset, stream_id, buffer_id))
 
             # check nb segment per nsx
-            nb_segments_for_nsx = [len(self.nsx_data[nsx_nb]) for nsx_nb in self.nsx_to_load]
+            nb_segments_for_nsx = [len(self.nsx_datas[nsx_nb]) for nsx_nb in self.nsx_to_load]
             if not all(nb == nb_segments_for_nsx[0] for nb in nb_segments_for_nsx):
                 raise NeoReadWriteError("Segment nb not consistent across nsX files")
             self._nb_segment = nb_segments_for_nsx[0]
@@ -468,7 +468,7 @@ class BlackrockRawIO(BaseRawIO):
                         ts_res = 30_000
                     period = self.__nsx_basic_header[nsx_nb]["period"]
                     sec_per_samp = period / 30_000  # Maybe 30_000 should be ['sample_resolution']
-                    length = self.nsx_data[nsx_nb][data_bl].shape[0]
+                    length = self.nsx_datas[nsx_nb][data_bl].shape[0]
                     if self.__nsx_data_header[nsx_nb] is None:
                         t_start = 0.0
                         t_stop = max(t_stop, length / self.sig_sampling_rates[nsx_nb])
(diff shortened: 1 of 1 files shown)


neo/rawio/blackrockrawio.py
```diff
@@ -930,28 +937,32 @@ def __read_nsx_header_variant_b(self, nsx_nb):
             # filter settings used to create nsx from source signal
             ("hi_freq_corner", "uint32"),
             ("hi_freq_order", "uint32"),
-            ("hi_freq_type", "uint16"),  # 0=None, 1=Butterworth
+            ("hi_freq_type", "uint16"),  # 0=None, 1=Butterworth, 2=Chebyshev
             ("lo_freq_corner", "uint32"),
             ("lo_freq_order", "uint32"),
             ("lo_freq_type", "uint16"),
-        ]  # 0=None, 1=Butterworth
+        ]  # 0=None, 1=Butterworth, -2-Chebyshev
```
zm711, line 944:
```suggestion
        ]  # 0=None, 1=Butterworth, 2=Chebyshev
```
? or is it negative 2

zm711 (commented):
This seems fine to me.

neo/rawio/blackrockrawio.py
h-mayorquin: No, that was a typo.

zm711 (approved):
Good by me. I'll wait to see if @samuelgarcia or @alejoe91  want to take a quick look too.

neo/rawio/blackrockrawio.py
```diff
@@ -323,19 +323,21 @@ def _parse_header(self):
...
         for nsx_nb in self._avail_nsx:
-            spec = self.__nsx_spec[nsx_nb] = self.__extract_nsx_file_spec(nsx_nb)
+            spec_version = self.__nsx_spec[nsx_nb] = self.__extract_nsx_file_spec(nsx_nb)
             # read nsx headers
-            self.__nsx_basic_header[nsx_nb], self.__nsx_ext_header[nsx_nb] = self.__nsx_header_reader[spec](nsx_nb)
+            nsx_header_reader = self.__nsx_header_reader[spec_version]
+            self.__nsx_basic_header[nsx_nb], self.__nsx_ext_header[nsx_nb] = nsx_header_reader(nsx_nb)
 
-            # The only way to know if it is the PTP-variant of file spec 3.0
+            # The only way to know if it is the peak-to-peak-variant of file spec 3.0
```
cboulay, line 331: PTP is not peak-to-peak!
https://en.wikipedia.org/wiki/Precision_Time_Protocol

neo/rawio/blackrockrawio.py
```diff
@@ -381,27 +387,29 @@ def _parse_header(self):
         self.sig_sampling_rates = {}
         if len(self.nsx_to_load) > 0:
             for nsx_nb in self.nsx_to_load:
-                spec = self.__nsx_spec[nsx_nb]
-                # The only way to know if it is the PTP-variant of file spec 3.0
+                basic_header = self.__nsx_basic_header[nsx_nb]
+                spec_version = self.__nsx_spec[nsx_nb]
+                # The only way to know if it is the peak-to-peak-variant of file spec 3.0
```
cboulay, line 392: Same. Precision Time Protocol, not peak-to-peak.

neo/rawio/blackrockrawio.py
```diff
@@ -971,32 +982,44 @@ def __read_nsx_dataheader_variant_b(
...
+        filename = f"{self._filenames['nsx']}.ns{nsx_nb}"
 
         filesize = self.__get_file_size(filename)
 
         data_header = {}
-        index = 0
 
-        if offset is None:
-            offset = self.__nsx_basic_header[nsx_nb]["bytes_in_headers"]
+        offset_to_first_data_block = offset or int(self.__nsx_basic_header[nsx_nb]["bytes_in_headers"])
```
cboulay, line 991: I think this will still evaluate to the second condition even if someone explicitly wanted an offset of 0.
If you really want a one-liner, it would be `offset if offset is not None else ...`

cboulay: Also the input offset needs to be converted to `int`

h-mayorquin: Ah, thanks, indeed. That clarifies it, I did not read well above.

h-mayorquin: Fixed.

cboulay (approved):
Concerns addressed. Thanks.

zm711:
Just to recap. Sam at the maintenance meeting said that he would prefer to ensure all file variants are handled since this PR would be at a release boundary. So we will confirm the header flag for other versions and incorporate that (or since this is put off we can discuss merging later this week--as long as we have a plan for the other variants).

h-mayorquin:
> Just to recap. Sam at the maintenance meeting said that he would prefer to ensure all file variants are handled since this PR would be at a release boundary. So we will confirm the header flag for other versions and incorporate that (or since this is put off we can discuss merging later this week--as long as we have a plan for the other variants).

This request does not make sense. I just check and the other variant is empty:

https://github.com/h-mayorquin/python-neo/blob/2240a29c4366fb3ed5dacadb54837004651ee35e/neo/rawio/blackrockrawio.py#L967-L973

No validation can be performed there.

zm711 (approved):
@samuelgarcia I'll leave the merge to you. But if the other variants don't matter this should be ready finally.

samuelgarcia:
So for the delay.
Merging now

Issue #1743, opened 2025-07-04
Wrong offset when using synthetic noise for BioCAM's event-based compression


b-grimaud:
Since it's generally assumed that the recording will be converted from unsigned to signed with a 12 bit depth downstream, the baseline of raw signal should read at 2048.

Assuming `BiocamIO` is imported and `data_path` points to a compress .brw file : 

```
reader = BiocamIO(data_path, fill_gaps_strategy="zeros")
(reader.get_analogsignal_chunk(block_index=0, seg_index=0,
                                       i_start=1024, i_stop=4096,
                                       channel_indexes=[45])
                                       ).mean()
```
returns `2048.0`.

```
reader = BiocamIO(data_path, fill_gaps_strategy="synthetic_noise")
(reader.get_analogsignal_chunk(block_index=0, seg_index=0,
                                       i_start=1024, i_stop=4096,
                                       channel_indexes=[45])
                                       ).mean()
```
returns `0.0`.

If I "simulate" the 12-bits conversion by substracting 2048 : 

```
(reader.get_analogsignal_chunk(block_index=0, seg_index=0,
                                       i_start=1024, i_stop=4096,
                                       channel_indexes=[45])-2048
                                       ).mean()
```

`zeros` returns the correct `0.0` but `synthetic_noise` returns `63488.0`, presumably because the array is initialized as 16 bits ?

Doing the same on a SpikeInterface recording is pretty similar : 

```
recording = se.read_biocam(data_path,
                           mea_pitch=60,
                           electrode_width=21,
                           fill_gaps_strategy="zeros",
                           ) 
recording = spre.unsigned_to_signed(recording, bit_depth=12)
recording.get_traces(start_frame=1, end_frame=1000, channel_ids=['45']).mean()
```
returns `0.0`.

Using `synthetic_noise` returns `-2048.0`.

The implementation is largely lifted from 3Brain's own docs, not sure how they handle it internally but I assume the bit conversion is done before noise simulation ?

A very simple fix is to add a flat 2048 to `generate_synthetic_noise` in `biocamrawio.py`, but there might be a more elegant solution.

@alejoe91 @mahlzahn

PR #1746, opened 2025-07-07
Fix offset when using synthetic noise with BioCAM's sparse compression

b-grimaud:
Simple fix for #1743


zm711 (commented):
Thanks @b-grimaud. I think this makes sense since we need the offset. We are planning a release in the next couple days, so we will need to discuss this before merge. It may or may not make it in this release, but either way I think it would be more helpful if you put a comment in explaining why we need the offset of 2048 and you can even put in your github issue. something like
```python
# offset of 2048 necessary b/c ....
corrected_data = data + 2048
return corrected_data
```

b-grimaud:
I added some context, hope it's clear enough !

zm711:
> I added some context, hope it's clear enough !

That looks perfect to me. Like I said Alessio will have final say after review, but the logic makes sense to me.

alejoe91 (approved):

PR #1753, opened 2025-07-28
OpenEphysBinary: do not split ADC streams for OneBox

alejoe91:
Fixes ##1752

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -159,11 +159,14 @@ class OpenEphysBinaryRawIO(BaseRawWithBufferApiIO):
                     # We then set the stream_id to the sync stream id
                     channel_stream_id = sync_stream_id
 
-                if "ADC" in chan_id:
-                    # These are non-neural channels and their stream should be separated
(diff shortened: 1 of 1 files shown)


zm711:
To me it looks like we have a spikeglx version of data but not an openephys version. So I would prefer 
1) add data of this format so we actually test
2) we write up a note about data structure

But I think it is fine if you want to review @h-mayorquin. If you have an example at neuroconv you can still test the function without the official test.

alejoe91:
@zm711 @h-mayorquin can any of you upload the data on GIN? I'm running into the usual git-annex version issue...

The data from @jialeren2024 is here: https://github.com/SpikeInterface/spikeinterface/issues/4083#issue-3261423960

zm711:
@apdavison any word on ebrains? It seems like docs are failing because we are timing out with our connection to download the ebrains dataset?

zm711 (approved):

zm711:
I'll merge after tests pass!

h-mayorquin (commented):
JUst one question that I did not send by mistake yesterday.

PR #1754, opened 2025-07-29
Remove name mangling in blackrock

### LINKED ISSUE #1742: Remove name mangling on blackrock
Blackrockrawio uses the mangling pattern liberally:

https://github.com/NeuralEnsemble/python-neo/blob/1dad7e26ade90b071124f088c172225be23baebd/neo/rawio/blackrockrawio.py#L844-L846

There are many downsides to name mangling and the few advantages I know don't apply strongly here. 

Would you guys be OK with removing name mangling in blackrock? @samuelgarcia 

This will make debugging way easier when working with the format.

h-mayorquin:
Should close #1742

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -187,68 +187,68 @@ class BlackrockRawIO(BaseRawIO):
         # These dictionaries are used internally to map the file specification
         # revision of the nsx and nev files to one of the reading routines
         # NSX
-        self.__nsx_header_reader = {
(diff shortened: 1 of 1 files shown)


zm711 (approved):
Looks good to me and tests are passing. I say we leave this up for tomorrow to see if @alejoe91 or @samuelgarcia want to read, but otherwise we can merge then.

h-mayorquin:
Sam is on vacations but he already agreed with this, can we merge? I will have to work on this format as I have data of some users now that might have a bug and the removal of name mangling will make it easier for me to work into it.

Issue #1761, opened 2025-09-03
EBRAINS connection failing


zm711:
**Describe the bug**

I tried following this link:
https://object.cscs.ch/v1/AUTH_63ea6845b1d34ad7a43c8158d9572867/hbp-d000017_PatchClamp-GranuleCells_pub

which is where we get some example files from and even from my browser I get a dropped connection. When I went to the ebrains website they say we need to login to get access to anything. Is this a new policy at ebrains? Should we change the files we use? or does someone have a login that we could add to our secrets?

PR #1762, opened 2025-09-05
New URL for doc building from ebrains

zm711:


zm711:
Fixes #1761 @apdavison I think this is ready to go. Thanks for finding that!

apdavison:
If you're interested in the history, this change was part of a process of federating EBRAINS storage across different HPC centres. Unfortunately it wasn't possible to set up redirects for the old URLs.

Issue #1755, opened 2025-08-22
BlackrockIO can't read my nev files.


layne6689:
**Describe the bug**
There is an error: "UserWarning: Detected 177 undocumented segments within nev data after timestamps" and "NeoReadWriteError: Inconsistent ns6 and nev file. 178 segments present in .nev file, but 1 in ns6 file," which occurs when I use BlackrockIO to read my .nev and .ns6 files at the same time. This error occurs consistently on our .nev files. But BlackrockIO works fine when I only read the .ns6 file(use `load_nev=False`). Additionally, it works correctly when using NPMK (https://github.com/BlackrockNeurotech/NPMK) in MATLAB to read the .nev file. 

**To Reproduce**
code:
```
filepath = 'D:/Desktop/dummy file/test'
reader = BlackrockIO(filename=filepath, load_nev=True)
```
error:
```
UserWarning: Detected 177 undocumented segments within nev data after timestamps [   2   19   43   54   86   97  108  128  139  149  161  182  196  209
  226  234  306  315  329  378  401  414  438  453  473  509  559  583
  613  641  653  674  684  742  757  779  802  815  830  842  855  867
  900  910  922  946  982  995 1062 1098 1111 1123 1169 1237 1267 1315
 1327 1338 1369 1379 1416 1425 1469 1529 1548 1560 1582 1595 1683 1715
 1738 1772 1784 1801 1815 1825 1834 1857 1895 1905 1937 1982 1992 2007
 2048 2126 2165 2175 2186 2206 2241 2253 2264 2284 2296 2336 2344 2354
 2392 2438 2452 2486 2495 2506 2529 2545 2557 2568 2592 2647 2671 2689
 2701 2759 2780 2814 2819 2839 2849 2860 2881 2888 2900 2951 2956 2967
 2978 3000 3022 3045 3156 3163 3177 3200 3209 3252 3273 3300 3308 3341
 3364 3376 3396 3420 3432 3450 3484 3496 3508 3540 3551 3588 3682 3694
 3704 3716 3721 3733 3759 3781 3793 3810 3827 3837 3863 3871 3905 3928
 3942 3957 3975 3987 3997 4046 4067 4120 4245].


NeoReadWriteError: Inconsistent ns6 and nev file. 178 segments present in .nev file, but 1 in ns6 file.
```
I creat a dummy file which have the same error:
[dummy file.zip](https://github.com/user-attachments/files/21935892/dummy.file.zip)

 
[...]

PR #1763, opened 2025-09-05
Refactor BlackRock spec specific functions

h-mayorquin:
I worked on Blackrock today because of issue #1755 plus another user problem that we encountered on the past NWB workshop and there is an improvement I would like to make.

The current codebase uses generic variant names (`variant_a`, `variant_b`, `variant_c`) throughout the `BlackrockRawIO` implementation, making it difficult for developers to understand which Blackrock file specification version each function handles. When debugging or maintaining code, developers need to:

1. Look up mapping dictionaries to understand what variant_a means  
2. Cross-reference specification versions with variant letters  
3. Remember arbitrary letter-to-version associations  

I think is better to get rid of the arbitrary a, b and c and just include the specification directly on the name:

```
- variant_a → spec_v21 (handles specification 2.1)  
- variant_b → spec_v22_30 (handles specifications 2.2, 2.3, and 3.0)  
- variant_c → spec_v30_ptp (handles specification 3.0 with PTP timestamps)  
- Special cases: spec_v21_22 and spec_v23 for NEV data functions with specific version ranges  
```

Debugging then becomes easier because the function name directly shows the specification context with the intent clear from the function signature alone.

All of these are private functions so no problem with internal name changing (god bless private functions)

What do you think?

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -267,9 +267,11 @@ class BlackrockRawIO(BaseRawIO):
             self._nev_spec = self._extract_nev_file_spec()
 
             # read nev headers
-            self._nev_basic_header, self._nev_ext_header = self._nev_header_reader[self._nev_spec]()
+            header_reader_function = self._nev_header_reader[self._nev_spec]
+            self._nev_basic_header, self._nev_ext_header = header_reader_function()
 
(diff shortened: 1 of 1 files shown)


zm711 (approved):
This makes sense to me!

zm711:
You good with me merging this? Or any other tweaks you want to make?

PR #1769, opened 2025-09-08
Blackrock add summary of automatic data segmentation 

h-mayorquin:
See https://github.com/NeuralEnsemble/python-neo/issues/1770

The summary of the gaps look like this:
image

### DIFF
--- a/neo/test/rawiotest/test_spikegadgetsrawio.py
+++ b/neo/test/rawiotest/test_spikegadgetsrawio.py
@@ -56,13 +56,13 @@ class TestSpikeGadgetsRawIO(
     def test_opening_gibberish_file(self):
         """Test that parsing a file without </Configuration> raises ValueError instead of infinite loop."""
         # Create a temporary file with gibberish content that doesn't have the required tag
-        with tempfile.NamedTemporaryFile(mode='wb', suffix='.rec') as temp_file:
+        with tempfile.NamedTemporaryFile(mode="wb", suffix=".rec") as temp_file:
             # Write simple gibberish content without the required </Configuration> tag
             temp_file.write(b"gibberish\n")
             temp_file.flush()
-            
+
             reader = SpikeGadgetsRawIO(filename=temp_file.name)
             with self.assertRaises(ValueError) as cm:
                 reader.parse_header()
-            
+
             self.assertIn("xml header does not contain '</Configuration>'", str(cm.exception))


neo/rawio/blackrockrawio.py
```diff
@@ -323,13 +323,25 @@ def _parse_header(self):
...
+        # Read headers
         for nsx_nb in self._avail_nsx:
             spec_version = self._nsx_spec[nsx_nb] = self._extract_nsx_file_spec(nsx_nb)
             # read nsx headers
             nsx_header_reader = self._nsx_header_reader[spec_version]
             self._nsx_basic_header[nsx_nb], self._nsx_ext_header[nsx_nb] = nsx_header_reader(nsx_nb)
+            
+            # The Blackrock defines period as the number of  1/30_000 seconds between data points
+            # E.g. it is 1 for 30_000, 3 for 10_000, etc
+            nsx_period = self._nsx_basic_header[nsx_nb]["period"]
```
zm711, line 337: Is this 30_000 true for all versions? I vaguely remember that different versions had different resolutions but maybe I'm wrong?

And seems like this was hardcoded the same way below before.

neo/rawio/blackrockrawio.py
```diff
@@ -482,7 +494,7 @@ def _parse_header(self):
                             t_stop = max(t_stop, timestamps[-1] / ts_res + sec_per_samp)
                         else:
                             t_start = timestamps / ts_res
-                            t_stop = max(t_stop, t_start + length / self.sig_sampling_rates[nsx_nb])
+                            t_stop = max(t_stop, t_start + length / self._nsx_sampling_frequency[nsx_nb])
```
zm711, line 497: can this be explained to me why would the length lead to being longer than the the `t_stop` itself? Maybe I need to read the whole reader to understand this, but if there is a two second explanation that would be great!

neo/rawio/blackrockrawio.py
```diff
@@ -1051,7 +1063,7 @@ def _read_nsx_dataheader_spec_v30_ptp(
             ("reserved", "uint8"),
             ("timestamps", "uint64"),
             ("num_data_points", "uint32"),
-            ("samples", "int16", self._nsx_basic_header[nsx_nb]["channel_count"]),
+            ("samples", "int16", (self._nsx_basic_header[nsx_nb]["channel_count"],)),
```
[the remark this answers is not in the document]
zm711: And based on our testing this change is also backward compatible or should we test this over a greater range of numpy versions than our current skinny IO tests?

neo/rawio/blackrockrawio.py
```diff
@@ -1060,24 +1072,58 @@ def _read_nsx_dataheader_spec_v30_ptp(
...
+        # Segment data, at the moment, we segment, where the data has gaps that are longer
+        # than twice the sampling period.
+        sampling_rate = self._nsx_sampling_frequency[nsx_nb]
+        segmentation_threshold = 2.0 / sampling_rate
+
+        # The raw timestamps are the indices of an ideal clock that ticks at `timestamp_resolution` times per second.
+        # We convert this indices to actual timestamps in seconds
+        raw_timestamps = struct_arr["timestamps"]
+        timestamps_sampling_rate = self._nsx_basic_header[nsx_nb]["timestamp_resolution"]  # clocks per sec uint64 or uint32
+        timestamps_in_seconds = raw_timestamps / timestamps_sampling_rate
```
zm711, line 1085: Could this division lead to any issues uint/float should just be float right?

neo/rawio/blackrockrawio.py
```diff
@@ -1060,24 +1072,58 @@ def _read_nsx_dataheader_spec_v30_ptp(
...
+                gap_position_seconds = timestamps_in_seconds[sample_index] - timestamps_in_seconds[0]  # Time position relative to start
+                
+                segmentation_report_message += (
+                    f"| {sample_index:>15,} | {gap_position_seconds:>21.6f} | {gap_duration_ms:>21.3f} |\n"
+                )
+            
+            segmentation_report_message += "+-----------------+-----------------------+-----------------------+\n"
+            warnings.warn(segmentation_report_message)
+        
+        for seg_index, seg_start_index in enumerate(segment_start_indices):
```
zm711, line 1118: this is a little unclear. why do you have index index for your naming. Could we make this naming a little clearer for the future :)

zm711 (commented):
Few initial comments/questions.

neo/rawio/blackrockrawio.py
h-mayorquin: Good point, let me double-check the casting rules to see if nasty surprises could appear.

h-mayorquin: Yes, I think that the unsigned integers will be cast to float before dividing so I don't expect anything here, can you think on something?

neo/rawio/blackrockrawio.py
zm711: Yeah I agree because we define `t_stop` in all branches right? So we don't need to guard with an initial `t_stop`. That was definitely my confusion. We could always update that later. It's not crucial.

h-mayorquin:
> > > And based on our testing this change is also backward compatible or should we test this over a greater range of numpy versions than our current skinny IO tests?
> > 
> > 
> > I think that shapes should have always been specified as tuples and using a scalar was the odd case that they are now banning. We can check if that notation works for the smallest version of numpy that we support (which is?) if you want to be extra careful.
> 
> It's probably fine. I'm just being overly cautious. We go down to 1.24, but we only test on 1.26 for the IOs. You're probably right that we were doing it incorrectly before. So switching to correct shouldn't lead to problems.

Ok.

neo/rawio/blackrockrawio.py
```diff
@@ -1042,42 +1054,80 @@ def _read_nsx_dataheader_spec_v30_ptp(
 
         data_header = {}
         index = 0
-
-        if offset is None:
-            # This is read as an uint32 numpy scalar from the header so we transform it to python int
-            offset = int(self._nsx_basic_header[nsx_nb]["bytes_in_headers"])
+        # This is read as an uint32 numpy scalar from the header so we transform it to python int
+        header_size = offset or int(self._nsx_basic_header[nsx_nb]["bytes_in_headers"])
```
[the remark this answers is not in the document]
h-mayorquin: No, I made the same mistake, dammit.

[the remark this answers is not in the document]
h-mayorquin: changed.

zm711:
Small delay, but this is on my list to do final review asap.

neo/rawio/blackrockrawio.py
cboulay: Can confirm, the nominal base rate has always been 30 kHz, and this is the rate used by sample groups 5 and 6. Other sample groups are just sub-sampling this base rate (hopefully with anti-aliasing built in).

zm711:
After tests I will merge.

PR #1774, opened 2025-09-12
Improve docstings of `OpenEphysBinaryRawIO`

h-mayorquin:
While working on fixing [a neuroconv issue](https://github.com/catalystneuro/neuroconv/issues/1498), I noticed that the `OpenEphysBinaryRawIO` class lacks clear documentation about how the folder structure of the Open Ephys binary format is parsed and use internally. 

This PR improves the situation in two ways:  

1. The docstring now more explicitly describes the type of structure being parsed and clarifies that `dirname` can point to different levels within the folder tree, each yielding different parsing results.  

2. The current `explore_folder` function mixes two responsibilities:  
   - Parsing the folder structure.  
   - Mapping that structure into Neo.  

   To separate concerns, I split this logic into two private functions. Each now has a dedicated docstring that clearly explains its role.  **There is no change in implementation** the function is just split in two.

### DIFF
--- a/neo/rawio/openephysbinaryrawio.py
+++ b/neo/rawio/openephysbinaryrawio.py
@@ -1096,24 +1096,6 @@ def explore_folder(dirname, experiment_names=None):
 
 
 def check_folder_consistency(folder_structure, possible_experiment_names=None):
-    """
-    Validate consistency of the discovered OpenEphys folder structure.
-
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):
Nice rewrite @h-mayorquin! This added docs will make it much easier to maintain for us too! :)

Issue #1773, opened 2025-09-10
General API for handling sample gaps on rawio


h-mayorquin:
@zm711, @samuelgarcia, and I had a discussion today about two current Blackrock issues (#1770 and #1755).

The main problem is that the gaps automatically detected by the code are much smaller than what would reasonably qualify as a segment in Neo (for example, someone deliberately pausing the recording). These gaps are more likely artifacts of the system (in #1770, it looks like a case of dropped samples).

Creating format-specific heuristics for each format would be a large maintenance burden. A simpler, more maintainable, and general solution is to provide users (who know their data best) with both information and control. We came up with the following proposal:

1. **Default behavior**: If gaps are detected, loading should error out. RawIO readers should raise an error when timestamp gaps are found and display a report showing the number, size, and characteristics of the gaps. This way, users can make an informed decision. Examples of this approach can be seen in #1769 and #1484.
2. **Opt-in behavior**: RawIO readers should provide an argument (e.g. `gap_tolerance_ms`, naming TBD per reader) that lets users explicitly load data with gaps if they wish. Gaps smaller than the value will be ignored, gaps larger than the value will be segmented.
3. **Timestamps from acquisition system**: To make data with gaps more useful, we should provide the original timestamps from the acquisition system when available (see #1652 for Intan, #1816 for Blackrock).
4. **`ignore_integrity_checks` as the global escape hatch.** A boolean kwarg at instantiation that bypasses every integrity check the reader performs, including gap detection and any format-specific file-validity checks. Default is `False` (strict). This serves two audiences: users who know their data is fine and want faster loads (integrity checks can be expensive, as @samuelgarcia pointed out), and users who need to open corrupt files to recover what they can (e.g. when the person who ran the experiment is no longer available). Examples: #1470 (Intan corrupt-file opening), #1740 (other integrity checks).

This design would allow us to implement a consistent API across RawIO, so gaps are handled with a common interface. The plan is as follows:

1. Implement the solution first for Blackrock (since there are open issues and users currently cannot read the data). This will also let us to discuss the types and naming of the interface.
2. Extend the interface to other popular readers (e.g. Intan, OpenEphys, SpikeGLX, Plexon) to uncover potential difficulties.
3. Once the API is stable for the main readers, integrate it into the abstract/parent classes as a general RawIO API.
4. If possible, deprecate the current gap-handling mechanisms already in place (e.g. in OpenEphys) to simplify the codebase.

[...]

PR #1777, opened 2025-09-18
Fix  `NeuralynxRawIO`  reading in one-dir mode with nested directories

h-mayorquin:
Users sometimes place extra directories and files under the Neuralynx root. Today, we exclude by extension, but in “one-dir” mode the reader asserts that every entry must be a file, which raises an error:

https://github.com/NeuralEnsemble/python-neo/blob/72ec76a0cc30ebdaf321a91460b51632959ee5cb/neo/rawio/neuralynxrawio/neuralynxrawio.py#L217-L231

We should fix this to (1) make the workflow more forgiving for users and (2) align behavior with `include_filenames`, which already allows nested folders.

Proposed change: restructure file discovery into three steps.

1) Gather candidates by mode:
   - **multiple-files**: validate existence only for the explicitly provided filenames.
   - **one-dir**: list directory entries and keep only regular files via `Path.is_file()` (silently ignore subdirectories).

2) Apply the excluded-filenames filter.

3) Keep only paths with valid Neuralynx extensions.

This ensures directories are ignored in “one-dir” mode while preserving strict validation in “multiple-files” mode, and it removes redundant extension checks from the main processing loop.

cc @weiglszonja

### DIFF
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -160,7 +160,7 @@ class NeuralynxRawIO(BaseRawIO):
             else:
                 exclude_filenames = exclude_filename
             warnings.warn(
-                "`exclude_filename` is deprecated and will be removed in v1.0. Please use `exclude_filenames` instead"
+                "`exclude_filename` is deprecated and will be removed in version 1.0. Please use `exclude_filenames` instead"
             )
 
         if include_filenames is None:
--- a/neo/rawio/neuralynxrawio/neuralynxrawio.py
+++ b/neo/rawio/neuralynxrawio/neuralynxrawio.py
@@ -56,6 +56,7 @@ from ..baserawio import (
 import numpy as np
 import os
 import pathlib
+from pathlib import Path
 import copy
 import warnings
 from collections import namedtuple, OrderedDict
@@ -151,7 +152,7 @@ class NeuralynxRawIO(BaseRawIO):
 
         if filename is not None:
             include_filenames = [filename]
-            warnings.warn("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+            warnings.warn("`filename` is deprecated and will be removed in v1.0. Please use `include_filenames` instead")
 
         if exclude_filename is not None:
             if isinstance(exclude_filename, str):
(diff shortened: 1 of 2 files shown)


neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -151,15 +152,15 @@ def __init__(
 
         if filename is not None:
             include_filenames = [filename]
-            warnings.warn("`filename` is deprecated and will be removed. Please use `include_filenames` instead")
+            warnings.warn("`filename` is deprecated and will be removed in v1.0. Please use `include_filenames` instead")
```
h-mayorquin, line 155: From here:
https://github.com/NeuralEnsemble/python-neo/pull/1440#issuecomment-2010810682

zm711:
@PeterNSteinmetz, we are trying to keep you in the loop. We are planning on making some adjustments in gap detection across our RawIOs (see #1773), which means we are working on various IOs in anticipation of our more general BaseRawIO level changes. If you have time we would love you to check out this PR and #1779. If you don't have time we can move forward but we appreciate all your help so if you have discussion points as we make these changes feel free to let us know.

PeterNSteinmetz:
I think this change, #1777, will cause it to throw a ValueError is a directory name is provided in include_filenames. It will not look for files in sub-directories.

h-mayorquin:
> I think this change, #1777, will cause it to throw a ValueError is a directory name is provided in include_filenames. It will not look for files in sub-directories.

I don't understand, the current behavior on master is that to throw an error in both "one-dir" and "multiple-files" when there is a nested folder:

https://github.com/NeuralEnsemble/python-neo/blob/72ec76a0cc30ebdaf321a91460b51632959ee5cb/neo/rawio/neuralynxrawio/neuralynxrawio.py#L217-L232

PeterNSteinmetz:
Yes, I guess it throws an error with sub-directories and one-dir mode. Sorry I thought you were quoting the proposed fix above. I see in the commit list the changes. Those appear that they should work. Is there a test added for this?

h-mayorquin:
> Yes, I guess it throws an error with sub-directories and one-dir mode. Sorry I thought you were quoting the proposed fix above. I see in the commit list the changes. Those appear that they should work. Is there a test added for this?

Ah, that makes sense. 

Regarding the test,
Yes, I added a test to ensure that random nested directories do not throw the reader off.

Issue #1778, opened 2025-09-18
[Bug]   `NeuralynxRawIO` doesn't differentiate  signals per stream


h-mayorquin:
Currently the neuralynx code is creating streams based on the sampling rate, the number of packages and the first package's timestamp:

https://github.com/NeuralEnsemble/python-neo/blob/72ec76a0cc30ebdaf321a91460b51632959ee5cb/neo/rawio/neuralynxrawio/neuralynxrawio.py#L385

In addition of not being user friendly this has caused problems before [has been problematic before](https://github.com/NeuralEnsemble/python-neo/issues/1678) because it depends on the version of numpy.

A bigger problem with defining and naming streams this way is that it does not correctly differentiate signals of different nature. I have an example session with the following files:

```
.
├── CSC145.ncs
├── CSC146.ncs
└── csc23_100.ncs

1 directory, 3 files

```
Two of them are eye tracking and one of them is an extracellular electrophysiology signals. If we check the headers, they have different gains (e.g. InputRange) and signal filtering parameters which matches their nature, but, as I mentioned above, the `NeuralynxRawIO` is incorrectly identifying the three of them as the same stream. The reason for this is because they share the same sampling frequency, number of records and timestamp of the first block.

<details>
```

Command: find . -type f -iname "*.ncs" -exec sh -c 'echo ">>> $1"; head -c 16384 "$1"' _ {} \;
Output:
>>> ./CSC145.ncs ######## Neuralynx Data File Header
 -FileType NCS
 -FileVersion 3.4
 -FileUUID fb4d4862-546b-426e-9af0-db1316e21c12
 -SessionUUID 845b834f-9def-4465-a3be-672ffa54a575
 -ProbeName 
 -OriginalFileName "D:\CheetahData\2024-09-27_08-49-25\CSC145.ncs"
 -TimeCreated 2024/09/27 08:49:53
 -TimeClosed 2024/09/27 15:12:31
 
 -RecordSize 1044
 -ApplicationName Cheetah "6.5.0 "
 -AcquisitionSystem AcqSystem1 DigitalLynxSX
 -ReferenceChannel "Source 05 Panel Ground"
 -SamplingFrequency 32000
 -ADMaxValue 32767
 -ADBitVolts 0.000003051757812500000169
 -AcqEntName CSC145
 -NumADChannels 1
 -ADChannel 144
 -InputRange 100000
 -InputInverted False
 
 -DSPLowCutFilterEnabled False
[...]

PR #1779, opened 2025-09-18
Fix, Neuralynx encoding error

h-mayorquin:
This is an issue with the same data discussed in #1778 

```
.
├── CSC145.ncs
├── CSC146.ncs
└── csc23_100.ncs
```

The problem is that the `DspFilterDelay_µs` is sometimes encoded in latin-1 and sometimes encoded in utf8. This means that we decoding with the current method (latin-1) sometimes this specific property is read without the micro symbol and creates a key error here:

https://github.com/NeuralEnsemble/python-neo/blob/72ec76a0cc30ebdaf321a91460b51632959ee5cb/neo/rawio/neuralynxrawio/neuralynxrawio.py#L543-L548

when aggregating annotations over the channels.

This patch fixes this by using a regex to match what Latin-1 will read if the data is encoded as UTF-8.

The deeper problem is that Neuralynx does not really maintain its promise of ASCII headers.

@weiglszonja 

Screenshot from 2025-09-18 17-11-18

### DIFF
--- a/neo/rawio/neuralynxrawio/nlxheader.py
+++ b/neo/rawio/neuralynxrawio/nlxheader.py
@@ -55,7 +55,13 @@ class NlxHeader(OrderedDict):
         ("DspHighCutNumTaps", "", None),
         ("DspHighCutFilterType", "", None),
         ("DspDelayCompensation", "", None),
-        ("DspFilterDelay_µs", "", None),
+        # DspFilterDelay key with flexible µ symbol matching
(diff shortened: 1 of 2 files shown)


PeterNSteinmetz:
I think this patch is fine and should cover both cases. 

There is no official documentation of how the Neuralynx software writes the ascii header and it has varied over versions in a somewhat unpredictable manner. Thus all the regex'es being used to parse the headers.

PR #1789, opened 2025-10-08
Fix over-segmentation in BlackRock

### LINKED ISSUE #1770: `BlackrockRawIO` is over-segmenting data
I have user data that, according to them, has no true segments at all. They did not stop and restart the recording session. Nevertheless, the current `BlackrockRawIO` implementation creates thousands of segments.

The reason is the following logic in the code:  
https://github.com/NeuralEnsemble/python-neo/blob/3b592a8509c6d76a205cae9d6198190782b4881c/neo/rawio/blackrockrawio.py#L1066-L1072  

Here, if the difference in timestamps is larger than twice the expected difference based on the sampling rate of the stream, a new segment is created. For the user’s ns4 and ns6 files, these thresholds are as small as 0.2 ms and 0.067 ms, which is overly strict. This causes recordings with tiny millisecond gaps (Which I think are buffer and/or jitter artifacts, not real breaks) to be split into  even though the user insists the recording was continuous.

The Matlab NPMK implementation lets users control this threshold with the max_tick_multiple parameter:

https://github.com/BlackrockNeurotech/NPMK/blob/a5b3e3b25b6e2f4594ecbb99d3e0e5e517530959/NPMK/openNSx.m#L182-L186

My suggestion is to improve this in three ways:

- Provide warnings when gaps are found, including their size and type, similar to how Intan handling works. See https://github.com/NeuralEnsemble/python-neo/pull/1769
- Add a parameter (e.g. `segmentation_threshold``segment_threshold_s`) to the constructor so users can control how large a gap must be before a new segment is created.
- Give users direct access to the raw timestamps so they can analyze the gaps themselves, perform custom interpolation or drift correction, and align with other systems.

h-mayorquin:
This will fix #1770

Testing data is on https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/167

I will add tests once that is merged.

This PR fixes #1770 using the methodology discussed on #1773. To do this, we do three things:
1) We separate the parsing of the data blocks from the segementation of the data
2) I also continue the refactor in #1772 and #1771 to not have dynamically loaded functions but to have the headers specified outside of the code in a global variable.
3) the reporting is now added to not only the PTP but also to the older formats that have timestamps.
4) I used a buffered version of the memmaps so we don't need to create a memamp per block but only views.

This is a large PR but I think that @samuelgarcia prefers them that way and he is the person that might end up reviewing it. I am happy to break it apart if whoever is gona review this prefers it that way.

### DIFF
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -849,11 +849,8 @@ class BlackrockRawIO(BaseRawIO):
         nsx_ext_header : numpy memmap
             Extended header information
         """
-        # Construct filename based on spec version
-        if spec == "2.1":
-            filename = ".".join([self._filenames["nsx"], f"ns{nsx_nb}"])
-        else:  # 2.2, 2.3, 3.0
-            filename = f"{self._filenames['nsx']}.ns{nsx_nb}"
+        # Construct filename
+        filename = f"{self._filenames['nsx']}.ns{nsx_nb}"
         
         # Get basic header structure for this spec
         basic_header_dtype = NSX_BASIC_HEADER_TYPES[spec]
--- a/neo/rawio/blackrockrawio.py
+++ b/neo/rawio/blackrockrawio.py
@@ -187,12 +187,6 @@ class BlackrockRawIO(BaseRawIO):
 
         # These dictionaries are used internally to map the file specification
         # revision of the nsx and nev files to one of the reading routines
-        self._nsx_params = {
-            "2.1": self._get_nsx_param_spec_v21,
-            "2.2": self._get_nsx_param_spec_v22_30,
-            "2.3": self._get_nsx_param_spec_v22_30,
-            "3.0": self._get_nsx_param_spec_v22_30,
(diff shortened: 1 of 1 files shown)


neo/rawio/blackrockrawio.py
```diff
@@ -853,354 +853,549 @@ def _extract_nev_file_spec(self):
...
-            }
 
-            # Jump to the next data block, the data is encoded as int16
-            data_block_size_bytes = num_data_points * channel_count * np.dtype("int16").itemsize
-            current_offset_bytes = offset_to_data_block_start + data_block_size_bytes
+            num_samples = int(header["nb_data_points"])
+            data_offset = current_offset + header.dtype.itemsize
+            timestamp = header["timestamp"]
+
+            # Create data view into memmap for this block
```
h-mayorquin, line 1128: Here, we use views on a linear memmaped buffer to avoid creating a memmap per data block. This is good because OSes usually limit the ammount of memamps you can create.

PR #1795, opened 2025-10-15
SpikeGLX: only check for bin file size if it exists

alejoe91:
@h-mayorquin 

For IBLExtractors we use the SpikeGLX machinery, but the bin file is not available locally


h-mayorquin (approved):
Yes, this reflects a real tension we’ve been grappling with in the SpikeGLX metadata.

On one hand, it’s ideal to obtain metadata directly from the accompanying metadata files. These are meant to serve as the source of truth and can be accessed even without the corresponding data files. This is the approach used in the IBL case.

On the other hand, users frequently modify the files themselves. This includes stub files, renamed files, and even cases where IBL members use their own ad-hoc SpikeGLX variants with custom compression. Because of these inconsistencies, instead of relying solely on the metadata, we began querying the data directly. This was the motivation behind my change that broke this.

Your proposed solution strikes me as a reasonable compromise between the two approaches. Query the data directly for size but if not, use the metadata.

PR #1797, opened 2025-10-17
Add support for VECTORIZED orientation in BrainVisionRawIO

Kkuntal990:
## Description
[Github Issue](https://github.com/NeuralEnsemble/python-neo/issues/1799)

This PR adds support for VECTORIZED data orientation in BrainVisionRawIO, which was previously causing a `NeoReadWriteError`.

## Problem
BrainVision files can store data in two orientations:
- **MULTIPLEXED**: ch1_s1, ch2_s1, ch3_s1, ch1_s2, ch2_s2, ... (interleaved)
- **VECTORIZED**: all ch1 samples, then all ch2 samples, etc. (sequential)

Previously, Neo only supported MULTIPLEXED orientation, causing imports to fail on VECTORIZED files such as those in OpenNeuro dataset ds004621.

## Solution
- Modified the orientation check to accept both MULTIPLEXED and VECTORIZED
- Implemented custom `_get_analogsignal_chunk()` method that:
  - Uses the default parent implementation for MULTIPLEXED files (no behavior change)
  - For VECTORIZED files, reads each channel's data from its sequential location in the binary file
- Maintains full backward compatibility with existing MULTIPLEXED files

## Testing
✅ **Validation against MNE-Python**: Tested on real-world VECTORIZED dataset (ds004621, 127 channels × 740,360 samples)
- Correlation: 1.0
- Maximum absolute difference: 0.0
- Results match MNE-Python exactly

✅ **Synthetic tests**: Created test files in both orientations and verified correct data reading

✅ **Backward compatibility**: MULTIPLEXED files continue to work as before

## Changes
- `neo/rawio/brainvisionrawio.py`:
  - Added `self._data_orientation` to track orientation type
  - Implemented custom `_get_analogsignal_chunk()` for VECTORIZED reading
  - Updated error message to reflect support for both orientations

## Related Issue
- Dataset: OpenNeuro ds004621
- [Github Issue](https://github.com/NeuralEnsemble/python-neo/issues/1799)

### DIFF
--- a/neo/rawio/baserawio.py
+++ b/neo/rawio/baserawio.py
@@ -1577,14 +1577,11 @@ class BaseRawWithBufferApiIO(BaseRawIO):
     def _get_signal_size(self, block_index, seg_index, stream_index):
         buffer_id = self.header["signal_streams"][stream_index]["buffer_id"]
         buffer_desc = self.get_analogsignal_buffer_description(block_index, seg_index, buffer_id)
-        # For "raw" type, shape is always (time, channels) regardless of file layout
-        # For "hdf5" type, shape can be (time, channels) or (channels, time) based on time_axis
-        if buffer_desc["type"] == "raw":
-            return buffer_desc["shape"][0]
-        else:
-            # some hdf5 revert the buffer
-            time_axis = buffer_desc.get("time_axis", 0)
-            return buffer_desc["shape"][time_axis]
+        # time_axis indicates which dimension is time:
+        # time_axis=0: shape is (time, channels)
+        # time_axis=1: shape is (channels, time)
+        time_axis = buffer_desc.get("time_axis", 0)
+        return buffer_desc["shape"][time_axis]
 
     def _get_analogsignal_chunk(
         self,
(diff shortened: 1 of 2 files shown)


Kkuntal990:
> Can you open an issue here in neo? I can't see the jira board.

Done

Kkuntal990:
## Latest Updates: Refactored to use time_axis pattern in BaseRawIO

Thanks @samuelgarcia for the suggestion! I've refactored the implementation to move the logic into BaseRawIO using the `time_axis` parameter, making it consistent with the existing HDF5 pattern.

---

###  Changes Made

#### 1. Replaced np.fromfile() with np.memmap() 
- Avoids multiprocessing issues with shared file handles
- More efficient and safer

#### 2. Extended BaseRawIO with time_axis support for raw buffers 
- Added `time_axis=1` handling in `BaseRawIO._get_analogsignal_chunk()`
- Follows the same convention as HDF5:
  - `time_axis=0`: shape is `(time, channels)` - MULTIPLEXED
  - `time_axis=1`: shape is `(channels, time)` - VECTORIZED
- Updated `_get_signal_size()` to use `shape[time_axis]` consistently

#### 3. Simplified BrainVisionRawIO 
- Sets `time_axis` in buffer_description based on DataOrientation
- Adjusts shape to match time_axis convention (for VECTORIZED: `(channels, time)`)
- No custom override needed - relies entirely on BaseRawIO

#### 4. Fixed shape convention consistency 
- Initially had inconsistency where raw buffers used `(time, channels)` even with `time_axis=1`
- Now fully consistent: `time_axis=1` means shape is `(channels, time)` for both raw and HDF5



- **More general**: Other readers can now use `time_axis` for similar layouts
- **Cleaner**: Removed BrainVision-specific override logic
- **Consistent**: Same pattern throughout BaseRawIO (raw & HDF5)

Please review when you can

samuelgarcia:
Hi @Kkuntal990.
Really sorry for the dealy.
I will try my best next week

arnodelorme:
Could we merge this branch because it is a blocker for us. Tx

zm711:
@Kkuntal990,

once we fix our testing I'll be doing the next release (early January). Could you let me know if you want to be included in our author list? I would need the name you want to appear and any affiliations that should be included.

PR #1807, opened 2025-12-30
Move lower limit of NumPy to 1.25 

zm711:
Per NEP29 + 1 yr

https://endoflife.date/numpy

1.24 went EOL 18DEC24. 

Testing failing due to issue over on quantities in the linked issue below. We can work on fixing that and then this should pass fine :)

PR #1808, opened 2025-12-30
Update license specifier for PEP639

### LINKED ISSUE #1793: Update license specifier in pyproject.toml
Please note that as of PEP639 we need to specify our license differently 

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

min version of Setuptools for this is 77 so we should discuss whether we want to update our build dependency to that.

zm711:
Fixes #1793 

As of [PEP369](https://peps.python.org/pep-0369/), we should specify our license like this. 

Testing might fail due to quantities (see #1807) for more info.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.15.0.dev0"
 authors = [{name = "Neo authors and contributors"}]
 description = "Neo is a package for representing electrophysiology data in Python, together with support for reading a wide range of neurophysiology file formats"
 readme = "README.rst"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 license = "BSD-3-Clause"
 classifiers = [
     "Development Status :: 4 - Beta",


zm711:
In addition the RTD is failing. It seems like it is not using an appropriate way of reading the license specifier.... I'll dig into this in a bit.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ authors = [{name = "Neo authors and contributors"}]
 description = "Neo is a package for representing electrophysiology data in Python, together with support for reading a wide range of neurophysiology file formats"
 readme = "README.rst"
 requires-python = ">=3.9"
-license = {text = "BSD 3-Clause License"}
+license = "BSD 3-Clause"
```
alejoe91:
```suggestion
license = "BSD-3-Clause"
```
I think it should be this!

PR #1815, opened 2026-02-06
fix TdtRawIO multi-block SEV file lookup (wrong path + case mismatch)

### LINKED ISSUE #1814: TdtRawIO: multi-block SEV files not found (wrong path + case mismatch)
**Describe the bug**
TdtRawIO._parse_header fails to find .sev files in **multi-block** TDT datasets on case-sensitive filesystems (Linux). Two bugs cause a silent fallback to .tev, resulting in all channels reading identical (wrong) data.

1. **Wrong search directory**: The `path` variable at

https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L290

never points to the block subdirectory. With default `sortname=""`, it's a stale loop variable from line 

https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L102-L106

With a sortname set, it gets reassigned to `self.dirname (the tank root)` at line

https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L165-L175

 In both cases, SEV files live in `self.dirname / segment_name` (e.g., _tank/Block-1/_), not where path points.

2. **Case mismatch**: The multi-block path constructs filenames with lowercase _ch (_ch1.sev)_, 

https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L286-L290

but for the data that I have TDT writes it with uppercase _Ch (_Ch1.sev)_. On Linux, the file is not found. The single-block code path already handles this with a `[cC]h` glob patter,  the multi-block path just missed it.

Both bugs cause .sev lookup to fail silently. Neo falls back to `.tev` and all channels read identical wrong data. No error or warning is raised.

**To Reproduce**
I don't have a publicly shareable multi-block dataset with `.sev` files yet, but the bug is visible from the code. The existing test datasets don't cover this path: `aep_05` is multi-block but has no `.sev` files; `dataset_0_single_block` has `.sev` files but goes through the (correct) single-block code path.

I can also contribute a trimmed multi-block test dataset with `.sev` files to `ephy_testing_data` once I have explicit permission from the data provider, so this code path gets regression-tested.

**Expected behaviour**
Each column of the returned array should contain distinct per-channel data read from the corresponding` .sev `file.

**Environment:**
- OS: Linux (Debian 13, case-sensitive ext4)
- Python version: 3.12
- Neo version: 0.14.3
- NumPy version: 1.26

**Additional context**

I've verified a minimal fix locally. Will open a PR with it soon.

ree-gupta:
Fixes multi-block SEV file lookup in TdtRawIO. Two bugs caused silent fallback to TEV, resulting in all channels reading identical data:

1. `path` variable at L290 was stale (from `iterdir()` loop at L103 or reassigned to tank root at L170),  never pointed to the block subdirectory where SEV files live
2. Filename used lowercase `ch` but TDT writes `Ch`,  not found on case-sensitive filesystems (Linux)

Fix: use `self.dirname / segment_name` for the correct block directory, and `[cC]h` glob to match both cases (same pattern the single-block path already uses).

Not caught by tests because `aep_05` (the only multi-block test dataset) has no SEV files.

Closes #1814

### DIFF
--- a/neo/rawio/tdtrawio.py
+++ b/neo/rawio/tdtrawio.py
@@ -288,12 +288,12 @@ class TdtRawIO(BaseRawIO):
                         block_path = self.dirname / segment_name
                         sev_regex = f"{tankname}_{segment_name}_{stream_name}_[cC]h{chan_id}.sev"
                         sev_filename = list(block_path.glob(sev_regex))
-if len(sev_filename) == 1:
-    sev_filename = sev_filename[0]
-elif len(sev_filename) == 0:
-    sev_filename = None   # Indirect flag for TEV Format, see issue 1087
-else:
-    raise ValueError(f"Multiple SEV files matched for channel {chan_id}: {sev_filename}")
+                        if len(sev_filename) == 1:
+                            sev_filename = sev_filename[0]
+                        elif len(sev_filename) == 0:
+                            sev_filename = None  # Indirect flag for TEV Format, see issue 1087
+                        else:
+                            raise ValueError(f"Multiple SEV files matched for channel {chan_id}: {sev_filename}")
                     else:
                         # for single block datasets the exact name of sev files is not known
                         sev_regex = f"*_[cC]h{chan_id}.sev"


pauladkisson:
I can confirm that this fix also works for my use case (hidden ._ files/folders)

neo/rawio/tdtrawio.py
```diff
@@ -283,11 +283,12 @@ def _parse_header(self):
...
                     # data buffer test if SEV file exists otherwise TEV
-                    # path = self.dirname / segment_name
                     if self.tdt_block_mode == "multi":
                         # for multi block datasets the names of sev files are fixed
-                        sev_stem = f"{tankname}_{segment_name}_{stream_name}_ch{chan_id}"
-                        sev_filename = (path / sev_stem).with_suffix(".sev")
+                        block_path = self.dirname / segment_name
+                        sev_regex = f"{tankname}_{segment_name}_{stream_name}_[cC]h{chan_id}.sev"
+                        sev_filename = list(block_path.glob(sev_regex))
+                        sev_filename = sev_filename[0] if len(sev_filename) == 1 else None
```
h-mayorquin: I find the None strange here. Did you copied this from the segment branch, correct?

I think this is a mistake (see here https://github.com/NeuralEnsemble/python-neo/issues/1817) 

Given that you are using a fully qualified name here finding two files should throw an error.

Maybe something like this would be better

```
sev_filename = list(block_path.glob(sev_regex))
if len(sev_filename) == 1:
    sev_filename = sev_filename[0]
elif len(sev_filename) == 0:
    sev_filename = None   # Indirect flag for TEV Format, see issue 1087
else:
    raise ValueError(f"Multiple SEV files matched for channel {chan_id}: {sev_filename}")
```

h-mayorquin (approved):
LGTM.

One small comment and I am fine with this if the tests pass : )

h-mayorquin:
@alejoe91 @samuelgarcia @zm711 
This looks good to me. I think is a simple fix to an obvious error. Let me know if you want to take a look otherwise it would be good to merge so I can work on:

https://github.com/NeuralEnsemble/python-neo/issues/1817

PR #1820, opened 2026-02-17
Move Blackrock nsx handling closer to the buffer API pattern

h-mayorquin:
PR #1513 introduced `BaseRawWithBufferApiIO`, a base class that stores lightweight buffer descriptions (file paths, offsets, dtypes) instead of persistent `np.memmap` objects, creating short-lived mmap views on demand through cached file descriptors. This PR moves BlackrockRawIO toward that same pattern. We cannot fully adopt `BaseRawWithBufferApiIO` because: (a) Blackrock's standard format (v2.2/2.3/3.0) stores multiple data blocks with explicit headers in a single file, requiring block-aware offset tracking that the base class's one-buffer-per-stream model does not support; (b) the PTP variant interleaves per-sample timestamps with sample data in fixed-size packets, requiring strided mmap access (`np.ndarray` with custom `strides`) rather than contiguous views; and (c) the segmentation logic for PTP files needs to read timestamps from the file to detect gaps, which couples parsing and segmentation in a way the base class does not allow. However, we can adopt the core idea: replace stored memmaps with on-demand creation.

This has the added benefit of reducing the number of memmaps that are open at any time and we have seen that this pattern is both 1) more performant memory wise, 2) less likely to lead to leaks. Importantly, this PR does not change any behavior of the reader, it is an internal optimization only. It lays the groundwork for a follow-up PR adding gap detection on the standard format.


neo/rawio/blackrockrawio.py
```diff
@@ -650,22 +642,117 @@ def _segment_t_stop(self, block_index, seg_index):
...
+
+        return np.ndarray(
+            shape=shape,
+            dtype=dtype,
+            buffer=raw_mmap,
+            offset=start_remainder,
+            strides=strides,
+        )
+
+    def _get_nsx_fid(self, nsx_nb):
```
zm711, line 728: Could you walk me through how this construction is going to happen? Shouldn't we grab all nsx_fids at one time to construct them. This calls and caches the fids at it goes right? creating the dict on the first call?

neo/rawio/blackrockrawio.py
```diff
@@ -61,6 +61,7 @@
 """
 
 import datetime
+import mmap
```
zm711, line 64: the benefit of mmap vs np.memmap?

neo/rawio/blackrockrawio.py
```diff
@@ -650,22 +642,117 @@ def _segment_t_stop(self, block_index, seg_index):
...
+            self._nsx_fids[nsx_nb] = open(filename, "rb")
+        return self._nsx_fids[nsx_nb]
+
     def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes):
         stream_id = self.header["signal_streams"][stream_index]["id"]
         nsx_nb = int(stream_id)
-        memmap_data = self.nsx_datas[nsx_nb][seg_index]
+        seg = self._nsx_data_header[nsx_nb][seg_index]
+        fid = self._get_nsx_fid(nsx_nb)
+        kw = seg["memmap_kwargs"]
```
zm711: Could we get a better variable name for readability?

neo/rawio/blackrockrawio.py
```diff
@@ -1218,26 +1309,47 @@ def _parse_nsx_data_v30_ptp(self, nsx_nb):
...
-        if not np.all(file_memmap["num_data_points"] == 1):
+        if not np.all(temp_memmap["num_data_points"] == 1):
             # Not actually PTP! Fall back to standard format
+            del temp_memmap
             return self._parse_nsx_data_v22_v30("3.0", nsx_nb)
 
-        # Extract data and timestamps from structured array
-        data = file_memmap["samples"]
-        timestamps = file_memmap["timestamps"]
+        del temp_memmap
```
zm711: why are we deleting the temp_memmap twice?

neo/rawio/blackrockrawio.py
```diff
@@ -1369,29 +1494,48 @@ def _segment_nsx_data(self, data_blocks_dict, nsx_nb):
...
+                    })
 
             # V2.1 or single block standard format: no segmentation needed
             else:
-                segments[0] = {
-                    "data": data,
-                    "timestamp": timestamps,  # Use singular for backward compatibility
-                    "nb_data_points": data.shape[0],
+                segments.append({
+                    "timestamp": block_info["timestamps"],  # Use singular for backward compatibility
```
zm711: Do we know what this backward compatibility is?

neo/rawio/blackrockrawio.py
```diff
@@ -1996,36 +2140,25 @@ def _delete_empty_segments(self):
...
+            for nsx_nb in self.nsx_to_load:
+                del self._nsx_data_header[nsx_nb][seg_index]
 
-                # Also remap nev data, ev_ids are the equivalent to keys above
-                if self._avail_files["nev"]:
-                    for k, (data, ev_ids) in self.nev_data.items():
-                        ev_ids[ev_ids == j] -= 1
+            # Remap nev segment ids: shift down all ids above the removed segment
+            if self._avail_files["nev"]:
+                for _key, (data, ev_ids) in self.nev_data.items():
```
zm711: where is _key even being used?

neo/rawio/blackrockrawio.py
```diff
@@ -1690,7 +1834,7 @@ def _match_nsx_and_nev_segment_ids(self, nsx_nb):
             for k, (data, ev_ids) in self.nev_data.items():
 
                 # Check all nonempty nsX segments
-                for i, seg in enumerate(list_nonempty_nsx_segments[:]):
+                for i, seg in enumerate(nonempty_nsx_segments[:]):
```
zm711: If you're already changing names of stuff want to change `i` into something more interpretable for future us?

neo/rawio/blackrockrawio.py
```diff
@@ -1996,36 +2140,25 @@ def _delete_empty_segments(self):
...
-        # Discard empty segments
-        removed_seg = []
-        for data_bl in range(self._nb_segment):
+        # Find empty segments (fewer than 2 samples across all nsx files)
+        empty_indices = []
+        for seg_index in range(self._nb_segment):
             keep_seg = True
             for nsx_nb in self.nsx_to_load:
-                length = self.nsx_datas[nsx_nb][data_bl].shape[0]
+                length = self._nsx_data_header[nsx_nb][seg_index]["nb_data_points"]
```
zm711, line 2150: Something I keep seeing. Why didn't the original implementation use this nb_data_points value? It seems to have been present before. They seem to want to use the shape of the data. Is it possible that the dict value is ever off from the actual data shape?

neo/rawio/blackrockrawio.py
```diff
@@ -1996,36 +2140,25 @@ def _delete_empty_segments(self):
...
-                # remap nsx seg index
-                for nsx_nb in self.nsx_to_load:
-                    data = self.nsx_datas[nsx_nb].pop(j)
-                    self.nsx_datas[nsx_nb][j - 1] = data
+                empty_indices.append(seg_index)
 
-                    data_header = self._nsx_data_header[nsx_nb].pop(j)
-                    self._nsx_data_header[nsx_nb][j - 1] = data_header
+        # Remove empty segments in reverse order to preserve indices
+        for seg_index in reversed(empty_indices):
```
zm711, line 2156: Nice switch to reversed

neo/rawio/blackrockrawio.py
```diff
@@ -1996,36 +2140,25 @@ def _delete_empty_segments(self):
...
+                del self._nsx_data_header[nsx_nb][seg_index]
 
-                # Also remap nev data, ev_ids are the equivalent to keys above
-                if self._avail_files["nev"]:
-                    for k, (data, ev_ids) in self.nev_data.items():
-                        ev_ids[ev_ids == j] -= 1
+            # Remap nev segment ids: shift down all ids above the removed segment
+            if self._avail_files["nev"]:
+                for _key, (data, ev_ids) in self.nev_data.items():
+                    ev_ids[ev_ids > seg_index] -= 1
```
zm711, line 2163: This was the other bit I was getting a lost at. So you check for > seg_index, but they seem to be only checking that j is equal. So let me double check that I understand correctly. Previously they were doing a nested loop with the j and the ev_ids and then basically it would iterate one value per loop. What you are doing is doing only one loop through the data and changing all applicable values during that one loop, right? If that's not the correct intuition than I might need you to explain this bit to me.

zm711 (commented):
Initial round of comments. I'm not as familiar with this reader so might need a bit more explanation in general @h-mayorquin.

neo/rawio/blackrockrawio.py
```diff
@@ -1218,26 +1309,47 @@ def _parse_nsx_data_v30_ptp(self, nsx_nb):
...
         channel_count = int(self._nsx_basic_header[nsx_nb]["channel_count"])
 
-        # Create structured memmap (timestamp + samples per packet)
+        # Create structured memmap for verification only
         ptp_dt = NSX_DATA_HEADER_TYPES["3.0-ptp"](channel_count)
-        npackets = int((filesize - header_size) / np.dtype(ptp_dt).itemsize)
-        file_memmap = np.memmap(filename, dtype=ptp_dt, shape=npackets, offset=header_size, mode="r")
+        ptp_dtype = np.dtype(ptp_dt)
+        npackets = int((filesize - header_size) / ptp_dtype.itemsize)
+        temp_memmap = np.memmap(filename, dtype=ptp_dt, shape=npackets, offset=header_size, mode="r")
```
zm711: Should this be changed to `mmap`?

neo/rawio/blackrockrawio.py
h-mayorquin: As mentioned in our in-person discussion this is the pattern that we use in the buffer array API and we use it for a reason: it behaves better with the python garbage collector.

I don't think either me or Sam understand that one that well but my guess is (with emphasis on guess) np.memmap slices keep references to the original memmap/file whereas np.ndarray(buffer=...) does not.

neo/rawio/blackrockrawio.py
h-mayorquin: Done.

neo/rawio/blackrockrawio.py
h-mayorquin: You've got it right, it's lazy. On the first call it creates the fid dict, then for each nsx number it opens the file and caches the handle on the first miss, returning the cached handle on every subsequent call. So the handles are built up one at a time as they're first needed, not all at once.

I kept it lazy deliberately to mirror the buffer API's own file-handle pattern, where the multiplexed path opens each buffer's fid on demand, caches it, and reuses it on later calls rather than opening everything upfront. This is the same convention keyed by nsx number.

It also genuinely defers work here. Only PTP files open during header parsing; non-PTP files don't open until the first data read. So a reader that just inspects the header, or only reads one stream, never opens files it doesn't touch. The fail-fast case is already covered by the missing-file check during parsing, which validates the requested set exists before we reach this point.

neo/rawio/blackrockrawio.py
h-mayorquin: Should be done.

neo/rawio/blackrockrawio.py
h-mayorquin: Good point, and here, I was using this to check for ptp and I actually don't think we need this. So changing this.

neo/rawio/blackrockrawio.py
h-mayorquin: no longer here.

neo/rawio/blackrockrawio.py
h-mayorquin: Thanks for checking this.

I dug into this and there was no real backward compatibility being preserved, the comment was misleading. The value under that key has always been a single scalar (one timestamp per block), and the only thing consuming the plural `"timestamps"` key was the segmentation step, which immediately renamed it back to the singular `"timestamp"` that the rest of the reader already uses. It never crossed a module boundary or fed the per-sample timestamps path. So it was just an inconsistent name from the refactor, and I collapsed it to the singular key that matches the value's actual cardinality and dropped the comment.

neo/rawio/blackrockrawio.py
h-mayorquin: Hes. Good idea. Done.

neo/rawio/blackrockrawio.py
h-mayorquin: The original used `.shape[0]` because the old reader eagerly memmapped every segment into `self.nsx_datas`, so the array was already in hand and `.shape[0]` was the natural way to get the sample count. The buffer-API refactor doesn't eager-map anymore, it stores `memmap_kwargs` and maps on demand, so there's no materialized array here to measure, and reading the stored `nb_data_points` is the equivalent.

My thinking:
It can't drift from the actual shape, because both come from the same source: the block header's `nb_data_points` field for standard formats, the filesize computation for v2.1, or the PTP gap-segment length. The data view's shape is derived from that same number rather than measured independently, so they're equal by construction. On master it actually round-tripped, `nb_data_points = data.shape[0]` where the memmap was created with that same count, so `.shape[0]` was never independent ground truth either.

neo/rawio/blackrockrawio.py
h-mayorquin: Let's switch to values, this is not used.

neo/rawio/blackrockrawio.py
h-mayorquin: Your intuition is exactly right. The old code had the inner `j` loop and decremented `ev_ids[ev_ids == j]` one index per iteration; the new code does it in a single vectorized `ev_ids[ev_ids > seg_index] -= 1`. The reason it can is that the segment container changed from a dict keyed by segment index to a list. On master, removing a segment meant manually re-keying every higher segment `j -> j-1` in an inner loop, and the nev remap was interleaved into that loop, hence `== j` per pass. With a list, `del [seg_index]` shifts everything down automatically, so there's no inner loop and the nev ids above the removed segment can all be shifted at once. Same result either way: every id above the removed segment drops by one.

h-mayorquin:
All the comments should be addressed now @zm711 . Let me know if thereis anything else.

zm711:
Let me reread this tomorrow and then I'll merge barring anything I find.

neo/rawio/blackrockrawio.py
```diff
@@ -710,19 +797,27 @@ def _get_blackrock_timestamps(self, block_index, seg_index, i_start, i_stop, str
...
+                fid=fid,
+                dtype=ts_kw["dtype"],
+                offset=ts_kw["offset"],
+                num_samples=ts_kw["num_samples"],
+                packet_size=ts_kw.get("packet_size"),
+                item_offset=ts_kw.get("item_offset", 0),
+            )
             ts_res = float(self._nsx_basic_header[nsx_nb]["timestamp_resolution"])
-            return raw_timestamps[i_start:i_stop].astype("float64") / ts_res
+            return timestamps[i_start:i_stop].astype("float64") / ts_res
```
zm711, line 820: Can't we just store timestamps as int's? I don't know the format, so maybe they need to be floats.

zm711 (approved):
One last question, but I don't think it is blocking for now.

neo/rawio/blackrockrawio.py
h-mayorquin: The reason this store this as float64 seconds is to keep one consistent return shape across all Blackrock variants, which store timing very differently:

- PTP (Gemini): one uint64 nanosecond timestamp per sample.
- Standard (v2.2/2.3/3.0): one timestamp per data block in 30 kHz ticks, not per sample.
- FileSpec 2.1: no stored timestamps at all.

I thought about returning integer but there is no single common axis for this to be done so I went for within-format consistency in float-seconds.

h-mayorquin: Let me know if I can merge this.

zm711:
Go ahead and merge when you're ready @h-mayorquin.

neo/rawio/blackrockrawio.py
```diff
@@ -650,22 +642,117 @@ def _segment_t_stop(self, block_index, seg_index):
...
-        return memmap_data.shape[0]
+        return self._nsx_data_header[nsx_nb][seg_index]["nb_data_points"]
 
     def _get_signal_t_start(self, block_index, seg_index, stream_index):
         stream_id = self.header["signal_streams"][stream_index]["id"]
         nsx_nb = int(stream_id)
         return self._sigs_t_starts[nsx_nb][seg_index]
 
+    @staticmethod
+    def _create_mmap_view(fid, dtype, offset, num_samples, num_channels=None,
```
samuelgarcia, line 653: very smart

PR #1822, opened 2026-02-18
Add `gap_tolerance_ms` API to Neuralynx reader

h-mayorquin:
Adds the `gap_tolerance_ms` parameter to `NeuralynxRawIO` and `NeuralynxIO`, following the pattern established by the Blackrock reader in #1789. When gaps are detected and no tolerance is set, a `ValueError` is raised with a detailed gap report. When a tolerance is provided, the data is automatically segmented at gaps exceeding that threshold. The old `strict_gap_mode` parameter is deprecated with a warning. By default (`gap_tolerance_ms=None`), only gaps of at least one sample period are reported as errors; sub-sample timestamp deviations from rounding or clock jitter are silently ignored.

A `_get_neuralynx_timestamps()` method exposes the original per-record hardware timestamps in microseconds, giving users access to the raw timing information for drift analysis or precise alignment with external systems.

The test infrastructure gains `rawio_kwargs` on `BaseTestRawIO` and `io_kwargs` on `BaseTestIO`, allowing any format's tests to pass constructor arguments without overriding test methods or subclassing the IO class.


samuelgarcia:
This is heroic!
I will try to have a look soon ! ( i am a lier )

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -245,7 +259,7 @@ def _buildNcsSections(ncsMemMap, sampFreq, gapTolerance=0):
         return ncsSects
 
     @staticmethod
-    def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=True):
+    def build_for_ncs_file(ncsMemMap, nlxHdr, gap_tolerance_us=None, **kwargs):
```
zm711, line 262: above you start with milliseconds and now you switch to microseconds.

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -256,11 +270,37 @@ def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=Tru
...
+        # Handle deprecated parameters
+        gapTolerance = kwargs.pop("gapTolerance", None)
+        strict_gap_mode = kwargs.pop("strict_gap_mode", None)
+        if kwargs:
+            raise TypeError(f"Unexpected keyword arguments: {list(kwargs.keys())}")
+
+        if gapTolerance is not None:
+            warnings.warn(
+                "The `gapTolerance` parameter is deprecated and will be removed in version 0.16. "
+                "Use `gap_tolerance_us` instead.",
```
zm711, line 289: again how will this get exposed if the top-level function is in milliseconds?

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -269,14 +309,13 @@ def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=Tru
             # restriction arose from the sampling being based on a master 1 MHz clock.
             microsPerSampUsed = math.floor(NcsSectionsFactory.get_micros_per_samp_for_freq(freq))
             sampFreqUsed = NcsSectionsFactory.get_freq_for_micros_per_samp(microsPerSampUsed)
-            if gapTolerance is None:
-                if strict_gap_mode:
-                    # this is the old behavior, maybe we could put 0.9 sample interval no ?
-                    gapTolerance = 0
+            if gap_tolerance_us is None:
+                if strict_gap_mode is not None and not strict_gap_mode:
```
zm711: Shouldn't we instead of using the deprecated variable name in the code we should just convert the deprecated code into the current code at the top somewhere with like two lines of code. Then we only check on the new code lines so when we are done deprecating it we delete the couple lines at the top rather than a big rewrite here.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -878,6 +907,76 @@ def _rescale_event_timestamp(self, event_timestamps, dtype, event_channel_index)
...
+            f"Gap Report for {os.path.basename(filename)}:\n"
+            f"Found {len(detected_gaps)} timestamp gaps "
+            f"(detection threshold: {NcsSectionsFactory._maxGapSampFrac} sample intervals)\n\n"
+            "Gap Details:\n"
+            "+-----------------+-----------------------+-----------------------+\n"
+            "| Record Index    | Record at (Seconds)   | Gap Size (ms)         |\n"
+            "+-----------------+-----------------------+-----------------------+\n"
+            + "".join(gap_detail_lines)
+            + "+-----------------+-----------------------+-----------------------+\n"
+        )
```
zm711, line 954: I have a hard time reading this, but it looked fine in Intan so I trust you for this bit of code.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -878,6 +907,76 @@ def _rescale_event_timestamp(self, event_timestamps, dtype, event_channel_index)
...
+        np.ndarray
+            Timestamps in microseconds from the NCS records, one per 512-sample record.
+        """
+        stream_id = self.header["signal_streams"][stream_index]["id"]
+        stream_mask = self.header["signal_channels"]["stream_id"] == stream_id
+        channel = self.header["signal_channels"][stream_mask][0]
+        chan_uid = (channel["name"], channel["id"])
+
+        data = self._sigs_memmaps[seg_index][chan_uid]
+        return data["timestamp"].copy()
```
zm711, line 983: How expensive is this?

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -906,6 +1005,17 @@ def scan_stream_ncs_files(self, ncs_filenames):
         if len(ncs_filenames) == 0:
             return None, None, None
 
+        # Determine gap tolerance in microseconds for NcsSectionsFactory
+        if self._use_legacy_gap_mode:
```
zm711, line 1014: Again I'm not sure if I like this strategy overall is one I like. Rather than decide between legacy vs modern why not just make "modern" handle the legacy case so that we can slowly delete all legacy based code. You have this as a private variable which is better than making it public, but I think the strategy I would prefer would be:
```
def (a, b, c)
 if c is not None:
    b= c if c is x else y
```

This way right at the top we handle the legacy case by converting the b variable based on the deprecated c. Then when the deprecation window is over we just delete c and the top little bit of code.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -906,6 +1005,17 @@ def scan_stream_ncs_files(self, ncs_filenames):
...
+            # Legacy mode: let NcsSectionsFactory use strict_gap_mode defaults
+            factory_kwargs = {"strict_gap_mode": self.strict_gap_mode}
+        elif self.gap_tolerance_ms is not None:
+            # New API: convert ms to us
+            factory_kwargs = {"gap_tolerance_us": self.gap_tolerance_ms * 1000.0}
```
zm711, line 1019: This is what I'm trying to understand. Won't it be confusing for the user if they are requesting a gap in ms but then internal we do everything on us? If we give them an error in us they might be confused no?

neo/test/iotest/common_io_test.py
```diff
@@ -80,6 +80,7 @@ class BaseTestIO:
 
     entities_to_test = []  # list of files to test compliance
     entities_to_download = []  # when files are at gin
+    io_kwargs = {}  # extra kwargs passed to ioclass constructor
```
zm711, line 83: Yeah this seems good. If we are trying to add gap tolerance everywhere then it makes sense to allow us to slowly out io_kwargs. Or to play with the tolerance as needed. This looks good.

zm711 (commented):
I think my biggest question is the overall design choice here. could you explain 
1) why handle legacy vs modern as two different if/else pathways deep in the code vs just taking the legacy values and encapsulating them in the modern version
2) Why bounce back and further between ms and us? (mostly just to remind me). If you do switch back and further I think you'll need to make sure anything that reaches the user will show up as ms (not us)

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: The microseconds isn't a choice I'm making here, it's the factory's existing convention. `NcsSectionsFactory` has always worked in microseconds because NCS record timestamps are natively integer microseconds, and its gap math was already in us before this PR (the old parameter was just an unlabeled `gapTolerance`). I'm respecting that internal unit and only renamed it to `gap_tolerance_us` to make it explicit. The single ms->us conversion happens once at the reader boundary, so the user never leaves milliseconds; only the internal layer that was always in us stays in us. This is basically to respect the other contributor way of thinking inside the internals.

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: `gap_tolerance_us` is the factory-level parameter, not the user-facing one. End users go through `NeuralynxRawIO(gap_tolerance_ms=...)`; this deprecation message is only reachable by code that calls the factory's old `gapTolerance` directly, which is internal. Fair point on consistency though: I can make the message point end users at the reader's `gap_tolerance_ms` so there's no ambiguity about which unit belongs where.

neo/rawio/neuralynxrawio/neuralynxrawio.py
h-mayorquin: Cheap. It copies only the timestamp column: one `uint64` per 512-sample record, so roughly 1/512 of the data size (tens of KB for a typical file, not the sample array). The `.copy()` is deliberate so the returned array owns its memory and isn't a live view into the file memmap.

neo/rawio/neuralynxrawio/neuralynxrawio.py
h-mayorquin: The conversion here is `ms -> us` only to feed the factory. The user never sees us: the gap report is explicitly emitted in ms (gap size) and seconds (position in the recording), so an error is already in the unit they passed. I'll double-check every user-facing string stays in ms.

neo/test/iotest/common_io_test.py
h-mayorquin: Thanks.

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: Done.

neo/rawio/neuralynxrawio/neuralynxrawio.py
h-mayorquin: I consolidated the value translation into the factory (see above), but the reader-level `_use_legacy_gap_mode` flag can't be collapsed the same way, because it isn't only routing: it also guards a genuine behavior difference. In legacy mode (a user passed `strict_gap_mode`) the reader deliberately skips the new error-on-gaps default so deprecated code keeps its old behavior; in modern mode it applies error-by-default. Collapsing the flag would silently change what deprecated calls do. The reader also can't pre-translate `strict_gap_mode` into a numeric tolerance up front, because the value depends on per-file acquisition type and frequency that only the factory sees. So the reader forwards the deprecated parameter to the factory, where it's now handled in one place, and keeps the flag only to preserve legacy behavior.

h-mayorquin:
@zm711 can I merge this one?

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -256,27 +270,57 @@ def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=Tru
...
+                "Use `gap_tolerance_us` instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            if gap_tolerance_us is None:
+                gap_tolerance_us = gapTolerance
+
+        if strict_gap_mode is not None:
+            warnings.warn(
+                "The `strict_gap_mode` parameter is deprecated and will be removed in version 0.16. "
```
zm711: Our numbering for Neo is different than SI. So I'm not sure how to warn this since we might have ten more versions before 0.16 or at minimum we would have 2 more so it's hard for people to plan. I guess you're arguing we should keep this for two major versions instead of minor (since we aren't at 1.0 we are treating it as 0.major.minor in my terminology).

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -256,27 +270,57 @@ def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=Tru
...
         acqType = nlxHdr.type_of_recording()
         freq = nlxHdr["sampling_rate"]
 
+        # Deprecation shim for strict_gap_mode (the boolean predecessor of gap_tolerance_us).
+        # strict_gap_mode True/None already match the modern per-type defaults below; only
+        # strict_gap_mode=False differed, tolerating a quarter-packet gap (PRE4 was 0 either way).
+        # Translating that single case here keeps the per-type branches on gap_tolerance_us only.
+        # Remove this block when strict_gap_mode is dropped in v0.16.
+        if gap_tolerance_us is None and strict_gap_mode is not None and not strict_gap_mode and acqType != AcqType.PRE4:
+            gap_tolerance_us = round(0.25 * NcsSection._RECORD_SIZE * 1e6 / freq)
```
zm711, line 316: Do we need to control the round precision or just let it do what it wants with float?

neo/rawio/neuralynxrawio/ncssections.py
```diff
@@ -256,27 +270,57 @@ def build_for_ncs_file(ncsMemMap, nlxHdr, gapTolerance=None, strict_gap_mode=Tru
...
             sampFreqUsed = NcsSectionsFactory.get_freq_for_micros_per_samp(microsPerSampUsed)
-            if gapTolerance is None:
-                if strict_gap_mode:
-                    # this is the old behavior, maybe we could put 0.9 sample interval no ?
-                    gapTolerance = 0
-                else:
-                    gapTolerance = 0
-
-            ncsSects = NcsSectionsFactory._buildNcsSections(ncsMemMap, sampFreqUsed, gapTolerance=gapTolerance)
+            if gap_tolerance_us is None:
```
zm711, line 323: Should we have some protection to ensure gap_tolerance_us is positive? What if someone puts in -10 will that error or pass silently with nonsense?

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -196,11 +204,32 @@ def __init__(
         else:
             self.rawmode = "one-dir"
 
+        # Handle gap_tolerance_ms and deprecated strict_gap_mode
+        if strict_gap_mode is not None:
+            warnings.warn(
+                "`strict_gap_mode` is deprecated and will be removed in version 0.16. "
```
zm711: at minimum we should put version 0.16.0 to be clear what this means.

neo/rawio/neuralynxrawio/neuralynxrawio.py
```diff
@@ -196,11 +204,32 @@ def __init__(
...
+                "Use `gap_tolerance_ms` instead to control gap handling. "
+                "See issue #1773 for details.",
+                DeprecationWarning,
```
zm711: Based on Chris' discovery at the hackathon should we just a future warning rather than a dep warning since this is user facing?

zm711 (commented):
Few more comments @h-mayorquin that I'm just thinking about.

neo/rawio/neuralynxrawio/neuralynxrawio.py
h-mayorquin: Yes, I changed that.

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: No, it was a mistake, I changed it to 0.16.0 now.

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: That line is verbatim from master's strict_gap_mode=False branch, so keeping round() reproduces the deprecated behavior exactly. Precision isn't really at stake: at 30 kHz it's 4266.67 against 4267 us, and the comparison is on int64 microseconds, so rounding can only flip the verdict for that one integer.

neo/rawio/neuralynxrawio/ncssections.py
h-mayorquin: I think this makes sense and I added the assertion and a test.

neo/rawio/neuralynxrawio/neuralynxrawio.py
h-mayorquin: Also changed.

PR #1829, opened 2026-03-28
fixed overflow error in axonarawio when loading long recordings

LizzyMcKay:
changed lines 389 and 405 in axonarawio.py so that sample indices are cast as np.uint64


alejoe91 (approved):

zm711 (approved):
This looks fine to me too. I feel like we did a bunch of overflow squashing a while back, but I'm sure we are still missing a bunch. Thanks for the PR.

PR #1830, opened 2026-04-17
Check `num_chan_per_chip` to avoid `ZeroDivisionError`

MGAMZ:
In some scenarios, spikegadgets' recording have missing `chanPerChip` values, neo will default to 32 in this case.
In our practice, this value can be mistakenly recorded to zero, and will cause `ZeroDivisionError` while parsing header:

```python
# around line:224
if num_ephy_channels % num_chan_per_chip == 0
```

This PR adds a simple check for `num_chan_per_chip`, when it appears to be non-positive, raise KeyError which will be caught and replace with the original default value for Intan chips.

```python
try:
    num_chan_per_chip = int(sconf.attrib["chanPerChip"])
    if num_chan_per_chip <= 0:  # invalid value
        raise KeyError
    if num_chan_per_chip > 100000:  # abnormal large value
        raise KeyError
except KeyError:
    num_chan_per_chip = 32  # default value for Intan chips
```

# Further Discussion

The related codes are introduced in #1496. In that PR, the @khl02007 assumes the `num_chan_per_chip` is reliable and it influces the channel mapping. But in fact this value can be wrong due to Trodes BUGs, It can be zero or an abormally large value.

If the value is mistakenly recorded as zero, a `ZeroDivisionError` will be triggered (as seen in this PR). If it is recorded as a large value (such as 1645402192), the condition `num_ephy_channels % num_chan_per_chip == 0` is evaluated to True. Consequently, `python-neo` will treat it as a valid value and trigger the remapping of `channel_ids`. This is potentially dangerous because the data could be inaccurate without the user's awareness.

This PR only introduces some simple checks and overwrite the wrong value to defaults. However, the default values may still deviate from the actual situation. The judgment `num_ephy_channels % num_chan_per_chip == 0` seems to be not good for robustness. But I am not sure how to improve this. Also hoping @jnjnnjzch can give some suggestions.


jnjnnjzch:
Well, I think it may be OK to add a warning letting users know something is wrong... but I doubt the necessarity to “replace with the original default value for Intan chips”, replacing is a dangerous action.

h-mayorquin:
Test data incoming:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/171

h-mayorquin:
The gin data is merged, we should add a test concerning it.

h-mayorquin:
This was fixed by https://github.com/NeuralEnsemble/python-neo/pull/1847 I think. Please confirm if you can or we will close in a month if not activity.

MGAMZ:
Yes, I confirmed that my issue is solved! Thanks for your efforts! @h-mayorquin

h-mayorquin:
That's great, @MGAMZ 

One more ask as in the other thread

> One more ask while I have you both here: would either of you be willing to share a short recording from your rig to add to our test set? That would be really helpful. Our current NP1 fixture is from Trodes 2.3.2 (where chanPerChip="32" for Neuropixels), so we do not have any fixture that exercises the exact configuration that has been reported (Trodes 2.8.x with chanPerChip="0"). In fact, anything you can spare would help: NP1, NP2, single-probe, single-shank, whatever your default rig produces, in any way you can send it. I can take care of stubbing and anonymizing the data.

MGAMZ:
@h-mayorquin I'll see if I can provide one and may email you.

h-mayorquin:
Thanks a bunch.

PR #1835, opened 2026-04-20
TdtRawIO: detect SEV vs TEV using the UCF flag instead of silent per-channel fallback

### LINKED ISSUE #1817: TdtRawIO: Detect SEV vs TEV format using the UCF flag instead of per-channel silent fallback
I got here thinking on #1815 

Currently, `TdtRawIO._parse_header()` determines whether to read from SEV or TEV on a [per-channel basis](https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L308-L311): for each channel it globs for an SEV file, and if not found, [silently falls back to the TEV buffer](https://github.com/NeuralEnsemble/python-neo/blob/c672dbd5976a89ecaea980f4831304a12d1ffa5a/neo/rawio/tdtrawio.py#L285-L314). However, TEV vs SEV is a per-store configuration choice made at recording time via the "Discrete Files" checkbox in the Synapse Stream Data Storage:

Image

When enabled, stream data is written only to per-channel `.sev` files; the `.tev` file still exists (it is required for block validation) but does not contain valid stream data for that store. Because this is a deliberate configuration choice, there is no valid fallback from SEV to TEV: if an SEV file is missing, TEV will not have the data either.

The TDT format encodes the storage mode in the event type field of the TSQ header via the `EVTYPE_UCF` bit (`0x10`). The official TDT Python SDK (`tdt.TDTbin2py`, [line 1896](https://pypi.org/project/tdt/)) checks this flag and refuses to fall back to TEV when it is set:

```python
# from official TDT SDK TDTbin2py.py (you will need to install the package and look at the source)
if current_name in sev_names:
    d = read_sev(block_path, event_name=current_name, ...)
else:
    if data[current_type_str][current_name].ucf == 1:
        warnings.warn('Expecting SEV files but none were found, skipping...')
        continue  # do NOT fall back to TEV
```

Neo could adopt a similar approach: check the `EVTYPE_UCF` bit (already available in the TSQ `evtype` field that neo reads) once per store, and if set, require SEV files to be present rather than silently falling back to TEV. The advantage of this would be to replace the current per-channel globbing with an explicit format flag and turn a silent data corruption into a clear error.

h-mayorquin:
Closes #1817.

In `TdtRawIO._parse_header`, storage mode (SEV vs TEV) is a per-store configuration choice set at recording time via Synapse's "Discrete Files" option, and encoded in the TSQ event type via the `EVTYPE_UCF` bit (`0x10`). The previous implementation ignored this flag and instead globbed for an SEV file per channel, silently falling back to the TEV buffer when one was not found. Because TEV does not contain valid stream data for a UCF-flagged store, that fallback produced wrong data rather than an error.

The check now runs once per store using `info["TankEvType"] & EVTYPE_UCF`. When the bit is set, a missing SEV file raises `NeoReadWriteError` with the store name, channel, and segment, making the failure mode explicit. When the bit is not set, the existing TEV fallback path is preserved so non-UCF recordings continue to work unchanged. `EVTYPE_MASK` already excludes the UCF bit, so none of the existing stream filtering is affected.


h-mayorquin:
@JuliaSprenger we should add a flag that keeps the data loading for a while before we merge this.

zm711:
So we need to wait for you to update before we merge?

PR #1842, opened 2026-04-22
Spikeglx: use probe features from ProbeTable to infer Neuropixels type

alejoe91:
Following #1839, the SpikeGLX reader is updated to infer the 1.0/2.0 probe type from the `datasheet` field of the associated entry in the `noe/resources/neuropixels_probe_features.json`.

In case the datasheet is not 1.0/2.0, a warning is pronted and unitary gains are added.


chrishalcrow (approved):
This looks good to me, and also allows the new NP2020 data to load successfully.
(I know my approval has little weight here - but I'm giving it anyway!!)

alejoe91:
> This looks good to me, and also allows the new NP2020 data to load successfully. (I know my approval has little weight here - but I'm giving it anyway!!)

I'll take it!

neo/rawio/spikeglxrawio.py
```diff
@@ -679,36 +679,60 @@ def extract_stream_info(meta_file, meta):
...
-            # https://github.com/billkarsh/SpikeGLX/blob/15ec8898e17829f9f08c226bf04f46281f106e5f/Markdown/Metadata_30.md#imec
-            # We allow also LF streams for NP2.0 because CatGT can produce them
-            # See: https://github.com/SpikeInterface/spikeinterface/issues/1949
-            if "imChan0apGain" in meta:
-                per_channel_gain[:-1] = 1 / float(meta["imChan0apGain"])
+        probe_part_number = meta.get("imDatPrb_pn", None)
+        with open(neuropixels_probe_features_file, "r") as f:
+            probe_features = json.load(f)
+
+        if probe_part_number is not None:
```
h-mayorquin: The two outer `if probe_part_number is not None:` and `if features is not None:` wrappers can collapse if we mirror probeinterface's Phase3A remap and raise instead of warning. This also picks up the Phase3A case (`imProbeOpt` instead of `imDatPrb_pn`), which probeinterface handles [here](https://github.com/SpikeInterface/probeinterface/blob/2789811ce49c71dc684a0bf2fdf35729fb2f55b3/src/probeinterface/neuropixels_tools.py#L808-L811) but #1842 currently drops to unitary gain.

```suggestion
        probe_part_number = meta.get("imDatPrb_pn", None)
        if probe_part_number is None and meta.get("imProbeOpt") is not None:
            probe_part_number = "NP1010"  # Phase3A remap, matches probeinterface
        if probe_part_number is None:
            raise ValueError("Could not determine probe part number from metadata.")

        features = probe_features["neuropixels_probes"].get(probe_part_number)
        if features is None:
            raise ValueError(f"Probe part number {probe_part_number} not found in ProbeTable.")
```

This replaces the warn-and-unitary-gain fallbacks with hard failures. We were doing this on master already  with `NotImplementedError` for unsupported probes, and silently scaled-wrong traces are harder to catch downstream.

neo/rawio/spikeglxrawio.py
```diff
@@ -679,36 +679,60 @@ def extract_stream_info(meta_file, meta):
...
+                    # See: https://github.com/SpikeInterface/spikeinterface/issues/1949
+                    if "imChan0apGain" in meta:
+                        per_channel_gain[:-1] = 1 / float(meta["imChan0apGain"])
+                    else:
+                        per_channel_gain[:-1] = 1 / 80.0
+                    max_int = int(meta["imMaxInt"]) if "imMaxInt" in meta else 8192
+                    gain_factor = float(meta["imAiRangeMax"]) / max_int
+                    channel_gains = gain_factor * per_channel_gain * 1e6
+                else:
+                    warn(
```
h-mayorquin: Same argument applies inside the gain branch: I think we should raise an error Raising matches master's behaviour which I think is better than letting wrong gains silently propagate.

```suggestion
                else:
                    raise NotImplementedError(
                        f"Probe {probe_part_number} has datasheet {datasheet!r}, "
                        f"which is not currently supported by the SpikeGLX gain calculation. \n"
                        "Please open an issue at python-neo repo"
                    )
```

h-mayorquin: This will become easier with a logic like the one in:

https://github.com/SpikeInterface/probeinterface/pull/434

h-mayorquin (changes requested):
Thanks @chrishalcrow for testing this and sorry for taking so long @alejoe91. Actually, I was not sure if it was ready so I delayed it, sorry for assuming wrong. I have some requests regarding the handling of warnings vs error but otherwise LGTM.

Metacommentary:

After #1839 and #1842 we have our own copy of the json of the probe library. This is fine as long as both releases stay in lockstep with billkarsh's ProbeTable, but the cadence is controlled by two independent actions and two independent release cycles. A user with `neo==X` and `probeinterface==Y` where X and Y bracket a ProbeTable update might see diferences depending on versios. Honestly, it would be been simpler to just use probeinterface here rather than python-neo carrying its own copy. I do understand the French reason you probably decided against it though.

neo/rawio/spikeglxrawio.py
alejoe91: done in last commit

neo/rawio/spikeglxrawio.py
alejoe91: done in last commit

alejoe91:
TODO: find a more robust way to assess probe technology from probe features (maybe crossing description/databus/datasheet)

alejoe91:
@h-mayorquin I think this is ready now: I improved the logic to spot 1.0 probes and in case of fixed gain we use the probe features gain as fallback in case the gain is not in the meta file

neo/rawio/spikeglxrawio.py
```diff
@@ -68,6 +65,21 @@
 from .utils import get_memmap_shape
 
 
+neuropixels_probe_features_file = Path(__file__).parents[1] / "resources" / "neuropixels_probe_features.json"
+
+
+def is_1_0_probe(features):
```
h-mayorquin: I think this should be private.

h-mayorquin (approved):
This looks good. I would prefer the helper method to be private to keep the API clean. Specially that method as maybe a better way might come alone.

alejoe91:
> This looks good. I would prefer the helper method to be private to keep the API clean. Specially that method as maybe a better way might come alone.

private now! I'll merge as soon as tests are done

Issue #1844, opened 2026-05-07
chanPerChip is equal to 0 or random number in Trodes 2.8.2 config for Neuropixels 1.0 or 2.0


dkassybayeva:
When using spikeinterface (version 0.104.3) and neo (version 0.14.4) with recording done in Trodes 2.8.2 I ran into an error at this line in spikegadgetsrawio.py :
if num_ephy_channels % num_chan_per_chip == 0
where error was that I cannot divide by 0
looking into it I saw that in Trodes config file chanPerChip = 0, while in previous Trodes versions this chanPerChip would change from session to session to have values 2 or 6 or 32 (random). The solution previously was to put this in the script of spikegadgetsrawio.py:
```
# as spikegadgets change we should follow this
        try:
            num_chan_per_chip = int(sconf.attrib["chanPerChip"])
        except KeyError:
            num_chan_per_chip = 32  # default value for Intan chips
```
However, I don't think Neurpixels 1.0 or 2.0 probes even use Intan chips and there must be a reason why in new Trodes version chanPerChip is consistently 0.  num_ephy_channels is taken from 'numChannels' in Trodes configuration file and is 768. Since there are no Intan chips, should the calculation rely on hwChan? As you can see from the code below, code never gets to hwChan because num_chan_per_chip is 0. 
        ```
```
if num_ephy_channels > 0:
            stream_id = "trodes"
            stream_name = stream_id
            buffer_id = ""
            signal_streams.append((stream_name, stream_id, buffer_id))
            self._mask_channels_bytes[stream_id] = []
```

            # we can only produce these channels for a subset of spikegadgets setup. If this criteria isn't
            # true then we should just use the raw_channel_ids and let the end user sort everything out
            if num_ephy_channels % num_chan_per_chip == 0:
                all_hw_chans = [int(schan.attrib["hwChan"]) for trode in sconf for schan in trode]
                missing_hw_chans = set(range(num_ephy_channels)) - set(all_hw_chans)
                channel_ids = self._produce_ephys_channel_ids(
                    num_ephy_channels_xml, num_chan_per_chip, missing_hw_chans
                )
                raw_channel_ids = False
            else:
                raw_channel_ids = True
```

h-mayorquin:
Can you test: #1847

Issue #1810, opened 2026-01-15
Can not read the trodes .rec file using spikeinterface.extractors.read_spikegadgets()


Sachuriga:
I'm using an HH128 and the MCU for tethered recording in Trodes. The signal looks good, and I can play back the .rec file within Trodes without issues.
However, when I try to load the .rec file into SpikeInterface using spikeinterface.extractors.read_spikegadgets(), I get an error stating that the file does not contain the channel mapping header.

Trodes version: 2.6.0
spikeinterface: 0.103

errors:
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
Cell In[13], line 4
      2 from probeinterface import generate_tetrode, ProbeGroup
      3 # 1. Load the raw data (it will have NO geometry initially)
----> 4 recording = se.read_spikegadgets(r'C:\Users\gl_pc\Downloads\20260115_110218\20260115_110218\20260115_195453.rec/20260115_195453.rec')
      5 recording
File c:\Users\gl_pc\.conda\envs\si_trodes\Lib\site-packages\spikeinterface\extractors\neoextractors\spikegadgets.py:50, in SpikeGadgetsRecordingExtractor.__init__(self, file_path, stream_id, stream_name, all_annotations, use_names_as_ids)
     41 def __init__(
     42     self,
     43     file_path,
   (...)     47     use_names_as_ids: bool = False,
     48 ):
     49     neo_kwargs = self.map_to_neo_kwargs(file_path)
---> [50](file:///C:/Users/gl_pc/.conda/envs/si_trodes/Lib/site-packages/spikeinterface/extractors/neoextractors/spikegadgets.py:50)     NeoBaseRecordingExtractor.__init__(
     51         self,
     52         stream_id=stream_id,
     53         stream_name=stream_name,
     54         all_annotations=all_annotations,
     55         use_names_as_ids=use_names_as_ids,
     56         **neo_kwargs,
     57     )
     58     self._kwargs.update(dict(file_path=str(Path(file_path).absolute()), stream_id=stream_id))
     60     probegroup = None  # TODO remove once probeinterface is updated to 0.2.22 in the pyproject.toml
[...]

PR #1847, opened 2026-05-11
Fix Neuropixels handling in SpikeGadgetsRawIO

h-mayorquin:
SpikeGadgetsRawIO crashes with ZeroDivisionError when opening any Neuropixels .rec file from Trodes 2.8.2 onwards (closes issue #1844 and maybe #1810). Trodes writes chanPerChip="0" on those headers to signal that the recording has no Intan chip layout, and the existing reader divides by that value without a guard. While investigating, I found a separate silent bug in the existing NP1 data on gin: it writes chanPerChip="32" despite using a Neuropixels probe, so the old divisibility check passed and the file went through `_produce_ephys_channel_ids`, producing fabricated chip-style channel ids that did not match any real hardware channel.

The fix switches the discriminator from chanPerChip to the `SpikeConfiguration.device` attribute, which Trodes writes as "intan", "neuropixels1", or "neuropixels2" (absent on legacy files, which were always Intan). For Neuropixels recordings the reader now assigns each channel id `f"hwChan{i}"` and name `f"probe{spikeSortingGroup}_chan{i}"`, where `i` is the channel index in the trodes stream (which equals the hwChan the SpikeGadgets firmware writes at byte position `i`) and `spikeSortingGroup` is the value of the SpikeChannel XML attribute that Trodes uses to bucket channels for per-probe spike-sorting export. The `hwChan` prefix keeps the id unambiguously non-numeric so consumers cannot coerce it to an array index. For Intan recordings the existing chip-interleaved synthesis runs when chanPerChip is consistent. Unknown device strings raise `NeoReadWriteError`. That route fixes both the NP2 crash and the silent NP1 misrouting in one change.

The fix is exercised by adding the NP2 stub from gin (`spikegadgets/neuropixels2_4shank/...`) to `entities_to_test`, and by an updated test (`test_neuropixels_uses_hwchan_ids`) that pins the NP1 fixture's channel ids and names to the new convention so the labelling cannot regress.

[EDITED]

### DIFF
--- a/neo/rawio/spikegadgetsrawio.py
+++ b/neo/rawio/spikegadgetsrawio.py
@@ -140,12 +140,6 @@ class SpikeGadgetsRawIO(BaseRawIO):
                 "than the number of channels in the hardware configuration"
             )
 
-        # as spikegadgets change we should follow this
-        try:
-            num_chan_per_chip = int(sconf.attrib["chanPerChip"])
-        except KeyError:
-            num_chan_per_chip = 32  # default value for Intan chips
-
         # explore sub stream and count packet size
         # first bytes is 0x55
         packet_size = 1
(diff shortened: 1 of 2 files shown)


dkassybayeva:
If in old version channel ids were incorrect, does it mean the whole pre-processing, including kilosort, needs to be rerun for those sessions to get correct ids and probe positions? 
I.e. does it affect channelmap.prb construction?

alejoe91 (approved):

alejoe91:
> If in old version channel ids were incorrect, does it mean the whole pre-processing, including kilosort, needs to be rerun for those sessions to get correct ids and probe positions? I.e. does it affect channelmap.prb construction?

I don't think it will, but @h-mayorquin can confirm!

RobertoDF:
> > If in old version channel ids were incorrect, does it mean the whole pre-processing, including kilosort, needs to be rerun for those sessions to get correct ids and probe positions? I.e. does it affect channelmap.prb construction?
> 
> I don't think it will, but @h-mayorquin can confirm!

Having sorted many sessions of the older version, I can say the channel ordering looked fine

h-mayorquin:
Hi @dkassybayeva, let me be clear about what changed.

For Neuropixels 1, the ids and names in `signal_channels` are changing. They were incorrect: the reader was passing every recording through the same chip-interleaved-id path that the tetrode/Intan recordings need, and that path produces meaningless labels for Neuropixels (which has no Intan chips). Crucially, the data extraction path does not change. The output of `get_analogsignal_chunk` is byte-for-byte the same as before, and the same is true if you were extracting data via SpikeInterface's `get_traces()`. The only workflows that will be disrupted are those that were relying on specific channel id or name strings; workflows that select channels by positional index are not affected.

For your concrete question about Kilosort: as far as I am aware the channel map (`channelmap.prb` or the equivalent `chanMap.mat`) is purely positional, so if you are running Kilosort directly the sorting should not be affected. Are you running Kilosort directly, or through SpikeInterface? @RobertoDF you mentioned the channel ordering looked fine on the older code, does your pipeline go through SpikeInterface?

h-mayorquin:
One more ask while I have you both here: would either of you be willing to share a short recording from your rig to add to our test set? That would be really helpful. Our current NP1 fixture is from Trodes 2.3.2 (where `chanPerChip="32"` for Neuropixels), so we do not have any fixture that exercises the exact configuration that has been reported (Trodes 2.8.x with `chanPerChip="0"`). In fact, anything you can spare would help: NP1, NP2, single-probe, single-shank, whatever your default rig produces, in any way you can send it. I can take care of stubbing and anonymizing the data.

h-mayorquin:
All right, I am going to merge but I wanted to make a final clarification of what is changing regaridng names and ids:

| channel index | before id | before name | after id | after name |
|---|---|---|---|---|
| 0 | `'0'` | `hwChan0` | `hwChan0` | `probe0_chan0` |
| 1 | `'32'` | `hwChan32` | `hwChan1` | `probe0_chan1` |
| 32 | `'257'` | `hwChan257` | `hwChan32` | `probe1_chan32` |
| 767 | `'767'` | `hwChan767` | `hwChan767` | `probe1_chan767` |

And the reason we needed to change is because the previous ids for neurpixel were misleading as they were using a path that was inteded for intan. As I said above, positional based mappings should still work but id based ones would break (as they were probably broken or working by mistake).

PR #1857, opened 2026-05-22
Add linting to the CI setup, using ruff

apdavison:
To improve overall code quality, I suggest adding a linter to the CI setup. I've gone with ruff, because I've had good experiences with it (in particular, it is fast), but I'm open to consider a different linter. (Note that ruff also does formatting, it's a drop-in replacement for black)

For now the setup is quite forgiving, but we can toughen it up later.

The CI job first tests there are no regressions in code quality, then produces a report on things that could be improved (linter warnings and errors). As we gradually fix these, the rules can be made stricter.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,49 @@
+name: Ruff lint
+
+# Ruff runs on every PR and every push to master.
+#
+# Two steps:
+#   * Strict: runs the rule subset for which the codebase is clean today.
+#     Failures block the workflow, so regressions of these rules cannot be
+#     merged.
+#   * Backlog (non-blocking): runs the full configured ruleset (E, F, I) and
+#     surfaces the existing backlog of cleanups (import sorting, unused imports,
+#     long lines, etc.). It is marked `continue-on-error: true`; as the backlog
+#     is cleared the rules can be migrated into the strict step.
+
+on:
+  pull_request:
+    branches: [master]
+    types: [synchronize, opened, reopened, ready_for_review]
+  push:
+    branches: [master]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ruff-strict:
+    name: Ruff (strict — real-bug rules)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Run ruff (strict subset)
+        uses: astral-sh/ruff-action@v3
+        with:
+          args: "check --select F821,F403,F405 ."
+
+  ruff-backlog:
+    name: Ruff (full ruleset — non-blocking)
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/ruff.yml
```diff
@@ -0,0 +1,49 @@
...
+  ruff-strict:
+    name: Ruff (strict — real-bug rules)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Run ruff (strict subset)
+        uses: astral-sh/ruff-action@v3
+        with:
```
h-mayorquin, line 35:
```suggestion
        # Strict subset — only rules the repo already passes, so they can block merges:
        #   F821 - undefined name (real NameError-class bugs, typos, missing imports)
        #   F403 - `from module import *` used (hides which names are in scope)
        #   F405 - name may be undefined, or defined from a star import (F821 that F403 made unprovable)
        with:
          args: "check --select F821,F403,F405 ."
```

pyproject.toml
```diff
@@ -141,3 +141,14 @@ neo = ["resources/*.json"]
...
 [tool.black]
 line-length = 120
+
+[tool.ruff]
+line-length = 120
+target-version = "py310"
+extend-exclude = ["doc/old_stuffs", "doc/build", "dist", "tmp.py"]
+
+[tool.ruff.lint]
+select = ["E", "F", "I"]
```
h-mayorquin, line 151:
```suggestion
# See https://docs.astral.sh/ruff/rules/
select = [
    "E",  # pycodestyle errors — style/whitespace (line length, blank lines); large backlog
    "F",  # Pyflakes — real bugs (undefined names, unused/star imports, redefinitions)
    "I",  # isort — import sorting and grouping
]
```

h-mayorquin (approved):

h-mayorquin (approved):
This looks like a good idea to me. First, we check what we are enforcing so no regression and then we can eventually move others. 

Could we:
1) Have pre-commit configuration on the repo following this one so developers who use it can avoid this automatically (note this is not enforced as  installing is opt in) 
2) Unify this with the black action? this will reduce the number of workflows and ruff works as a formater anyway. 

Plus, it will reduce PR churn.

Both can be done in a separate PR.

.github/workflows/ruff.yml
h-mayorquin: I always forget what those rules mean so I like to document on the spot:

https://github.com/catalystneuro/neuroconv/blob/8d8348f2cbd0cdee74f2eec48d92db469c4c5888/pyproject.toml#L434-L442

Feel free to ignore.

zm711:
Also in favor of this. I think we remove black and unify to ruff for both. At this point Sam said he is fine for whatever as long as he doesn't have to install something. So this workflow is typically fine for him.

Although in discussion it is not clear the full value added since we already to black once a week we aren't gaining too much other than "more correct" PRs?

JuliaSprenger:
Hi @apdavison! As far as I see ruff will not provide automatic fixes, but will only flag inconsistent code layout. Won't this be as annoying in the long run as during the pep8speaks times? Or are you planning to switch also from black autoformatting to ruff autoformatting?

PR #1858, opened 2026-05-29
Intergrate changes to add micromed V5 reading

lmartinet:
I just added Version 5 of micromedio in micromedrawio and manage some encoding problems in the same file.


zm711:
Do we have a test file we can add?

zm711:
@lmartinet just wanted to ping you one more time. We are happy to allow this if you we can have a test file to make sure the reader will work.

PR #1866, opened 2026-06-24
Don't crash parsing ABF files with an invalid acquisition date

h-mayorquin:
Some ABF files store a "no date" sentinel (all bits set) in their date header fields, and `AxonRawIO` currently raises while building the recording datetime from those values, which blocks reading the file entirely even though the date is non-essential annotation. Both the ABF v1 and ABF v2 paths produce out-of-range values from this sentinel, so parsing now falls back to a `None` recording datetime rather than failing, keeping the signal accessible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,4 @@ B95.zip
 grouped_ephys
 uv.lock
 .venv
+.python-version
--- a/neo/rawio/axonrawio.py
+++ b/neo/rawio/axonrawio.py
@@ -665,21 +665,23 @@ def _parse_abf_v1(f, header_description):
     header["sProtocolPath"] = header["sProtocolPath"].replace(b"\\", b"/")
 
     # date and time
+    # ABF1 does not reliably store the calendar date here, so it is left at 1900-01-01 and only the
+    # time-of-day is read. A "no date" file writes the 0xFFFFFFFF sentinel, which reads as a
+    # negative time; a valid time-of-day is always in [0, 86400) seconds. Guard that range and fall
+    # back to rec_datetime=None, leaving any other value to build normally so real errors surface.
     YY = 1900
     MM = 1
     DD = 1
-    hh = int(header["lFileStartTime"] / 3600.0)
-    mm = int((header["lFileStartTime"] - hh * 3600) / 60)
-    ss = header["lFileStartTime"] - hh * 3600 - mm * 60
-    ms = int(np.mod(ss, 1) * 1e6)
-    ss = int(ss)
-    try:
-        header["rec_datetime"] = datetime.datetime(YY, MM, DD, hh, mm, ss, ms)
-    except (ValueError, OverflowError):
-        # Date/time header fields hold an out-of-range or "no date" sentinel
-        # (e.g. 0xFFFFFFFF). The acquisition date is non-essential annotation,
-        # so fall back to None rather than blocking the read of the signal.
+    seconds_per_day = 24 * 3600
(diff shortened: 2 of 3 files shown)


neo/rawio/axonrawio.py
```diff
@@ -673,7 +673,13 @@ def _parse_abf_v1(f, header_description):
     ss = header["lFileStartTime"] - hh * 3600 - mm * 60
     ms = int(np.mod(ss, 1) * 1e6)
     ss = int(ss)
-    header["rec_datetime"] = datetime.datetime(YY, MM, DD, hh, mm, ss, ms)
+    try:
+        header["rec_datetime"] = datetime.datetime(YY, MM, DD, hh, mm, ss, ms)
+    except (ValueError, OverflowError):
+        # Date/time header fields hold an out-of-range or "no date" sentinel
```
zm711: Should we warn people? I don't know this system well enough to know if warning a person that they didn't have an acquisition time is a marker of a "poor" acquisition or is something that the user should know they forgot to set up? What do you think? It's not actionable, but it might be valuable for the user to know. I just don't know the reader.

zm711 (commented):
Just one general question @h-mayorquin.

neo/rawio/axonrawio.py
h-mayorquin: I am honestly not sure a missing date is a real problem, so I would rather follow pyABF precedent here to just skip it. As far as I can tell it happens because the date field holds an unset/sentinel value (`0xFFFFFFFF`) that does not parse into a real date, and pyABF does not treat that as fatal either, it just falls back.

In addition, most users of neo do not really use this field so I don't think the warning will be useful for their purposes. Most of the datetime work was done to parse metadata for neuroconv where this is required and there it fails loudly so I think is Ok.

Btw, thanks for making me check this, I found another error that I will address.

h-mayorquin: https://github.com/NeuralEnsemble/python-neo/pull/1870

zm711: Comment from this PR actually applies to this one
https://github.com/NeuralEnsemble/python-neo/pull/1870#discussion_r3536504681

.gitignore
```diff
@@ -82,4 +82,5 @@ doc/*.nwb
 B95.zip
 grouped_ephys
 uv.lock
-.venv
\ No newline at end of file
+.venv
+.python-version
```
zm711, line 86: Interesting where did this come from?

h-mayorquin: this is a uv thing.

h-mayorquin:
At the end:

For version 2 I replaced the `try/except` with an explicit check for the two "no date" sentinel values (`0` and `0xFFFFFFFF`) on `uFileStartDate`, falling back to `rec_datetime = None` only for those and letting any other out-of-range value raise as before.

For version 1 I did the equivalent, but on the time field rather than the date field. That path hardcodes the date to 1900 and only reads `lFileStartTime`, where `0xFFFFFFFF` comes through as `-1` and `0` is a legitimate midnight. So instead of matching the two literal values (which would wrongly reject a real midnight), I check that the value is a valid time-of-day (`0 <= lFileStartTime < 86400`). Same effect: the sentinel falls back to `None`, everything else builds normally, and there's no `try/except`.

Both are covered by a test asserting that `invalid_date_abf1.abf` and `invalid_date_abf2.abf` parse with `rec_datetime = None`.

h-mayorquin:
This should be ready.

zm711 (approved):
Good by me.

PR #1868, opened 2026-06-24
ABF: Deduplicate non-unique signal channel ids instead of refusing the file

h-mayorquin:
`BaseRawIO` requires channel ids to be unique within a stream so a channel can be addressed by id, and currently raises `signal_channels do not have unique ids for stream {i}` the moment two channels collide. That turns a metadata defect into an unreadable file: some recordings, for example ABF files re-saved by Clampex, carry duplicate channel ids while the original export loads fine, and there is no way for a caller to get at the perfectly good signal behind the collision. The check lives in core `rawio`, so this affects every format, not just ABF.

This change makes the duplicates unique on the spot by suffixing them, warns that it did so, and proceeds, so the data stays readable and id-based channel selection keeps working. Well-formed files are untouched: the path only runs when ids actually collide, so the common case is unchanged and cannot newly warn or fail. Regression coverage uses a purpose-crafted fixture on gin whose channels all share one id, asserting the file now parses with a warning and that every channel ends up with a unique id.

### DIFF
--- a/neo/rawio/axonrawio.py
+++ b/neo/rawio/axonrawio.py
@@ -190,11 +190,23 @@ class AxonRawIO(BaseRawWithBufferApiIO):
                 t_start = t_start * fSynchTimeUnit * 1e-6
             self._t_starts[seg_index] = t_start
 
-        # Create channel header
+        # Create channel header. By default assume channels 0..nbchannel-1 were sampled in order,
+        # which is always the case for version >= 2.0. For version < 2.0 the channel ids come from
+        # nADCSamplingSeq (the ADC sampling sequence) and are also used to index the per-channel
+        # metadata (name, units, gain). Some re-saved exports corrupt this sequence so every entry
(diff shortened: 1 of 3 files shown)


zm711:
Could you walk me through why this happens with ABF? Why are the ids the same?

zm711:
We discussed this together and we have decided not to alter `baserawio` instead this appears to be an export error that corrupts the IDs such that they are all:

```python
info['nADCSamplingSeq']
array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
```

We want to deal with this locally in `axonrawio`. Since for version > 2.0 we already just assume `chan_ids` are `range(1, nb_channel)` we can add a check for version < 2.0 and if `len(np.unique(['nADCSamplingSeq']))==1` then we will plan to just do the same default of just creating a sequence of values.

h-mayorquin:
Done.

zm711 (approved):

PR #1870, opened 2026-06-25
ABF Fix v1 date parsing

h-mayorquin:
I realized that we were hard-coding this.

### DIFF
--- a/neo/rawio/axonrawio.py
+++ b/neo/rawio/axonrawio.py
@@ -665,15 +665,24 @@ def _parse_abf_v1(f, header_description):
     header["sProtocolPath"] = header["sProtocolPath"].replace(b"\\", b"/")
 
     # date and time
-    YY = 1900
-    MM = 1
-    DD = 1
+    # lFileStartDate is a YYYYMMDD-packed integer, parsed the same way as uFileStartDate in ABF2.
+    # An unset/sentinel value (0 or 0xFFFFFFFF) yields an out-of-range date that the try/except
+    # below turns into rec_datetime=None.
+    YY = int(header["lFileStartDate"] / 10000)
+    MM = int((header["lFileStartDate"] - YY * 10000) / 100)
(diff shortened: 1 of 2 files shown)


neo/rawio/axonrawio.py
```diff
@@ -665,15 +665,24 @@ def _parse_abf_v1(f, header_description):
...
+    MM = int((header["lFileStartDate"] - YY * 10000) / 100)
+    DD = int(header["lFileStartDate"] - YY * 10000 - MM * 100)
     hh = int(header["lFileStartTime"] / 3600.0)
     mm = int((header["lFileStartTime"] - hh * 3600) / 60)
     ss = header["lFileStartTime"] - hh * 3600 - mm * 60
     ms = int(np.mod(ss, 1) * 1e6)
     ss = int(ss)
-    header["rec_datetime"] = datetime.datetime(YY, MM, DD, hh, mm, ss, ms)
+    try:
+        header["rec_datetime"] = datetime.datetime(YY, MM, DD, hh, mm, ss, ms)
```
zm711: I'm wondering if we could check for the two bad values instead of try-except which can mask errors. What if there is another error mode that we miss because of this try-except that we would catch if we explicitly check for 0 or 0xFFFF... before attempting the date time.datetime.

zm711:
This is ready for review now @h-mayorquin ?

h-mayorquin:
Yes.

h-mayorquin:
The original goal here was small: stop hardcoding the version 1 acquisition date (it always came back as 1900-01-01) and read the real `lFileStartDate` field instead. The "no date" sentinel (`0xFFFFFFFF`) was crashing `parse_header()` outright. Fixed for both version 1 and version 2 in #1866, using the explicit two-value check you suggested rather than a try/except that could mask other errors.

zm711 (approved):

PR #1878, opened 2026-07-15
OpenEphysBinaryRawIO: detect timestamp gaps and add gap_tolerance_ms

h-mayorquin:
`OpenEphysBinaryRawIO` previously had no gap detection (there was a literal `# TODO for later : gap checking`), so a within-recording discontinuity, for example from a dropped USB packet, was silently treated as contiguous and every sample after the gap got the wrong time. This adds gap detection using the per-sample sample indices the format already stores in `sample_numbers.npy` (v0.6+) or `timestamps.npy` (pre-v0.6): any non-unit increment is a gap. Detection is exact, since these are integer sample counters with no jitter.

The interface matches the cross-reader gap API from #1773 that is already in Blackrock and Neuralynx. `gap_tolerance_ms=None` (the default) raises a `ValueError` with a per-gap report, so the reader never hands back a silently mis-timed recording; a float segments the recording at gaps larger than the threshold and absorbs smaller ones (`0.0` segments on every gap); and `ignore_integrity_checks=True` bypasses the check for fast loads or corrupt-file recovery. It also adds `_get_openephysbinary_timestamps` for direct access to the raw per-sample indices, matching `_get_blackrock_timestamps` and `_get_neuralynx_timestamps`.

One intended behavior change: loading a recording that contains a gap now errors by default instead of returning a single silently-concatenated segment. Clean recordings are unaffected; code that was unknowingly reading a gapped recording will need to pass `gap_tolerance_ms` or `ignore_integrity_checks=True`. The tests synthesize gaps from a real clean fixture (`openephysbinary/v0.6.x_neuropixels_with_sync`), since no public OpenEphys recording with a real gap exists.

@alejoe91


h-mayorquin:
1) Think about how are segments right now (they correspond to experiments)
2) Think about the `ignore_integrity_checks`

Issue #1865, opened 2026-06-24
Wrong offset in Biocam sparse reader ?


b-grimaud:
When the sparse reader for 3Brain file was implemented (see #1446), the "synthetic noise" option had an offset of +2048 since noise levels were stored around 0. It was assumed that, after a 12-bits unsigned to signed conversion, the generated noise would end up around the baseline (see issue #1743).

However, I've recently noticed odd-looking units if I run sorters on compressed data.
Checking out the traces, it looks like the offset isn't valid anymore.

Through SI : 

```
recording = se.read_biocam(path,
                            mea_pitch=60,
                            electrode_width=21, 
                            fill_gaps_strategy="zeros"
                            ) 
recording = spre.unsigned_to_signed(recording, bit_depth=12)
recording.get_traces(start_frame=10*fs, end_frame=11*fs, channel_ids=['421'])
```
Plots the following : 

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/3856ef6e-3bf2-42fe-a5e1-948300946044" />

Whereas with "synthetic_noise" : 

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/8160cb27-1e4d-4c53-88fe-5924c62e29de" />

Looking through `NoiseMean` in the HDF5 shows that indeed the noise is now stored around 2048.

I think they might have changed something internally about how noise is encoded. I've also noticed that re-exporting compressed recording to "raw" through BrainWave 5 results in a different HDF5 structure than with BrainWave 6, and the latter cannot be read with neo. I'll open a separate PR, but there's definitely  something different on the way they save data.

@mahlzahn Have you noticed anything different with recent recordings ? We keep our BrainWave up to date, and I started noticing the issue on recordings from January 2026.

Running latest on Ubuntu 24.04.

zm711:
It seems like at minimum we should have a check saying that we can't read version 6 for now while you both work on a fix?

b-grimaud:
The issue with BrainWave 6 looks pretty simple, I will open a PR here as well as a small one for ProbeInterface.

Regarding the offset, I've reached out to 3Brain to know when the change took place on their end. If we can pinpoint a specific version, this can be checked from the metadata and the offset can adapt accordingly. The other option is to look at the average of `NoiseMean`, earlier recordings should be ~0 and newer ones ~2048, but setting some kind of threshold around those values doesn't feel very stable.

PR #1882, opened 2026-07-22
Add support for BioCAM files recorded with BrainWave 6

b-grimaud:
As mentionned in #1865, the newer versions of 3Brain's recording software (BrainWave 6) stores metadata slightly differently than before (BrainWave 5).

I found this tweak in a local branch, but I don't remember writing it.
@mahlzahn is this yours ? I don't know where else I could have found it.

I'll also open a corresponding PR in probeinterface.


zm711:
Looks good to me. Do you have any tiny test files so we can make sure it works moving forward?

PR #1885, opened 2026-07-23
BiocamRawIO: Fix a performance regression (and address #1532)

kevindoran:
Previous commits (starting around 1814bc1e) relating to #1532 made a performance regression by forcing an array copy. It causes ~25x slowdown on my machine. 

Here we remove the regression, and properly chunk-read in the event of smaller subsets of large arrays being requested, which is a fix for the problem from #1532.

Edit: I previously got confused with a related issue in SpikeingInterface, [#3303](https://github.com/SpikeInterface/spikeinterface/issues/3303), which I think is connected.

### DIFF
--- a/neo/rawio/biocamrawio.py
+++ b/neo/rawio/biocamrawio.py
@@ -25,6 +25,10 @@ import warnings
 from neo.core import NeoReadWriteError
 
 
+# Maximum bytes to read at once.
+_MAX_READ_BYTES = 64 * 1024 * 1024  # 64 MB
+
+
 class BiocamRawIO(BaseRawIO):
     """
     Class for reading data from a Biocam h5 file.
@@ -136,12 +140,8 @@ class BiocamRawIO(BaseRawIO):
             raise ValueError("`stream_index must be 0")
         return self._segment_t_start(block_index, seg_index)
 
-    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes):
-        if i_start is None:
-            i_start = 0
-        if i_stop is None:
-            i_stop = self._num_frames
-
+    def _read_block(self, i_start, i_stop):
+        """Read frames [i_start, i_stop) for all channels, as (n_frames, n_channels)."""
         # read functions are different based on the version of biocam
         if self._read_function is readHDF5t_brw4_sparse:
             if self._fill_gaps_strategy is None:
@@ -163,34 +163,45 @@ class BiocamRawIO(BaseRawIO):
         else:
             data = self._read_function(self._filehandle, i_start, i_stop, self._num_channels)
 
-        # older style data returns array of (n_samples, n_channels), should be a view
-        # but if memory issues come up we should doublecheck out how the file is being stored
-        if data.ndim > 1:
-            if channel_indexes is None:
-                channel_indexes = slice(None)
-            sig_chunk = data[:, channel_indexes]
+        # Newer style data is read as a flat (n_samples * n_channels) array. The h5py read is
+        # C-contiguous, so this reshape is a free view: it copies nothing and allocates nothing.
+        # Older style data is already (n_samples, n_channels).
+        if data.ndim == 1:
+            data = data.reshape(i_stop - i_start, self._num_channels)
+        return data
 
-        # newer style data returns an initial flat array (n_samples * n_channels)
-        # we iterate through channels rather than slicing
-        # Due to the fact that Neo and SpikeInterface tend to prefer slices we need to add
-        # some careful checks around slicing of None in the case we need to iterate through
-        # channels. First check if None. Then check if slice and only if slice check that it is slice(None)
+    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes):
+        if i_start is None:
+            i_start = 0
+        if i_stop is None:
+            i_stop = self._num_frames
+        n_frames = i_stop - i_start
+        # Read a single sample to determine the number of selected channels.
+        if channel_indexes is None:
+            channel_indexes = slice(None)
+        n_ch = self._read_block(i_start, i_start + 1)[:, channel_indexes].shape[1]
+        dtype = self.header["signal_channels"]["dtype"][0]
+        itemsize = np.dtype(dtype).itemsize
+        read_bytes = n_frames * self._num_channels * itemsize
+        out_bytes = n_frames * n_ch * itemsize
+
+        # Reading a subset of channels from a very large recording.
+        # The BioCam hdf5 layout is such that all channels are chunked together, so they cannot be
+        # read independently. In the event that only a subset of channels of a very large recording
+        # is needed, we cannot slice a single channel. Instead, we must loop and read all channels,
+        # slicing as we go. We will do this chunking if the read is large and the output to be
+        # returned is smaller.
+        read_by_chunks = read_bytes > _MAX_READ_BYTES and out_bytes < read_bytes // 2
+        if read_by_chunks:
+            # Read in chunks, copying desired channels into a preallocated array.
+            out = np.empty((n_frames, n_ch), dtype=dtype)
+            max_frames = max(1, _MAX_READ_BYTES // (itemsize * self._num_channels))
+            for start in range(i_start, i_stop, max_frames):
+                stop = min(start + max_frames, i_stop)
+                out[start - i_start : stop - i_start] = self._read_block(start, stop)[:, channel_indexes]
         else:
-            if channel_indexes is None:
-                channel_indexes = [ch for ch in range(self._num_channels)]
-            elif isinstance(channel_indexes, slice):
-                start = channel_indexes.start or 0
-                stop = channel_indexes.stop or self._num_channels
-                step = channel_indexes.step or 1
-                channel_indexes = [ch for ch in range(start, stop, step)]
-
-            sig_chunk = np.zeros((i_stop - i_start, len(channel_indexes)), dtype=data.dtype)
-            # iterate through channels to prevent loading all channels into memory which can cause
-            # memory exhaustion. See https://github.com/SpikeInterface/spikeinterface/issues/3303
-            for index, channel_index in enumerate(channel_indexes):
-                sig_chunk[:, index] = data[channel_index :: self._num_channels]
-
-        return sig_chunk
+            out = self._read_block(i_start, i_stop)[:, channel_indexes]
+        return out
 
 
 def open_biocam_file_header(filename) -> dict:
--- a/neo/test/rawiotest/test_biocamrawio.py
+++ b/neo/test/rawiotest/test_biocamrawio.py
@@ -4,6 +4,11 @@ Tests of neo.rawio.BiocamRawIO
 
 import unittest
 
+import numpy as np
+import pytest
+import h5py
+
+from neo.rawio import biocamrawio
 from neo.rawio.biocamrawio import BiocamRawIO
 from neo.test.rawiotest.common_rawio_test import BaseTestRawIO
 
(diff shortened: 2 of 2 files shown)


zm711:
Thanks for this @kevindoran. Seems like in trying to fix the original issue I may have kept it the same/made it worse. Appreciate the fix. Could we have you post some sort of benchmark in a comment about the improvement in performance (more like a terminal screenshot or copy of output from the terminal rather than just saying 25x)? 

I would propose that you

1) keep the performance improvement in this PR
2) split off the performance regression test into a separate PR

I suggest this because we haven't discussed as maintainers if we want to support performance testing specifically and in general we try to avoid monkey patching as much as we can. So if you open that test as a separate PR that will give us an opportunity to discuss it as a concept for Neo without preventing us from merging this PR (so long as tests pass).  How's that sound?

kevindoran:
The test is a logical test; it doesn't measure execution time or memory usage. The monkey patch just adds a counter: it counts how many read calls there is, which is something we can deduce and assert logically, for example, it shouldn't be called multiple times when the whole stimulus is requested, but should be called multiple times when a small number of channels are requested.

If it was a performance test that measure execution time (and differ between machines), I agree that it wouldn't really belong in the codebase at the moment, as there isn't a dedicated machine they could run on.

kevindoran:
Regarding the 25x. I was doing preprocessing and a spike sort that required 3 full reads of the recording. With the fix, subsequent reads hdf5 can use read cache in ram (I think). So exact speed-ups will depend on how many reads, and the size of the recording relative to available ram (and disk and cpu).
```

Phase_01.brw, 8 x 65536-frame chunks = 4.29 GB per pass
timeit -n 1 -r 3

script        time                          throughput   multiple
stock_warm    ~3.6 min (prime + 3 passes)    0.080 GB/s   1x
fixed_warm    ~8 s (prime + 3 passes)        2.03 GB/s    25.4x
stock_cold    ~2.7 min (3 passes)            0.079 GB/s   1x
fixed_cold    ~13 s (3 passes)               0.97 GB/s    12.3x
```

zm711:
Thanks for the performance stats that's good enough for my purposes. We just like to have records associated with the PR so we can go back and try to assess things in the future. I think I would still prefer the monkey patch to be a separate proposal. But I'm willing to be overruled if @alejoe91 and @h-mayorquin are cool with it.

neo/rawio/biocamrawio.py
```diff
@@ -163,34 +163,45 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
+    def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, stream_index, channel_indexes):
+        if i_start is None:
+            i_start = 0
+        if i_stop is None:
+            i_stop = self._num_frames
+        n_frames = i_stop - i_start
+        # Read a single sample to determine the number of selected channels.
+        if channel_indexes is None:
+            channel_indexes = slice(None)
+        n_ch = self._read_block(i_start, i_start + 1)[:, channel_indexes].shape[1]
```
alejoe91:
```suggestion
        if channel_indexes is None:
            channel_indexes = slice(None)
            n_ch = self._num_channels
        else:
            n_ch = len(list(channel_indexes)
```
`n_ch` is available from the parse header, or is the length of channel indices. The only thing this logic doesn't cover is if we explicitly pass a slice as `channel_indexes`

kevindoran: Yea, if you don't pre-fetch, you need to handle all cases of slice (or don't support slice). Seems a shame to drop support for slice.

neo/rawio/biocamrawio.py
```diff
@@ -25,6 +25,10 @@
 from neo.core import NeoReadWriteError
 
 
+# Maximum bytes to read at once.
+_MAX_READ_BYTES = 64 * 1024 * 1024  # 64 MB
```
alejoe91: @kevindoran what is the rationale behinf fixing this to 64MB? 
It seems quite small...

neo/test/rawiotest/test_biocamrawio.py
```diff
@@ -23,5 +28,142 @@ class TestBiocamRawIO(
     ]
 
 
+def write_minimal_brw(path, n_ch, n_frames, version=102):
```
alejoe91: I see where these tests are coming from, but in general we tend to only have tests on real data. 

I would remove these tests and port the `test_values` and `test_chunked_read_matches_direct_read` using the existing test files: https://github.com/kevindoran/python-neo/blob/989423851488cb91a5cc639d32ba50bc801d077b/neo/test/rawiotest/test_biocamrawio.py#L26-L27

Could you do that?

neo/rawio/biocamrawio.py
alejoe91: Maybe this could be a class attribute and we could have a class method `set_max_read_bytes` to increase it/decrease it?

How is performance affected by it? The smaller this `_MAX_READ_BYTES` is, the lower the temporary RAM usage (final RAM usage is dictated by start/end frames and number of channels requested), but also the lower the speed. Could you test a couple of lager values to measure this?

alejoe91 (changes requested):
Thanks @kevindoran 

I think this is the right way to go!! I just added a couple of comments columns that I think will make the implementaion more performant and easier to maintain (test-wise)

neo/rawio/biocamrawio.py
kevindoran: I tried a sweep, and there was no improvement past 64 MB. I thought of adding as an attribute, but opted to just keep the change small and simple.

neo/rawio/biocamrawio.py
kevindoran: You will be definitely reading that [start, start+1] bit anyway, in the very next read, so I don't think there is any  performance implications of doing the pre-fetch to check shape.

neo/test/rawiotest/test_biocamrawio.py
kevindoran: Following the spec, it's a real brw file. It's much easier to follow unit tests that are self contained and don't rely on the format of some non-human readable file. If the project maintainers wish a different style, go for it. I'll tap out at this point.

neo/rawio/biocamrawio.py
h-mayorquin: Unresolving here.

For the first point:

`slice.indices(n)` is stdlib and does exactly this job. One line covers every form.

For the second point, I don't understand, are we caching the response somehow?

neo/test/rawiotest/test_biocamrawio.py
h-mayorquin: There are three reason for our preference for files. This can admit edge cases but:
* The files on the test library are a public good and can be used for other libraries. This is good and make our job easier.
* Encapsulation. The test carries the format's internals inline, so understanding a failure means reading the writer too. I'd rather reason about "a file with these characteristics fails" than learn a second implementation before I can read the test.
* Independence. A file built in the test code shares the reader's assumptions, so it can't contradict them.

If you want to tap out I am fine with deferring this. I can add the file with these characteristics to gin, do a PR with the requested modifications and credit you.

neo/rawio/biocamrawio.py
kevindoran: Only remaining case is if a bool mask is passed.

kevindoran: For the second point, hdf5 has a read cache, and below that the OS cache.

neo/rawio/biocamrawio.py
h-mayorquin: I agree with @alejoe91 that this should be an attribute rather than a module-level constant. Beyond giving users an escape hatch, the sweep below says one value cannot serve both layouts, so the knob ends up doing real work rather than being configurable for its own sake.

I generated a large synthetic file using the same approach as the writer in your tests, 4096 channels by 80000 frames, so a 625 MiB full all-channel read, and swept the budget for a whole-recording request of 3 channels. Same sweep against the event-based sparse file already in the test data.

image

Two things come out of it.

The budget is too large rather than too small. On the dense layout the minimum sits around 4 MiB at 102 ms, against 381 ms at 64 MiB, so the current value is about 3.7x off. The flat region above 64 MiB is real and matches what you found, but the improvement is below it, not above. Peak allocation is exactly `budget + output` at every point I measured, so a smaller budget is also a tighter guarantee: 4.5 MiB rather than 64.5 MiB on the same request.

The sparse reader wants the opposite, because its cost is per call rather than per byte. `decode_event_based_raw_data` re-reads `TOC` and `EventsBasedSparseRawTOC` and re-walks the event blob for the covering blocks on every call, so splitting multiplies a fixed cost: 1159 ms at 1 MiB against 183 ms unchunked. It does still need the ceiling, though. `readHDF5t_brw4_sparse` allocates `np.zeros((nch, num_frames))` before it returns, so the sparse path has the same out-of-memory exposure as dense and the compression only helps on disk.

That suggests two defaults keyed on the read function:

```python
class BiocamRawIO(BaseRawIO):
    # Ceiling on the temporary all-channel buffer. The dense layouts read a contiguous
    # slice, so cost scales with bytes and a small buffer stays in cache. The event-based
    # sparse decoder re-reads both tables of contents and re-walks the event blob on every
    # call, so its cost is per call rather than per byte and it wants a buffer large enough
    # that it almost never splits.
    max_read_bytes_dense = 4 * 1024 * 1024
    max_read_bytes_sparse = 512 * 1024 * 1024
```

```python
budget = (
    self.max_read_bytes_sparse
    if self._read_function is readHDF5t_brw4_sparse
    else self.max_read_bytes_dense
)
```

neo/rawio/biocamrawio.py
h-mayorquin: Bool masks never reach `_get_analogsignal_chunk`. The base class converts them to integer indices with `np.nonzero` first, and raises if the length is wrong: [baserawio.py#L988-L995](https://github.com/NeuralEnsemble/python-neo/blob/35cbce7f44cb01e22fdb43d9a8803e0d0a6bed4f/neo/rawio/baserawio.py#L988-L995). So what arrives is only `None`, a slice, or an integer array. If you want it robust to direct calls on the private method anyway, `np.arange(self._num_channels)[channel_indexes].size` covers every case including bool in one expression.

On the caching, the bytes are cached but that is not where the cost is:

* `Raw` is stored contiguously in every BioCam file (`chunks=None` on all three in the test data), so HDF5's chunk cache never applies. Only the OS page cache does, and that saves the disk read, not the h5py call, the allocation, or the copy out of the page cache.
* My numbers were already warm. Warmup call, then 11 to 15 reps, median. On the v3.x flat file the probe costs 0.101 ms against 0.101 ms for a full 1000-frame read. On the sparse file it costs a flat 1.72 ms against 2.81 ms for a 100-frame read.
* The sparse cost is Python-level decoding: reading both TOCs and walking every record in the covering block. No cache touches that.

Either way, the probe has to go for a reason that is not about speed. On `i_start == i_stop == num_frames` it reads one frame past the end, gets an empty array back, and [`_read_block`](https://github.com/kevindoran/python-neo/blob/989423851488cb91a5cc639d32ba50bc801d077b/neo/rawio/biocamrawio.py#L169-L170) reshapes it to `(1, n_channels)`:

```
ValueError: cannot reshape array of size 0 into shape (1, 4096)
```

Master returns `(0, n_ch)` there. The real read handles the empty case on its own, since `reshape(0, n_ch)` is legal, so computing `n_ch` arithmetically fixes the crash and the overhead together.

neo/test/rawiotest/test_biocamrawio.py
kevindoran: Please feel free to do as you wish! I just want the performance improvement on master.

h-mayorquin: @alejoe91 @zm711 here is the data:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/182

alejoe91: Merged on GIN ;)

h-mayorquin:
Ok, I changed the tests and the other points of discussion. This can be merged here so @kevindoran gets credit.

zm711:
@alejoe91 could you do the review on this now. I'm going to be pure clinical for the next month.

alejoe91 (approved):

alejoe91:
> @alejoe91 could you do the review on this now. I'm going to be pure clinical for the next month.

LGTM! After @h-mayorquin change to test on data from GIN I thinkw e can merge :)

h-mayorquin:
I am merging then.

PR #1886, opened 2026-07-27
fix: create signal and waveform views with reshape

stanbot8:
Signal and waveform creation changes NumPy array dimensions through attribute assignment. Create 2D signal and 3D waveform views with reshape. This keeps both data paths on the supported NumPy array API and avoids the NumPy 2.5 deprecation. All tests pass.

### DIFF
--- a/neo/core/analogsignal.py
+++ b/neo/core/analogsignal.py
@@ -211,7 +211,7 @@ class AnalogSignal(BaseSignal):
         obj = pq.Quantity(signal, units=units, dtype=dtype).view(cls)
 
         if obj.ndim == 1:
-            obj.shape = (-1, 1)
+            obj = obj.reshape(-1, 1)
 
         if t_start is None:
             raise ValueError("`t_start` cannot be None")
--- a/neo/rawio/alphaomegarawio.py
+++ b/neo/rawio/alphaomegarawio.py
@@ -911,8 +911,7 @@ class AlphaOmegaRawIO(BaseRawIO):
                         offset=file_position,
                     )
                     i += 1
-        waveforms.shape = nb_spikes, 1, spike_length
-        return waveforms
+        return waveforms.reshape(nb_spikes, 1, spike_length)
 
     def _event_count(self, block_index, seg_index, event_channel_index):
         event_id = int(self.header["event_channels"]["id"][event_channel_index])


zm711 (approved):
This seems okay to me Andrew. Our previous strategy forces a reshape as a view rather than as a copy (the old behavior would error if a copy would be required for a particular reshape), but if NumPy is dropping support for using that trick I think it is fine to do just do the reshape and if a copy of the array is required then do it and take a computational hit. If you have a different idea feel free to suggest that.

alejoe91:
LGTM too, just a minor comment for @stanbot8

neo/test/rawiotest/test_alphaomegarawio.py
```diff
@@ -28,13 +28,16 @@ class and a full test will be done to test if the new coded IO
...
+import pytest
+
 from neo.rawio.alphaomegarawio import AlphaOmegaRawIO
 
 from neo.test.rawiotest.common_rawio_test import BaseTestRawIO
 
 logging.getLogger().setLevel(logging.INFO)
 
 
+@pytest.mark.filterwarnings("error:Setting the shape on a NumPy array has been deprecated:DeprecationWarning")
```
alejoe91: these should not warn anymore, right?
```suggestion
```

neo/test/coretest/test_analogsignal.py
```diff
@@ -65,6 +66,7 @@ def test__create_from_2d_list(self):
         self.assertEqual(signal.t_stop, len(data) / rate)
         self.assertEqual(signal[9, 0], 9000 * pq.uV)
 
+    @pytest.mark.filterwarnings("error:Setting the shape on a NumPy array has been deprecated:DeprecationWarning")
```
alejoe91: these should not warn anymore, right?
```suggestion
```

alejoe91 (changes requested):
Thanks @stanbot8 

Just a minor comment with suggestion. We can remofe the pytest filter warings right?

stanbot8:
@alejoe91 Right, they don't warn anymore. The `error:` filter is a regression guard that turns the shape-assignment `DeprecationWarning` into a test failure if it ever comes back, since the suite doesn't escalate warnings globally. I can drop both if you'd rather keep the tests clean.

alejoe91:
> @alejoe91 Right, they don't warn anymore. The `error:` filter is a regression guard that turns the shape-assignment `DeprecationWarning` into a test failure if it ever comes back, since the suite doesn't escalate warnings globally. I can drop both if you'd rather keep the tests clean.

yes please!

stanbot8:
Done, dropped both.

alejoe91 (approved):

zm711:
Thanks @stanbot8. We typically add people to our author list with an affiliation. If you'd like to be included in the next release could you provide the name and affiliation you'd like to include (if you have an academic/work affiliation that you feel should be included). Thanks!

stanbot8:
Thanks! I'd be happy to be included. Please list me as Stanley C., Independent Contributor.

PR #1891, opened 2026-07-30
Fix CedRawIO with sonpy >= 1.9.12 and re-enable the CED tests

### LINKED ISSUE #1890: CedRawIO is broken with sonpy >= 1.9.12: the `lib` namespace no longer exists
## Summary

`neo/rawio/cedrawio.py` reaches the sonpy API through `sonpy.lib`:

```python
import sonpy
self.smrx_file = sonpy.lib.SonFile(sName=str(self.filename), bReadOnly=True)
```

sonpy 1.9.12 (released 2026-06-09) reorganised its package layout and no longer defines
`lib`. Every call site in `cedrawio.py` therefore raises:

```
AttributeError: module 'sonpy' has no attribute 'lib'
```

The compiled library itself is fine — it opens and reads `.smrx` files normally. Only the
namespace moved. SpikeInterface's `read_ced` is affected transitively.

## Reproduction

```bash
uv venv --python 3.14 && uv pip install sonpy neo
```

```python
import sonpy
print(sonpy.__version__ if hasattr(sonpy, "__version__") else "n/a")
sonpy.lib.SonFile          # AttributeError: module 'sonpy' has no attribute 'lib'

from neo.rawio import CedRawIO
CedRawIO(filename="m365_1sec.smrx").parse_header()   # same AttributeError
```

## Evidence

Tested with `neo 0.14.5` and `spikeinterface 0.104.8`.

| OS | Python | sonpy resolved | wheel tag | `import sonpy` | `sonpy.lib`? | `CedRawIO.parse_header` | native read of a real `.smrx` |
|---|---|---|---|---|---|---|---|
| Windows | 3.10 – 3.14 | 1.9.12 | `cp3xx-win_amd64` | OK | **absent** | `AttributeError` | **works** |
| Linux x86_64 | 3.14 | 1.9.12 | `cp314-manylinux_2_39_x86_64` | OK | **absent** | `AttributeError` | not tested |
| Linux x86_64 | 3.13 | 1.9.5 (fallback) | `py3-none-any` | `ImportError` | n/a | n/a | n/a |
| macOS | — | — | — | not tested | not tested | not tested | not tested |

On Windows the file read succeeds through the new namespace: `GetOpenError() == 0`,
`MaxChannels() == 101`, `GetVersion() == 257`, plausible timebase. So this is purely an
import-path problem, not a functional regression in sonpy's reader.

The Linux 3.13 row is a different failure mode worth noting separately: no 1.9.12 wheel
exists for cp310–cp313 outside Windows, the resolver falls back to sonpy 1.9.5, and its
bundled `sonpy/linux/sonpy.so` targets the CPython 3.9 ABI —
`ImportError: undefined symbol: _PyThreadState_UncheckedGet`.

## Root cause

sonpy's packaging changed shape between 1.9.5 and 1.9.12.

Every release from 1.1 through 1.9.5 shipped a single `py3-none-any` wheel with a 331-byte
`__init__.py` that dispatched per platform and bound the binary to `lib`:

```python
if system() == 'Windows':
    import sonpy.amd64.sonpy as lib
elif system() == 'Darwin':
    import sonpy.darwin.sonpy as lib
elif system() == 'Linux':
    import sonpy.linux.sonpy as lib
```

1.9.12 ships per-interpreter binary wheels instead, and `lib` is gone from all eight of them.
The replacement `__init__.py` is inconsistent across platforms in the same release:

| wheel | `__init__.py` | content |
|---|---|---|
| `cp39` – `cp314` win_amd64 | 103 bytes | docstring + `from .sonpy import *` + `__version__` |
| `cp314` macosx universal2 | 21 bytes | `from .sonpy import *` |
| `cp314` manylinux_2_39_x86_64 | **0 bytes** | empty |

The empty Linux `__init__.py` looks like a build-side mistake and is worth reporting to CED
separately, but it matters here: on Linux, neither `sonpy.lib` nor `sonpy.SonFile` resolves.
Only `sonpy.sonpy.SonFile` does. Any fix therefore needs three fallbacks, not two.

## Why pinning `sonpy < 1.9.12` will not work

The version metadata makes the two mutually exclusive:

- `neo 0.14.5` declares `Requires-Python: >=3.10`
- `sonpy 1.9.5` declares `Requires-Python: >=3.9, <3.10`

There is no sonpy release before 1.9.12 that is installable on any Python that Neo supports.
CED support has effectively been unreachable since Neo dropped 3.9; 1.9.12 made sonpy
installable again on modern Python, which is what surfaced the namespace mismatch.

## Proposed fix

Resolve the namespace once, then use it at the three call sites. Full diff at the bottom
of this issue; it applies cleanly to `main` (`35cbce7`) and touches only `cedrawio.py`.

```python
def _get_sonpy_namespace():
    import importlib
    import sonpy

    for candidate in (
        lambda: sonpy.lib,                                    # <= 1.9.5
        lambda: sonpy,                                        # >= 1.9.12, Windows / macOS
        lambda: importlib.import_module("sonpy.sonpy"),       # >= 1.9.12, Linux
    ):
        try:
            namespace = candidate()
        except (AttributeError, ImportError):
            continue
        if hasattr(namespace, "SonFile"):
            return namespace
    raise ImportError(...)
```

Verified on Linux / Python 3.14 / sonpy 1.9.12: the helper resolves to `sonpy.sonpy`,
`SonFile`, `DataType.Adc`, `DataType.AdcMark` and `MarkerFilter` are all reachable, and
`CedRawIO.parse_header()` runs to completion instead of raising.

The existing skip guard in `neo/test/iotest/test_cedio.py` also needs updating — it imports
`sonpy.amd64.sonpy` / `sonpy.linux.sonpy`, so `HAVE_SONPY` is silently `False` with 1.9.12
and the CED tests never run.

## Not covered here

- macOS has not been tested. It is the one platform where the 1.9.12 wheel exists only for
  cp314, so 3.10–3.13 would hit the fallback path instead. Happy to run it if someone has
  a machine available.
- Whether a full read (not just `parse_header`) returns identical data before and after the
  change has not been checked on Linux for lack of a test file.

## Side observation

`CedRawIO.parse_header()` completes without error on a nonexistent filename, yielding zero
channels — sonpy returns an error code rather than raising, and it is not checked. Minor and
unrelated, but possibly worth its own issue.

## Patch

Against `main` (`35cbce7`). Only `neo/rawio/cedrawio.py` is modified.

```diff
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -31,6 +31,42 @@
 )
 
 
+def _get_sonpy_namespace():
+    """Return the sonpy module namespace that exposes ``SonFile``.
+
+    The layout of the sonpy package changed in 1.9.12:
+
+    * <= 1.9.5  ``sonpy/__init__.py`` binds the platform binary to ``lib``
+    * >= 1.9.12 the top level does ``from .sonpy import *`` (Windows, macOS),
+      but the Linux wheel ships an empty ``__init__.py``, so only the
+      ``sonpy.sonpy`` extension module is populated.
+
+    Probing all three keeps both old and new installs working.
+    """
+    import importlib
+
+    import sonpy
+
+    candidates = (
+        lambda: sonpy.lib,
+        lambda: sonpy,
+        lambda: importlib.import_module("sonpy.sonpy"),
+    )
+    for candidate in candidates:
+        try:
+            namespace = candidate()
+        except (AttributeError, ImportError):
+            continue
+        if hasattr(namespace, "SonFile"):
+            return namespace
+
+    raise ImportError(
+        "The installed sonpy package exposes no 'SonFile' symbol "
+        "(tried sonpy.lib, sonpy and sonpy.sonpy). "
+        "See https://pypi.org/project/sonpy/ for supported versions."
+    )
+
+
 class CedRawIO(BaseRawIO):
     """
     Class for reading data from CED (Cambridge Electronic Design) spike2.
@@ -67,9 +103,9 @@
         return self.filename
 
     def _parse_header(self):
-        import sonpy
+        sonpy_ns = _get_sonpy_namespace()
 
-        self.smrx_file = sonpy.lib.SonFile(sName=str(self.filename), bReadOnly=True)
+        self.smrx_file = sonpy_ns.SonFile(sName=str(self.filename), bReadOnly=True)
         smrx = self.smrx_file
 
         self._time_base = smrx.GetTimeBase()
@@ -82,7 +118,7 @@
         for chan_ind in range(smrx.MaxChannels()):
             chan_type = smrx.ChannelType(chan_ind)
             chan_id = str(chan_ind)
-            if chan_type == sonpy.lib.DataType.Adc:
+            if chan_type == sonpy_ns.DataType.Adc:
                 physical_chan = smrx.PhysicalChannel(chan_ind)
                 divide = smrx.ChannelDivide(chan_ind)
                 if self.take_ideal_sampling_rate:
@@ -105,13 +141,13 @@
                 buffer_id = ""
                 signal_channels.append((ch_name, chan_id, sr, dtype, units, gain, offset, stream_id, buffer_id))
 
-            elif chan_type == sonpy.lib.DataType.AdcMark:
+            elif chan_type == sonpy_ns.DataType.AdcMark:
                 # spike and waveforms : only spike times is used here
                 ch_name = smrx.GetChannelTitle(chan_ind)
                 first_time = smrx.FirstTime(chan_ind, 0, max_time)
                 max_time = smrx.ChannelMaxTime(chan_ind)
                 divide = smrx.ChannelDivide(chan_ind)
-                # here we don't use filter (sonpy.lib.MarkerFilter()) so we get all marker
+                # here we don't use filter (sonpy_ns.MarkerFilter()) so we get all marker
                 wave_marks = smrx.ReadWaveMarks(chan_ind, int(max_time / divide), 0, max_time)
 
                 # here we load in memory all spike once because the access is really slow
```

I have this ready as a PR if the approach looks right to you — I did not want to open one
before checking whether you would rather raise a clear error than probe namespaces, and
whether the `test_cedio.py` skip guard should move in the same change.

AxelNoun:
Closes #1890.

sonpy 1.9.12 reorganised its package layout and dropped the `lib` namespace that
`cedrawio.py` relies on, so every `sonpy.lib.*` access raises
`AttributeError: module 'sonpy' has no attribute 'lib'`.

While preparing the fix I found that the CED tests could not have caught this, for two
independent reasons — so this PR fixes the reader and restores the test coverage that would
have flagged it.

### 1. `neo/rawio/cedrawio.py` — resolve the namespace

A cached `_get_sonpy_namespace()` helper probes three candidates in order:

| candidate | covers |
|---|---|
| `sonpy.lib` | `<= 1.9.5`, the old per-platform dispatch |
| `sonpy` | `>= 1.9.12` on Windows and macOS (`from .sonpy import *`) |
| `sonpy.sonpy` | `>= 1.9.12` on Linux, whose wheel ships an empty `__init__.py` |

The third is not redundant: the `cp314-manylinux_2_39_x86_64` wheel has a 0-byte
`__init__.py`, so neither `sonpy.lib` nor `sonpy.SonFile` resolves there. If none of the
three exposes `SonFile`, an `ImportError` naming all three is raised, rather than letting an
`AttributeError` surface from the middle of `_parse_header`.

### 2. Both CED test guards

There are two, and each fails differently with 1.9.12:

- `neo/test/iotest/test_cedio.py` replicated the old per-platform dispatch
  (`import sonpy.linux.sonpy`, …), which no longer exists → `HAVE_SONPY = False` → skipped.
- `neo/test/rawiotest/test_cedrawio.py` used a bare `import sonpy`, which *succeeds* with
  1.9.12 → `HAVE_SONPY = True` → the test would run and fail with the `AttributeError`.

Both now delegate to `_get_sonpy_namespace()`, so "sonpy is usable" has one definition.

### 3. `pyproject.toml` — the `test` extra never installed sonpy

```toml
"sonpy;python_version<'3.10'",
```

with `requires-python = ">=3.10"`. The marker cannot be satisfied, so sonpy is absent from
every CI run and both guards were moot regardless of how they were written. This is why the
breakage went unnoticed.

Changed to match where sonpy actually publishes usable wheels:

```toml
"sonpy; platform_system=='Windows' or python_version>='3.14'",
```

1.9.12 ships `win_amd64` wheels for cp39–cp314, but `manylinux` and `macosx` only for cp314.
The marker is deliberately not just `"sonpy"`: on Linux 3.10–3.13 pip would fall back to the
source distribution, which ships a Windows `.pyd` (verified: `PE32+ executable (DLL) […] for
MS Windows`) and produces an unusable install.

The practical effect is that the automatic `ubuntu-latest` / Python 3.14 CI job will now
install sonpy and actually exercise `CedRawIO`.

### Verification

The existing test suite goes from failing to passing. Windows, Python 3.12, sonpy 1.9.12,
against `master` (`35cbce7`):

| | `pytest neo/test/rawiotest/test_cedrawio.py -v` |
|---|---|
| unpatched | `FAILED` — `AttributeError: module 'sonpy' has no attribute 'lib'` at `cedrawio.py:72` |
| patched | `PASSED` (1.07 s) — `test_read_all` across all three spike2 entities |

That covers both `.smrx` and the two `.smr` entities, so the reader is exercised end to end
and not just at import time.

Additionally:

- Windows, Python 3.10–3.14, sonpy 1.9.12: `sonpy.lib` absent on all five; the compiled
  library reads `spike2/m365_1sec.smrx` correctly through the new namespace
  (`GetOpenError() == 0`, `MaxChannels() == 101`), confirming an import-path problem rather
  than a functional regression in sonpy.
- Linux x86_64, Python 3.14, sonpy 1.9.12: the helper resolves to `sonpy.sonpy`; `SonFile`,
  `DataType.Adc`, `DataType.AdcMark` and `MarkerFilter` are all reachable, and
  `CedRawIO.parse_header()` runs to completion instead of raising.
- Both guards checked in both directions: `HAVE_SONPY` is `True` with sonpy 1.9.12 present
  and `False` when sonpy cannot be imported.
- The proposed marker evaluates `True` exactly where a usable wheel exists, across the six
  platform/version combinations in Neo's support range.
- `black --line-length 120 --check` clean on all changed files.
- After the three commits: `pytest neo/test/rawiotest/test_cedrawio.py neo/test/iotest/test_cedio.py -v` → **8 passed**.

### Not covered

macOS is untested. It is the one platform where the 1.9.12 wheel exists only for cp314, so
3.10–3.13 are excluded by the marker there. Happy to run it if someone has a machine.

### Open questions

1. Is `platform_system=='Windows' or python_version>='3.14'` acceptable, or would you rather
   keep the `test` extra minimal and add sonpy only to the CI workflow?
2. Should the `ced` extra (line 101, currently a bare `"sonpy"`) get the same marker for
   consistency? I left it alone to keep the diff focused.

### DIFF
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -19,6 +19,9 @@ Please note that the SONPY package:
 Author : Samuel Garcia
 """
 
+import functools
+import importlib
+
 import numpy as np
 
 from .baserawio import (
@@ -29,6 +32,41 @@ from .baserawio import (
     _spike_channel_dtype,
     _event_channel_dtype,
 )
+
+
+@functools.cache
+def _get_sonpy_namespace():
+    """Return the sonpy module namespace that exposes ``SonFile``.
+
+    The layout of the sonpy package changed in 1.9.12:
+
+    * ``<= 1.9.5`` shipped a single pure-Python wheel whose ``__init__.py``
+      dispatched on the platform and bound the binary to ``lib``.
+    * ``>= 1.9.12`` ships one binary wheel per interpreter. The top level does
+      ``from .sonpy import *`` on Windows and macOS, but the Linux wheel ships
+      an empty ``__init__.py``, so only the ``sonpy.sonpy`` extension module is
+      populated there.
+
+    Probing all three keeps old and new installations working.
+    """
+    import sonpy
+
+    candidates = [getattr(sonpy, "lib", None), sonpy]
+    try:
+        candidates.append(importlib.import_module("sonpy.sonpy"))
+    except ImportError:
+        pass
+
+    for namespace in candidates:
+        if namespace is not None and hasattr(namespace, "SonFile"):
+            return namespace
+
+    raise ImportError(
+        "The installed sonpy package exposes no 'SonFile' symbol "
+        "(tried sonpy.lib, sonpy and sonpy.sonpy). "
+        "Note that sonpy only publishes wheels for Windows, and for Linux and macOS "
+        "on Python 3.14 and above; the source distribution is not usable."
+    )
 
 
 class CedRawIO(BaseRawIO):
@@ -67,9 +105,9 @@ class CedRawIO(BaseRawIO):
         return self.filename
 
     def _parse_header(self):
-        import sonpy
+        sonpy_ns = _get_sonpy_namespace()
 
-        self.smrx_file = sonpy.lib.SonFile(sName=str(self.filename), bReadOnly=True)
+        self.smrx_file = sonpy_ns.SonFile(sName=str(self.filename), bReadOnly=True)
         smrx = self.smrx_file
 
         self._time_base = smrx.GetTimeBase()
@@ -82,7 +120,7 @@ class CedRawIO(BaseRawIO):
         for chan_ind in range(smrx.MaxChannels()):
             chan_type = smrx.ChannelType(chan_ind)
             chan_id = str(chan_ind)
-            if chan_type == sonpy.lib.DataType.Adc:
+            if chan_type == sonpy_ns.DataType.Adc:
                 physical_chan = smrx.PhysicalChannel(chan_ind)
                 divide = smrx.ChannelDivide(chan_ind)
                 if self.take_ideal_sampling_rate:
@@ -105,13 +143,13 @@ class CedRawIO(BaseRawIO):
                 buffer_id = ""
                 signal_channels.append((ch_name, chan_id, sr, dtype, units, gain, offset, stream_id, buffer_id))
 
-            elif chan_type == sonpy.lib.DataType.AdcMark:
+            elif chan_type == sonpy_ns.DataType.AdcMark:
                 # spike and waveforms : only spike times is used here
                 ch_name = smrx.GetChannelTitle(chan_ind)
                 first_time = smrx.FirstTime(chan_ind, 0, max_time)
                 max_time = smrx.ChannelMaxTime(chan_ind)
                 divide = smrx.ChannelDivide(chan_ind)
-                # here we don't use filter (sonpy.lib.MarkerFilter()) so we get all marker
+                # here we don't use filter (sonpy_ns.MarkerFilter()) so we get all marker
                 wave_marks = smrx.ReadWaveMarks(chan_ind, int(max_time / divide), 0, max_time)
 
                 # here we load in memory all spike once because the access is really slow
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -21,7 +21,6 @@ Author : Samuel Garcia
 
 import importlib
 import importlib.util
-from functools import cache
 
 import numpy as np
 
@@ -35,7 +34,6 @@ from .baserawio import (
 )
 
 
-@cache
 def _get_sonpy_namespace():
     """Return the sonpy namespace exposing SonFile, whatever the installed layout."""
     if importlib.util.find_spec("sonpy") is None:
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -21,6 +21,7 @@ Author : Samuel Garcia
 
 import functools
 import importlib
+import importlib.util
 
 import numpy as np
 
@@ -51,14 +52,16 @@ def _get_sonpy_namespace():
     """
     import sonpy
 
-    candidates = [getattr(sonpy, "lib", None), sonpy]
-    try:
-        candidates.append(importlib.import_module("sonpy.sonpy"))
-    except ImportError:
-        pass
-
-    for namespace in candidates:
+    for namespace in (getattr(sonpy, "lib", None), sonpy):
         if namespace is not None and hasattr(namespace, "SonFile"):
+            return namespace
+
+    # Only reached with the >= 1.9.12 Linux wheels, whose __init__.py is empty.
+    # find_spec imports the parent package, and raises if it is not a package,
+    # so guard on __path__ before probing the submodule.
+    if hasattr(sonpy, "__path__") and importlib.util.find_spec("sonpy.sonpy") is not None:
+        namespace = importlib.import_module("sonpy.sonpy")
+        if hasattr(namespace, "SonFile"):
             return namespace
 
     raise ImportError(
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -19,9 +19,9 @@ Please note that the SONPY package:
 Author : Samuel Garcia
 """
 
-import functools
 import importlib
 import importlib.util
+from functools import cache
 
 import numpy as np
 
@@ -35,41 +35,31 @@ from .baserawio import (
 )
 
 
-@functools.cache
+@cache
 def _get_sonpy_namespace():
-    """Return the sonpy module namespace that exposes ``SonFile``.
+    """Return the sonpy namespace exposing SonFile, whatever the installed layout."""
+    if importlib.util.find_spec("sonpy") is None:
+        raise ImportError("sonpy is not installed. Install it with `pip install sonpy`.")
 
-    The layout of the sonpy package changed in 1.9.12:
+    sonpy = importlib.import_module("sonpy")
 
-    * ``<= 1.9.5`` shipped a single pure-Python wheel whose ``__init__.py``
-      dispatched on the platform and bound the binary to ``lib``.
-    * ``>= 1.9.12`` ships one binary wheel per interpreter. The top level does
-      ``from .sonpy import *`` on Windows and macOS, but the Linux wheel ships
-      an empty ``__init__.py``, so only the ``sonpy.sonpy`` extension module is
-      populated there.
+    # <= 1.9.5 binds the extension module as an attribute, not a submodule,
+    # so find_spec("sonpy.lib") cannot see it.
+    lib = getattr(sonpy, "lib", None)
+    if lib is not None and hasattr(lib, "SonFile"):
+        return lib
 
-    Probing all three keeps old and new installations working.
-    """
-    import sonpy
+    # >= 1.9.12 on Windows/macOS re-exports into the package namespace.
+    if hasattr(sonpy, "SonFile"):
+        return sonpy
 
-    for namespace in (getattr(sonpy, "lib", None), sonpy):
-        if namespace is not None and hasattr(namespace, "SonFile"):
-            return namespace
+    # >= 1.9.12 on Linux ships an empty __init__.py.
+    if importlib.util.find_spec("sonpy.sonpy") is not None:
+        nested = importlib.import_module("sonpy.sonpy")
+        if hasattr(nested, "SonFile"):
+            return nested
 
-    # Only reached with the >= 1.9.12 Linux wheels, whose __init__.py is empty.
-    # find_spec imports the parent package, and raises if it is not a package,
-    # so guard on __path__ before probing the submodule.
-    if hasattr(sonpy, "__path__") and importlib.util.find_spec("sonpy.sonpy") is not None:
-        namespace = importlib.import_module("sonpy.sonpy")
-        if hasattr(namespace, "SonFile"):
-            return namespace
-
-    raise ImportError(
-        "The installed sonpy package exposes no 'SonFile' symbol "
-        "(tried sonpy.lib, sonpy and sonpy.sonpy). "
-        "Note that sonpy only publishes wheels for Windows, and for Linux and macOS "
-        "on Python 3.14 and above; the source distribution is not usable."
-    )
+    raise ImportError("sonpy is installed but exposes no SonFile.")
 
 
 class CedRawIO(BaseRawIO):
--- a/neo/rawio/cedrawio.py
+++ b/neo/rawio/cedrawio.py
@@ -89,6 +89,13 @@ class CedRawIO(BaseRawIO):
 
     * This IO reads smr and smrx files
 
+    * sonpy is installed by the ``ced`` extra, but upstream only publishes wheels for Windows,
+      and for Linux and macOS from Python 3.14 onwards. Elsewhere the extra resolves to nothing
+      installable and this class raises an ImportError naming the constraint on first use; the
+      PyPI source distribution ships a Windows binary and is not usable.
+
+    * Old smr files can be read without sonpy using Spike2RawIO. Only smrx requires this class.
+
     """
 
     extensions = ["smr", "smrx"]
--- a/neo/test/iotest/test_cedio.py
+++ b/neo/test/iotest/test_cedio.py
@@ -1,17 +1,10 @@
 import unittest
-from platform import system
-from sys import maxsize
 
 try:
-    if system() == "Windows":
-        if maxsize > 2**32:
-            import sonpy.amd64.sonpy
-        else:
-            import sonpy.win32.sonpy
-    elif system() == "Darwin":
-        import sonpy.darwin.sonpy
-    elif system() == "Linux":
-        import sonpy.linux.sonpy
+    from neo.rawio.cedrawio import _get_sonpy_namespace
+
+    # Raises ImportError if sonpy is missing or exposes no usable namespace.
+    _get_sonpy_namespace()
     from neo.io import CedIO
 except ImportError:
     HAVE_SONPY = False
--- a/neo/test/rawiotest/test_cedrawio.py
+++ b/neo/test/rawiotest/test_cedrawio.py
@@ -5,7 +5,10 @@ from neo.rawio.cedrawio import CedRawIO
 from neo.test.rawiotest.common_rawio_test import BaseTestRawIO
 
 try:
-    import sonpy
+    from neo.rawio.cedrawio import _get_sonpy_namespace
+
+    # Raises ImportError if sonpy is missing or exposes no usable namespace.
+    _get_sonpy_namespace()
 
     HAVE_SONPY = True
 except ImportError:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ test = [
     "coverage",
     "coveralls",
     "pillow",
-    "sonpy;python_version<'3.10'",
+    "sonpy; platform_system=='Windows' or python_version>='3.14'",
     "pynwb",
     "probeinterface",
     "zugbruecke>=0.2",


neo/rawio/cedrawio.py
```diff
@@ -31,6 +34,41 @@
...
+
+    Probing all three keeps old and new installations working.
+    """
+    import sonpy
+
+    candidates = [getattr(sonpy, "lib", None), sonpy]
+    try:
+        candidates.append(importlib.import_module("sonpy.sonpy"))
+    except ImportError:
+        pass
```
zm711: I don't think this try-except is necessary. you could use `importlib.util.find_spec` to verify submodules without importing and without the try except (you just have to be careful of the logic to prevent errors before you've got to the deepest nesting. Then you can just do the import at the end with `import_module`.

zm711 (commented):
I like this overall, though I would prefer a tiny rewrite in the logic of the new function you've created. Your other concerns about the testing were on purpose from our perspective. When we no longer have active contributors (or if one of us doesn't have the familiarity/time) for an `io` we do a slow deprecation process where we slowly reduce testing (but keep it accessible for people) until it completely dies. Your work appears to revive this`io` which means that we are happy to fully test things again within your constraints.

neo/rawio/cedrawio.py
AxelNoun: Good call, thanks, pushed in a follow-up commit. Two details worth flagging:

- `import importlib` doesn't pull in `importlib.util`, so the import is now explicit.
- `find_spec("sonpy.sonpy")` imports the parent and raises `ModuleNotFoundError` if `sonpy` has no `__path__`. Guarding on that first keeps the failure on the explicit `ImportError` below rather than surfacing from `find_spec`.

Reordering also means the submodule is only imported when the first two candidates come up empty, i.e. the Linux >= 1.9.12 case, instead of eagerly on every platform. Verified against four fake sonpy layouts: right namespace on the three real ones, explicit `ImportError` on a package exposing nothing.

AxelNoun:
> I like this overall, though I would prefer a tiny rewrite in the logic of the new function you've created. Your other concerns about the testing were on purpose from our perspective. When we no longer have active contributors (or if one of us doesn't have the familiarity/time) for an `io` we do a slow deprecation process where we slowly reduce testing (but keep it accessible for people) until it completely dies. Your work appears to revive this`io` which means that we are happy to fully test things again within your constraints.

Thanks, that context helps and it reframes part of my description. I read the `test` extra marker and the two divergent guards as accidents; if the reduced testing was deliberate, then "this is why the breakage went unnoticed" is the wrong framing on my side. Happy to reword that section so it doesn't read as an oversight report.

On the revival: I'm happy to take CedRawIO on going forward. Ping me on anything CED and I'll pick it up.

Taking "within your constraints" as a green light on the two open questions:

1. I'll keep `sonpy; platform_system=='Windows' or python_version>='3.14'` in the `test` extra. The practical effect is that the existing ubuntu-latest / 3.14 job installs sonpy and exercises the reader end to end, so coverage comes back without a new matrix entry.

2. The `ced` extra is worth the same marker, I think. Today `pip install neo[ced]` on Linux 3.10 to 3.13 falls back to the sdist and installs a Windows `.pyd`, so users get an install that looks fine and fails later. With the marker they get nothing installed plus the explicit `ImportError` from `_get_sonpy_namespace()` at first use, which at least names the constraint. Say the word and I'll add it here, otherwise I'll keep this diff focused and open a separate PR.

zm711:
I've approved our work flows so let's make sure everything passes but this looks good to me. I'll do a final read through once tests pass.

zm711:
> Taking "within your constraints" as a green light on the two open questions:

> I'll keep sonpy; platform_system=='Windows' or python_version>='3.14' in the test extra. The practical effect is that the existing ubuntu-latest / 3.14 job installs sonpy and exercises the reader end to end, so coverage comes back without a new matrix entry.

> The ced extra is worth the same marker, I think. Today pip install neo[ced] on Linux 3.10 to 3.13 falls back to the sdist and installs a Windows .pyd, so users get an install that looks fine and fails later. With the marker they get nothing installed plus the explicit ImportError from _get_sonpy_namespace() at first use, which at least names the constraint. Say the word and I'll add it here, otherwise I'll keep this diff focused and open a separate PR.

I would say separate PR. A lot of of the team is away on vacation this month and I'm only intermittently able to respond. So multiple small PRs will be easier on us for now. Thanks for updating this reader.

AxelNoun:
Applied, with one adjustment I want to flag rather than slip in.

`find_spec` works for two of the three candidates but not for `sonpy.lib`. In
1.9.5 there is no `sonpy/lib.py` - `__init__.py` does platform dispatch and binds
the extension module as an attribute (`import sonpy.linux.sonpy as lib`), so
`find_spec("sonpy.lib")` returns None, the candidate is skipped, and since
`SonFile` isn't on the top-level namespace in 1.9.5 either, resolution falls
through to ImportError. I caught this building fixtures from the published
wheels rather than from assumption - my first synthetic fixture invented a
`sonpy/lib.py` and passed happily.

So: `find_spec` for the top-level guard and for `sonpy.sonpy`, `getattr` for
`lib`, where find_spec is structurally blind.

Layouts, from unpacking the wheels:

| Distribution | `__init__.py` | `SonFile` |
|---|---|---|
| 1.9.5 py3-none-any | 331 B, platform dispatch | `sonpy.lib` (attribute) |
| 1.9.12 manylinux cp314 | 0 B | `sonpy.sonpy` |
| 1.9.12 macosx cp314 | 21 B, `from .sonpy import *` | `sonpy` |
| 1.9.12 win_amd64 cp39-314 | 103 B, `from .sonpy import *` | `sonpy` |

One thing worth deciding rather than me deciding it: **the `sonpy.lib` branch is
dead code for us.** Every pre-1.9.12 release pins `Requires-Python` to a single
minor (1.9.5 is `>=3.9, <3.10`), so on anything from 3.10 up the only installable
sonpy is 1.9.12. I've kept the branch as defensive handling - it costs one
`getattr` and protects against sonpy repackaging again - but I'm equally happy to
drop it and simplify to the two live layouts. Your call.

Same metadata also explains the extras marker, though not the way I first
assumed. Below 3.14 on Linux/macOS there's no 1.9.12 wheel and no older
installable version, so pip falls to the sdist - and the sdist *succeeds*. It
declares no extension to compile, ships `sonpy/sonpy.pyd` (a Windows binary) as
its only package data, and pip happily builds
`sonpy-1.9.12-cp312-cp312-linux_x86_64.whl` from it. The install exits 0;
`import sonpy` then raises `ModuleNotFoundError: No module named 'sonpy.sonpy'`
from the `from .sonpy import *` in `__init__.py`. Verified on Linux / Python
3.12. So the `python_version >= '3.14'` gate isn't guarding against a build
failure, it's guarding against a silently broken install.

Worth noting what that does to the resolver: `find_spec("sonpy")` returns a valid
spec there, and `import_module("sonpy")` raises. It surfaces as an ImportError so
callers still behave, but the message points at `sonpy.sonpy` when the failure is
really in `__init__.py`. No amount of `find_spec` can catch that - the package
exists, it's just unusable. Happy to wrap the top-level import for a clearer
message if you think it's worth it, or leave it.

Local verification (accurate scope):

- layout matrix: 5/5 PASS against fixtures built from the real wheels (resolver
  extracted from `neo/rawio/cedrawio.py`)
- real install: Windows 11 / Python 3.14.2 / sonpy 1.9.12 -> resolves to `sonpy`
- CED pytest suite: **not run** (no working git-annex locally)
- neither the Linux nested path nor the 1.9.5 attribute path was exercised
  against a real install here - fixtures only; and 1.9.5 cannot install on 3.10+

On the red IO job - it isn't this PR. `ubuntu-latest / 3.14 / 2.3.3` fails with
95 errors spread across alphaomega, axon, blackrock, edf, spikeglx, tdt and ced
alike, all the same `IncompleteResultsError` behind a 403 on
`gin.g-node.org/NeuralEnsemble/ephy_testing_data`. sonpy 1.9.12 installs fine and
there is no sonpy-related `ImportError`, `AttributeError` or `ModuleNotFoundError`
anywhere in the log. datalad 1.6.0 marked their own `test_gin_cloning` as xfail on
`TimeoutError` and `IncompleteResultsError` - the same exception we're seeing
here - citing HTTP 403 from GitHub-hosted IPs, TCP timeouts and slow responses
(PR #7876).

Worth flagging separately: the dataset-hash step stays green through this.
`echo "dataset_hash=$(git ls-remote ...)" >> $GITHUB_OUTPUT` takes its exit status
from `echo`, so the 403 yields an empty hash and the cache key silently degrades
to `Linux-datasets-`. Happy to open a separate issue for that.

That said, it does mean the CED tests haven't yet gone green in CI - which is the
one claim in this PR I can't currently back with a check.

neo/rawio/cedrawio.py
```diff
@@ -31,6 +35,33 @@
 )
 
 
+@cache
```
zm711: What's the benefit of caching here? It seems like this function is run 1x-2x max. Happy to be educated here if this is valuable. I need to rerun our IO tests so I'll do that as soon as I have a moment.

AxelNoun: You're right, it should go : I checked rather than guessing.

After the first call the uncached function costs ~1.3 µs against ~0.03 µs cached:
by then find_spec and import_module are both just sys.modules lookups, so the
import system has already done the caching for us.

More to the point, functools.cache doesn't memoize exceptions, so it does nothing
for the one path that actually costs something - find_spec on a missing module is
a full sys.path scan, ~30 µs, and it re-runs on every call regardless. It
optimises the free path and can't touch the expensive one.

It also adds friction to testing: any test of this function would need
cache_clear() between cases. Dropped in the commit above.

And thanks for rerunning the IO tests.

zm711: Thanks for checking. That was my intuition, but always better to verify. Testing are running now. If they pass I'll try to merge this when I'm back to my computer.

zm711:
Just as quick summary:

3.10 < python < 3.14 our CI/reader CEDIO was previously broken. This fixes the problem in general and adds testing for sonpy/CEDIO for python >= 3.14.

So when checking tests for python < 3.14 we will skip the CED testing (in case someone checks in the future and is confused). Once 3.14 is our minimum this problem will go away.

zm711 (approved):
Looks great. Thanks.

zm711:
@AxelNoun when we merge a PR from someone we give the option to be added to our author list. Would you like us to add you before our next release? If so what name would you like to appear and what affiliation (if you have one) should be included?

AxelNoun:
> @AxelNoun when we merge a PR from someone we give the option to be added to our author list. Would you like us to add you before our next release? If so what name would you like to appear and what affiliation (if you have one) should be included? 

Thanks a lot, and thanks for the quick review!
Yes, I’d be glad to be added. My name is Axel Coiffard-Donaty ; no affiliation to include.
Happy to contribute again in the future !

PR #1893, opened 2026-08-05
BiocamRawIO: reshape flat datasets instead of gathering channel by channel

### LINKED ISSUE #1892: BiocamRawIO: per-channel loop over already-loaded data makes BRW4 reads ~18x slower than necessary
**Affects:** current `neo/rawio/biocamrawio.py` (checked against the version
shipped with SpikeInterface 0.104.8).
**Severity:** performance only — output is correct.

---

## Summary

`BiocamRawIO._get_analogsignal_chunk` handles BRW4 files whose raw dataset is a
flat 1-D array by looping over channels:

```python
# neo/rawio/biocamrawio.py:187-191
sig_chunk = np.zeros((i_stop - i_start, len(channel_indexes)), dtype=data.dtype)
# iterate through channels to prevent loading all channels into memory which can cause
# memory exhaustion. See https://github.com/SpikeInterface/spikeinterface/issues/3303
for index, channel_index in enumerate(channel_indexes):
    sig_chunk[:, index] = data[channel_index :: self._num_channels]
```

The stated rationale — avoiding loading all channels into memory — does not
apply, because `data` has **already been fully read into memory** by
`self._read_function(...)` on the line immediately above. The loop therefore
buys no memory saving and costs `num_channels` strided gathers over an array
that is already resident. On a 4096-channel BioCam that is 4096 cache-hostile
passes per chunk.

Since the array is contiguous and frame-major, a single `reshape` gives the same
result as a view.

## Measured

3Brain BioCam, 4096 channels @ 19,753.775 Hz, BRW4 (`Well_A1/Raw`, flat uint16),
reading 100,000 frames:

| | Windows (native h5py) | WSL2 over DrvFs |
|---|---|---|
| stock neo | 93 MB/s | 77 MB/s |
| with the patch below | — | **214 MB/s (2.8×)** |
| raw `h5py` + `reshape` ceiling | 1,647 MB/s | 328 MB/s |

Output verified **bit-identical** (`np.array_equal`) between stock and patched
paths.

## Suggested fix

```python
else:
    nch = self._num_channels
    if data.size % nch:
        # unexpected layout - fall back rather than mis-shape the data
        ...existing per-channel loop...
    else:
        arr = data.reshape(-1, nch)
        if channel_indexes is None or (
            isinstance(channel_indexes, slice) and channel_indexes == slice(None)
        ):
            sig_chunk = arr
        else:
            sig_chunk = arr[:, channel_indexes]
```

The event-based-compressed branch (`readHDF5t_brw4_sparse`) is untouched.

If the memory concern in issue #3303 is really about `_read_function` loading a
whole chunk, that is worth addressing separately — but it should be fixed there,
by reading a channel subset from HDF5, rather than by re-gathering data that is
already in RAM.

## Working patch

A monkey-patch implementing exactly this, with a `verify()` that asserts
bit-identical output and reports the speed-up, is available and can be turned
into a PR on request.

## Environment

neo as shipped with SpikeInterface 0.104.8, h5py 3.16.0, Python 3.12,
Ubuntu 24.04 (WSL2) and Windows 11.

adityasingh2400:
`BiocamRawIO._get_analogsignal_chunk` has two branches. Older files hand back a 2D `(n_samples, n_channels)` array and are indexed directly. Newer files (BRW 4.x `Well_*/Raw`, and the v3 `101`/`102` formats) hand back a flat frame-major 1D array, and that branch first expands the caller's selection into a list with `range(start or 0, stop or n_ch, step or 1)`, then rebuilds the chunk with one strided gather per channel.

The `range(...)` expansion is wrong for any slice carrying a zero or negative bound, because `or` treats `0` as absent and `range` has no notion of negative indices. On a 16 channel file, `slice(0, -1)` returns 0 channels instead of 15, `slice(None, None, -1)` returns 0 instead of 16, `slice(0, 0)` returns all 16 instead of none, and `slice(-2, None)` returns 18 columns, which is more columns than the file has channels, with the first two silently filled from the wrong end of the flat buffer. None of these raise. Every one of them is correct today on an older 2D file, so the same reader answers the same question two different ways depending on which firmware wrote the recording.

The per-channel loop is also the reason issue #1892 was filed. Its comment says it avoids loading all channels into memory, but `self._read_function(...)` on the line above has already read the whole span into a numpy array, so the loop buys no memory and costs `n_channels` cache-hostile passes plus a second full-size allocation for the destination.

The fix reshapes the flat array to `(i_stop - i_start, n_channels)`, which is a view onto the buffer that was just read, and then lets both layouts share one indexing expression. numpy applies the caller's slice, index list, or array exactly as it does everywhere else in Neo, so the slice forms above stop disagreeing with the 2D branch. A frame range that runs past the end of the dataset used to surface as a numpy broadcast error from inside the loop, and now raises `NeoReadWriteError` naming the range that was asked for.

Verified on the two BioCam files in the gin test data, `biocam_hw3.0_fw1.6.brw` (36 channels, `readHDF5t_101`) and `biocam_hw3.0_fw1.7.0.12_raw.brw` (100 channels, `readHDF5t_brw4`). Against a verbatim copy of the old branch the output is bit identical for `None`, `slice(None)`, `slice(0, 4)`, `slice(2, None)`, `slice(None, None, 3)` and an explicit index list, and reading 2000 frames is 9x faster on the first file and 68x on the second. Full `rawio_compliance` passes on both, as does a `BiocamIO.read_block()`. The new tests build a minimal BRW 4.x file with h5py in the style of the existing `test_biocamrawio_gain`, so they need no downloaded data. Against 35cbce7f they are 5 failed 9 passed before the source change and 14 passed after, and `neo/test/rawiotest` plus `neo/test/coretest` is 637 passed 130 skipped 0 failed with the change in place.

I could not test the event-based sparse path against a real compressed recording, but that path is untouched: `readHDF5t_brw4_sparse` returns a 2D array and so has always taken the other branch.

Fixes #1892

### DIFF
--- a/doc/source/authors.rst
+++ b/doc/source/authors.rst
@@ -103,6 +103,7 @@ and may not be the current affiliation of a contributor.
 * Reema Gupta [50]
 * Sai Asish Yamani [51]
 * Kevin Doran [52]
+* Aditya Singh (github)
 
 1. Centre de Recherche en Neuroscience de Lyon, CNRS UMR5292 - INSERM U1028 - Université Claude Bernard Lyon 1
 2. Unité de Neuroscience, Information et Complexité, CNRS UPR 3293, Gif-sur-Yvette, France
--- a/neo/rawio/biocamrawio.py
+++ b/neo/rawio/biocamrawio.py
@@ -163,34 +163,27 @@ class BiocamRawIO(BaseRawIO):
         else:
             data = self._read_function(self._filehandle, i_start, i_stop, self._num_channels)
 
-        # older style data returns array of (n_samples, n_channels), should be a view
+        # Newer style data comes back as a flat array of length (n_samples * n_channels)
+        # laid out frame-major, i.e. every channel of frame 0, then every channel of
+        # frame 1, and so on. The read functions materialize it in memory already, so
+        # reshaping is a free view onto the same buffer and puts both file layouts on the
+        # same (n_samples, n_channels) footing.
+        if data.ndim == 1:
+            expected_size = (i_stop - i_start) * self._num_channels
+            if data.size != expected_size:
+                raise NeoReadWriteError(
+                    f"Read {data.size} values for frames {i_start}:{i_stop} of {self._num_channels} "
+                    f"channels but expected {expected_size}. The requested frame range is most likely "
+                    "out of bounds."
+                )
+            data = data.reshape(i_stop - i_start, self._num_channels)
+
+        # older style data is already (n_samples, n_channels), should be a view
         # but if memory issues come up we should doublecheck out how the file is being stored
-        if data.ndim > 1:
-            if channel_indexes is None:
-                channel_indexes = slice(None)
-            sig_chunk = data[:, channel_indexes]
+        if channel_indexes is None:
+            channel_indexes = slice(None)
 
-        # newer style data returns an initial flat array (n_samples * n_channels)
-        # we iterate through channels rather than slicing
-        # Due to the fact that Neo and SpikeInterface tend to prefer slices we need to add
-        # some careful checks around slicing of None in the case we need to iterate through
-        # channels. First check if None. Then check if slice and only if slice check that it is slice(None)
-        else:
-            if channel_indexes is None:
-                channel_indexes = [ch for ch in range(self._num_channels)]
-            elif isinstance(channel_indexes, slice):
-                start = channel_indexes.start or 0
-                stop = channel_indexes.stop or self._num_channels
-                step = channel_indexes.step or 1
-                channel_indexes = [ch for ch in range(start, stop, step)]
-
-            sig_chunk = np.zeros((i_stop - i_start, len(channel_indexes)), dtype=data.dtype)
-            # iterate through channels to prevent loading all channels into memory which can cause
-            # memory exhaustion. See https://github.com/SpikeInterface/spikeinterface/issues/3303
-            for index, channel_index in enumerate(channel_indexes):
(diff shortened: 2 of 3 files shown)


zm711:
#1885 works to address this. I would suggest testing it out first to see if that fixes your performance issue as we are about to merge that PR.

adityasingh2400:
Checked it, and yes, #1885 fixes this. I think it should land instead of mine, so I am closing this.

It carries the same core change, dropping the per-channel loop for a reshape onto the flat buffer and a plain slice:

```python
data = data.reshape(i_stop - i_start, self._num_channels)
```

It is also the better version. The loop existed to keep memory bounded when only a few channels are wanted from a large recording, which is SpikeInterface#3303. My PR removes the loop and gives that bound up. #1885 removes it too but replaces it with an explicit ceiling, `max_read_bytes_dense` and `max_read_bytes_sparse`, and splits the read when the buffer would be large and the output small. That keeps the property I dropped, and the separate sparse ceiling is a detail I had not considered, since the event decoder re-walks its tables per call so it wants fewer, bigger reads.

One small thing you may or may not want from mine. I added a guard that raises when the flat read comes back the wrong length, which turns an out of bounds frame range into a clear error rather than a reshape traceback:

```python
expected_size = (i_stop - i_start) * self._num_channels
if data.size != expected_size:
    raise NeoReadWriteError(...)
```

Happy to send that as a small separate PR against #1885 once it merges if you think it is worth having. Otherwise no need, and thanks for pointing me at the right PR.

PR #1895, opened 2026-08-05
Fix CircularRegionOfInterest.pixels_in_region omitting boundary pixels

### LINKED ISSUE #1889: CircularRegionOfInterest.pixels_in_region omits boundary pixels that its own is_inside accepts
**Describe the bug**

`CircularRegionOfInterest.is_inside()` accepts pixels that `pixels_in_region()`
never returns, so the region is not symmetric about its own centre. It looks
like a regression from b43026db rather than a deliberate choice, and the
example in `examples/plot_imageseq.py` is affected.

`is_inside()` tests a closed disc (`<=`), so a pixel exactly `radius` away is
inside. `pixels_in_region()` scans `range(int(floor(c - r)), int(ceil(c + r)))`,
and `range` stops one short of its endpoint, so pixels at the `+x` and `+y`
extremes are never passed to `is_inside()` at all. They aren't rejected, they
are never asked about.

This only bites when a pixel centre lands at exactly distance `radius`, which
for a whole-number centre means a whole-number radius. `radius=2.5` and
`radius=1.01` are unaffected. Whole numbers are the common case though, and
both ROIs in `examples/plot_imageseq.py` use them.

**To Reproduce**

```python
import numpy as np
import quantities as pq
from neo.core import ImageSequence
from neo.core.regionofinterest import CircularRegionOfInterest

seq = ImageSequence(np.zeros((2, 120, 120)), sampling_rate=1 * pq.Hz,
                    spatial_scale=1 * pq.micrometer, units="V")

roi = CircularRegionOfInterest(seq, 10, 10, 5)      # from examples/plot_imageseq.py
pixels = roi.pixels_in_region()

print(roi.is_inside(15, 10), [15, 10] in pixels)    # True False
print(roi.is_inside(5, 10), [5, 10] in pixels)      # True True

xs = [p[0] for p in pixels]
print(min(xs), max(xs))                             # 5 14, for a circle centred on 10

accepted = sum(1 for y in range(120) for x in range(120) if roi.is_inside(x, y))
print(len(pixels), accepted)                        # 79 81
```

(15, 10) sits exactly 5 pixels from the centre. `is_inside()` says it is in the
region. It is never enumerated. The same holds for (10, 15). The mirror pixels
(5, 10) and (10, 5) are both returned, so the region reaches 5 pixels to the
left of the centre and 4 to the right.

The other ROI in the same example, `CircularRegionOfInterest(seq, 50, 50, 25)`,
returns 1959 pixels where `is_inside()` accepts 1961.

**Expected behaviour**

`pixels_in_region()` should return every pixel `is_inside()` accepts. For
`radius=5` centred on 10 that is 81 pixels spanning x 5 to 15, symmetric about
the centre.

`RectangularRegionOfInterest` already behaves this way. Its `is_inside()` uses a
half-open interval, which matches what its own `range()` bounds enumerate, and
sweeping 1056 centre/width/height combinations turns up no disagreement between
its predicate and its enumeration.

**Environment:**
 - OS: Linux 6.17, Ubuntu
 - Python version: 3.12.3
 - Neo version: 0.15.dev0 at 35cbce7f, and 0.14.5 (nothing has touched
   `regionofinterest.py` since that tag, so a pip install reproduces it)
 - NumPy version: 2.5.1

**Additional context**

The history suggests this was collateral damage rather than a decision, and
there is one thing only you can tell me.

Before b43026db the loop read:

```python
for y in range(self.y - self.radius, self.y + self.radius + 1):
```

Inclusive, matching the `<=` that `is_inside()` has carried since the file was
added. b43026db ("Make RegionOfInterest work with float arguments",
2019-09-26 10:55) rewrote the bounds with `floor`/`ceil` and the `+1` went with
it. Twenty-nine minutes later 52977d38 ("fix tests; pep8") changed the
radius-1 assertion in `test_regionofinterest.py` to the three-pixel list and
added a radius-1.01 case asserting the five-pixel plus shape.

So: was that assertion change a deliberate call that a pixel exactly on the
circle is outside it, or was it getting the suite green after the bounds
changed? If it was deliberate then `is_inside()` is what needs fixing, not the
enumeration, and the fix is the opposite of what I would have written.

`PolygonRegionOfInterest` has the same bounding box under-scan. A square with
vertices at (10,10), (20,10), (20,20), (10,20) enumerates x 10 to 19 and y 10
to 19. I haven't investigated whether that one matters, since polygon has no
`is_inside()` to disagree with.

I have a fix and regression tests ready, using
`range(int(ceil(c - r)), int(floor(c + r)) + 1)` so the scan covers the closed
interval exactly rather than over-scanning past it. Across 44 centre and radius
combinations, including half-integer centres and fractional radii, it returns
exactly the set `is_inside()` accepts, with nothing missing and nothing extra.

Against 35cbce7f, with those two lines changed, the radius-1 assertion restored
to the five-pixel list and two regression tests added to
`Test_CircularRegionOfInterest`, `neo/test/coretest` runs 621 passed, 12
skipped, 0 failed, up from 619 passed, 12 skipped, 0 failed. Reverting the two
lines and leaving everything else gives 3 failures. Nothing else in the suite
moves either way.

Happy to open a PR, once you tell me which side of it you want changed.

adityasingh2400:
Fixes #1889.

`CircularRegionOfInterest.is_inside()` tests a closed disc with `<=`, so a pixel centre exactly `radius` away is inside. But `pixels_in_region()` enumerated `range(int(floor(c - r)), int(ceil(c + r)))`, and `range` excludes its stop value, so pixels at the `+x` and `+y` extremes were never passed to `is_inside()` at all. They were not rejected, they were never asked about.

The region therefore came out asymmetric about its own centre. For `CircularRegionOfInterest(seq, 10, 10, 5)` it returned 79 pixels spanning x 5 to 14, while `is_inside()` accepts 81 spanning 5 to 15.

The fix adds 1 to both upper bounds, so every pixel `is_inside()` accepts is offered to it.

The repo's own test file contradicted itself on this. Line 11 expected 3 pixels at `radius=1` while line 13 expected 5 at `radius=1.01`. Under a closed disc both radii enclose exactly the same pixel centres, so those two lines could not both be right. **Updating the `radius=1` expectation to the same 5 pixels is an intentional part of this fix, not an accident.** That assertion is what encoded the bug.

I swept the other two ROI classes for the same off-by-one and they are unaffected, so this leaves them alone. `RectangularRegionOfInterest.is_inside()` uses a half-open interval that exactly matches what its `range()` bounds enumerate, with 0 disagreements across 11664 centre, width and height combinations. The polygon ray caster excludes its top and right edges to match its bounding box, with 0 disagreements across 400 random polygons.

Tests: the corrected `test_result`, plus `test_pixels_in_region_matches_is_inside`, which locks in the real invariant that the enumeration must equal the predicate's accept set, and `test_pixels_in_region_is_symmetric_about_centre`, which is the issue's exact reproduction.

Restoring `neo/core/regionofinterest.py` from master makes all three fail, with `[[6, 5], [5, 6], [6, 6]] != [[6, 5], [5, 6], [6, 6], [7, 6], [6, 7]]`. All five pass after. The core suite is green at 617 passed and 16 skipped, and the circle sweep goes from 486 of 4536 configurations disagreeing to 0 of 4536.

Disclosure: this change was prepared with AI assistance. I have reviewed and tested it.

### DIFF
--- a/neo/core/regionofinterest.py
+++ b/neo/core/regionofinterest.py
@@ -79,9 +79,13 @@ class CircularRegionOfInterest(RegionOfInterest):
 
     def pixels_in_region(self):
         """Returns a list of pixels whose *centres* are within the circle"""
+        # `is_inside` tests a closed disc, so a pixel centre exactly `radius` away is
+        # inside. `range` excludes its stop value, so the upper bounds need a `+ 1` for
+        # those extremes to be offered to `is_inside` at all. Without it the region is
+        # not symmetric about its own centre.
         pixel_in_list = []
-        for y in range(int(floor(self.y - self.radius)), int(ceil(self.y + self.radius))):
-            for x in range(int(floor(self.x - self.radius)), int(ceil(self.x + self.radius))):
+        for y in range(int(floor(self.y - self.radius)), int(ceil(self.y + self.radius)) + 1):
+            for x in range(int(floor(self.x - self.radius)), int(ceil(self.x + self.radius)) + 1):
                 if self.is_inside(x, y):
                     pixel_in_list.append([x, y])
 
--- a/neo/test/coretest/test_regionofinterest.py
+++ b/neo/test/coretest/test_regionofinterest.py
@@ -8,10 +8,47 @@ class Test_CircularRegionOfInterest(unittest.TestCase):
 
     def test_result(self):
         seq = ImageSequence([[[]]], spatial_scale=1, frame_duration=20 * pq.ms)
-        self.assertEqual((CircularRegionOfInterest(seq, 6, 6, 1).pixels_in_region()), [[6, 5], [5, 6], [6, 6]])
+        # `is_inside` tests a closed disc, so radius 1 and radius 1.01 enclose the same
+        # pixel centres. The radius 1 expectation used to be [[6, 5], [5, 6], [6, 6]],
+        # which contradicted both the line below it and `is_inside` itself.
+        self.assertEqual(
+            (CircularRegionOfInterest(seq, 6, 6, 1).pixels_in_region()), [[6, 5], [5, 6], [6, 6], [7, 6], [6, 7]]
+        )
         self.assertEqual(
             (CircularRegionOfInterest(seq, 6, 6, 1.01).pixels_in_region()), [[6, 5], [5, 6], [6, 6], [7, 6], [6, 7]]
         )
+
+    def test_pixels_in_region_matches_is_inside(self):
+        """Every pixel `is_inside` accepts must be enumerated by `pixels_in_region`."""
+        seq = ImageSequence([[[]]], spatial_scale=1, frame_duration=20 * pq.ms)
+        for x, y, radius in ((6, 6, 1), (10, 10, 5), (6, 6, 2.5), (7, 4, 3), (5, 5, 1.01)):
+            roi = CircularRegionOfInterest(seq, x, y, radius)
+            enumerated = {tuple(pixel) for pixel in roi.pixels_in_region()}
+            margin = int(radius) + 2
+            accepted = {
+                (px, py)
+                for py in range(y - margin, y + margin + 1)
+                for px in range(x - margin, x + margin + 1)
+                if roi.is_inside(px, py)
+            }
+            self.assertEqual(enumerated, accepted, f"disagreement for x={x}, y={y}, radius={radius}")
+
+    def test_pixels_in_region_is_symmetric_about_centre(self):
+        """The disc must reach as far right and up as it does left and down."""
+        seq = ImageSequence([[[]]], spatial_scale=1, frame_duration=20 * pq.ms)
+        roi = CircularRegionOfInterest(seq, 10, 10, 5)
+        pixels = roi.pixels_in_region()
+
+        # (15, 10) sits exactly `radius` from the centre, so the closed disc includes it.
+        self.assertTrue(roi.is_inside(15, 10))
+        self.assertIn([15, 10], pixels)
+        self.assertIn([10, 15], pixels)
+
+        xs = [pixel[0] for pixel in pixels]
+        ys = [pixel[1] for pixel in pixels]
+        self.assertEqual((min(xs), max(xs)), (5, 15))
+        self.assertEqual((min(ys), max(ys)), (5, 15))
+        self.assertEqual(len(pixels), 81)
 
 
 class Test_RectangularRegionOfInterest(unittest.TestCase):


adityasingh2400:
The readthedocs check here is a builder-side network failure, not something this branch does.

The build dies in `doc/source/rawio.rst`, whose IPython block calls `urlretrieve` against `web.gin.g-node.org` to pull the ephy_testing_data sample files. The traceback bottoms out in `ssl.wrap_socket` with a `ConnectionResetError`, so the download was cut off partway rather than the docs failing to build.

This PR only touches `neo/core/regionofinterest.py` and its test, nothing under `doc/` and nothing that runs during the docs build. The same check is green on my other two open PRs against this repo, #1893 and #1894, which is what you would expect if the cause were the fetch rather than the branch.

A rebuild should clear it. I have left it alone rather than pushing an empty commit to force one, but say the word if you would rather I did.

adityasingh2400:
The Read the Docs check is red here and it is a network failure on the builder rather than anything in this change.

The build dies while sphinx is reading `doc/source/rawio.rst`, inside the IPython directive that executes a live block:

```
File .../urllib/request.py:1351, in AbstractHTTPHandler.do_open
    raise URLError(err)
URLError: <urlopen error [Errno 104] Connection reset by peer>
...
RuntimeError: Unexpected exception in `.../doc/source/rawio.rst` line None
```

This PR touches `neo/core/regionofinterest.py` and `neo/test/coretest/test_regionofinterest.py` only, with no documentation changes, so nothing here reaches `rawio.rst`.

I also checked the version angle before assuming it. The two neo builds earlier the same day, at 12:07 and 12:40, resolved to the same `ipython-9.16.1` as this one and both passed, so this is not the IPython regression that has been breaking docs builds elsewhere. It is a connection reset partway through a download.

A rebuild should clear it. I have not pushed an empty commit to force one, since that seemed like the wrong way to retrigger, but I am happy to if that is easier than clicking rebuild.

Arthur031221:
This resolves what I reported in #1889, so I went over the change.

The bounds here are better than the tighter ones I suggested in that issue.
I compared both against is_inside over a tenths grid of 40,000 centre and
radius combinations: the padded form used in this PR agreed with is_inside
everywhere, while the ceil and floor plus one form I proposed drops a
boundary pixel in about 0.2 percent of cases through float rounding. Centre
1.8, radius 3.8 is one of them: 1.8 - 3.8 evaluates to -1.9999999999999998,
ceil turns that into -1, and x = -2 is never tested although is_inside
accepts it. The padding costs a perimeter band of rejected candidates, up to
44 extra visits at radius 5 with a half integer centre, which is nothing at
these sizes. So the shape of this fix is the right one, not just an
equivalent one.

The history of how the plus one disappeared is already written up in #1889,
so I will not repeat it here.

adityasingh2400:
Thanks for going and checking it rather than taking my word for it, and for the counterexample. I reproduced it before replying.

```
1.8 - 3.8   = -1.9999999999999998
ceil(...)   = -1     the tighter lower bound
floor(...)  = -2     what this PR uses
is_inside(-2) is True, so the tighter form never offers it
```

I ran my own sweep to see the size of it, one dimensional over 39800 centre and radius pairs on a tenths grid. The padded form disagreed with `is_inside` zero times, the `ceil` and `floor` form disagreed 216 times. That is 0.54 percent on my grid rather than your 0.2 percent, but we are sweeping different shapes so I would not read anything into the gap. The direction is the same and it is the direction that matters.

Your point about cost is the one I would have worried about, so it is good to see it quantified. The extra visits are rejected by `is_inside` immediately and the loop is already quadratic in the radius, so a perimeter band does not change the shape of the cost.

The only red check is `docs/readthedocs.org:neo`, which is the builder failing to fetch the ephy_testing_data samples in `doc/source/rawio.rst`. This PR touches `neo/core/regionofinterest.py` and its test only.

adityasingh2400:
The `call-iotests` job went red on the re-run, and it is the same remote-data problem as the docs check rather than anything in this branch.

All 94 entries are errors, not failures, and every one is datalad failing to fetch the test corpus:

```
datalad.support.exceptions.IncompleteResultsError
  'https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/': The requested URL returned error: 403
```

They are spread across unrelated readers, `test_alphaomegarawio.py`, `test_axographrawio.py`, `test_axonarawio.py` and so on, which is what a missing corpus looks like rather than a code change.

`NeoCoreTest` is the job that actually exercises this change, and it passed:

```
neo/test/coretest/test_regionofinterest.py .....
621 passed, 12 skipped in 30.79s
```

`regionofinterest.py` shows up in the iotest log only in the coverage table, never in a traceback.

adityasingh2400:
Update on the red I described above, since it is out of date.

A later re-run of `call-iotests` finished green at 22:40 UTC that same evening, about twenty minutes after I posted. This branch is now passing all 74 checks, including both `call-iotests` jobs and the readthedocs build. The gin.g-node.org 403s were transient after all.

Nothing in the branch changed in between, so this is ready for review whenever you have time.

zm711:
Thanks @adityasingh2400, I've tagged the appropriate maintainer to review this. Unfortunately for speediness August is a big vacation month for many of the team members on the project. I will keep on the team to review this as soon as folks are back from their time off.

adityasingh2400:
Thanks for tagging someone and for the heads up about August. No rush from my side.

The branch is green and I will leave it alone unless something upstream moves under it, so it should be ready whenever your colleagues are back.

PR #1898, opened 2026-08-12
Read NeuroExplorer files above 2 GB and fix spike waveform offsets

h-mayorquin:
Rodrigo Paz from the Nelson lab can't convert their NeuroExplorer sessions as most of their files are above 2 GB.
This PR reads `DataOffset` as unsigned and that recovers the true offset for any file below 4 GB. Above that the low bits are really gone and the file has to be re-exported as `.nex5`, which uses 64 bit offsets. I also fixed the spike waveforms which were read from `offset + n * 2` when the timestamps that precede them are 4 bytes each.

I tested both on a 2.5 GB recording of theirs with 196 variables, 69 of them past the boundary. Every array lands where the headers say and the first samples of `AD18` read `-208, -241, -282` instead of `193, 215, 224`.

### DIFF
--- a/neo/rawio/neuroexplorerrawio.py
+++ b/neo/rawio/neuroexplorerrawio.py
@@ -101,7 +101,7 @@ class NeuroExplorerRawIO(BaseRawIO):
                 sig_channels.append((name, _id, sampling_rate, dtype, units, gain, offset, stream_id, buffer_id))
                 self._sig_lengths.append(entity_header["NPointsWave"])
                 # sig t_start is the first timestamp if datablock
-                offset = entity_header["offset"]
+                offset = int(entity_header["offset"])
                 timestamps0 = self._memmap[offset : offset + 4].view("int32")
                 t_start = timestamps0[0] / self.global_header["freq"]
                 self._sig_t_starts.append(t_start)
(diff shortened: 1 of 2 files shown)


neo/rawio/neuroexplorerrawio.py
```diff
@@ -158,13 +158,13 @@ def _get_analogsignal_chunk(self, block_index, seg_index, i_start, i_stop, strea
...
-        n = entity_header["n"]
-        nb_sample = entity_header["NPointsWave"]
+        n = int(entity_header["n"])
+        nb_sample = int(entity_header["NPointsWave"])
         # offset = entity_header['offset']
         # timestamps = self._memmap[offset:offset+n*4].view('int32')
         # offset2 = entity_header['offset'] + n*4
         # fragment_starts = self._memmap[offset2:offset2+n*4].view('int32')
-        offset3 = entity_header["offset"] + n * 4 + n * 4
+        offset3 = int(entity_header["offset"]) + n * 4 + n * 4
```
zm711: could we get a comment here for the offset logic. Maybe more for my benefit. whenever I have to go back and read one of these readers I always have to re-teach myself why the values are the way they are. A quick comment about the `4*n` would be super useful.

zm711: Also why offset and offset3. What is the difference (if you can fit that in a comment I mean).

zm711 (commented):
Looks good to me. Just a light request for a couple comments for ease in case I have to review this in the future. Not a blocker for merging, though.

neo/rawio/neuroexplorerrawio.py
h-mayorquin: I used variables to try to make this more clear, check out the last discussion.

zm711 (approved):
This looks good to me. Thanks for the better variable names.

